PR #24, opened 2017-09-19
fix refresh of contexts and executions

ebiiii:


ebiiii:
Ok for me, you can merge

PR #25, opened 2017-09-20
docker: separate server and nodejs package

jirikuncar:
closes SwissDataScienceCenter/renga#87

### DIFF
--- a/.travis/deploy-docker.sh
+++ b/.travis/deploy-docker.sh
@@ -1,8 +1,9 @@
 #!/bin/bash
 
 docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD" $DOCKER_REGISTRY
-docker build . -t "${DOCKER_REGISTRY}${DOCKER_REPOSITORY}renga-ui:$(node -p "require('./package.json').version")"
-docker push "${DOCKER_REGISTRY}${DOCKER_REPOSITORY}renga-ui:$(node -p "require('./package.json').version")"
-docker tag "${DOCKER_REGISTRY}${DOCKER_REPOSITORY}renga-ui:$(node -p "require('./package.json').version")" "${DOCKER_REGISTRY}${DOCKER_REPOSITORY}renga-ui:latest"
-docker push "${DOCKER_REGISTRY}${DOCKER_REPOSITORY}renga-ui:latest"
 
+docker-compose build
+docker-compose push
+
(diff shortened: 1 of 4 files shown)


.travis/deploy-docker.sh
```diff
@@ -1,8 +1,5 @@
 #!/bin/bash
 
 docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD" $DOCKER_REGISTRY
-docker build . -t "${DOCKER_REGISTRY}${DOCKER_REPOSITORY}renga-ui:$(node -p "require('./package.json').version")"
-docker push "${DOCKER_REGISTRY}${DOCKER_REPOSITORY}renga-ui:$(node -p "require('./package.json').version")"
-docker tag "${DOCKER_REGISTRY}${DOCKER_REPOSITORY}renga-ui:$(node -p "require('./package.json').version")" "${DOCKER_REGISTRY}${DOCKER_REPOSITORY}renga-ui:latest"
-docker push "${DOCKER_REGISTRY}${DOCKER_REPOSITORY}renga-ui:latest"
-
+docker-compose build
+docker-compose push
```
ebiiii: we may still want to have the tagged version in addition to the "latest". I think we can just add it after the docker-compose push

docker-compose.yml
```diff
@@ -0,0 +1,34 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+version: '3'
+volumes:
+  ui_dist:
+services:
+  ui:
```
ebiiii: would this configuration of services (ui and ui-server) then need to be replicated in the docker-compose of renga?

jirikuncar: Yes

.travis/deploy-docker.sh
jirikuncar: Sure, I was showing this as proof of concept since the PR on renga was sleeping.

ebiiii:
What I had in mind is more like https://github.com/SwissDataScienceCenter/renga-ui/pull/26.
It only needs one docker image to be pushed.
We can close this PR I suppose.

jirikuncar:
We should have separate Docker for the static files that can be directly served using nginx and traefik. We can discuss it IRL.

rokroskar:
should this be closed since #26 was merged?

PR #29, opened 2017-09-25
Made webpack output files with hashes

leafty:
Note: if the following error appears when loading the app in dev mode (at http://localhost:5000)
> RuntimeError: Flask-Webpack requires 'WEBPACK_MANIFEST_PATH' to be set and it must point to a valid json file.

just wait a bit for webpack to compile the app


Makefile
```diff
@@ -2,16 +2,16 @@ ifndef PLATFORM_VERSION
 	PLATFORM_VERSION = latest
 endif
 
-npm-targets = \
-	dist/css/main.css \
-	dist/css/material-components-web.css \
-	dist/js/main.js \
-	dist/js/material-components-web.js
+npm-targets += dist/manifest.json
```
jirikuncar: is there a reason why do you prefer `+=` over single multi-line definition?

leafty: Edits generate clean git diffs

PR #34, opened 2017-09-29
Properly set app root

### LINKED ISSUE #33: Set cookie only for /ui
PR coming soon

leafty:
Closes #33

### DIFF
--- a/config/webpack.config.base.js
+++ b/config/webpack.config.base.js
@@ -49,7 +49,7 @@ let config = {
     ]),
     new ManifestPlugin({
       fileName: 'manifest.json',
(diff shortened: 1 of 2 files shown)


config/webpack.config.base.js
```diff
@@ -49,7 +49,7 @@ let config = {
     ]),
     new ManifestPlugin({
       fileName: 'manifest.json',
-      seed: { assets: {}, publicPath: '/static/' },
+      seed: { assets: {}, publicPath: '/ui/static/' },
```
jirikuncar: we should make this configurable if possible

PR #39, opened 2017-10-16
add project selector to UI

ebiiii:


ebiiii:
You can proceed as suggested. At the moment the project selector doesn't affect the views, so it is not critical for the behavior of the UI.

PR #54, opened 2017-10-24
Add tutorial

ebiiii:
Not all features are fully functional...


src/components/tutorial/tutorial.html
```diff
@@ -0,0 +1,223 @@
...
+        <div>
+          <h3 class="headline mb-0">Launch an execution</h3>
+        </div>
+        <v-layout>
+          <v-flex xs12 sm4>
+            <v-card class="ma-1 pa-3">
+              <v-list two-line subheader>
+                <v-subheader inset>From an IDE</v-subheader>
+              </v-list>
+              <v-select label="base image" required :items="[{text: 'jupyter notebook', value: 'jupyter/scipy-notebook'}, {text: 'R studio', value:'rocker/rstudio'}]" v-model="imageIDE"></v-select>
```
ableuler: Actually, would it help here to specify a tag for the docker image, like rocker/rstudio:latest? I hat tons of versions of this image after launching R studio through the ui.

leafty: I reported this bug already: https://github.com/SwissDataScienceCenter/renga-deployer/issues/31. You can bump it up I guess.

ebiiii:
Regarding the timeout, this is a "hardcoded" value somewhere in the flask app. We can think about what is a good value for there, but in general either the service is quick or it is asynchronous, but it should not use long awaiting HTTP queries for that. (As far as I understood it will anyway change with the architecture refactoring)

The provided notebook images will be modified to directly login. I just took the plain jupyter and rstudio for the moment.

PR #56, opened 2017-10-25
parametrize deploy ip

ebiiii:


leafty:
Thanks! Will test and merge if it works.

Issue #38, opened 2017-10-12
Add file modal window disappears when selecting file on chrome/opera


leafty:
Small bug which can be annoying for chrome/opera users and should be fixed for the datajam.

ableuler:
Same happens on chrome.

PR #60, opened 2017-11-01
add interactions to graph view

### LINKED ISSUE #48: Add interaction options to each node in the graph
Make it possible to trigger the actions related to a node directly from the UI.

ableuler:
Resolves #48 (and #38)
- Extracted file upload, file version upload, execution launching and bucket creation to separate dialog components. Those components should also be used in tutorial view (new issue)
- Reused those components in graph tooltip
- Started factoring out api calls into separate module, respective issue remains open


src/components/deploy/execution/index.ts
```diff
@@ -18,25 +18,28 @@
...
+import { router } from '../../../main'
 import { GraphItem } from '../../graph-item-list/graph-item'
 
+
 Component.registerHooks([
   'beforeRouteEnter',
   'beforeRouteLeave',
   'beforeRouteUpdate'
-])
+]);
```
ebiiii: I think it is not necessary

src/components/graph/index.ts
```diff
@@ -18,25 +18,64 @@
...
+    activeVertex: DisplayVertex = null
+    activeEdge: DisplayEdge = null
+
+    get tooltipVertex () {
+        return (this.selectedVertex ? this.selectedVertex : this.activeVertex)
+    }
+
+    dialog: string = null
+
+    closeDialog() {
```
ebiiii: Would it be easy to trigger a reload of the graph data when an action is performed (through a dialog), without reloading the whole page? I was using that as a shortcut for the other component, but I think on the long term, it is better to just refresh the content.

src/components/deploy/execution/index.ts
ableuler: Yes, I've switched to no-semi-colon too. This one must have slipped.

src/components/deploy/execution/index.ts
```diff
@@ -51,6 +54,11 @@ export class ExecutionComponent extends Vue {
           { text: 'Namespace', value: 'deployer:execution_namespace' }
         ]
 
+    dialog: string = null;
```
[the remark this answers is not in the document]
ableuler: I think this is fine. It will interpret any non-null and non-empty string as true and null or '' as false. Like this I can use dialog='bucket', 'execution', etc to distinguish

src/components/graph/index.ts
ableuler: Agree 100%. Can we create a separate issue for all the reloads? I thought it was already quite a lot of change for one PR and for the time being I just kept the reloads where they had been before.

src/components/deploy/execution/execution.html
```diff
@@ -4,21 +4,11 @@
...
-          <v-spacer></v-spacer>
-          <v-btn secondary flat @click.native="execDialog = false">Cancel</v-btn>
-          <v-btn secondary @click.native="addExec($event)">Start</v-btn>
-        </v-card-actions>
-      </v-card>
+    <v-dialog v-model="dialog">
+        <v-btn secondary fixed bottom right fab slot="activator"><v-icon>play_arrow</v-icon></v-btn>
+        <execution-dialog
+                v-on:closeDialog="closeDialog"
+                v-bind:contextUUID="contextUUID"
```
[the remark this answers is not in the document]
ableuler: Yes, that's much nicer indeed.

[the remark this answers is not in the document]
ableuler: Is this more or less what you also had in mind?

src/components/graph/index.ts
ebiiii: Yep, let's move this to an issue.

src/components/graph/index.ts
```diff
@@ -69,10 +69,14 @@ export class GraphComponent extends Vue {
 
     dialog: string = null
 
-    closeDialog() {
+    cancel() {
         this.dialog = null
     }
 
+    success() {
+        location.reload()
```
ebiiii: do you need to add `this.dialog = null` here too ?

src/components/deploy/execution/index.ts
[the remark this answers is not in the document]
ebiiii: ok, make sense. (it is consistent with the other cases where you use the string value for selecting the dialog)

src/components/graph/index.ts
ableuler: Yes, otherwise the cancel button in a dialog triggered from the graph view doesn't do anything. We could also do it directly inside the `graph.html` template (`v-on:cancel="dialog = null"`) but I think it's cleaner like this.

PR #68, opened 2017-11-07
Various improvements of the graph display

ableuler:


src/components/graph/graph.styl
```diff
@@ -3,6 +3,11 @@
 .links line {
   stroke: $grey.lighten-1
   stroke-width: 3.5px
+  marker-end: url(#arrowhead)
```
jirikuncar: 🎆

src/components/graph/index.ts
```diff
@@ -109,30 +120,19 @@ export class GraphComponent extends Vue {
...
-                        }
-                    }
-                }
+            // Use this temporarily to enrich executions with the context that has produced them in order to construct
+            // the correct url for this execution (contextUUID should be part of the API response)
+
+            let newEdge = this.edges[this.edges.length - 1]
+
+            if (newEdge.display_name === 'launch') {
+                let context = this.vertices.find( vertex => {
```
ciyer: You can write these as oneliners. (Applies to the others below as well.)

src/components/graph/index.ts
```diff
@@ -209,7 +219,12 @@ export class GraphComponent extends Vue {
         }
     }
 
-
+    dispatcher = d3.dispatch('collapse')
```
ciyer: Need a let.

src/components/graph/index.ts
```diff
@@ -450,40 +488,65 @@ export class GraphComponent extends Vue {
...
                     link.target.mergedTo = null
                 }
             }
 
-            // Change path of clicked vertices.
-            d3.select(this)
+            // Check for other zero-length links and make them (or their heads) invisible
+            for (let link of edges) {
+                let linkC = linkCoords(link)
+                if (linkC.start.x === linkC.end.x && linkC.start.y === linkC.end.y) {
```
ciyer: Extract the condition to a variable:
const isInvisible = linkC.start.x === linkC.end.x && linkC.start.y === linkC.end.y
d3.select(`#${link.id}`).classed('invisible', isInvisible)

src/components/graph/index.ts
ableuler: Didn't see it either when we looked at it, but this is actually still part of the GraphComponent class declaration. I moved it further up so it comes before all the methods (where one would expect it).

PR #73, opened 2017-11-09
Context dialog

ableuler:
- Add context-dialog component along the lines of those added in #48.
- Make context- and bucket creation respect the selected project.


src/components/dialogs/index.ts
```diff
@@ -59,6 +59,40 @@ export class ProjectDialogComponent extends Vue {
...
+
+    addContext() {
+        this.progress = true
+
+        createContext(this.context_image, this.context_ports.split(/\s*,\s*/), this.projectId)
+            .then(response => {
+                    return response.json()
+                }
+            ).then(response => {
+                console.log('create', response)
```
ciyer: Remove this log statement or add a TODO comment before to clarify intention.

Issue #58, opened 2017-10-27
Integrating existing/external code/algorithms


ableuler:
To exploit the features built in SwissDataScienceCenter/renga-python#63, I suggest:

- [x]  Add option to select a notebook from the list of files as input when creating a context.
- [x]  Add option to specify the number of input/output slots when creating a context.
- [x]  Add option to select files as input and for writing output to (as new versions?) when running an execution from a context.
- [ ]  Adapt visual representation of knowledge graph: (Use different icons for files with a 'code'-ending, create vertices for the input/output slots, relink edges)

jirikuncar:
see also SwissDataScienceCenter/renga-python#63

ableuler:
To exploit the features built in SwissDataScienceCenter/renga-python#63, I suggest:
- [ ] Add option to select a notebook from the list of files as input when creating a context.
- [ ] Add option to specify the number of input/output slots when creating a context.
- [ ] Add option to select files as input and for writing output to (as new versions?) when running an execution from a context.
- [ ] Adapt visual representation of knowledge graph:
 - Use different icon for files with a 'code'-ending
 - Create vertices for the input/output slots, relink edges

PR #75, opened 2017-11-10
Enable context input/output slots in ui.

ableuler:
Addresses the first 3 bullet points of #58.
- Select an existing '.ipynb' file and add it as notebook-input when creating a context
- Add input output slots when creating a context
- Add default files for the input/output slots when creating a context
- Choose files for non-default input/output slots when creating an execution.


src/components/deploy/details/index.ts
```diff
@@ -53,7 +64,17 @@ export class DetailExecutionComponent extends Vue {
...
+
+        // For rengahub/minimal-notebook we adapt the url and add the token.
+        let tokenLabel = this.context.spec.labels
+            .find( label => label.includes('renga.notebook.token='))
+
+        if (tokenLabel) {
+            url += tokenLabel.replace('renga.notebook.', '/notebooks/current_context/inputs/notebook?')
+        }
+
+        window.open(url)
```
ebiiii: Just mentioning so that we don't forget, this allows anybody who has access to the "context" vertex to open the notebook of somebody else while he might be still working on it. It is an interesting sharing feature, but could also be exploited with less good intentions ;-)

src/components/deploy/context/context.html
```diff
@@ -1,10 +1,9 @@
 <div>
-    <v-progress-linear v-bind:indeterminate="true" v-show="progress" ></v-progress-linear>
     <graph-item-table v-bind:headers="headers" name_key="resource:file_name" v-bind:url_list="'./api/deployer/contexts'" v-on:selected="onSelect" v-bind:parser="parser">
       <span slot="title">Execution Contexts</span>
     </graph-item-table>
     </v-navigation-drawer>
-    <v-dialog v-model="dialog">
+    <v-dialog v-model="dialog" width="700px">
```
jirikuncar: IMHO this should be done in css so we can have a responsive version.

src/components/deploy/details/index.ts
ableuler: I agree! But that's more a question of whether the notebook token should be stored as a context label than a UI topic, right?

src/components/deploy/context/context.html
ableuler: Yes, you're right it should be done via css, but I've found the behaviour of the v-dialog to be a bit unexpected and didn't want to burn too much time on it at this point. However, the responsiveness is ok. The "width" property seems to act more like "max-width", and the dialog adapts nicely to a narrow screen.

src/components/dialogs/index.ts
```diff
@@ -66,27 +121,93 @@ export class ProjectDialogComponent extends Vue {
...
+    onFilesChanged() {
+        this.notebooks = this.files.filter((file: any) => file.text.slice(-6) === '.ipynb')
+    }
 
     addContext() {
         this.progress = true
 
-        createContext(this.context_image, this.context_ports.split(/\s*,\s*/), this.projectId)
+        // Leave out the dummy element '' of the inputSlot / outputSlot arrays and create the labels.
+        let labels: string[] = []
```
ciyer: Let's simplify this code a bit. 

```
function collectLabels(names, files, path) {
    const labels = []
    for (let i = 0; i < names.length - 1; i++) {
        let label = `renga.context.${path}.${names[i]}`
        if (files[i]) label += `=${files[i].file.id}`
        labels.push(label)
    }
    return labels
}

labels = collectLabels(this.inputSlotNames, this.inputSlotFiles, "inputs")
labels = labels.concat(collectLabels(this.outputSlotNames, this.outputSlotFiles, "outputs"))
```

PR #76, opened 2017-11-12
file upload directly from url

### LINKED ISSUE #59: Integrating existing/external data sets

ableuler:
Resolves #59 on the UI without involvement of the storage service:
- Add "file-from-url" option to file- and file version upload dialog.
- Add simple web-proxy to flask server app to avoid CORS problems.

### DIFF
--- a/src/utils/renga-api.ts
+++ b/src/utils/renga-api.ts
@@ -17,8 +17,6 @@
  */
 
 
-
-
 export function createBucket(bucketName: string, bucketBackend: string, projectId?: number) {
 
     let payload = JSON.stringify({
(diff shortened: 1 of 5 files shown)


src/utils/renga-api.ts
```diff
@@ -129,30 +126,58 @@ export function addFileVersion(fileId: number, fileInput: any) {
...
+                        postRequest['body'] = reader.result
+                        resolve( fetch('./api/storage/io/write', postRequest) )
                     }
-                )
+                    reader.readAsArrayBuffer(e.files[0])
+                })
+                return readerPromise
+
+            } else if (fileUrl) {
+                return fetch('./webproxy', getRequest)
```
ebiiii: for small datasets why not, but if it could be transfered directly from the flask server, without getting the data in the browser, it could be also quicker ?

ableuler: You're right, that would have been the better option once the flask server needed to be involved. But still, I would rather leave out the UI completely and implement it as part of the storage service so that the same functionality can be accessed from the CLI/python-client too.

ebiiii: Make sense. Let's go with that as a first step.

PR #87, opened 2017-11-21
Datajam getting ready

ableuler:


src/components/dialogs/index.ts
```diff
@@ -110,10 +110,16 @@ export class ProjectDialogComponent extends DialogBaseComponent {
         createBucket(this.bucketName, this.bucketBackend, this.projectId)
             .then(response => {
                 console.log('create', response)
+                this.bucketName = 'bucket'
+                this.bucketBackend = 'local'
```
ebiiii: maybe empty string would be better here. BTW you can also take the opportunity of this PR to change the default values in the tutorial to 's3' for the storage and 'k8s' for the execution ?

ableuler: Good point, adapted it. For the s3 and k8s defaults, maybe we can set those in a data-jam specific branch?

PR #98, opened 2017-12-06
Datajam changes into master

ebiiii:
Let's use this pull request to review the changes (it's easier this way to see the actual diff)

### DIFF
--- a/src/components/tutorial/index.ts
+++ b/src/components/tutorial/index.ts
@@ -24,7 +24,7 @@ import fetchItemList from '../graph-item-list'
 import { UserState, NoUser, LoggedUser } from '../user'
 import { Project }  from '../project'
 import { FileObj, Bucket }  from '../storage'
-import { addFile, addFileVersion, createBucket, createContext,
+import { addFile, addFileVersion, createBucket, createContext, duplicateFile,
     getContext, getBucketBackends, getProjects, getProjectFiles, runContext, updateFile } from '../../utils/renga-api'
 
 
@@ -322,6 +322,21 @@ export class TutorialComponent extends Vue {
     executeIDE() {
         this.progress = true
 
+        if (this.nbIDE > 0) {
+            duplicateFile(this.nbIDE, 0, 'Notebook copy.ipynb', this.project.id).then(r => {
+                console.log('duplicate', r)
+                return r.json()
+            }).then( result => {
+                this.nbIDE = result.id
+                this.do_executeIDE()
+            })
+        } else {
+            this.do_executeIDE()
+        }
+    }
+
+    do_executeIDE() {
+
         createContext(this.imageIDE, ['8888'], [],
             this.project.id, this.nbIDE
         ).then(response => {
@@ -338,7 +353,6 @@ export class TutorialComponent extends Vue {
                 }
             )
         })
-
     }
 
     executeDocker() {
--- a/src/components/tutorial/index.ts
+++ b/src/components/tutorial/index.ts
@@ -70,6 +70,7 @@ export class TutorialComponent extends Vue {
 
     imageIDE = ''
     nbIDE = 0
+    nbCopy = true
     notebooks: any[] = []
     imageDocker = ''
     portsDocker = ''
@@ -209,7 +210,7 @@ export class TutorialComponent extends Vue {
                 if (response.status === 201) {
                     let payload = JSON.stringify({
                         name: this.project_name,
-                        backend: 's3',
+                        backend: STORAGE_DEFAULT_BACKEND,
                         request_type: 'create_bucket'
                     })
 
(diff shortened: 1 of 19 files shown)


src/components/tutorial/index.ts
```diff
@@ -222,7 +209,7 @@ export class TutorialComponent extends Vue {
                 if (response.status === 201) {
                     let payload = JSON.stringify({
                         name: this.project_name,
-                        backend: 'local',
+                        backend: 's3',
```
ebiiii: let's parametrize the default backends

src/components/tutorial/index.ts
```diff
@@ -328,63 +319,63 @@ export class TutorialComponent extends Vue {
...
-                }
-            ).then(response2 => {
-                return response2.json()
-                }
-            ).then(response2 => {
-                console.log('executing', response2)
-                this.progress = false
-                router.push(`/deploy/context/${response.identifier}/execution/${response2.identifier}`)
+        if (this.nbIDE > 0) {
+            duplicateFile(this.nbIDE, 0, 'Notebook copy.ipynb', this.project.id).then(r => {
```
ebiiii: make the copy optional with a checkbox

src/components/tutorial/index.ts
```diff
@@ -328,63 +319,63 @@ export class TutorialComponent extends Vue {
...
-        this.do_execution(this.dockers()[this.imageIDE], '')
+    do_executeIDE() {
 
+        createContext(this.imageIDE, ['8888'], [],
+            this.project.id, this.nbIDE
+        ).then(response => {
+            console.log('create', response)
+            return response.json()
+        }).then(responsej => {
+            runContext('k8s', '', responsej.identifier).then(r => {
```
ebiiii: parametrize backend

src/components/tutorial/tutorial.html
```diff
@@ -43,23 +30,22 @@ <h3 class="headline mb-0">Choose an existing one</h3>
...
                 <v-spacer></v-spacer>
-                <v-btn secondary @click="chooseProject()">Next</v-btn>
+                <v-btn secondary @click="chooseProject()">Select and continue</v-btn>
               </v-card-actions>
             </v-card>
           </v-flex>
         </v-layout>
       </v-stepper-content>
       <v-stepper-content :step="2">
-        <div>
```
ebiiii: Should we keep the "import" option or not ?

src/components/user/index.ts
```diff
@@ -116,7 +116,7 @@ export class UserMenuComponent extends Vue {
             if (state.logged_in) {
                 this.items.push({ title: 'Execution contexts', icon: 'settings_applications', path: '/deploy/context' })
                 this.items.push({ title: 'Storage', icon: 'folder', path: '/storage' })
-                this.items.push({ title: 'Graph', icon: 'share', path: '/graph' })
+//                this.items.push({ title: 'Graph', icon: 'share', path: '/graph' })
```
ebiiii: do we keep it for the moment?

src/main.ts
```diff
@@ -34,6 +34,15 @@ import { GraphItemListComponent, GraphItemTableComponent } from './components/gr
 import { BucketDialogComponent, ContextDialogComponent, ExecutionDialogComponent,
     ProjectDialogComponent, VersionDialogComponent } from './components/dialogs'
 
+// Activate sentry if a DSN is provided through the environment variables
+import Raven from 'raven-js';
+import RavenVue from 'raven-js/plugins/vue.js';
+
+Raven
+    .config('https://250bfa79136d4bc2b19d4f8d977b4f5f@sentry.io/246730')
```
ebiiii: secret url ?

ableuler: I think the goal here should be to respect the SENTRY_DSN environment variable like all the other services. Will be a bit more tricky than for the backend stuff, but I'm trying to find a not too hacky way to do this.

src/components/user/index.ts
ableuler: Should we leave it out and retry with the graph as soon as the graph view displays only the current project?

src/components/tutorial/tutorial.html
ableuler: Having the collaborative aspects of renga in mind, I think it's actually a nice feature. But it means that I will have to loosen on the assumption that every file is in a bucket which belongs to the project.

src/components/tutorial/index.ts
ableuler: 👍

src/components/tutorial/index.ts
ableuler: For the storage we can do a query to get the available backends. The same would be nice for the deployer.

src/components/tutorial/index.ts
ableuler: ok

PR #119, opened 2018-02-22
Travis

rokroskar:
* run tests on travis
* build and push images on `development` and `master` branches

Note that tests are failing now -- that will be fixed in another PR

---
addresses https://github.com/SwissDataScienceCenter/renga/issues/117 because it will allow pushing images to registry from travis

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,7 @@
 FROM node:alpine
 
 WORKDIR /app
-COPY package.json ./
-COPY ./src /app/src
-COPY ./public /app/public
-COPY docker-entrypoint.sh ./
+COPY . /app
 # We rename index.html so we can add the right source for the keyckoak apdaptor into the script tag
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,9 +1,7 @@
 FROM node:alpine
 
 WORKDIR /app
-COPY package.json ./
-COPY ./src /app/src
-COPY ./public /app/public
+COPY . /app
```
ableuler: Are you just trying to reduce the number of lines here? The problem is that you don't want to copy the (potentially very big) node_modules directory into the docker container.

rokroskar: indeed -- see `.dockerignore`

ableuler: Ah sorry, I missed that...

rokroskar:
ok I'll fix the conflict, rebase and merge

PR #144, opened 2018-03-18
Added Schema and Model base classes

ableuler:
...and an example for possible renga models.

### DIFF
--- a/src/model/Model.js
+++ b/src/model/Model.js
@@ -1,0 +1,293 @@
+/*!
+ * Copyright 2017 - Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ *  renga-ui
+ *
+ *  Model.js
+ *
+ *  A schema is comprised of property definitions which are of the form:
+ *    {[property name]: {initial: [initial value], mandatory: [true/false], ...}
+ *
+ *
+ *  A model is a class created from a schema definition. Instances of a model have
+ *  methods to create, initialize, validate, update, etc. the object properties
+ *  according to the schema definitions.
+ *
+ *  Models are created from the Model base class by overriding the getSchema method
+ *
+ *  class modelName extends Model {
+ *    getSchema() {return someSchema}
+ *  }
+ *
+ *  or by using a convenience method:
+ *
+ *  const modelName = someSchema.toModel();
+ *
+ *  The latter variant creates instances with a nameless constructor
+ *  (instance.constructor.name = "").
+ */
+
+// TODO: Maybe use [jsdoc](http://usejsdoc.org/) here?
+
+import immutableUpdate from 'immutability-helper';
+import uuid from 'uuid';
+
+
+const SCHEMA_PROP = 'schema';
+const INITIAL_PROP = 'initial';
+const MANDATORY_PROP = 'mandatory';
+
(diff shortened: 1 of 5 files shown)


ableuler:
This is still work in progress and should only be merged after the definition of the RengaModels is more complete.

src/model/Model.js
```diff
@@ -0,0 +1,263 @@
...
+
+import immutableUpdate from 'immutability-helper';
+import uuid from 'uuid';
+
+
+const SCHEMA_PROP = 'schema';
+const INITIAL_PROP = 'initial';
+const MANDATORY_PROP = 'mandatory';
+
+class Schema {
```
ciyer: I like the Schema class, but I think, for extensibility, we should distinguish between how the field specifications are defined (which is fine as is), and how they are stored. I think we should create a 

```
class FieldSpec {
   constructor(spec) {
    this.initial = spec.initial;
    this.mandatory = spec.mandatory;
    this.schema = [...]
    // Possibly in the future
    // this.validator = spec.validator;
  }
}
```

src/model/Model.js
```diff
@@ -0,0 +1,263 @@
...
+      result = result && subValidation.result;
+      errors = errors.concat(subValidation.errors);
+    }
+  });
+  return {result, errors}
+}
+
+// Check if a single value is valid if the corresponding property in the schema
+// is marked as mandatory.
+function validateSingleValue(value) {
```
ciyer: This should have a different signature:

    function validateField(fieldSpec, fieldValue)

src/model/Model.test.js
```diff
@@ -0,0 +1,180 @@
...
+ *  renga-ui
+ *
+ *  Model.test.js
+ *  Tests for models.
+ */
+
+import { Schema, modelUpdateReducer } from './Model';
+import { createStore, combineReducers } from 'redux';
+
+const SimpleSchema = new Schema({
```
ciyer: To follow conventions, the instances of Schema should have lower-case names: simpleSchema, complexSchema.

src/model/Model.test.js
```diff
@@ -0,0 +1,180 @@
...
+  it('complex instance updates correctly to new instane using original syntax', () => {
+    const complexInstance = new ComplexModel();
+    const updatedInstance = complexInstance.update({subthing: {age: {$set: 1}}});
+    expect(updatedInstance).toEqual({...complexInstance, subthing: {age: 1}});
+    expect(updatedInstance).not.toBe(complexInstance);
+  });
+});
+
+describe('update redux store', () => {
+  const simpleReducer = (state=new SimpleModel(), action) => modelUpdateReducer(state, action);
```
ciyer: This is very nice. But I think it would be more natural to say 
```
const model = new SimpleModel();
const simpleReducer = simpleSchema.reducer(state=model, action);
const store = createStore(simpleReducer);
```

src/model/Model.test.js
```diff
@@ -0,0 +1,180 @@
...
+  const complexReducer = (state=new ComplexModel(), action) => modelUpdateReducer(state, action);
+  const reducer = combineReducers({simple: simpleReducer, complex: complexReducer});
+
+  it('updates simple instance in redux store', () => {
+    const store = createStore(simpleReducer);
+    const state = store.getState();
```
ciyer: I think this is clearer as:
```
const model = store.getState();
store.dispatch(model.setAction('name', 'John Doe'));
...
```

src/model/Model.test.js
ableuler: Agree - done!

src/model/Model.test.js
ableuler: Actually the reducer is almost independent of the schema. The only thing that is specific to a schema is the default state input in case none is provided. Now there are 3 possible ways to do this:
1. Define the default state in place (what I originally did).
2. Tie the reducer to the schema (as you suggested).
3. Tie the reducer to the model as static method (what I instinctively would have done).

2) - 3) are enabled too now and tested.

src/model/Model.test.js
ableuler: Agree - state is not the best name. Renamed it to `simple` to be consistent with the nested test case.

src/model/Model.js
ableuler: So do I understand correctly, that the calling sequence 
```
const someSchema = new Schema({
  name: {initial: '', mandatory: true},
  purpose: {initial: '', mandatory: false}
});
```
will remain exactly the same, but internally the schema instance will look something like:
```
{
  name: nameFieldSpec,
  purpose: purposeFieldSpec
}
```
where `nameFieldSpec` and `purposeFieldSpec` are both instance of the `FieldSpec` class?

ciyer: Yes, that is correct. This keep the API minimal and nice for the clients of Schema, but allows us to have a unified API for the internals, since all user-provides field spec will be converted into FieldSpec objects.

PR #146, opened 2018-03-20
141 project page

### LINKED ISSUE #141: Improve layout of Project page
- [x] The top bar of the project should show the summary and have pills for the sub content: Overview, Kus, Data, Code, Contributors project-nav
- [x] ~~Subpages should include breadcrumbs~~ -- Not necessary after restructuring

ciyer:
Restructure the project page to show Kus under the project. Closes #141.

### DIFF
--- a/src/project/Project.js
+++ b/src/project/Project.js
@@ -33,6 +33,7 @@ import { createStore } from '../utils/EnhancedState'
 import Present from './Project.present'
 import State from './Project.state'
 import Ku from '../ku/Ku'
+import Notebook from '../file/Notebook'
 
 class New extends Component {
   constructor(props) {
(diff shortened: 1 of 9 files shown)


src/project/Project.js
```diff
@@ -111,13 +113,25 @@ class View extends Component {
...
     const description = displayMetadataValue(state.core, 'description', 'no description');
     const readmeText = state.data.readme.text;
-    return {title, description, displayId, internalId, visibilityLevel, externalUrl, readmeText}
+    const lastActivityAt = state.core.last_activity_at;
+    const overviewUrl = `/projects/${internalId}/`;
+    const kusUrl = `/projects/${internalId}/kus`;
+    /* eslint-disable no-useless-escape */
+    const kuUrl = `/projects/${internalId}/kus/:kuIid(\\d+)`;
+    const kuList = <Ku.List key="kus" projectId={internalId} {...ownProps} client={this.props.client} />
+    const kuView = (p) => <Ku.View key="ku" projectId={internalId}
```
ableuler: Is there an advantage in defining all these variables depending on `internalId` already here and passing them down as props rather than just passing down `internalId` and constructing the values inside `ProjectView`? (I was a bit confused at first to find all these things in `mapStateToProps`)  It would definitely be an advantage if there was a use case for the `ProjectView` component where the relation of `kuUrl`, `kuList`, etc was a different one than what's defined above, but I don't see where we would need this kind of flexibility. Or am I missing something here?

src/project/Project.js
```diff
@@ -111,13 +113,25 @@ class View extends Component {
     const title = displayMetadataValue(state.core, 'title', 'no title');
     const description = displayMetadataValue(state.core, 'description', 'no description');
     const readmeText = state.data.readme.text;
-    return {title, description, displayId, internalId, visibilityLevel, externalUrl, readmeText}
+    const lastActivityAt = state.core.last_activity_at;
+    const overviewUrl = `/projects/${internalId}/`;
+    const kusUrl = `/projects/${internalId}/kus`;
+    /* eslint-disable no-useless-escape */
+    const kuUrl = `/projects/${internalId}/kus/:kuIid(\\d+)`;
+    const kuList = <Ku.List key="kus" projectId={internalId} {...ownProps} client={this.props.client} />
```
ableuler: I think `{...ownProps}` can be omitted here as `<VisibleProjectView />` doesn't get any props.

src/project/Project.js
ciyer: Yes, the routing structure is not yet really finished. I have been trying to get a nicer pattern for this, but it has been a battle. I didn't want this to hold up the merge, since I've made a lot of changes, but I will create an issue to define a  structure for routing.

src/project/Project.js
```diff
@@ -111,13 +113,25 @@ class View extends Component {
     const title = displayMetadataValue(state.core, 'title', 'no title');
```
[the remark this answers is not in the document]
ciyer: Good suggestion, this is nicer.

ciyer:
I rewrote some commits to make the history easier to follow.

PR #150, opened 2018-03-27
132 minor ui improvments

ciyer:
- Retrieve and show project tags from gitlab
- Retrieve and show stars from gitlab
- Added parameter to provide a name to a redux store, useful with the debugging tools
- Work on improving the styling of notebooks, but this is still WIP.

### DIFF
--- a/src/gitlab/client.js
+++ b/src/gitlab/client.js
@@ -164,7 +164,7 @@ export default class GitlabClient {
 
 
 function carveProject(projectJson) {
-  const result = {metadata: {core: {}, visibility: {}}, all: projectJson};
+  const result = {metadata: {core: {}, visibility: {}, system: {}}, all: projectJson};
   result['metadata']['visibility']['level'] = projectJson['visibility'];
 
   result['metadata']['core']['created_at'] = projectJson['created_at'];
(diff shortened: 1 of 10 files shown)


src/gitlab/client.js
```diff
@@ -175,6 +175,10 @@ function carveProject(projectJson) {
   result['metadata']['core']['title'] = projectJson['name'];
   result['metadata']['core']['external_url'] = projectJson['web_url'];
 
+  result['metadata']['system']['tag_list'] = projectJson['tag_list'];
+  result['metadata']['system']['star_count'] = projectJson['star_count'];
+  result['metadata']['system']['forks_count'] = projectJson['forks_count'];
```
ableuler: Maybe use `fork_count` everywhere, `star_count` is singular too.

ciyer: Both the gitlab and github APIs present this as forks_count. Search for ```forks_count``` here:

    https://developer.github.com/v3/repos/forks/

Github uses ```stargazers_count,``` which we could use instead of the gitlab ```star_count.```

ableuler: Ok, sorry - didn't check how GitLab does it. I'm fine with following GitLab.

PR #152, opened 2018-03-28
Improved react state and redux model bindings

ableuler:
Still to be brushed a little bit here and there, but basic ideas are there. Tests which involve API calls are still missing, the corresponding file is only a stub.

### DIFF
--- a/src/model/Model.js
+++ b/src/model/Model.js
@@ -65,8 +65,17 @@ class FieldSpec {
       // We ignore properties which are not part of the known field specification properties.
       if (FIELD_SPEC_PROPS.indexOf(prop) < 0) return;
 
+      // Handle arrays in Field spec definitions
+      if (prop === SCHEMA_PROP && spec[prop] instanceof Array) {
+        if (spec[prop][0] && !(spec[prop] instanceof Schema)) {
+          this[prop] = [new Schema(spec[prop][0])]
+        }
+        else {
+          this[prop] = []
+        }
+      }
       // Sub-objects in field spec definitions are turned into schema definitions.
-      if (prop === SCHEMA_PROP && !(spec[prop] instanceof Schema)) {
+      else if (prop === SCHEMA_PROP && !(spec[prop] instanceof Schema)) {
         this[prop] = new Schema(spec[prop]);
       }
       else {
@@ -155,7 +164,7 @@ class StateModel {
   }
 
   set(obj, callback) {
-    const updateObj = updateObjectFromObject(obj);
+    const updateObj = updateObjectFromObject(obj, this.get());
     this.immutableUpdate(updateObj, callback);
   }
 
@@ -218,8 +227,10 @@ class StateModelComponent extends Component {
 // where all values are undefined
 function createEmpty(schema, newObj={}) {
   Object.keys(schema).forEach((prop) => {
-
-    if (schema[prop].hasOwnProperty(SCHEMA_PROP)) {
+    if (schema[prop].hasOwnProperty(SCHEMA_PROP) && schema[prop][SCHEMA_PROP] instanceof Array) {
+      newObj[prop] = []
+    }
+    else if (schema[prop].hasOwnProperty(SCHEMA_PROP)) {
       newObj[prop] = createEmpty(schema[prop][SCHEMA_PROP])
     }
     else {
@@ -229,7 +240,7 @@ function createEmpty(schema, newObj={}) {
   return newObj;
 }
 
-// Apply the defaults defined in a schema to a generic object. WE don't overwrite
+// Apply the defaults defined in a schema to a generic object. We don't overwrite
 // already existing values, defaults are only applied to undefined values.
 function applyDefaults(schema, obj) {
   Object.keys(schema).forEach((prop) => {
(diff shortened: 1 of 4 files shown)


src/model/Model.js
```diff
@@ -25,36 +25,38 @@
...
 
 import immutableUpdate from 'immutability-helper';
-import uuid from 'uuid';
-
+import { Component } from 'react';
+// Todo: Resolve dependency from our custom store
+import { createStore } from '../utils/EnhancedState';
 
+// Property names for the field specs.
 const SCHEMA_PROP = 'schema';
```
ciyer: I think an enum is cleaner for this:
```
const PropertyName = { SCHEMA, INITIAL, MANDATORY }
```

src/model/Model.js
```diff
@@ -25,36 +25,38 @@
...
 const INITIAL_PROP = 'initial';
 const MANDATORY_PROP = 'mandatory';
-
 const FIELD_SPEC_PROPS = [SCHEMA_PROP, INITIAL_PROP, MANDATORY_PROP];
```
ciyer: If we use an enum as above, then this can become
```
Object.keys(PropertyName)
```
but we don't need to store it globally anymore, I would say.

src/model/Model.js
```diff
@@ -25,36 +25,38 @@
...
 
+// Named consts for the bindings to the store.
+const REDUX_STORE = 'redux_store_binding';
```
ciyer: Another case for an enum.
```
const StateKind = { REDUX, REACT }
```

src/model/Model.js
```diff
@@ -25,36 +25,38 @@
...
+const REACT_STATE = 'react_state_binding';
+
+// We need only one action type. The information about which
+// part of the state has to be modified is contained in the action payload.
+const UPDATE_ACTION_TYPE = 'update';
```
ciyer: Using an enum here will allow cleaner future extensibility:
```
const ActionType = { UPDATE }
```

src/model/Model.js
```diff
@@ -25,36 +25,38 @@
...
+
+// Fields which are updating are set to this value.
+const UPDATING_PROP_VAL = 'is_updating';
```
ciyer: I think we should have an enum:
```
const SpecialPropVal = { UPDATING }
```

src/model/Model.js
```diff
@@ -93,108 +95,125 @@ class Schema {
...
+      stateObject = this.reduxStore.getState();
+    }
+    else if (this.stateBinding === REACT_STATE) {
+      stateObject = this.reactComponent.state;
+    }
 
-  // Immutable update: create new modified instance.
-  update(updateObj){
-    return new this.constructor(immutableUpdate(this, updateObj))
+    if (!propertyAccessorString) {
```
ciyer: This should be moved to the top of the method for clarity and efficiency. 
The return value should be ```undefined```.

src/model/Model.js
```diff
@@ -93,108 +95,125 @@ class Schema {
...
+    else {
+      return nestedPropertyAccess(propertyAccessorString, stateObject);
+    }
   }
 
-  // Immutably set a value to a property using dot-notation syntax for
-  // the property accessor.
-  set(propertyAccessor, value) {
-    return this.update(createUpdateObject(propertyAccessor, value))
+  setOne(propertyAccessorString, value, callback) {
```
ciyer: Why is this called ```setOne```?

src/model/Model.js
ableuler: My idea here was a different one: `model.get()` without any argument does not return `undefined` but the entire state object. I found this convenient.

src/model/Model.js
ableuler: Because 
```
model.set('prop.subprop.subsubprop', value)
```
 is used to update only one value, while 
```
model.set({
  prop: {
    subprop: {
      subsubprop: value1
    }
  }, 
  prop2: value2
})
```
 can be used to update multiple values at the same time.

src/model/Model.js
ableuler: Done.

src/model/Model.js
ableuler: Done.

src/model/Model.js
ableuler: Done.

src/model/Model.js
ableuler: Done.

src/model/Model.js
ableuler: Done.

src/model/Model.js
```diff
@@ -210,34 +299,65 @@ function createEmpty(schema, newObj={}) {
...
-// Apply the defaults defined in a schema to a generic object. WE don't overwrite
+// Apply the defaults defined in a schema to a generic object. We don't overwrite
 // already existing values, defaults are only applied to undefined values.
 function applyDefaults(schema, obj) {
   Object.keys(schema).forEach((prop) => {
 
-    if (schema[prop][INITIAL_PROP] !== undefined) {
-      if (schema[prop][INITIAL_PROP] instanceof Function) {
-        obj[prop] = schema[prop][INITIAL_PROP]()
+    if (schema[prop][PropertyName.INITIAL] !== undefined) {
```
ciyer: Would be clearer (and potentially more efficient) if the re-used value were stored in variables
```
const schemaProp = schema[prop];
const initialPropValue = schemaProp[PropertyName.INITIAL];
```

But this is not a show stopper.

ableuler: You're right, I've quickly changed this.

PR #154, opened 2018-04-04
129 new project

ableuler:
Introduce very basic project creation through the ui:
- Use the new state schema / models
- Change some minor aspects of the model implementation / validation

### DIFF
--- a/src/model/RengaModels.js
+++ b/src/model/RengaModels.js
@@ -26,29 +26,30 @@
 import { Schema } from './Model'
 
 const userSchema = new Schema({
-  name: {initial: '', mandatory: true}
+  name: {initial: '', mandatory: false},
+  username: {initial: '', mandatory: true},
+  avatarUrl: {initial: '', mandatory: false}
 });
-const User = userSchema.toModel();
 
 const metaSchema = new Schema({
-  author: {schema: userSchema, mandatory: true},
-  description: {initial: '', mandatory: true},
-  displayId: {initial: '', mandatory: true},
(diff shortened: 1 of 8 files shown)


src/model/RengaModels.js
```diff
@@ -26,29 +26,30 @@
...
-const User = userSchema.toModel();
 
 const metaSchema = new Schema({
-  author: {schema: userSchema, mandatory: true},
-  description: {initial: '', mandatory: true},
-  displayId: {initial: '', mandatory: true},
-  createdAt: {initial: () => new Date()}
+  id: {initial: '', mandatory: false},
+  // author: {schema: userSchema, mandatory: false},
+  createdAt: {initial: () => new Date()},
```
ciyer: I think we should add another category, system: true/false, for these.

src/project/Project.js
```diff
@@ -24,61 +24,52 @@
...
     }
-  }
+    else {
+      // This should be done by proper form validation.
+      console.error('Can not create new project - insufficient information: ', validation.errors)
+    }
+  };
+  onTitleChange = (e) => {
+    this.model.setOne('display.title', e.target.value);
+    this.model.setOne('display.slug', slugFromTitle(e.target.value));
```
ciyer: It looks like these two calls dispatch two actions. We may want to (later) make it possible to update both values in one call.

ableuler: Agree - actually it already is possible and I also agree to your proposed renaming of `setOne` to `set` and `set` to `setAll`. I'll address this in an upcoming PR.

PR #164, opened 2018-04-09
Extend project information

ableuler:


src/gitlab/client.js
```diff
@@ -35,15 +42,33 @@ export default class GitlabClient {
...
-    return fetch(this._baseUrl + `projects/${projectId}`, {
+    const projectPromise = fetch(this._baseUrl + `projects/${projectId}`, {
       method: 'GET',
       headers: headers
     })
       .then(response => response.json())
-      .then(d => carveProject(d))
+      .then(d => carveProject(d));
+
+    const treePromise = this.getRepositoryTree(projectId, undefined, true);
```
ciyer: Why not call ```this.getRepositoryTree(projectId, recursive=true)```

ableuler: You are right, named parameters are better here. Guess I didn't do it because I don't like the JS syntax for it. I changed it.

PR #165, opened 2018-04-09
Minor fixes

ableuler:


src/gitlab/client.js
```diff
@@ -25,25 +32,46 @@ export default class GitlabClient {
...
-    return fetch(this._baseUrl + `projects/${projectId}`, {
+    const projectPromise = fetch(this._baseUrl + `projects/${projectId}`, {
       method: 'GET',
       headers: headers
     })
       .then(response => response.json())
-      .then(d => carveProject(d))
+      .then(d => carveProject(d));
+
+    const treePromise = this.getRepositoryTree(projectId, undefined, true);
```
ciyer: This would be clearer as ```this.getRepositoryTree(projectId, recursive=true)```

ableuler: I agree that named parameters are better - addressed this in previous PR.

PR #190, opened 2018-04-20
Make project overview in case of a missing readme a bit friendlier.

ableuler:


src/gitlab/client.js
```diff
@@ -135,7 +135,17 @@ export default class GitlabClient {
 
   getProjectReadme(projectId) {
     return this.getRepositoryFile(projectId, 'README.md', 'master', 'raw')
-      .then(text => ({text}))
+      .then(text => {
+        // TODO: This is a temporary fix until better error handling is implemented.
+        if(text === '{"message":"404 Commit Not Found"}') {
+          return {
+            text: 'This repository seems to be empty. Why don\'t you start by committing a README.md file?'
```
rokroskar: doesn't have to be empty, just README isn't there...

ableuler: agree, this is really temporary for demo purposes.

PR #214, opened 2018-05-18
Switch to @nteract/notebook-preview to display notebooks.

### LINKED ISSUE #107: Issues in Jupyter Notebook Rendering
react-jupyter has moved to [@kyso/react-jupyter](https://www.npmjs.com/package/@kyso/react-jupyter) and  is now based on [@nteract/notebook-preview](https://www.npmjs.com/package/@nteract/notebook-preview). May as well just move to the nteract package directly, as this is a big project that is likely to continue in the future.

_From @ciyer on February 7, 2018 14:11_

The renderer causes warnings when running tests. I think switching to 

  https://github.com/MikeShi42/react-jupyter

Instead of the version that is in npm would fix this problem. May need to refer to this information to get this to work.

  https://stackoverflow.com/questions/17509669/how-to-install-an-npm-package-from-github-directly

I'm also not sure that the tests are actually checking anything.

_Copied from original issue: SwissDataScienceCenter/incubator-renga-ui#30_

ciyer:
Closes #107.


src/App.css
```diff
@@ -82,16 +82,6 @@ body {
   outline: none;
 }
 
-.deployButton {
```
ableuler: Are you sure that these stiles are obsolete?

ciyer:
The main motivation for moving to a new library for displaying Jupyter Notebooks is that the library that we were using had not been updated in 1y. The current version of that library is based on the nteract component, and so I decided to just switch to nteract instead. It is being maintained and developed.

Issue #196, opened 2018-05-05
logout does not log out


rokroskar:
It doesn't seem like the logout from the drop-down menu does anything anymore?

leafty:
Probable cause: no 'deep' logout, i.e. clearing of session from the third party login provider.

ableuler:
This should have been improved by #198, yet I'm leaving this issue open as some things (for example logging out with the actual provider of the login (see the comment of @leafty) remain to be done.

ableuler:
Will be resolved once the gateway is integrated into renku.

rokroskar:
closing for now, waiting on #228

Issue #127, opened 2018-03-01
Integrate gateway into renku


ableuler:
A proxy should be used to validate a request with keycloak and - if permitted - exchange the token before forwarding the request to gitlab.

For a more detailed description of this, please see the issues in https://github.com/SwissDataScienceCenter/incubator-proxy

rokroskar:
@ableuler and @3C111 can you please expand the description of this as the definition becomes more concrete? Perhaps the title should also be changed to something more along the lines of what has been discussed

PR #228, opened 2018-05-30
Gateway integration

### LINKED ISSUE #211: change "logging in with gitlab"
Could it just say "logging in" or maybe not even anything at all?

ableuler:
Merge after https://github.com/SwissDataScienceCenter/renku-gateway/pull/20
Merge together with https://github.com/SwissDataScienceCenter/renku/pull/205
closes #211, #196, #127


src/index.js
```diff
@@ -23,11 +23,11 @@ configPromise.then((res) => {
     const store = createStore(reducer);
 
     const client = new GitlabClient(
-      params.GITLAB_URL + '/api/v4/',
-      cookies.get('gitlab_token'),
+      params.GATEWAY_URL + '/v4/',
```
rokroskar: this is a gitlab-specific API version -- why is this appended directly to the `GATEWAY_URL`?

ableuler: I agree, the decision which gitlab api version to use should be delegated from the client to the gateway configuration in order to be more future-proof. I simply didn't bother changing this yet.

rokroskar: would be good to do https://github.com/SwissDataScienceCenter/renku-gateway/issues/9 asap then this is a non-issue

rokroskar: yes, that's what I meant. You may specify an api version - the default is `v4`.

ableuler: 👍

ableuler:
@rokroskar I changed the corresponding PR on the Renku repo.

PR #230, opened 2018-05-31
Fix url behind "launch notebook server" button

### LINKED ISSUE #226: "Launch Notebook Server" URL is broken
This might be by design? It doesn't seem to yield the correct URL, however:

https://testing.datascience.ch/jupyterhub/services/notebooks/undefined/221d52a93f083ba12cc2f2382eef67748f256304/undefined?branch=master

image

ableuler:
Closes #226

### DIFF
--- a/src/gitlab/client.js
+++ b/src/gitlab/client.js
@@ -224,11 +224,11 @@ class GitlabClient {
 
   // TODO: Once the gateway is up and running, the client should not need to be aware of the
   // TODO: JUPYTERHUB_URL anymore but simply query the notebook url from the gateway
-  async getNotebookServerUrl(projectId, projectPath, filePath, commitSha='latest', ref='master') {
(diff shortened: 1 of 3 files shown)


rokroskar:
Does it now also show up when there are already notebooks there?

ableuler:
No, it doesn't. I wasn't sure what exactly the expected behaviour would be (and where this button should show up), so for now I just fixed the thing that is obviously broken.

rokroskar:
Ok, we should add clicking this button to the integration test in renku.

ableuler:
@rokroskar Can you quickly add this button click or approve the PR and create a new issue for more thorough testing of the UI in the integration tests? I'm happy to have a look at splinter at some point, but I don't want a small an simple bugfix PR to be stalled because of this.

rokroskar:
we can make an issue for the integration test -- I was actually just about to try out @leafty's staging set up to deploy this PR...

PR #232, opened 2018-06-01
Deactivate SW

leafty:
The service worker gobles up too many paths.

As long as it does this, I would suggest to deactivate it like it is done here, or to configure it to only serve static assets.


leafty:
Sure, didn't know it was generated by react scripts

PR #235, opened 2018-06-04
Improve landing page

### LINKED ISSUE #221: Landing Page
In the interest of getting something better than the status quo in a short amount of time, the landing page should follow the following logic:

- If the user is not logged in, show a page about Renku in general.
- If the user is logged in, apply the following logic to decide what to show:
    - If the user has no starred projects, show information about how to star a project and how to create a project
    - If the user has some starred projects, show them

ciyer:
Fix #221.

### DIFF
--- a/src/gitlab/client.js
+++ b/src/gitlab/client.js
@@ -461,6 +461,7 @@ function carveProject(projectJson) {
   result['metadata']['core']['title'] = projectJson['name'];
   result['metadata']['core']['external_url'] = projectJson['web_url'];
   result['metadata']['core']['path_with_namespace'] = projectJson['path_with_namespace'];
+  result['metadata']['core']['owner'] = projectJson['owner'];
 
   result['metadata']['system']['tag_list'] = projectJson['tag_list'];
   result['metadata']['system']['star_count'] = projectJson['star_count'];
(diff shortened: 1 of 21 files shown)


src/utils/Errors.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+/**
+ *  renku-ui
+ *
+ *  Errors.js
+ *  Utility functions for handling errors
+ */
+
+function alertError(text) {
```
ableuler: Good idea to put this shell so we can later replace the alert with something else. It would make sense to use it everywhere where we currently just alert the user.

src/gitlab/client.js
```diff
@@ -461,6 +461,7 @@ function carveProject(projectJson) {
   result['metadata']['core']['title'] = projectJson['name'];
   result['metadata']['core']['external_url'] = projectJson['web_url'];
   result['metadata']['core']['path_with_namespace'] = projectJson['path_with_namespace'];
+  result['metadata']['core']['owner'] = projectJson['owner'];
```
ableuler: The owner property should also be added to the project schema.

src/utils/Errors.js
ciyer: Agreed. I will make this change as I touch code where I see alert.

src/gitlab/client.js
ciyer: I did this.

PR #252, opened 2018-06-20
Notebook image build status notification

ableuler:
Add a notification if notebook image build has failed or is still on-going.

addresses https://github.com/SwissDataScienceCenter/renku/issues/263

### DIFF
--- a/src/gitlab/client.js
+++ b/src/gitlab/client.js
@@ -471,6 +471,15 @@ class GitlabClient {
       })
   }
 
+  getJobs(projectId) {
+    let headers = this.getBasicHeaders();
+    return renkuFetch(`${this._baseUrl}/projects/${projectId}/jobs`, {
+      method: 'GET',
+      headers,
+      queryParams: {per_page: 100}
+    }, 'json', false)
+  }
+
   mergeMergeRequest(projectId, mrIid) {
     let headers = this.getBasicHeaders();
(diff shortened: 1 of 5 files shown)


rokroskar:
Is the display of the banner on all pages intentional? I assumed it would only show up on the notebooks page.

ableuler:
Yes it is. I know it's not ideal but as there are multiple ways to launch a notebook (ku tab, notebook tab) i chose the more radical approach for now.

rokroskar:
Instead of looping through all the jobs, you might find it better to find the pipeline that corresponds to the commit sha and then look at the jobs in the pipeline. If there are a lot of jobs this could save some resources. I also wonder what happens when you have e.g. two jobs for the same sha and one fails but the other one passes.

ableuler:
Better? yes, of course! easier to quickly implement? nope! In fact I'm not even paging through all results which is a bit worrying because the jobs endpoint (as opposed to the pipeline endpoint) doesn't allow you to enforce inverse chronological ordering of the results (which mostly seems to be there anyway).
>I also wonder what happens when you have e.g. two jobs for the same sha and one fails but the other one passes.

But those wouldn't have the same name, right?

rokroskar:
Yes, the two jobs would have the same name. One might fail because e.g. the registry is down so the push fails. When it's restarted, the new job will have a different ID but be otherwise identical.

ableuler:
I sort them chronologically in the ui and pick the latest with right name.

PR #266, opened 2018-07-12
Post request for nb server

rokroskar:
Merge at the same time/after https://github.com/SwissDataScienceCenter/renku-notebooks/pull/25

addresses https://github.com/SwissDataScienceCenter/renku-notebooks/issues/24

### DIFF
--- a/src/file/File.container.js
+++ b/src/file/File.container.js
@@ -51,7 +51,8 @@ class JupyterNotebookContainer extends Component {
   render() {
     return <JupyterNotebookPresent
       notebook={this.props.notebook}
-      deploymentUrl={this.state.deploymentUrl}/>
+      deploymentUrl={this.state.deploymentUrl}
+      gitlabToken={this.props.client._token}/>
   }
 }
 
--- a/src/file/File.container.js
+++ b/src/file/File.container.js
@@ -51,8 +51,7 @@ class JupyterNotebookContainer extends Component {
   render() {
     return <JupyterNotebookPresent
       notebook={this.props.notebook}
-      deploymentUrl={this.state.deploymentUrl}
-      gitlabToken={this.props.client._token}/>
+      deploymentUrl={this.state.deploymentUrl}/>
   }
 }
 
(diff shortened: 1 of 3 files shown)


leafty:
I get:
```
Mixed Content: The page at 'https://staging.testing.datascience.ch/projects/1/files/notebooks' was loaded over HTTPS, but requested an insecure resource 'http://staging.testing.datascience.ch/jupyterhub/hub/api/oauth2/authorize?client_id=service-notebooks&redirect_uri=%2Fjupyterhub%2Fservices%2Fnotebooks%2Foauth_callback&response_type=code&state=eyJ1dWlkIjogIjM4NzRhZWJiZDE5MDRiYzNiNzRhOGQ5OWIzMzAzZDc3IiwgIm5leHRfdXJsIjogIi9qdXB5dGVyaHViL3NlcnZpY2VzL25vdGVib29rcy9kZW1vL3Rlc3QvZTBlYjJiYmJiNGE5YzNlNmQ2ODRjMTQ3NTM5YzkxMTNmNDE4ODM0MCJ9'. This request has been blocked; the content must be served over HTTPS.
```

rokroskar:
@ableuler clicking the `Launch Notebook` button tries to make a pop-up using this PR -- could this be fixed to just open a new tab? Also, we shouldn't wait for the `POST` to return before calling the second `GET`.

ableuler:
@rokroskar From stackoverflow: 
"The general rule is that popup blockers will engage if window.open or similar is invoked from javascript that is not invoked by direct user action. That is, you can call window.open in response to a button click without getting hit by the popup blocker, but if you put the same code in a timer event it will be blocked."
This also includes callbacks after `fetch`. This means the new tab has to be opened right on button click.

leafty:
@ableuler @rokroskar Would it work then to send the post and to open the new tab at the same time? Would could fix `renku-notebooks` later if there is a chance of a 404.

leafty:
Note 1: Will probably fail the launch notebook button test.
Note 2: After merge, start working on a fix for the popup.

leafty:
@ableuler would be great if we can push this quick

ableuler:
> Note 1: Will probably fail the launch notebook button test.
As far as I know it is still failing if a user clicks the launch notebook button, at least the first time a button is clicked (and therefore a new notebook-server started). Or did I miss somethin?

leafty:
it's about the test running in travis / helm test

PR #282, opened 2018-08-09
Various bugfixes and improvements

ableuler:


src/gitlab/initRenkuProject.js
```diff
@@ -16,36 +16,70 @@
...
+      'branch': 'master',
+      'commit_message': 'init renku repository',
+      'actions': resolvedActions
+    }
+  });
+}
+
+
+function evaluateTemplate(content, projectName) {
+  console.log(content, projectName);
```
ciyer: Remove debug logging.

src/gitlab/initRenkuProject.js
```diff
@@ -16,36 +16,70 @@
...
+  const now = new Date();
+  const templatedVariables = {
+    'name': projectName,
+    'date-updated': now.toISOString(),
+    'date-created': now.toISOString(),
+  };
+
+  const newContent = content.replace(/{{\s?([^\s]*)\s?}}/g, (match, group) => {
+    console.log(match);
```
ciyer: Remove debug logging.

src/gitlab/initRenkuProject.js
```diff
@@ -16,36 +16,70 @@
  * limitations under the License.
  */
 
-const payload = {
-  'branch': 'master',
-  'commit_message': 'init renku repository',
-  'actions': [
+function getPayload(projectName){
+
+  const TEMPLATE_URL = 'https://raw.githubusercontent.com/SwissDataScienceCenter/renku-project-template/master/latest/';
```
ciyer: This there a reason this goes directly to GitHub and not through the gateway to get the initial project contents?

ableuler: I agree that the entire logic of project creation could (and should) be moved to the gateway. In this case, the UI would simply send a project POST to the gateway which would translate it into a project POST followed by the POST of the initial commit. I think a POST to `/api/v1/projects` should do exactly that (see https://github.com/SwissDataScienceCenter/renku-gateway/issues/29 for this).

On the other hand, I don't see any added value in simply funnelling the current calls to GitHub through the gateway (just like we're loading other things like gravatars etc directly from the web too).

src/gitlab/initRenkuProject.js
ableuler: Removed.

src/gitlab/initRenkuProject.js
ableuler: Removed - thanks for noticing!

src/gitlab/initRenkuProject.js
ciyer: Gravitars are public resources. The new project template can be dependent on many things (version, environment, etc.), and it would be nice to have a point of control to be able to inject logic to determine where the templates should come from. But since this is a temporary solution, it's ok.

PR #298, opened 2018-08-22
Make renku info part of landing page a templateable markdown file.

### LINKED ISSUE #294: Make welcome page templateable
We should make it easy to have custom text on the welcome page and be able to change this via the chart values. 

Related to #293

ableuler:
Closes #294.


ableuler:
This is currently deployed on https://staging.testing.datascience.ch, see https://gitlab.com/renku/stage-renku-deploy/blob/master/values.yaml for how the templating can be done in deployments.

PR #301, opened 2018-08-23
Improve token handling

### LINKED ISSUE #295: fail gracefully if an expired token is presented
at the moment, if an expired token is present, the landing page shows the popup telling the user they don't have the necessary permissions to view the page. Instead, it should fail gracefully and remove the expired token, leaving the user with the standard landing page.

ableuler:
- expired access_token -> refresh token and redo request
- expired refresh_token -> send browser to login route (keycloak eventually)
- don't send cookies in regular fetch requests
- determine logged-in status based on cookie presence

Merge after https://github.com/SwissDataScienceCenter/renku-gateway/pull/46
Closes #295

### DIFF
--- a/src/gitlab/client.js
+++ b/src/gitlab/client.js
@@ -27,27 +27,65 @@ class GitlabClient {
   // ku    -->  issue
 
 
-  constructor(baseUrl, token, tokenType, jupyterhub_url) {
+  constructor(baseUrl, cookies, jupyterhub_url) {
     this._baseUrl = baseUrl;
-    this._token = token;
-    this._tokenType = tokenType;
+    // TODO: Which GitLab API version to use should actually be a
+    //       a concern of the Gateway.
+    this._apiUrl = baseUrl + '/v4';
+    this._cookies = cookies;
+    this._accessToken = cookies.get('access_token');
+    this._refreshToken = cookies.get('refresh_token');
     this._jupyterhub_url = jupyterhub_url;
+    this._doLogin = () => {
+      this._cookies.remove('access_token');
+      this._cookies.remove('refresh_token');
+      this._cookies.remove('id_token');
+      window.location = `${this._baseUrl}/auth/login?redirect_url=${encodeURIComponent(window.location.href)}`;
+    };
   }
 
-  getBasicHeaders() {
+  _clientFetch(url, options, returnType='json', alertOnErr=true) {
+    return renkuFetch(url, options, returnType, alertOnErr)
+      .catch((error) => {
+        if (error.case === API_ERRORS.tokenExpired) {
+          return this._getRefreshedTokens()
+            .then(() => {
+              // Use the refreshed token and try again
+              options.headers.set('Authorization', `Bearer ${this._accessToken}`);
+              return renkuFetch(url, options, returnType, alertOnErr)
+            })
+        }
+      })
+  }
+
+  _getRefreshedTokens() {
+    let headers = this._getBasicHeaders();
+    // Overwrite access_token with refresh_token
+    headers.set('Authorization', `Bearer ${this._refreshToken}`);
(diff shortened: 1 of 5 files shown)


src/gitlab/client.js
```diff
@@ -27,37 +27,72 @@ class GitlabClient {
...
+
+        this._accessToken = responseData.access_token;
+        this._refreshToken = responseData.refresh_token;
+        this._cookies.set('access_token', responseData.access_token);
+        this._cookies.set('refresh_token', responseData.refresh_token);
+      })
   }
 
-  getBasicHeaders() {
+  _getBasicHeaders() {
```
leafty: python conventions lurking their way in?

src/gitlab/client.js
```diff
@@ -27,37 +27,72 @@ class GitlabClient {
...
+            .then(() => {
+              // Use the refreshed token and try again
+              options.headers.set('Authorization', `Bearer ${this._accessToken}`);
+              return renkuFetch(url, options, returnType, alertOnErr)
+            })
+        }
+      })
+  }
+
+  _getRefreshedTokens() {
```
leafty: I guess the next improvement is to make so that there is only one of these in flight at a time

ableuler: I don't think that's necessary. Under normal circumstances this would prevent only a few extra requests every 30 minutes and not have any functional advantages.

src/gitlab/client.js
ableuler: Yeah, there seems to be some debate weather this convention should be adopted for ES6 classes... I kind of like the reasoning behind it, but I don't have a strong opinion about it.

src/gitlab/client.js
leafty: What happens if we turn this on? :stuck_out_tongue: 
screenshot from 2018-08-24 07-59-21

src/gitlab/client.js
leafty: let's hope they don't use Go-lang conventions

src/gitlab/client.js
```diff
@@ -27,37 +27,72 @@ class GitlabClient {
...
     this._jupyterhub_url = jupyterhub_url;
+    this._doLogin = () => {
+      this._cookies.remove('access_token');
+      this._cookies.remove('refresh_token');
+      this._cookies.remove('id_token');
+      window.location = `${this._baseUrl}/auth/login?redirect_url=${encodeURIComponent(window.location.href)}`;
+    };
+  }
+
+  _clientFetch(url, options, returnType='json', alertOnErr=true) {
```
leafty: Is this now the method to use everywhere in the UI?

src/gitlab/client.js
```diff
@@ -27,37 +27,72 @@ class GitlabClient {
   // ku    -->  issue
 
 
-  constructor(baseUrl, token, tokenType, jupyterhub_url) {
+  constructor(baseUrl, cookies, jupyterhub_url) {
     this._baseUrl = baseUrl;
```
ciyer: I see you needed to change the structure baseUrl to make it possible to access the "auth" endpoint. But I do not like that it caused "/v4/" to have to be added everywhere. Could you add a new property, _apiUrl that would be ```${baseUrl}/v4``` to avoid the duplication?

ableuler: @ciyer good point, followed your suggestion. As stated in the comment this should actually soon be moved to the gateway, while we specify which version of the gateway to use (which will also apply for the auth endpoints).

src/gitlab/client.js
leafty: just wanted to make sure that's what you want the code to look like :wink:

ableuler: Sorry, misunderstood your question above. This method is used in all api calls, but never called directly from outside the GitlabClient class. So in the reactJS application you're still going to see things like `client.getProjectKus(projectId)`.

PR #306, opened 2018-08-24
Always show welcome page

ableuler:
Make sure the content of the templateable welcome page is accessible to all users, also the users who are logged in and have already starred a project.

### DIFF
--- a/src/landing/Landing.present.js
+++ b/src/landing/Landing.present.js
@@ -152,13 +152,13 @@ class Starred extends Component {
   }
 }
 
-class AnonymousWelcome extends Component {
+class Welcome extends Component {
   render() {
     return (<Row>
       <Col md={8} lg={6} xl={4}>
         <RenkuIntroText welcomePage={this.props.welcomePage}/>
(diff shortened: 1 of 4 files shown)


ableuler:
@ciyer @rokroskar up on staging now!

rokroskar:
tested, looks fine to me

src/landing/Landing.present.js
```diff
@@ -203,16 +203,17 @@ class Home extends Component {
     let selected = this.props.ui.selected;
     let nav = null;
     const urlMap = this.props.urlMap;
-    let welcome = <AnonymousWelcome urlMap={urlMap} welcomePage={this.props.welcomePage} />;
+    let welcome;
```
ciyer: I would prefer the following:
```
const anonymous = this.props.user != null && this.props.user.id != null;
const welcome = <Welcome welcomePage={this.props.welcomePage} anonymous={anonymous}/>;
if (anonymous) { 
  nav = [...]
} else {
  [...]
}
```

ableuler:
@ciyer Good point. Adapted and thereby also used the `loggedIn` property already defined higher up in the tree.

PR #309, opened 2018-09-01
Enable quick navigation through nav bar

ableuler:
Enable project search through a reactive search bar displaying auto-suggestions. This makes it possible to jump to a project quickly.

---
Closes #264.

PR #312, opened 2018-09-05
Update project creation

ableuler:
Update project creation such that:
 - New files added to the project template are autmatically added.
 - The correct version of the project template is used. The version is read from the the global chart values or from the version of the renku-ui Chart.

Builds on top of the changes in #309.


rokroskar:
If this envisioned for 0.2.0 please add it to that project -- if not, add it to the 0.2.1 milestone.

PR #320, opened 2018-09-18
Select namespace by project creation

### LINKED ISSUE #242: Create a new project in a group namespace
New projects are created in the user namespace. It should be possible to create them in a group namespace as well.

ciyer:
Fix #242.


ciyer:
There is a problem when creating a project in a subgroup (e.g., a group that is within another group). This seems to be caused by the gateway decoding url-encoded data in the url. This may have been fixed by a more recent gateway commit.

PR #335, opened 2018-10-19
Set server options

ableuler:
Please note that the majority of the code changes proposed in this PR will be revisited once the connection between the ui and jupyterhub us routed through the gateway.

### DIFF
--- a/src/file/File.present.js
+++ b/src/file/File.present.js
@@ -22,7 +22,8 @@ import NotebookPreview from '@nteract/notebook-preview';
 // Do not import the style because this does not work after webpack bundles things for production mode.
 // Instead define the style below
 //import './notebook.css'
-import { Button, Row, Col } from 'reactstrap';
+import { Button, Row, Col, Tooltip } from 'reactstrap';
+import Cookies from 'universal-cookie';
 import '../../node_modules/highlight.js/styles/atom-one-light.css'
 
 class StyledNotebook extends React.Component {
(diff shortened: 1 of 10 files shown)


src/file/File.present.js
```diff
@@ -60,29 +61,99 @@ class StyledNotebook extends React.Component {
...
+    };
+    this.cookies = new Cookies();
+    this.state.showTooltip = false;
+  }
+
+  componentDidMount() {
+    this.componentDidUpdate()
+  }
+
+  componentDidUpdate() {
```
ciyer: This code is almost the same as  src/notebooks/Notebooks.container.js::LaunchNotebookServer::setServerStatus. Is there a reason to not extract the common code into a method?

src/notebooks/Notebooks.container.js
```diff
@@ -31,5 +33,156 @@ class NotebookAdmin extends Component {
...
+      this.serverOptionsSet = false;
+    }
+    this.previousNotebookServerUrl = this.props.core.notebookServerUrl;
+    this.setServerStatus()
+    this.setServerOptions()
+  }
+
+
+  setServerStatus() {
+    if (this.serverStatusSet) return;
```
ciyer: See comment above in File.present.js.

src/file/File.present.js
ableuler: I completely agree, in fact there are many other aspects which are non-satisfactory about this PR, like the fact that a presentational container issues API requests. Most of this will be refactored once the gateway handles the JH token for the user.

ableuler: Created a new issue to keep track of the requested changes:https://github.com/SwissDataScienceCenter/renku-ui/issues/336

PR #349, opened 2018-11-02
Changes to graph visualization to support new dot file.

ciyer:
UI changes to support https://github.com/SwissDataScienceCenter/renku-python/pull/340 / https://github.com/SwissDataScienceCenter/renku-python/issues/339.

Should be merged after the above.

### DIFF
--- a/src/file/Lineage.present.js
+++ b/src/file/Lineage.present.js
@@ -53,7 +53,9 @@ class FileLineageGraph extends Component {
 
     const graph = this.graph();
     const {nodeIds, edges, centralNode} = this.nodesAndEdges();
-    if (nodeIds.length < 2) {
+    if (centralNode == null) {
+      subGraph.setNode("0", {id: "0", label: "No lineage information." });
+    } else if (nodeIds.length < 2) {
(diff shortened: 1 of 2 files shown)


rokroskar:
This seems to work -- see `zurich-bikes-analysis` project on staging: 

image

rokroskar:
This seems to work for the moment though I'm not sure about the display of the commit hash -- in any case, we should move away from providing the dot graph and use JSON instead asap so you can do whatever you want with that data.

ciyer:
The commit hash is interesting because it tells you which things happened together, but it is certainly not easy to read. I will think about a better way to present this.

Why were there two renku updates done here? I would have expected the first update to have updated everything downstream as well.

src/file/Lineage.present.js
```diff
@@ -28,9 +28,11 @@ import { ExternalLink } from '../utils/UIComponents'
 import './Lineage.css';
 
 
-function nodeIdToPath(nodeId) { return nodeId.split(',')[1].slice(2, -2) }
+function nodeIdToPath(nodeId) { return nodeId.split(':')[1].slice(1,) } // Strip the initial / from the path
```
rokroskar: ah, we should've changed this in the dot export instead...

PR #352, opened 2018-11-26
Fix/improve quick-nav bar:

ableuler:
- Fix processing of the server return promise after recent changes to the API client library.
- Query search API only after no letter has been typed for one second to reduce numbe of API calls.


ableuler:
Starting search not already after first letter input. The rest of the changes I'd prefer to tackle in a separate issue so that this PR can be merged now. Currently, the entire functionality is broken.

PR #366, opened 2019-01-08
QuickNav jump improvements and support for searching for projects

### LINKED ISSUE #353: Improve quick-nav bar
There is no point in searching all of renku with the quick-nav bar. Instead, it should

- [x] Provide quick nav with suggestions for the user's own and starred repos
- [x] Offer searching all of renku, but take the user to another page for this
- [x] Limit number of suggestions being shown

The user's own and starred repos should be cached in the state object for quick access.

ciyer:
* The QuickNav jump function is restricted to member and starred projects. These are cached, so display of results is immediate.
* QuickNav also allows searching for projects.

Fix #353.

### DIFF
--- a/src/api-client/test-client.js
+++ b/src/api-client/test-client.js
@@ -25,6 +25,7 @@ const methods = {
       data: samples.projects,
       pagination: {
         currentPage: 1,
+        totalItems: 1,
       }
     }
   },
--- a/src/landing/Landing.js
+++ b/src/landing/Landing.js
@@ -72,7 +72,7 @@ class Home extends Component {
   mapDispatchToProps(dispatch, ownProps) {
     return {
       onStarred: (e) => { dispatch(State.Home.Ui.selectStarred()) },
-      onYourActivity: (e) => { dispatch(State.Home.Ui.selectYourActivity()) },
+      onMember: (e) => { dispatch(State.Home.Ui.selectMember()) },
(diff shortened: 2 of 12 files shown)


ciyer:
I don't really like the way this looks either, but I think it is a big usability improvement over the status quo. I think the better solution is to have 3 tabs on the project page: Yours, Starred, Explore (Yours would be projects where you are a dev or higher).

I think the New Project button is important for usability, esp for new users, and the tabs would resolve the aesthetic issues with the current solution.

ableuler:
> clicking on `Projects` on the top navbar twice gives me the full project list (i.e. it seems to be the equivalent of clicking the search button without a search text).

Noticed the same, not sure if this is a feature or a bug...

I agree with @ciyer that the caching of "your projects" and the ability to quickly switch between them adds some value. I would suggest to merge this PR and then move the search functionality into a separate tab later. Actually, the landing page is very close to what I would imagine the project page to look like. screenshot 2019-02-06 at 16 37 15

While moving the project search into a separate tab, It would also be nice to make it a bit more powerful, for example enable searching for all projects of a given user (although there doesn't seem to be a straighforward api for this: https://gitlab.com/gitlab-org/gitlab-ce/issues/23185).

rokroskar:
Ah yes agreed - apologies, I thought I approved this earlier. Can we make an issue for these improvements?

ableuler:
> Ah yes agreed - apologies, I thought I approved this earlier. Can we make an issue for these improvements?
I think #377 is about this.

Issue #376, opened 2019-02-06
script to run telepresence not working in linux


lorenzo-cavazzi:
The `run-telepresence.sh` script file doesn't seem to work properly on Ubuntu 18.04.

- The `base64` command seems to wrap line after a fixed number of chars
- telepresence can't run properly with the `inject-tcp` proxying method

PR #378, opened 2019-02-06
run-telepresence.sh fixes for linux enviroment

lorenzo-cavazzi:


lorenzo-cavazzi:
Reference for the 2 fixes:

- base64 `-w` option: https://linux.die.net/man/1/base64
- telepresence proxying methods: https://www.telepresence.io/reference/methods

run-telepresence.sh
```diff
@@ -18,27 +18,40 @@
 
 set -e
 
-MINIKUBE_IP=`minikube ip`
+
 CURRENT_CONTEXT=`kubectl config current-context`
 WELCOME_PAGE=`echo "## Welcome to Renku through telepresence
-Some deployment-specific information will be read from the your values.yaml file and be displayed as markdown file." | base64`
+Some deployment-specific information will be read from the your values.yaml file and be displayed as markdown file." | base64 -w 0`
```
ciyer: This is an annoying difference between macOS (BSD?) and linux. `base64 -w0` returns an error on macOS. You'll need to test for linux or macOS and call the appropriate command, either `base64 -w 0` or `base64` respectively.

run-telepresence.sh
```diff
@@ -50,4 +63,9 @@ EOF
 # echo "BROWSER=none npm start"
 # echo "================================================================================================================="
 
-BROWSER=none telepresence --swap-deployment renku-ui --namespace renku --method inject-tcp --expose 3000:80 --run npm start
+# The `inject-tcp` proxying switch helps when running multiple instances of telepresence but creates problems when
+# suid bins need to run. Please switch the following two lines when trying to run multiple telepresence.
+# Reference: https://www.telepresence.io/reference/methods
+
+# BROWSER=none telepresence --swap-deployment ${SERVICE_NAME} --namespace ${K8S_NAMESPACE} --method inject-tcp --expose 3000:80 --run npm start
+BROWSER=none telepresence --swap-deployment ${SERVICE_NAME} --namespace ${K8S_NAMESPACE} --expose 3000:80 --run npm start
```
ciyer: I think there should be a newline at the end of this line.

ableuler: It's good practice to configure your editor to always add a newline at the end of the file.

run-telepresence.sh
lorenzo-cavazzi: done, could you check if it looks better now and test if it runs on a mac?

run-telepresence.sh
lorenzo-cavazzi: done, and thanks for the suggestion

run-telepresence.sh
```diff
@@ -63,4 +73,9 @@ EOF
 # echo "BROWSER=none npm start"
 # echo "================================================================================================================="
 
-BROWSER=none telepresence --swap-deployment ${SERVICE_NAME} --namespace ${DEV_NAMESPACE} --method inject-tcp --expose 3000:80 --run npm start
+# The `inject-tcp` proxying switch helps when running multiple instances of telepresence but creates problems when
+# suid bins need to run. Please switch the following two lines when trying to run multiple telepresence.
+# Reference: https://www.telepresence.io/reference/methods
+
+# BROWSER=none telepresence --swap-deployment ${SERVICE_NAME} --namespace ${K8S_NAMESPACE} --method inject-tcp --expose 3000:80 --run npm start
```
ciyer: This comment can be deleted.

run-telepresence.sh
```diff
@@ -63,4 +73,9 @@ EOF
...
+BROWSER=none telepresence --swap-deployment ${SERVICE_NAME} --namespace ${K8S_NAMESPACE} --expose 3000:80 --run npm start
```
ciyer: Looks like this line got mixed up in the rebase. It should be 

```BROWSER=none telepresence --swap-deployment ${SERVICE_NAME} --namespace ${DEV_NAMESPACE} --method inject-tcp --expose 3000:80 --run npm start```

as in master.

run-telepresence.sh
ciyer: This works, thanks.

run-telepresence.sh
ciyer: I didn't realize that turning off inject-tcp was intentional. Just use ${DEV_NAMESPACE} for the namespace argument.

lorenzo-cavazzi: fixed the wrong variable name.
I leaved the comment because removing the `inject-tcp` method has some drawbacks, someone probably wants to use that command instead of the new one.
I want to go back to this in the future to see if I can get it working in Ubuntu

run-telepresence.sh
lorenzo-cavazzi: I leaved the comment because removing the `inject-tcp` method has some drawbacks, someone probably wants to use that command instead of the new one.
I want to go back to this in the future to see if I can get it working in Ubuntu

ciyer: Changed the logic to preserve the old behavior on macOS (inject-tcp), and only skip this on linux.

lorenzo-cavazzi:
closes #376

Issue #377, opened 2019-02-06
Projects search page should be finer-grained


ciyer:
The project search page currently just offers search. Instead, it should offer 3 options in tabs: Yours, Starred, Search.

The behavior of Yours and Starred should be similar to the landing page and show projects where the user is a a member or has starred, respectively. Search should be where the current functionality of the projects page goes.

# Details

Each tab can, but does not need to have its own URL. If it is easier to use the browser to track the state, do that, if it is easier to track the state of which tab is displayed in redux, that is also fine. 

In the first case, then the search URL in quick-nav bar will need to be updated. In the latter case, If a search query is provided, then the active tab should be the search tab, otherwise the Yours tab.

Current behavior is that clicking once on the projects top-nav shows an empty page, waiting for input, clicking twice runs the search with empty input. It should either always run the search or never.

# Mockup

<img width="523" alt="image" src="https://user-images.githubusercontent.com/1196411/52339402-4c9fe700-2a0d-11e9-9246-152b97d69a0a.png">

rokroskar:
hasn't this been done? can the issue be closed?

ciyer:
Yes, I had moved it to the "Done" column in our project board and assumed that that closes it, but apparently not.

PR #379, opened 2019-02-11
Tabs added to projects page (your projects, starred projects, search)

vfried:
#377

### DIFF
--- a/src/project/list/ProjectList.present.js
+++ b/src/project/list/ProjectList.present.js
@@ -69,25 +69,29 @@ class ProjectNavTabs extends Component {
     return (
       <Row key="nav">
         <Col md={12}>
-            {
-                <Nav pills className={'nav-pills-underline'}>
-                   <NavItem>
-                     <RenkuNavLink to={this.props.urlMap.projectsUrl} alternate={this.props.urlMap.yourProjects}  title="Your Projects" />
-                   </NavItem>
-                   <NavItem>
-                     <RenkuNavLink exact={false} to={this.props.urlMap.starred}  title="Starred Projects" />
-                   </NavItem>
-                   <NavItem>
-                     <RenkuNavLink exact={false} to={this.props.urlMap.projectsSearchUrl}  title="Search" />
-                   </NavItem>
-                </Nav>
-               :
-               <span></span>
-            }
+          {
             (this.props.loggedIn) ?
+              [
+                <div key="top-space">&nbsp;</div>,
+                <Nav key="nav" pills className={'nav-pills-underline'}>
+                  <NavItem>
+                    <RenkuNavLink to={this.props.urlMap.projectsUrl}
+                      alternate={this.props.urlMap.yourProjects}  title="Your Projects" />
+                  </NavItem>
+                  <NavItem>
+                    <RenkuNavLink exact={false} to={this.props.urlMap.starred}  title="Starred Projects" />
+                  </NavItem>
+                  <NavItem>
+                    <RenkuNavLink exact={false} to={this.props.urlMap.projectsSearchUrl}  title="Search" />
+                  </NavItem>
+                </Nav>,
+                <div key="bottom-space">&nbsp;</div>]
+              :
+              <span></span>
+          }
         </Col>
       </Row>
-      )
+    )
   }
 }
 
--- a/src/project/list/ProjectList.present.js
+++ b/src/project/list/ProjectList.present.js
@@ -123,16 +123,16 @@ class DisplayProjects extends Component{
       <Col md={{size: 2,  offset: 3}}><Loader /></Col> :
       <Col md={8}>{rows}</Col>;
 
-      if (rows.length > 0 ){
-        return (<Row key="projects">{projectsCol}</Row>);
-      }
-      else {
-        if(!loading){
-          return <DisplayEmptyProjects
-            projectsSearchUrl={projectsSearchUrl}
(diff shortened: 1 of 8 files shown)


src/project/list/ProjectList.present.js
```diff
@@ -63,17 +64,169 @@ class ProjectSearchForm extends Component {
...
+        visibleTab = <div>  <Row key="form">
+                <Col md={8}>
+                  <ProjectSearchForm searchQuery={this.props.searchQuery} handlers={this.props.handlers} />
+                </Col>
+              </Row>
+              <Row key="spacer2"><Col md={8}>&nbsp;</Col></Row>
+              <Row key="projects">{projectsCol}</Row>,
+              <Pagination key="pagination" {...this.props} /></div>
+      }
+      else if (selected ==='starred')
```
ableuler: I think it would be good practice to define constants for the tab names.

src/project/list/ProjectList.present.js
```diff
@@ -63,17 +64,169 @@ class ProjectSearchForm extends Component {
...
+      <NavItem>
+          &nbsp;&nbsp;
+          <Link className="btn btn-primary" role="button" to={this.props.urlMap.projectNewUrl}>New Project</Link>
+          <span></span>
+      </NavItem>
+    </Nav>
+  }
+}
+
+class StarredEmptyProjects extends Component {
```
ableuler: While it needs to be discussed what the distinction between the landing page and `projects` exactly should be, you can try to _reuse_ the components from the langing page rather than copy-pasting them

src/project/list/ProjectList.present.js
```diff
@@ -63,17 +64,169 @@ class ProjectSearchForm extends Component {
...
+        <NavLink href="#" active={selected === 'starred'}
+          onClick={this.props.handlers.onStarred}>Starred Projects</NavLink>
+      </NavItem>
+      <NavItem>
+        <NavLink href="#" active={selected === 'explore'}
+          onClick={this.props.handlers.onExplore}>Explore</NavLink>
+      </NavItem>
+      <NavItem>
+          &nbsp;&nbsp;
+          <Link className="btn btn-primary" role="button" to={this.props.urlMap.projectNewUrl}>New Project</Link>
```
ableuler: For me the `New Project` button could be taken away from this page as the same option is available through the "+" in the top bar.

src/project/list/ProjectList.present.js
```diff
@@ -63,17 +64,169 @@ class ProjectSearchForm extends Component {
...
+
+      if(selected===undefined  || selected === 'explore'){
+        selected = 'explore'
+        visibleTab = <div>  <Row key="form">
+                <Col md={8}>
+                  <ProjectSearchForm searchQuery={this.props.searchQuery} handlers={this.props.handlers} />
+                </Col>
+              </Row>
+              <Row key="spacer2"><Col md={8}>&nbsp;</Col></Row>
+              <Row key="projects">{projectsCol}</Row>,
```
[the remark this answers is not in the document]
vfried: Hi, I checked all your comments and I will work on them to solve them. There is also some commented code i left behind without realising.

src/project/list/ProjectList.present.js
ciyer: New and casual users do not know to look for the little "+" in the top bar. I like having multiple ways to create a new project.

vfried: I changed a bit the layout of this, re-added the big title (Projects) and the button like in the previous version, and under that there are the three tabs.

screenshot 2019-02-13 at 14 53 14

ableuler:
The routing seems to work fine 👍. I have one general question: Are you reusing the same component now for displaying the list of starred projects as in the search? I'm asking because there seems to be no pagination for starred project. I'm not saying that this is absolutely necessary right now, but if you're reusing components, pagination should be there automatically.
Also, please run `npm run test` on the branch which has your code changes and try to fix the failing test. This should make the red alerts below disappear :)

vfried:
> The routing seems to work fine 👍. I have one general question: Are you reusing the same component now for displaying the list of starred projects as in the search? I'm asking because there seems to be no pagination for starred project. I'm not saying that this is absolutely necessary right now, but if you're reusing components, pagination should be there automatically.
> Also, please run `npm run test` on the branch which has your code changes and try to fix the failing test. This should make the red alerts below disappear :)

I only managed do a generic component for the "your projects" and the "starred projects" but this doesn't apply to the search since it's slightly different, so for now only the search has pagination functionality.

I decided not to waste to much time in integrating the pagination to the your projects and starred since i think we should do a generic "Projects Container" where we can just "throw" any list of projects and display it, we could use this for the landing page also. If we decide to do that i could do it but it would take me a bit more of time. What do you think about this?
 
Regarding the tests, I will run npm and fix the failing ones.

vfried:
The test was fixed, if you just want to see how it looks it's deployed here: https://i377-projecttabs.dev.renku.ch/

src/project/list/ProjectList.present.js
vfried: The code changed a lot so this comment doesn't apply anymore.

src/project/list/ProjectList.present.js
vfried: There is a proposal for reusing the components under, i think we should decide if we want to do a generic component for projects outside of both the landing page and the projects list. If we don't i think at least the component should be in the projects list and not in the landing page.

Issue #351, opened 2018-11-22
Integrate notebooks admin page into UI


rokroskar:
With #331 merged we can make requests to the notebook service directly from the UI. We should now be able to greatly simplify the notebook server interactions. 

First steps:
- [x] Create a new "Notebook Servers" tab in the project view
- [x] Display the information currently presented in a separate browser tab served from `https://<renku-domain>/jupyterhub/services/notebooks/ui/` in the above tab
- [x] Add the "connect" and "stop" options per notebook server

Some ideas for further improvements: 
* the notebook server button itself could reflect the status of the running server for the selected project
* the project view page could include an icon to connect to a running server
* there could be a global page per user on the platform level to display running servers
* the notebook servers tab could display further relevant information like e.g. running commit, branch, time inactive etc.

lorenzo-cavazzi:
Would embedding the `https://<renku-domain>/jupyterhub/services/notebooks/ui/` page in the tab through an `<iframe>` be an acceptable solution? I believe this would easily satisfy the requirements in the first steps, but then we won't be able to customize the page further.

Anyway this would be the result

![image](https://user-images.githubusercontent.com/43481553/52716848-fba96900-2f9f-11e9-8b89-74b2ae116eff.png)

ciyer:
Cute idea, but this is not the solution I had in mind. :)

Issue #340, opened 2018-10-24
improve notebook launch


rokroskar:
- [x] Remove the "Connect to Jupyter" button (and the popup on "+" in the top navbar) and use the one in the "Notebook Servers" 
- [x] Add the "Launch Jupyter" button in the "Notebook Servers" tab
- ~~[ ] Allow using other branches for notebook launch: Add the option to specify a branch for launching the notebook server -- should be done by attaching the `?branch=<name>` query parameter to the notebook launch `POST` request URL.~~
- ~~[ ] Ensure that the latest commit is used when a notebook server is started (currently, we use the latest commit *at the time the page was loaded*. This causes problems if a user pushes between loading the view and launching a notebook server)~~

Next step will include fixing the list of Servers displayed in the "Notebook Servers" tab and fixing the "Notebooks" page from top navbar

lorenzo-cavazzi:
waiting for #406 for the last 2 points

lorenzo-cavazzi:
last 2 points --> #416

PR #380, opened 2019-02-18
Added Notebooks Server tab in Projects page

lorenzo-cavazzi:
First steps of #351

Not sure if I should move the `fetchNotebookServers` logic inside the `NotebookServers` component. I also have some doubts about a few choices I have made, I'll add some comment in the code

### DIFF
--- a/src/api-client/notebook-servers.js
+++ b/src/api-client/notebook-servers.js
@@ -1,0 +1,50 @@
+/*!
+ * Copyright 2019 - Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+function addNotebookServersMethods(client) {
+  client.getNotebookServers = () => {
+    let headers = client.getBasicHeaders();
+    headers.append('Content-Type', 'application/json');
+    const url = `${client.baseUrl}/notebooks/user`;
+
+    return client.clientFetch(url, {
+      method: 'GET',
+      headers: headers
+    }).then(resp => {
+      return {
+        "names": Object.keys(resp.data.servers),
+        "data": { ...resp.data.servers }
+      }
+    });
+  }
+
+  client.stopNotebookServer = (serverName) => {
+    let headers = client.getBasicHeaders();
+    const url = `${client.baseUrl}/notebooks/servers/${serverName}`;
+
+    return client.clientFetch(url, {
+      method: 'DELETE',
+      headers: headers
+    }, "text")
+      .then(resp => {
+        return true;
+      });
+  }
+}
+
+export default addNotebookServersMethods;
--- a/src/notebooks/Notebooks.container.js
+++ b/src/notebooks/Notebooks.container.js
@@ -21,6 +21,7 @@ import React, { Component } from 'react';
 
 import { NotebookServerOptions } from './Notebooks.present';
 import { ExternalLink } from '../utils/UIComponents';
+import { NotebookServers as NotebookServersPresent } from './Notebooks.present';
 
 class NotebookAdmin extends Component {
   render() {
@@ -28,7 +29,11 @@ class NotebookAdmin extends Component {
     // return <iframe width="100%" height="100%"
     //   style={{border:"none"}}
     //   src={adminUiUrl}></iframe>
-    return <ExternalLink url={adminUiUrl} title="Launch Notebook Admin UI" />
+    return <div>
+      <ExternalLink url={adminUiUrl} title="Launch Notebook Admin UI" />
+      {/* // TODO: attach the new interface Notebook Server display page also here instead of linking
+      to the external page */}
+    </div>
   }
 }
 
@@ -165,4 +170,10 @@ class LaunchNotebookServer extends Component {
   }
 }
 
-export { NotebookAdmin, LaunchNotebookServer };
+class NotebookServers extends Component {
+  render() {
+    return <NotebookServersPresent {...this.props} />
+  }
+}
+
+export { NotebookAdmin, LaunchNotebookServer, NotebookServers };
--- a/src/notebooks/Notebooks.container.js
+++ b/src/notebooks/Notebooks.container.js
@@ -26,13 +26,9 @@ import { NotebookServers as NotebookServersPresent } from './Notebooks.present';
 class NotebookAdmin extends Component {
   render() {
     const adminUiUrl = this.props.adminUiUrl;
-    // return <iframe width="100%" height="100%"
-    //   style={{border:"none"}}
-    //   src={adminUiUrl}></iframe>
+    // TODO: don't open an external popup but display the content here
     return <div>
       <ExternalLink url={adminUiUrl} title="Launch Notebook Admin UI" />
-      {/* // TODO: attach the new interface Notebook Server display page also here instead of linking
-      to the external page */}
     </div>
   }
 }
--- a/src/notebooks/Notebooks.present.js
+++ b/src/notebooks/Notebooks.present.js
@@ -17,7 +17,7 @@
  */
 
 import React, { Component } from 'react';
-import { Form, FormGroup, Label, Input, Button, Row, Col} from 'reactstrap';
+import { Form, FormGroup, Label, Input, Button, Row, Col, Table} from 'reactstrap';
 
 class NotebookServerOptions extends React.Component {
   render() {
@@ -60,7 +60,7 @@ class NotebookServerOptions extends React.Component {
     return (
       <div className="container">
         <Row key="header">
-            <Col sm={12} md={6}><h3>Launch new Jupyterlab server</h3></Col>
+          <Col sm={12} md={6}><h3>Launch new Jupyterlab server</h3></Col>
         </Row>
         <Row key="spacer"><Col sm={8} md={6} lg={4} xl={3}>&nbsp;</Col></Row>
         <Row key="form"><Col sm={8} md={6} lg={4} xl={3}>
@@ -71,7 +71,7 @@ class NotebookServerOptions extends React.Component {
             </Button>
           </Form>
         </Col></Row>
-    </div>
+      </div>
     );
   }
 }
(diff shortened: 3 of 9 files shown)


src/notebooks/Notebooks.container.js
```diff
@@ -165,4 +170,23 @@ class LaunchNotebookServer extends Component {
...
+class NotebookServers extends Component {
+  constructor(props){
+    super(props);
+
+    this.onStopServer = this.doStopServer.bind(this);
+  }
+
+  async doStopServer(serverName) {
+    const url = `${this.props.url}/services/notebooks/servers/${serverName}`;
+    await fetch(url, {credentials: 'include', method: 'DELETE'});
```
ableuler: It would be nice to do the actual fetch from the `api-client` module and have a `stopServer(serverName)` (or similar) method there.

src/api-client/notebook-servers.js
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+function addNotebookServersMethods(client) {
+  client.getNotebookServers = () => {
+    let headers = client.getBasicHeaders();
+    headers.append('Content-Type', 'application/json');
+
+    return client.clientFetch(`${client.jupyterhubUrl}/services/notebooks/user`, {
```
ableuler: This call should be routed through the gateway which will exchange the session cookie for an access token that jupyterhub accepts. `/api/notebooks/...` will be mapped to `/jupyterhub/services/notebooks/...` there, so it should be enough to adapt the URL in the above call accordingly.

src/project/Project.state.js
```diff
@@ -64,6 +64,15 @@ class ProjectModel extends StateModel {
       })
   }
 
+  fetchNotebookServers(client) {
+    this.set('core.notebookServersUpdating', true);
+    return client.getNotebookServers()
+      .then(resp => {
+        this.set('core.notebookServers', resp.data);
```
ableuler: We might want to revisit the schema for the internal representation of a project soon, but for the moment i think it's ok to store the information like this.

ableuler:
Nice job on your first issue! I already have a few requests for changes and I will take another look at it later today.

lorenzo-cavazzi:
@ableuler can't answer to your previous comment (now marked as outdated) but I have addressed both

ciyer:
This looks very good. The only thing is, it seems that the list over servers is not updated after the stop notebook server command returns. Could this be fixed?

lorenzo-cavazzi:
@ciyer Yes, I can poll to check the new status. At the moment we don't have a websocket-like API. Do you think it makes sense to continuously check the list of notebooks while on that page - let's say query the API every 2-5 seconds?

ciyer:
I think this makes sense when the user has triggered a "Stop Notebook" -- in this case, we know that there should be a change to the status coming. Once the status has been updated, we can also stop the polling.

lorenzo-cavazzi:
Apparently solving that problem is not as easy as I hoped... would it make sense to move it to another issue?

ciyer:
The problem is that the "Connect to Jupyter" button is still enabled after the server has stopped?

src/project/Project.state.js
```diff
@@ -64,6 +64,15 @@ class ProjectModel extends StateModel {
       })
   }
 
+  fetchNotebookServers(client) {
+    this.set('core.notebookServersUpdating', true);
```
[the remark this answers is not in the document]
ciyer: I think setUpdating is better -- it more clearly communicates the intention and is something that can easily be searched for, etc.

lorenzo-cavazzi:
> The problem is that the "Connect to Jupyter" button is still enabled after the server has stopped?

yes, and the logic to disable the button is in the `componentDidUpdate` method of the `LaunchNotebookButton` react component.
I have tried some quick workaround but without success. Should that logic be moved somewhere else so that the server status computation can be invoked from outside the component?

rokroskar:
There are several components at play here -- is the `Connect to Jupyter` button visible because Jupyterhub (resp. the notebooks service) says that it is running?

rokroskar:
> would it make sense to move it to another issue?

IMO yes because this might be a deeper problem involving the notebooks service/jupyterhub.

lorenzo-cavazzi:
> There are several components at play here -- is the `Connect to Jupyter` button visible because Jupyterhub (resp. the notebooks service) says that it is running? Or is it really a UI issue (some component not getting updated correctly)?

This is primarily a UI issue because the component is not updated correctly when the server is not running anymore, but if I invoke the API I see the server is not running anymore. On the other hand we don't have a websocket signaling for server status changes, at the moment we can only invoke the API every x seconds to check the real current status - maybe this can be improved.

lorenzo-cavazzi:
> > would it make sense to move it to another issue?
> 
> IMO yes because this might be a deeper problem involving the notebooks service/jupyterhub.

If I query the API to check the server status when the server has already stopped, it spawns again. This is a bit strange... I can check the status with another API but it's a bit more complicated. Not sure if it's a problem with jupyterhub

src/notebooks/Notebooks.present.js
```diff
@@ -117,4 +119,53 @@ class RangeOption extends Component {
...
+    const url = this.props.url;
+    const className = "btn btn-primary " + (!this.props.ready ? "disabled" : "");
+    const loader = !this.props.ready ? <Loader size="14" inline="true" margin="1" /> : null;
+    return <tr>
+      <td>
+        {name}
+        {loader}
+      </td>
+      <td>
+        <a className={className} role="button" href={url}>Connect</a>
```
ciyer: The "Connect" button should open JupyterLab in a new window/tab. Use ```utils/UIComponents.js:ExternalLink``` to get this behavior.

src/project/Project.state.js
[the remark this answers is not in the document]
lorenzo-cavazzi: changed it so that the first time I query the list of server I use `this.setUpdating`, while the next times I don't want any visual loading effect - and of course I got rid of  `core.notebookServersUpdating`.

src/notebooks/Notebooks.present.js
lorenzo-cavazzi: Using `<ExternalLink>` now. I had to slightly change the component so that it can be easily disabled

rokroskar:
> If I query the API to check the server status when the server has already stopped, it spawns again.

the spawn should only happen on a `POST` -- if it happens on a `GET` it's not intended and should be fixed.

lorenzo-cavazzi:
> > If I query the API to check the server status when the server has already stopped, it spawns again.
> 
> the spawn should only happen on a `POST` -- if it happens on a `GET` it's not intended and should be fixed.

I was using `POST` instead of `GET` also to query the status, which according to [the documentation](https://github.com/SwissDataScienceCenter/renku-notebooks#endpoints) is wrong. But using `GET` is not returning a response in `JSON` format, instead I get an HTML page with data available as a JS object I can easily identify by the id `jupyter-config-data` (see below for an example).
I can live with that, just want to be sure it's the intended behavior.

```
<!doctype html><html lang="en"><head><meta charset="utf-8"><title>JupyterLab</title><meta name="viewport" content="width=device-width, initial-scale=1"><script id="jupyter-config-data" type="application/json">{
    "buildAvailable": "True",
    "buildCheck": "True",
    "token": "",
    "devMode": "False",
    "bundleUrl": "/jupyterhub/user/lorenzo.cavazzi.tech/lorenzo-2e-zurich-2db-24d39f9/lab/static/",
    "notebookVersion": "[5, 6, 0]",
    "terminalsAvailable": "True",
    "ignorePlugins": "[]",
    "serverRoot": "~/zurich-bikes-tutorial",
    "mathjaxConfig": "TeX-AMS-MML_HTMLorMML-full,Safe",
    "mathjaxUrl": "/jupyterhub/user/lorenzo.cavazzi.tech/lorenzo-2e-zurich-2db-24d39f9/static/components/MathJax/MathJax.js",
    "appName": "JupyterLab",
    "appNamespace": "jupyterlab",
    "appSettingsDir": "/opt/conda/share/jupyter/lab/settings",
    "appVersion": "0.35.4",
    "cacheFiles": "True",
    "pageUrl": "/lab",
    "publicUrl": "/lab/static/",
    "schemasDir": "/opt/conda/share/jupyter/lab/schemas",
    "settingsUrl": "/lab/api/settings/",
    "staticDir": "/opt/conda/share/jupyter/lab/static",
    "templatesDir": "/opt/conda/share/jupyter/lab/static",
    "themesDir": "/opt/conda/share/jupyter/lab/themes",
    "themesUrl": "/lab/api/themes/",
    "treeUrl": "/lab/tree/",
    "userSettingsDir": "/home/jovyan/.jupyter/lab/user-settings",
    "workspacesApiUrl": "/lab/api/workspaces/",
    "workspacesDir": "/home/jovyan/.jupyter/lab/workspaces",
    "workspacesUrl": "/lab/workspaces/",
    "baseUrl": "/jupyterhub/user/lorenzo.cavazzi.tech/lorenzo-2e-zurich-2db-24d39f9/",
    "wsUrl": ""
  }</script><link rel="icon" type="image/x-icon" href="/jupyterhub/user/lorenzo.cavazzi.tech/lorenzo-2e-zurich-2db-24d39f9/static/base/images/favicon.ico" class="idle favicon"><link rel="" type="image/x-icon" href="/jupyterhub/user/lorenzo.cavazzi.tech/lorenzo-2e-zurich-2db-24d39f9/static/base/images/favicon-busy-1.ico" class="busy favicon">
  

</head><body><script type="text/javascript">
  /* Remove token from URL. */
  (function () {
    var location = window.location;
    var search = location.search;

    // If there is no query string, bail.
    if (search.length <= 1) {
      return;
    }

    // Rebuild the query string without the `token`.
    var query = '?' + search.slice(1).split('&')
      .filter(function (param) { return param.split('=')[0] !== 'token'; })
      .join('&');

    // Rebuild the URL with the new query string.
    var url = location.origin + location.pathname +
      (query !== '?' ? query : '') + location.hash;

    window.history.replaceState({ }, '', url);
  })();
</script><script type="text/javascript" src="/jupyterhub/user/lorenzo.cavazzi.tech/lorenzo-2e-zurich-2db-24d39f9/lab/static/vendors~main.57bd3cc0bd3cf189076d.js"></script><script type="text/javascript" src="/jupyterhub/user/lorenzo.cavazzi.tech/lorenzo-2e-zurich-2db-24d39f9/lab/static/main.6bdd2f1e3af6b3a3a469.js"></script></body></html>
```

rokroskar:
You need to ask for `application/json` in the header

lorenzo-cavazzi:
I have already added it
```
x-requested-with:XMLHttpRequest
Content-Type:application/json
Cookie:<mycookie>
```

rokroskar:
And you get html instead of json?

lorenzo-cavazzi:
yes

rokroskar:
@lorenzo-cavazzi apologies I didn't read carefully earlier -- your header should be `Accept: application/json` not `Content-Type: application/json`.

lorenzo-cavazzi:
I have expanded #340 so that the scope of this PR won't become way bigger than the original issue it references. The "Connect to Jupiter" button problem will be address there ~~*immediately*~~ soon

lorenzo-cavazzi:
closes #351

Issue #387, opened 2019-02-26
no info shown that a project already exists


rokroskar:
When trying to create a project with the same name as one that already exists, the button-click silently fails and a blank form is presented. We should validate the project name before sending the `POST`. 

(reported by @cchoirat)

PR #391, opened 2019-02-28
Adding new project now provides an alert when some errors occurs

lorenzo-cavazzi:
The user now gets an alert when some errors occur while creating a new project. This addresses #387 as well as other generic errors.
The "Create" button is also disabled after the first click until a response comes from the API, and a Loader is displayed in the meanwhile.


rokroskar:
Is there a reason this is not a part of https://github.com/SwissDataScienceCenter/renku-ui/milestone/3?

PR #374, opened 2019-01-30
UI for configuring KG integration

ciyer:
Fix #369.

Issue #362, opened 2018-12-07
Refactor lineage view to use graph service


ableuler:
**MVP**

- [x] Query the lineage API of the gateway rather than the artefacts API. (Needs the first two sub-issues of https://github.com/SwissDataScienceCenter/renku-gateway/issues/71 to be resolved).
Done in #374 and #392 

**post - MVP**

~Improve presentation and interactivity by making projects expandable on click.~

rokroskar:
@ciyer the necessary pieces are in place for this to be done. It seems to me that the changes required are pretty minimal -- could you have a look?

rokroskar:
closing this otherwise it could stay open indefinitely -- @ciyer @lorenzo-cavazzi could you please make another issue to address the `post-MVP` part?

lorenzo-cavazzi:
I think the "on click" interactivity is already covered by #256 . Not sure if the second point here was in the same direction, but with #400 it will be also easy to preview project files.

Issue #369, opened 2019-01-23
UI for configuring knowledge-graph / project integration


ciyer:
# Summary

The renku-graph service needs some support from the UI to work. 

- [x] When a project is created, call the gateway endpoint to register a web hook with gitlab
- [x] When a project is viewed, if the user is the maintainer of the project, check if a web hook has been registered. If not, show an alert with a button to register a web hook
- Provide a button on the settings page to force a reindex of the project. (Note that there is no API endpoint for this functionality so far). (Moved to #390)

<del>To turn on the webhook, make a service `POST` request to `http://webhook-service/projects/<project-id>/hooks` with the gitlab OAuth token in the `OAUTH-TOKEN` header. For the moment, webhooks only work with public projects.</del>

There is an open PR https://github.com/SwissDataScienceCenter/renku-gateway/pull/91 which allows connecting to the webhoos service through the gateway. Calls to `/api/graph/projects/<project_id>/webhooks` are forwarded to `/projects/<project_id>/webhooks` on the webhooks service. From the ui side there is no special authentication header needed (GW will exchange the session cookie for a oauth token).

PR #392, opened 2019-03-01
Use graph services for lineage display

ableuler:
- Remove dot-file parsing from lineage component
- Query graph api instead of artefact api
- Adapt to new graph serialization

Addresses #362 
Merge after #374

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,9 +5,9 @@
   "requires": true,
   "dependencies": {
     "@babel/cli": {
-      "version": "7.1.5",
-      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.1.5.tgz",
-      "integrity": "sha512-zbO/DtTnaDappBflIU3zYEgATLToRDmW5uN/EGH1GXaes7ydfjqmAoK++xmJIA+8HfDw7UyPZNdM8fhGhfmMhw==",
+      "version": "7.2.3",
+      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.2.3.tgz",
+      "integrity": "sha512-bfna97nmJV6nDJhXNPeEfxyMjWnt6+IjUAaDPiYRTBlm8L41n8nvw6UAqUCbvpFfU246gHPxW7sfWwqtF4FcYA==",
       "dev": true,
       "requires": {
         "chokidar": "^2.0.3",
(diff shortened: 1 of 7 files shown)


ableuler:
The PR is almost finished, some sanity checks and comparisons with the previous graph need to be done.

lorenzo-cavazzi:
This is the comparison between the old and the new lineage output for a sample file taken from the Renku tutorial. The new version is on the right.

Lineage comparison

lorenzo-cavazzi:
requires #369 to enable the webhooks in gitlab for lineage generation

lorenzo-cavazzi:
Preview temporary available at https://graphpreview.dev.renku.ch/

rokroskar:
This seems to work as advertised - @ableuler @jirikuncar the only thing it's missing is some more data from the sparql query

ciyer:
Code looks good! I'm surprised it's so few changes. The only thing I do not like (and I do not know if you can control this) is the content of the graph. This new graph is unreadable. As Rok requested, we need to get the others involved and make sure the same graph data is made available from the backend.

lorenzo-cavazzi:
New preview available at https://graphpreview.dev.renku.ch/

Now the content looks better (image 1) but we still have some extra nodes (image 2, `renku update` on top). I changed the commit nodes to show the renku command. If it looks good enough I would merge the PR and let content-related problems to be solved in the gateway

rokroskar:
I don't see any extra nodes here: https://graphpreview.dev.renku.ch/projects/72/files/lineage/figs/grid_plot.png

rokroskar:
OK I see now -- I think that's because the lineage query is not returning all the parents -- can you confirm @ableuler ?

ableuler:
> OK I see now -- I think that's because the lineage query is not returning all the parents -- can you confirm @ableuler ?

I'm not sure I understand what you mean. For me the main question is if the "forward lineage" query should also be able to change direction and go backwards. I would say yes, but only one step. So the zhbikes_parquet nodes should be there, but not the renku update notes above them. I'm not at all sure if and how this can be achieved in the SPARQL query itself. But I think doing some filtering in python after the query is also fine.

lorenzo-cavazzi:
This would be the updated version with a single step backward

Screenshot from 2019-04-04 11-06-37

lorenzo-cavazzi:
Let me know if you think it's better and I'll merge the fix on `gateway`.

Btw I am going to merge this PR

rokroskar:
> Let me know if you think it's better and I'll merge the fix on gateway.

you mean the fix is deployed?

lorenzo-cavazzi:
Not yet, I am creating a PR for that on `renku-gateway`. Could you approve this PR in the meanwhile?

lorenzo-cavazzi:
Updated on https://graphpreview.dev.renku.ch/

rokroskar:
Looks great! Thanks @lorenzo-cavazzi @ableuler and @jirikuncar! 🎉

PR #399, opened 2019-03-12
Switch to nginx-unprivileged base image to not run as root.

### LINKED ISSUE #396: the renku-ui container should not run as root user
The container should run as non-root user.

ciyer:
Fix #396

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,8 +20,8 @@ COPY docker-entrypoint.sh /app/docker-entrypoint.sh
 USER root
 RUN touch /usr/share/nginx/html/config.json
 RUN chmod a+r /usr/share/nginx/html/config.json
-RUN chown 1001 /usr/share/nginx/html/config.json
-USER 1001
+RUN chown nginx /usr/share/nginx/html/config.json
+USER nginx
 
 
 HEALTHCHECK --interval=20s --timeout=10s --retries=5 CMD test -e /var/run/nginx.pid
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,11 +10,19 @@ COPY src /app/src/
 RUN npm install --silent && \
     npm run-script build
 
-FROM nginx:1.13-alpine
+FROM nginxinc/nginx-unprivileged:1.14-alpine
 
 COPY --from=0 /app/build /usr/share/nginx/html
 COPY nginx.vh.default.conf /etc/nginx/conf.d/default.conf
 COPY docker-entrypoint.sh /app/docker-entrypoint.sh
+
+# Set up the config file written by docker-entrypoint
+USER root
+RUN touch /usr/share/nginx/html/config.json
+RUN chmod a+r /usr/share/nginx/html/config.json
+RUN chown 1001 /usr/share/nginx/html/config.json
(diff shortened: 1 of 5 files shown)


rokroskar:
did you try deploying this? I assume you need to change the ports in the chart: https://github.com/SwissDataScienceCenter/renku-ui/blob/master/helm-chart/renku-ui/templates/deployment.yaml#L21

ciyer:
I thought I had deployed this, but it turns out that I had deployed a different version.

Dockerfile
```diff
@@ -10,12 +10,20 @@ COPY src /app/src/
...
 COPY --from=0 /app/build /usr/share/nginx/html
 COPY nginx.vh.default.conf /etc/nginx/conf.d/default.conf
 COPY docker-entrypoint.sh /app/docker-entrypoint.sh
 
+# Set up the config file written by docker-entrypoint
+USER root
+RUN touch /usr/share/nginx/html/config.json
+RUN chmod a+r /usr/share/nginx/html/config.json
+RUN chown 1001 /usr/share/nginx/html/config.json
+USER 1001
```
jirikuncar: we should make sure we store the user id before switching to root

helm-chart/renku-ui/values.yaml
```diff
@@ -54,7 +54,7 @@ image:
 
 service:
   type: ClusterIP
-  port: 80
+  port: 8080
```
jirikuncar:
```suggestion
  nodePort: 80
  port: 8080
```

rokroskar: Are you sure you mean `nodePort`? Or just `targetPort`?

jirikuncar: It should have been a `targetPort`.

jirikuncar: see https://github.com/SwissDataScienceCenter/renku-ui/blob/master/helm-chart/renku-ui/templates/service.yaml#L14

Dockerfile
ciyer: I changed to using the user name specified in the nginx base image.

Dockerfile
```diff
@@ -10,12 +10,20 @@ COPY src /app/src/
...
 COPY --from=0 /app/build /usr/share/nginx/html
 COPY nginx.vh.default.conf /etc/nginx/conf.d/default.conf
 COPY docker-entrypoint.sh /app/docker-entrypoint.sh
 
+# Set up the config file written by docker-entrypoint
+USER root
+RUN touch /usr/share/nginx/html/config.json
+RUN chmod a+r /usr/share/nginx/html/config.json
+RUN chown nginx /usr/share/nginx/html/config.json
+USER nginx
```
jirikuncar: 👍

helm-chart/renku-ui/values.yaml
ciyer: I have been trying to get this to work, but I cannot. I do prefer this suggestion, but I feel that enough time has been spent trying to get it to work for now. I recommend switching to port 8080 and revisiting this later.

PR #402, opened 2019-03-20
Moving Jupyter interactions to Notebook Servers

lorenzo-cavazzi:
This addresses the first 2 points of #340

### DIFF
--- a/src/file/File.container.js
+++ b/src/file/File.container.js
@@ -21,7 +21,7 @@ import ReactMarkdown from 'react-markdown';
 import ReactDOM from 'react-dom';
 import hljs from 'highlight.js';
 
-import { JupyterNotebookPresent, LaunchNotebookButton } from './File.present';
+import { JupyterNotebookPresent } from './File.present';
 import { ACCESS_LEVELS } from '../api-client';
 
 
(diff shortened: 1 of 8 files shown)


lorenzo-cavazzi:
I realized I am not verifying the user access level, showing the "start" button to everybody. I'll fix it

lorenzo-cavazzi:
Ok now it's ready for the review

ciyer:
I think we should remove the + > Launch Jupyter button as well. I think this is too specific a feature to need to have access to there. Let me know if you disagree.

lorenzo-cavazzi:
I definitely agree.

Should we rename "Notebook Servers" tab? I think the definition is technically correct but I was wondering if something like "Jupyter Notebooks" or just "Notebooks" would be more user friendly.

ciyer:
> I definitely agree.
> 
> Should we rename "Notebook Servers" tab? I think the definition is technically correct but I was wondering if something like "Jupyter Notebooks" or just "Notebooks" would be more user friendly.

I think we should eventually rename it to "interactive environments" or similar, but that can wait until we have support for easily defining environments (e.g., R Studio in addition to jupyter lab).

Issue #370, opened 2019-01-23
the launch-jupyter page shows button when it shouldn't


rokroskar:
It might happen that the tokens are expired stale in which case the launch-jupyter dialog (in the image below) is not fully presented. Instead of the full form, the user just sees the "launch jupyter" button, which doesn't actually do anything. This should be corrected either from the gateway side (get a new token) or the UI side (not sure how). 

![image](https://user-images.githubusercontent.com/3353586/51623667-82d36600-1f39-11e9-9392-e592a4f65528.png)

ableuler:
I think the gateway can not do this independently of the UI because as far as I can see, Jupyterhub does not issue any refresh tokens. So the entire the entire oAuth dance with Jupyterhub will have to be performed in order to obtain a new token.

rokroskar:
I'm not sure -- the issue is more that the "Launch Server" button is there, but the form isn't. So there is probably a broken request somewhere that causes the form not to render, but the same failure doesn't prevent the button from being rendered, or better, to cause some sort of a re-login.

ableuler:
Yes, the the request for the server options is most likely failing due to and expired token and thus re-login with Jupyterhub is needed. I'm just wondering how this is best done.

rokroskar:
presumably it returns a 401 or 403 but I haven't tried. 

see `c.JupyterHub.cookie_max_age_days` here: https://jupyterhub.readthedocs.io/en/stable/api/app.html

It's a float so it should be possible to set it to something sufficiently small for testing.

ableuler:
I don’t think that `c.JupyterHub.cookie_max_age_days` sets the expiration time of the oAuth access token. But here's a way to reproduce the problem: 
- log into renku through the ui
- browse to `https://<renku-domain>/jupyterhub/hub/token` and revoke the access token
- delete all jupyterhub cookies in your browser for your renku-domain
- try to launch a notebook.

vfried:
> I don’t think that `c.JupyterHub.cookie_max_age_days` sets the expiration time of the oAuth access token. But here's a way to reproduce the problem:
> 
> * log into renku through the ui
> * browse to `https://<renku-domain>/jupyterhub/hub/token` and revoke the access token
> * delete all jupyterhub cookies in your browser for your renku-domain
> * try to launch a notebook
> 
> Does this work (well fail) @vfried ?

For some reason this only works when i'm not running telepresence :(

vfried:
[...]

PR #404, opened 2019-03-21
Logout and redirect to login when jupyterhub token expires

vfried:
This addresses the bug #370 (It solves the problem following the instructions that @ableuler describes in the bug) This is a way of "fake-replicating" the bug so we can't be 100% sure that it will work, but it should.

### DIFF
--- a/src/api-client/index.js
+++ b/src/api-client/index.js
@@ -26,7 +26,6 @@ import addUserMethods  from './user';
 import addKuMethods  from './ku';
 import addInstanceMethods from './instance';
 import addNotebookServersMethods from './notebook-servers';
-
 import testClient from './test-client'
 
 const ACCESS_LEVELS = {
@@ -74,9 +73,11 @@ class APIClient {
       .catch((error) => {
 
         // For permission errors we send the user to login
-        // TODO: forbidden should not be included here, but test further
-        if (error.case === API_ERRORS.unauthorizedError) {
-          return this.doLogin()
+        if (error.case === API_ERRORS.unauthorizedError){
+          //TODO: when jupyterhub has refresh tokens we should remove the following if and return promise.
+          if( error.response && error.response.url && error.response.url.includes("/api/notebooks"))
+            return Promise.reject(error);
+          return this.doLogin();
         }
 
         // Alert only if corresponding option is set to true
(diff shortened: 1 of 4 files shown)


src/api-client/index.js
```diff
@@ -58,7 +57,6 @@ class APIClient {
     addUserMethods(this);
     addKuMethods(this);
     addInstanceMethods(this);
-    addNotebookServersMethods(this);
```
ableuler: I don't understand why you are removing this. Won't this break the notebook admin tab in the projects page that @lorenzo-cavazzi created recently?

vfried: I know, i made a mistake when merging things... I will correct it tomorrow morning.

vfried: I just corrected this...

rokroskar:
@vfried just to be sure, this is using the response from the notebooks service implemented in https://github.com/SwissDataScienceCenter/renku-notebooks/pull/117 correct?

vfried:
> @vfried just to be sure, this is using the response from the notebooks service implemented in [SwissDataScienceCenter/renku-notebooks#117](https://github.com/SwissDataScienceCenter/renku-notebooks/pull/117) correct?

Yes, it's using #117

src/notebooks/Notebooks.container.js
```diff
@@ -50,25 +51,27 @@ class LaunchNotebookServer extends Component {
...
-    this.componentDidUpdate()
+    if (this.state.doLogOut === false)
+      this.componentDidUpdate()
   }
 
   componentDidUpdate() {
-    if (this.props.core.notebookServerAPI !== this.previousNotebookServerAPI){
-      this.serverStatusSet = false;
-      this.serverOptionsSet = false;
+    if (this.state.doLogOut === false) {
```
ciyer: I think this logic would be easier to follow if it were:
```
if (this.state.doLogOut === true) return;  // We are about to refresh anyway...
[same code as before]
```

vfried: You are right...

rokroskar:
If this works, can we approve, merge and make a bugfix release of the UI and the notebooks service?

ciyer:
> If this works, can we approve, merge and make a bugfix release of the UI and the notebooks service?

Yes, good idea.

PR #413, opened 2019-04-01
Release 0.4.1

ciyer:


ciyer:
> Removed a repetition (#387 had 2 links on the same line)

Good catch, thanks!

Issue #400, opened 2019-03-18
Improve file display / navigation


ciyer:
* Provide a hierarchical view of the files. Currently this is flatten. The hierarchy should be flattened if there only < 6 entries in a further level, otherwise show hierarchically.

* Files should be badged "LFS" if they are in LFS

* The folders can be expended by clicking on the triangle, or they can be entered by double click, which updates the file view to just children of the folder

<img width="708" alt="image" src="https://user-images.githubusercontent.com/1196411/54533219-f1f19780-4989-11e9-97b3-8eafec0f0185.png">

lorenzo-cavazzi:
closed by #420

PR #420, opened 2019-04-08
File display

vfried:
This is to address the issue #400, at the moment it looks like this... 

image

Some things will need to be addressed after this PR is merged:
- See if we can fetch LFS Files or if we can know if a file is from GIT LFS
- Being able to change between branches (now it shows master only)
- Add new formats of files to preview
- Better integration with launching JupyterLab

### DIFF
--- a/src/file/File.container.js
+++ b/src/file/File.container.js
@@ -21,7 +21,9 @@ import ReactMarkdown from 'react-markdown';
 import ReactDOM from 'react-dom';
 import hljs from 'highlight.js';
 
-import { JupyterNotebookPresent } from './File.present';
+import { atobUTF8 } from '../utils/Encoding';
+
+import { JupyterNotebookPresent, JupyterNotebookBody, JupyterNotebookButtonIcon } from './File.present';
 import { ACCESS_LEVELS } from '../api-client';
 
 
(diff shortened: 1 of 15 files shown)


src/file/Lineage.present.js
```diff
@@ -49,33 +57,33 @@ class FileLineageGraph extends Component {
   subGraph() {
     const subGraph = new graphlib.Graph()
       .setGraph({})
-      .setDefaultEdgeLabel(function() { return {}; });
+      .setDefaultEdgeLabel(function () { return {}; });
```
ciyer: I think there should be no space between `function` and `()`.

src/utils/treeview/data.js
```diff
@@ -0,0 +1,187 @@
+function buildTree(parts, treeNode, jsonObj) {
+  if (parts.length === 0) {
+    return;
+  }
+  for (var i = 0; i < treeNode.length; i++) {
```
ciyer: `var` -> `let`

src/utils/treeview/data.js
```diff
@@ -0,0 +1,187 @@
...
+    if (parts[0] == treeNode[i].name) {
+      buildTree(parts.splice(1, parts.length), treeNode[i].children, jsonObj);
+      return;
+    }
+  }
+  var newNode;
```
ciyer: `var` -> `let`

src/utils/treeview/data.js
vfried: I deleted this file, the data was just for testing

src/utils/treeview/data.js
vfried: This file was deleted, i just used it for testing reasons.

src/file/File.container.js
```diff
@@ -85,10 +110,12 @@ class FilePreview extends React.Component {
   render(){
     // File has not yet been fetched
     if (!this.props.file) {
-      return <p>Loading...</p>;
+      return "Loading...";
     }
     // Various types of images
     if (this.fileIsImage()) {
+      if(atobUTF8(this.props.file.content).includes("https://git-lfs.github.com/"))
```
ciyer: I think we should just use regular atob here, since the image is going to be binary content, not UTF-8 encoded.

vfried: I think you are right, i am not parsing the image in that line, I just check if the content contains the git lfs text but that will not have special characters anyway

Issue #406, opened 2019-03-22
Project page refactoring


lorenzo-cavazzi:
Refactor the project page so that:

- [x] project existence is tested first. If the project doesn't exists (or it is private and the user can't access) display a proper page and don't invoke any other project-related APIs.
- **see comment below** ~[ ] data used by multiple child components (e.g. `ProjectViewHeader` and `LaunchNotebookServer`) are collected only once by the parent component and stored in the Redux store. Refresh them continuously wherever it's needed~ 

This should prevent having inconsistent information displayed on the same page and prevent re-loading child components multiple times when it's not needed.

lorenzo-cavazzi:
#421 addresses the first point, I suggest to close this issue and to address the second point gradually -- it's easier/faster to move a component used in multiple project tabs as soon as working on an issue that directly involves that component

PR #421, opened 2019-04-09
Re-factoring of Project and UserModel

### LINKED ISSUE #389: trying to view a non-accessible project doesn't provide an adequate feedback
When a user is trying to access a non-existing or a private project some unhandled errors are thrown and the resulting page may be confusing.
A clear indication that the project doesn't exist could be a better solution.

(Maybe handle this with the react-router)

lorenzo-cavazzi:
This addresses the first part of #406 and closes #389 

The re-factor involves the following:

* `UserState` is initialized with the attribute `available` to verify if the user login has finished or not. This required a few changes in pages where user existence is checked and in unit tests -- these have also expanded to test for both logged in and anonymous user where meaningful.
* `Project` page first verifies if the user login is still ongoing, and it waits until this step is completed before fetching project data. This helps to avoid re-loading component multiple times in short timespan.
* A "404" page is displayed when the project is not available. **Info:** please feel free to suggest any update on both text and layout -- see more below.
* Webhooks variables stored in redux are moved to `webhook` property instead of polluting `core`

Temporary preview available here: https://lorenzotest.dev.renku.ch/
 **Info:** Second point can be tested by trying to access a non-existing project (tip: different feedback for anonymous and logged-in users) -- https://lorenzotest.dev.renku.ch/projects/9999/

### DIFF
--- a/src/app-state/UserState.test.js
+++ b/src/app-state/UserState.test.js
@@ -1,14 +1,16 @@
-import UserState from  './UserState'
+import UserState from './UserState'
+import { InitialState } from './UserState'
 import { createStore } from '../utils/EnhancedState';
 import { testClient as client } from '../api-client';
 import { user } from '../api-client/test-samples';
+import { doesNotReject } from 'assert';
 
 
 describe('fetch user', () => {
   // We use a redux store to hold some global application state.
   it('starts in initial state', () => {
     const store = createStore(UserState.reducer);
-    expect(store.getState()).toEqual(null);
+    expect(store.getState()).toEqual(InitialState);
   });
 
   // Load the user profile and dispatch the result to the store.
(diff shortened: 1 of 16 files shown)


src/app-state/UserState.test.js
```diff
@@ -18,6 +20,42 @@ describe('fetch user', () => {
...
+      const generatedUser = await generateFakeUser();
+      expect(generatedUser).toEqual(store.getState());
+    });
   });
 });
+
+/**
+ * Generate a fake user to be injected in the other tests
+ * 
+ * @param {string} anonymous  True to ask for an anonymous user (no id nor username)
```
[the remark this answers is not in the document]
lorenzo-cavazzi: right, fixed (and rebased on master).
Not updating the preview for this small change :)

src/project/Project.present.js
```diff
@@ -796,34 +794,96 @@ class ProjectSettings extends Component {
...
+      tip = <InfoAlert timeout={0}>
+        <p className="mb-0">
+          <FontAwesomeIcon icon={faInfoCircle} /> You should first try to login.
+        </p>
+      </InfoAlert>
+    }
+
+    return <Row>
+      <Col>
+        <h1>404 - project not found :(</h1>
```
ciyer: Let's keep this simple for now. We can make a cuter version later. :)
```
<h1>404</h1>
<h2>Project not found</h2>
<p>We could not find project #{this.props.id}.</p>
<p>It is possible that the project has been deleted by its owner or you don&apos;t have permission to access it.</p>
{ tip }

File: src/project/Project.present.js [hunk 3]
```diff
@@ -796,34 +794,96 @@ class ProjectSettings extends Component {
   }
 }
 
+class ProjectViewNotFound extends Component {
+  render() {
+    let tip;
+    if (this.props.logged) {
+      tip = <InfoAlert timeout={0}>
+        <p>
```
ciyer:
```
        <p>
           <FontAwesomeIcon icon={faInfoCircle} />If you are sure the project exists you, you may want to try the following:
         </p>
         <ul className="mb-0">
           <li>Do you have multiple accounts? Are you logged in with the right user?</li>
           <li>
             If you received this link from someone, ask that person to make sure you have access to the project.
           </li>
         </ul>
```

src/project/Project.present.js
```diff
@@ -796,34 +794,96 @@ class ProjectSettings extends Component {
...
+            If you received this link from someone, please contact him and ask to make the project
+            accessible to your user.
+          </li>
+        </ul>
+      </InfoAlert>
+    }
+    else {
+      tip = <InfoAlert timeout={0}>
+        <p className="mb-0">
+          <FontAwesomeIcon icon={faInfoCircle} /> You should first try to login.
```
ciyer: Text:
```You might need to be logged in to see this project. Please try to log in.```

src/project/Project.present.js
lorenzo-cavazzi: Done, but removed the repetition in **bold**
`If you are sure the project exists` **you** `, you may want to try the following:`

lorenzo-cavazzi:
Updated, also on https://lorenzotest.dev.renku.ch

As reference, these were the 2 previous screen for non-existing projects

Screenshot from 2019-04-10 15-59-19

Screenshot from 2019-04-10 15-59-40

Issue #434, opened 2019-04-18
Improve error messages for not found files inside the file viewer


vfried:
When a file is not found or can't be loaded from some reason there should be an error saying that the file doesn't exist or could not be loaded.

PR #435, opened 2019-04-18
Added message of not found or not loaded when there is an error when fetching a file

vfried:
When a file is not found or could not be loaded there is an error message:

image

Also deleted some console.log that where left from a previous commit.


lorenzo-cavazzi:
closes #434

PR #437, opened 2019-04-23
Update of the notebook viewer version

vfried:
It looks like an update on the version of the notebook previewer fixes this issue.

The other version was deprecated: https://github.com/nteract/notebook-preview 
Now using: https://github.com/nteract/notebook-render

### DIFF
--- a/src/file/File.present.js
+++ b/src/file/File.present.js
@@ -18,8 +18,8 @@
 
 import React from 'react';
 import ReactDOM from 'react-dom';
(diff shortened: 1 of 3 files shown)


src/file/File.present.js
```diff
@@ -127,14 +127,15 @@ class StyledNotebook extends React.Component {
   }
 
   render() {
-    const notebookStyle = `
+    let notebookStyle = `
```
lorenzo-cavazzi: minor detail: is there any reason to change this from `const` to `let`?

PR #439, opened 2019-04-24
fix jupyterhub API

### LINKED ISSUE #424: notebook servers list loading error
If jupyterhub is restarted, the jupyterhub token stored by the gateway is no longer valid. The UI receives a 401 in this case from the notebook service, and should force a re-login but it does not seem to do so. 

Image used: 0.4.2-4231768

image

lorenzo-cavazzi:
Closes #424 

Preview available at https://lorenzotest.dev.renku.ch

Please try it to be sure the problem solved by #404 doesn't come back.
If you wan't to play with rancher to try to generate 401 responses, the `lorenzotest` namespace is in the `UI-team` folder.

### DIFF
--- a/src/api-client/index.js
+++ b/src/api-client/index.js
@@ -78,11 +78,6 @@ class APIClient {
 
         // For permission errors we send the user to login
         if (reLogin && error.case === API_ERRORS.unauthorizedError){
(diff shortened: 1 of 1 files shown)


vfried:
I tried to break it and couldn't. I did the token revoke and cookie removal, etc and the error is not there anymore. Good work!

Issue #432, opened 2019-04-16
Improve the icons in the new files display


vfried:
- [x] Add tooltips to all the icons
- [x] Set as selected the file/graph depending on what is being displayed
- [x] Display the jupyterlab icon that is missing now on the graph view
- [x] Display gitlab link for not found files? -->Done, i also improved the error message
- [x] Add an LFS icon or label when the file is from git lfs
- [x] When a user selects "graph/content" keep this selected in case the user selects a new file

PR #447, opened 2019-05-02
Icons and functionality improvements

vfried:
This addresses the issue #432

### DIFF
--- a/src/api-client/project.js
+++ b/src/api-client/project.js
@@ -33,23 +33,34 @@ function groupedFiles(files, projectFiles) {
   return projectFiles
 }
 
-function buildTree(parts, treeNode, jsonObj, hash, currentPath) {
+function buildTree(parts, treeNode, jsonObj, hash, currentPath, gitattributes, openFilePath) {
   if (parts.length === 0) {
     return;
   }
 
   currentPath = currentPath === "" ? parts[0] : currentPath+'/'+parts[0];
+  let nodeName = parts[0];
+
+
+  let fileInsideIsSelected = openFilePath.length > 0 
+    && openFilePath.substring(1).startsWith(nodeName);  
+
+  let openFilePathParam =  fileInsideIsSelected 
+    ? openFilePath.substring(1).replace(nodeName,"") : "" ;
+
 
   for (let i = 0; i < treeNode.length; i++) {
     if (parts[0] === treeNode[i].text) {
-      buildTree(parts.splice(1, parts.length), treeNode[i].children, jsonObj, hash, currentPath+'/'+parts[0]);
+      buildTree(parts.splice(1, parts.length), treeNode[i].children, jsonObj, hash, currentPath+'/'+parts[0], gitattributes, openFilePathParam);
       return;
     }
   }
 
+  const isLfs = gitattributes ? gitattributes.includes(currentPath+" filter=lfs diff=lfs merge=lfs -text") : false;
+  
   let newNode;
   if (parts[0] === jsonObj.name)
-    newNode = { 'name': parts[0], 'children': [], 'jsonObj': jsonObj, 'path': currentPath  };
+    newNode = { 'name': parts[0], 'children': [], 'jsonObj': jsonObj, 'path': currentPath ,'isLfs': isLfs};
   else
     newNode = { 'name': parts[0], 'children': [], 'jsonObj': null, 'path': currentPath };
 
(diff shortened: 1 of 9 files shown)


src/api-client/project.js
```diff
@@ -33,49 +33,76 @@ function groupedFiles(files, projectFiles) {
...
   let tree = [];
   let hash = {};
-  for (let i = 0; i < list.length; i++) {
-    buildTree(list[i].path.split('/'), tree, list[i] , hash, "");
+  let lfs = files.filter((treeObj) => treeObj.path === '.gitattributes'); 
+
+  if(lfs.length >0){
+    return client.getRepositoryFile(projectId, lfs[0].path, 'master', 'raw')
+      .catch(e => {
+        console.log("error")
```
lorenzo-cavazzi: is there something useful that can be done inside this catch - like displaying a feedback to the user or triggering another action? Otherwise I would remove all the `catch`

src/file/File.present.js
```diff
@@ -54,7 +53,7 @@ class ShowFile extends React.Component {
     this.props.client.getRepositoryFile(this.props.projectId, filePath, branchName, 'base64')
       .catch(e => {
         if (e.case === API_ERRORS.notFoundError) {
-          this.setState({error:"The file with path '"+ this.props.filePath +"' does not exist or could not be loaded."})
+          this.setState({error:"ERROR 404: The file with path '"+ this.props.filePath +"' does not exist."})
```
lorenzo-cavazzi: would be nice if this message was displayed also when the user is in "lineage mode" when a file doesn't exist.

src/project/filestreeview/FilesTreeView.js
```diff
@@ -86,31 +105,76 @@ class FilesTreeView extends Component {
...
+      : this.props.projectUrl+redirectURL;
+    const toLineage = this.props.lineageUrl+redirectURL
 
     return (
       <div className="tree-container">
         <div className="tree-title">
-          <span>Repository</span>
+          <span className="tree-header-title text-truncate">Repository</span>
+          <span className="float-right throw-right-in-flex">
+            <Dropdown color="primary" size="sm" isOpen={this.state.dropdownOpen} toggle={this.toggle}>
```
lorenzo-cavazzi: Does it make sense to add a tooltip also here, as in the icons on the right of the page? Maybe some like "select view mode". I am not sure if this would be a good solution, but at the moment it is a bit difficult for the user to understand the purpose of this icon.

src/project/filestreeview/treeviewstyle.css
```diff
@@ -31,6 +31,10 @@
     cursor: pointer;
 }
 
+.throw-right-in-flex{
+    margin-left: auto;;
```
lorenzo-cavazzi: double semicolon

src/api-client/project.js
vfried: This error was there for testing reasons and i forgot to remove it, it's not there anymore.

src/project/filestreeview/FilesTreeView.js
vfried: I think it doesn't, too many tooltips could be a bit annoying. I just changed the "Repository" for "File View" / "Lineage View" to make it a bit more clear.

src/project/filestreeview/treeviewstyle.css
vfried: Fixed.

vfried:
> I have left a few comments, feel free to disagree where I propose changes if you think the current solution is better :)
> 
> There is another detail to fix. When the screen is medium sized, the navigation tree on the left is too small
> 
> Screenshot from 2019-05-03 08-48-49

I tried to improve a bit the layout here, I could also try adding a minimum width or something if this doesn't look good to you.

src/file/File.present.js
vfried: I made some changes for the error message to be displayed when it's a 404 but it looks like the error that comes from the lineage is not well formed.

lorenzo-cavazzi: The API response from the lineage endpoint is the same for non-existing files and files without lineage, you would need to use this endpoint also when the visualization is on Lineage. But this is a minor detail, we could create another issue for later.

lorenzo-cavazzi:
> I tried to improve a bit the layout here, I could also try adding a minimum width or something if this doesn't look good to you.

Not sure how the minimum width would behave with `Col`.
Maybe we could go from `sm={12} md={4} lg={2}` to `sm={12} md={4} lg={3} lx={2}`, or even `sm={12} md={4} lx={3}`. But I guess the perfect balance here is impossible :)

PR #455, opened 2019-05-10
Files tree view with lazy loading

vfried:
I implemented lazy loading for the file tree. 

1) The level 0 will be downladed
2) With each click on a folder the following level of the sub-tree will ve loaded in the folder

==> in case a link is sent with a file open all the subtrees to reach the selected files will be loaded too.

### DIFF
--- a/src/api-client/project.js
+++ b/src/api-client/project.js
@@ -33,71 +33,38 @@ function groupedFiles(files, projectFiles) {
   return projectFiles
 }
 
-function buildTree(parts, treeNode, jsonObj, hash, currentPath, gitattributes, openFilePath) {
-  if (parts.length === 0) {
+function buildTreeLazy(name, treeNode, jsonObj, hash, currentPath, gitattributes, openFilePath) {
+  if (name.length === 0) {
     return;
   }
-
-  currentPath = currentPath === "" ? parts[0] : currentPath+'/'+parts[0];
-  let nodeName = parts[0];
-
-
-  let fileInsideIsSelected = openFilePath.length > 0 
-    && openFilePath.substring(1).startsWith(nodeName);  
-
-  let openFilePathParam =  fileInsideIsSelected 
-    ? openFilePath.substring(1).replace(nodeName,"") : "" ;
-
-
-  for (let i = 0; i < treeNode.length; i++) {
-    if (parts[0] === treeNode[i].text) {
-      buildTree(parts.splice(1, parts.length), treeNode[i].children, jsonObj, hash, currentPath+'/'+parts[0], gitattributes, openFilePathParam);
-      return;
-    }
-  }
-
+  currentPath = jsonObj.path;
+  let nodeName = name;
+  let nodeType = jsonObj.type;  // "tree" "blob" "commit"
   const isLfs = gitattributes ? gitattributes.includes(currentPath+" filter=lfs diff=lfs merge=lfs -text") : false;
-  
-  let newNode;
-  if (parts[0] === jsonObj.name)
-    newNode = { 'name': parts[0], 'children': [], 'jsonObj': jsonObj, 'path': currentPath ,'isLfs': isLfs};
-  else
-    newNode = { 'name': parts[0], 'children': [], 'jsonObj': null, 'path': currentPath };
-
-  const currentNode = treeNode.filter(node => node.name === newNode.name);
-  
-  if (currentNode.length === 0) {
-    treeNode.push(newNode);
-    hash[newNode.path] = {'name': parts[0], 'selected': false, 'childrenOpen': fileInsideIsSelected , 'path': currentPath, 'isLfs': isLfs };
-    buildTree(parts.splice(1, parts.length), newNode.children, jsonObj, hash, currentPath, gitattributes, openFilePathParam );
-  } else {
-    for (let j = 0; j < newNode.children.length; j++) {
-      currentNode[0].children.push(newNode.children[j]);
-    }
-    buildTree(parts.splice(1, parts.length), currentNode[0].children, jsonObj, hash, currentPath, gitattributes, openFilePathParam ) ;
-  }
+  let newNode =  { 'name': nodeName, 'children': [], 'jsonObj': jsonObj, 'path': currentPath ,'isLfs': isLfs , 'type': nodeType};
+  hash[newNode.path] = {'name':nodeName, 'selected': false, 'childrenOpen': false , 'childrenLoaded':false , 'path': currentPath, 'isLfs': isLfs, 'type': nodeType , 'treeRef': newNode};
+  treeNode.push(newNode);
 }
 
-function getFilesTree(client, files, projectId, openFilePath) {
-  let list = files.filter((treeObj) => treeObj.type === 'blob');
+function getFilesTreeLazy(client, files, projectId, openFilePath, lfsFiles) {
   let tree = [];
   let hash = {};
-  let lfs = files.filter((treeObj) => treeObj.path === '.gitattributes'); 
+  let lfs = files.filter((treeObj) => treeObj.path === '.gitattributes');
 
-  if(lfs.length >0){
+  if(lfs.length > 0){
     return client.getRepositoryFile(projectId, lfs[0].path, 'master', 'raw')
       .then(json => {
-        for (let i = 0; i < list.length; i++) {
-          buildTree(list[i].path.split('/'), tree, list[i] , hash, "", json, openFilePath);
+        for (let i = 0; i < files.length; i++) {
+          buildTreeLazy(files[i].name, tree, files[i] , hash, "", json, openFilePath);
         }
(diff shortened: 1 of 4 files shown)


src/api-client/project.js
```diff
@@ -33,71 +33,39 @@ function groupedFiles(files, projectFiles) {
...
-  }
+  let newNode =  { 'name': nodeName, 'children': [], 'jsonObj': jsonObj, 'path': currentPath ,'isLfs': isLfs , 'type': nodeType};
+  hash[newNode.path] = {'name':nodeName, 'selected': false, 'childrenOpen': false , 'childrenLoaded':false , 'path': currentPath, 'isLfs': isLfs, 'type': nodeType , 'treeRef': newNode};
+  treeNode.push(newNode);
 }
 
-function getFilesTree(client, files, projectId, openFilePath) {
-  let list = files.filter((treeObj) => treeObj.type === 'blob');
+function getFilesTreeLazy(client, files, projectId, openFilePath, lfsFiles) {
+  //let list = files; //might need filtering
```
ciyer: Is this an out-of-date comment?

src/api-client/project.js
```diff
@@ -33,71 +33,39 @@ function groupedFiles(files, projectFiles) {
...
   let tree = [];
   let hash = {};
-  let lfs = files.filter((treeObj) => treeObj.path === '.gitattributes'); 
+  let lfs =  files.filter((treeObj) => treeObj.path === '.gitattributes') ;
```
ciyer: Minor: There is an extra space before `files.filter` and before `;`.

src/api-client/project.js
```diff
@@ -137,14 +105,13 @@ function addProjectMethods(client) {
     })
   }
 
-  client.getProjectFilesTree = (projectId, openFilePath) => {
-    return client.getRepositoryTree(projectId, { path: '', recursive: true }).then((tree) => {
-      const fileStructure = getFilesTree(client, tree, projectId, openFilePath);
+  client.getProjectFilesTree = (projectId, openFilePath, currentPath = '', lfsFiles) => {
+    return client.getRepositoryTree(projectId, { path: currentPath, recursive: false }).then((tree) => {
+      let fileStructure = getFilesTreeLazy(client, tree, projectId, openFilePath, lfsFiles);
```
ciyer: `fileStructure` looks like it should be `const`.

src/project/Project.state.js
```diff
@@ -110,18 +112,53 @@ class ProjectModel extends StateModel {
           this.setObject(updatedState);
           this.set('filesTree', d);
           return d;
+        })
+        .then(d=> {
+          if(openFilePath !== '' && openFilePath.length > 0 && openFilePath.split('/').length-1>0){
```
ciyer: 1. `openFilePath !== '' && openFilePath.length > 0` is redundant. ` openFilePath.length > 0` means `openFilePath !== ''`

2. I think `openFilePath.split('/').length > 1` is easier to read.

src/project/Project.state.js
```diff
@@ -98,9 +98,11 @@ class ProjectModel extends StateModel {
       });
   }
 
-  fetchProjectFilesTree(client, id, openFilePath) {
+  fetchProjectFilesTree(client, id, openFilePath, openFolder ) {
     if (this.get('transient.requests.filesTree') === SpecialPropVal.UPDATING) return;
     const oldTree = this.get('filesTree');
+    if(openFilePath.startsWith('/'))
+      openFilePath = openFilePath.substring(1);
     if(oldTree === null || oldTree === undefined){
```
ciyer: Could we extract a method for this branch of the if statement? Maybe `initialFetchProjectFilesTree(...)`?

src/project/Project.state.js
```diff
@@ -110,18 +112,53 @@ class ProjectModel extends StateModel {
...
+            const goto = openFolder !== undefined ?  openFolder + "/" +openFilePathArray[0] : openFilePathArray[0];                
+            return this.fetchProjectFilesTree(client, id, openFilePath.replace(openFilePathArray[0],''), goto);
+          } else {
+            return d;
+          } 
         });
-    } else {
-      return oldTree;
+    } else { 
+      if(openFolder !== undefined && oldTree.hash[openFolder].childrenLoaded === false) {
```
ciyer: Could we extract this code to a method as well. Maybe `deeperFetchProjectFilesTree(...)`

src/project/Project.state.js
```diff
@@ -98,9 +98,11 @@ class ProjectModel extends StateModel {
...
   }
 
-  fetchProjectFilesTree(client, id, openFilePath) {
+  fetchProjectFilesTree(client, id, openFilePath, openFolder ) {
     if (this.get('transient.requests.filesTree') === SpecialPropVal.UPDATING) return;
     const oldTree = this.get('filesTree');
+    if(openFilePath.startsWith('/'))
+      openFilePath = openFilePath.substring(1);
     if(oldTree === null || oldTree === undefined){
       this.setUpdating({transient:{requests:{filesTree: true}}});
```
ciyer: This line, `this.setUpdating({transient:{requests:{filesTree: true}}});`,  can be moved outside the `if`, since it is in both branches.

src/project/Project.state.js
```diff
@@ -110,18 +112,53 @@ class ProjectModel extends StateModel {
...
+            return d;
+          } 
         });
-    } else {
-      return oldTree;
+    } else { 
+      if(openFolder !== undefined && oldTree.hash[openFolder].childrenLoaded === false) {
+        this.setUpdating({transient:{requests:{filesTree: true}}});
+        return client.getProjectFilesTree(id, openFilePath, openFolder, oldTree.lfsFiles)
+          .then(resp => resp)
```
ciyer: Is this line necessary? (Same for the above).

src/project/Project.state.js
```diff
@@ -110,18 +112,53 @@ class ProjectModel extends StateModel {
...
+            oldTree.hash[openFolder].childrenLoaded=true;
+            oldTree.hash[openFolder].childrenOpen = true;
+            for (const node in d.hash) 
+              oldTree.hash[node] = d.hash[node];
+            const updatedState = { filesTree: oldTree, transient:{requests:{filesTree: false}} };
+            this.setObject(updatedState);
+            this.set('filesTree', oldTree);
+            return oldTree;
+          }).then(d=> {
+            if(openFilePath !== '' && openFilePath.length > 0 && openFilePath.split('/').length-1>0){
```
ciyer: This code duplicates code from the other branch of the if. Extract to a shared function/method.

src/api-client/project.js
vfried: No, i left it there in case there is some type of file we don't want like the 'commit' file i showed you.

src/api-client/project.js
vfried: You are right.

src/project/Project.state.js
vfried: It's in all cases but one, if i put it breaks the code.

src/api-client/project.js
vfried: I deleted it, i guess we can just add it if necessary later.

src/api-client/project.js
vfried: Done.

src/project/Project.state.js
vfried: Done

src/project/Project.state.js
vfried: Done

src/project/Project.state.js
vfried: Done

src/project/Project.state.js
vfried: Yes, the line is necessary... it prevents from trying to refetch the tree in case the tree is being updated.

src/project/Project.state.js
vfried: I made a new function for it now.

src/project/Project.state.js
```diff
@@ -98,30 +99,84 @@ class ProjectModel extends StateModel {
...
+        : openFilePathArray[0];
+      return this.fetchProjectFilesTree(client, id, openFilePath.replace(openFilePathArray[0],''), goto);
+    } else {
+      return tree;
+    } 
+  }
+
+  cleanFilePathUrl(openFilePath){
+    if(openFilePath.startsWith('/'))
+      openFilePath = openFilePath.substring(1);
```
ciyer: Should there be a `return` here?

src/project/Project.state.js
ciyer: Sorry -- it was not clear the way the comment was entered -- but I'm referring only to 
```.then(resp => resp)```. This is unnecessary.

src/project/Project.state.js
vfried: Yes, i think so... i added it

PR #460, opened 2019-05-10
Remove need for the Jupyterhub URL in the UI

### LINKED ISSUE #454: Remove Jupyterhub dependency
As soon as https://github.com/SwissDataScienceCenter/renku-notebooks/issues/95 is resolved, the UI should not be aware of the Jupyterhub URL anymore.

ableuler:
Closes #454

---

Merge together with https://github.com/SwissDataScienceCenter/renku-notebooks/pull/152


ableuler:
> I think some changes should be done also in `File.container.js`, otherwise the notebooks' preview in `Files` tab doesn't show the "Launch Jupyter" icon on top right anymore

Yes you're right. This is fixed now. Some more remains to be done here though (see issue #462 ).

ableuler:
> Now I see the Jupyterlab button while browsing a notebook in the `Files` tab of a project, but if I click it nothing happens. It should open the notebook in a new tab.

Yes, I fixed the deactivation of the button for the case when there's no running notebook server. There already was an attempt to do that using the `disabled` html attribute, but that only works for a `button` but not for a `span` element. Deactivating that button is the right thing to do - the tooltip tells you why the button is deactivated. Currently, on can click the button but then you get to a forever-polling tab which is just waiting for a notebook server to be launched. 
What needs to be fixed is an update of the state information about running notebook servers after the successful notebook launch. Currently, one has to switch projects or reload the page to make the tooltip disappear and activate the button. I've added a bullet point to #462 about this.

lorenzo-cavazzi:
waiting until the next release to be merged

rokroskar:
@lorenzo-cavazzi so this PR is not going into the current release?

lorenzo-cavazzi:
@rokroskar yes, the idea is to exclude it from `0.5.0` -- that will be a pretty big release

rokroskar:
@lorenzo-cavazzi ok then I will defer https://github.com/SwissDataScienceCenter/renku-notebooks/pull/152 to the next release there as well. Agreed @ableuler?

ableuler:
> @lorenzo-cavazzi ok then I will defer [SwissDataScienceCenter/renku-notebooks#152](https://github.com/SwissDataScienceCenter/renku-notebooks/pull/152) to the next release there as well. Agreed @ableuler?

ok for me...

ableuler:
> waiting until the next release to be merged

Now that the release is out I would like to get this in so I can continue cleaning up the charts. The concerns by @lorenzo-cavazzi have been addressed... (@vfried, @ciyer)

PR #461, opened 2019-05-10
deployment cleanup

ableuler:


rokroskar:
don't we need to be a bit careful here to do all of these related PRs at the same time?

ableuler:
This PR is ready for review.

PR #463, opened 2019-05-14
Fix vulnerable dependencies

ableuler:
Involves package updates with breaking changes! Please **test** before approving! The new image is deployed on https://deps.dev.renku.ch


lorenzo-cavazzi:
I have tested a little bit and I have not found any problem, but the gateway version is a bit old on `deps.dev.renku.ch` and I can't test graph related feature. Could you update it?

rokroskar:
I also tested and seems fine - I assumed the graph is just turned off and that's why it doesn't work?

ableuler:
> I have tested a little bit and I have not found any problem, but the gateway version is a bit old on `deps.dev.renku.ch` and I can't test graph related features. Could you update it?

That means we should probably update the gateway requirement in the main renku chart... Anyway, it's now using the latest gateway chart/image.

Issue #416, opened 2019-04-01
Start notebook from any branch/commit


rokroskar:
- [ ] Add option to choose branch when launching a notebook -- should be done by attaching the `?branch=<name>` query parameter to the notebook launch POST request URL.
- [ ] Ensure that the latest commit is used when a notebook server is started (currently, we use the latest commit at the time the page was loaded. This causes problems if a user pushes between loading the view and launching a notebook server)

OPTIONAL
- [ ] Add option to choose a specific commit from chosen branch when launching a notebook. Display date and (head of) commit message per commit.

info: requires #406

lorenzo-cavazzi:
the refactor should address also this #336

PR #466, opened 2019-05-17
time helper functions

lorenzo-cavazzi:
Should we have a standardized way to display exact dates? I have created a few helper functions to simplify handling the ISO date format returned by APIs. I will use this in #416 to display commit dates.

Feel free to disagree with this PR or to suggest alternatives.

### DIFF
--- a/src/utils/Time.js
+++ b/src/utils/Time.js
@@ -1,0 +1,63 @@
+/*!
+ * Copyright 2019 - Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


src/utils/Time.js
```diff
@@ -0,0 +1,63 @@
...
+ *  renku-ui
+ *
+ *  Time.js
+ *  Helper functions to handle date and time
+ */
+
+// TODO: expand this with {timeFormat} from 'd3' to properly handle human readable dates according to localization
+
+class Time {
+  static isValidDate(date) {
```
ciyer: I suggest a different name. When I see "isValidDate" I expect it to check if a string is a valid date. I prefer `isDate` or `isDateObject`.

src/utils/Time.js
```diff
@@ -0,0 +1,63 @@
...
+    return date instanceof Date && !isNaN(date);
+  }
+
+  static validateDate(date) {
```
ciyer: I think `parseDate` is a better name for this. We can also add an optional `format` argument later.

src/utils/Time.js
```diff
@@ -0,0 +1,63 @@
...
+      return date;
+    }
+    const convertedDate = new Date(date);
+    if (this.isValidDate(convertedDate)) {
+      return convertedDate;
+    }
+    throw("Invalid date");
+  }
+
+  static formatIsoReadable(inputDate, type="datetime") {
```
ciyer: I prefer the name `toISOString` here, since it has the same semantics, it's just an extension.

lorenzo-cavazzi:
function names updated with the suggestions

PR #469, opened 2019-05-23
opt-out from knowledge graph for new private projects

lorenzo-cavazzi:
Adds the possibility to opt-out from Knowledge Graph when creating a new project with `private` visibility.

Test available at https://lorenzotest.dev.renku.ch


lorenzo-cavazzi:
updated with a new warning for non indexed private projects

lorenzo-cavazzi:
yes, this PR will add it soon: SwissDataScienceCenter/renku#573

Issue #429, opened 2019-04-16
New Jupyter server with auto-saved data


lorenzo-cavazzi:
When the user starts a new Jupyter server, if there is a previous auto-saved state, a popup should ask for the next action -- keep the auto-saved state or discard it and start from a clean state

PR #472, opened 2019-05-24
feat(notebook): start jupyterlab servers from any branch and commit

### LINKED ISSUE #416: Start notebook from any branch/commit

lorenzo-cavazzi:
Add the possibility to start a Jupyterlab server form any branch and commit

Screenshot from 2019-05-24 18-41-19

If the server is already running, the user can connect directly from the launch page.
A preview is available here: https://lorenzotest.dev.renku.ch/
_It already includes full URL paths from this PR: SwissDataScienceCenter/renku-notebooks#181_

---

closes #416 
addresses https://github.com/SwissDataScienceCenter/renku/issues/437

### DIFF
--- a/src/notebooks/Notebooks.container.js
+++ b/src/notebooks/Notebooks.container.js
@@ -18,162 +18,278 @@
 
 import React, { Component } from 'react';
 import { connect } from 'react-redux'
-import { Col } from 'reactstrap';
 
-import { NotebookServerOptions, NotebookServers, LogOutUser } from './Notebooks.present';
+import { NotebookServers } from './Notebooks.present';
+import { StartNotebookServer as StartNotebookServerPresent } from './Notebooks.present';
 
 import NotebooksModel from './Notebooks.state';
 import { Notebooks as NotebooksPresent } from './Notebooks.present';
+import { StatusHelper } from '../model/Model'
 
-class LaunchNotebookServer extends Component {
+/**
+ * Displays a start page for new Jupiterlab servers.
+ * 
+ * @param {Object[]} branches   Branches as redurnet by gitlab "/branches" API
+ * @param {function} refreshBranches   Function to invoke to refresh the list of branches
+ * @param {number} projectId   id of the reference project
+ * @param {number} projectPath   path of the reference project
+ * @param {Object[]} client   api-client used to query the gateway
+ * @param {string} successUrl  Optional: url to redirect when then notebook is succesfully started
+ * @param {Object} history  Optional: used with successUrl to properly set the new url without reloading the page
+ */
+class StartNotebookServer extends Component {
   constructor(props) {
-    super(props)
-    this.state = {
-      serverOptions: {},
-      serverRunning: false,
-      serverStarting: false,
-      doLogOut: false
-    };
-    this._unmounting = false;
-  }
+    super(props);
+    this.model = new NotebooksModel(props.client);
 
-  componentWillUnmount() {
-    this._unmounting = true;
+    this.handlers = {
+      refreshBranches: this.refreshBranches.bind(this),
+      refreshCommits: this.refreshCommits.bind(this),
+      setBranch: this.setBranchFromName.bind(this),
+      setCommit: this.setCommitFromId.bind(this),
+      toggleMergedBranches: this.toggleMergedBranches.bind(this),
+      setDisplayedCommits: this.setDisplayedCommits.bind(this),
+      setServerOption: this.setServerOptionFromEvent.bind(this),
+      startServer: this.startServer.bind(this)
+    }
+
+    this.state = { 
+      first: true,
+      startring: false
+    };
   }
 
   componentDidMount() {
-    if (this.state.doLogOut === false)
-      this.componentDidUpdate()
+    this._isMounted = true;
+    this.startNotebookPolling();
+    this.refreshBranches();
   }
 
-  componentDidUpdate() {
-    if (this.state.doLogOut === true) return; // We are about to refresh anyway...
+  componentWillUnmount() {
+    this.stopNotebookPolling();
+    this._isMounted = false;
+  }
 
-    if (this.props.core.notebookServerAPI !== this.previousNotebookServerAPI) {
-      this.serverStatusSet = false;
-      this.serverOptionsSet = false;
+  componentDidUpdate(previousProps) {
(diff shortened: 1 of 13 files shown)


lorenzo-cavazzi:
The second commit adds a `Legacy start server` button with the old "new notebook" page. It's meant to simplify finishing the development, please remember to reset it

lorenzo-cavazzi:
This is now ready for the review

src/notebooks/Notebooks.container.js
```diff
@@ -18,163 +18,262 @@
...
+    }
+
+    this.setBranch({});
+    this.setCommit({});
+    return false;
+  }
+
+  refreshCommits(updatedBranch) {
+    const commits = this.model.get("data.commits");
+    if (!StatusHelper.isUpdating(commits)) {
```
ciyer: I think the following is easier to read
```
if (!StatusHelper.isUpdating(commits)) return;
...

File: src/notebooks/Notebooks.state.js [hunk 2]
```diff
@@ -40,27 +61,167 @@ class NotebooksModel extends StateModel {
...
+  verifyIfRunning(projectId, projectPath, servers) {
+    const notebooks = servers ?
+      servers :
+      this.get('notebooks.all');
+    const branch = this.get('filters.branch');
+    const commit = this.get('filters.commit');
+    for (let notebookName of Object.keys(notebooks)) { 
+      const notebook = notebooks[notebookName];
+      const annotations = cleanAnnotations(notebook["annotations"], "renku.io");
+      if (parseInt(annotations.projectId) === projectId) {
```
ciyer: The deep nesting of `if`s gets difficult to follow. I suggest:
```
if (parseInt(annotations.projectId) !== projectId) continue;
```

src/notebooks/Notebooks.state.js
```diff
@@ -40,27 +61,167 @@ class NotebooksModel extends StateModel {
...
+        status: false,
+        url: null
+      }});
+    }
     const oldPoller = this.get('notebooks.polling');
     if (oldPoller == null) {
       const newPoller = setInterval(() => {
-        return this.fetchNotebooks();
+        const fetchPromise = this.fetchNotebooks();
+        if (verifyRunningId) {
```
ciyer: Extract a function/method with this code, since it is duplicated
```
    if (verifyRunningId) {
         return fetchPromise.then((servers) => {
             return this.verifyIfRunning(verifyRunningId, projectPath, servers);
           });
         }
         else {
           return fetchPromise;
         }
```

ciyer:
First of all -- this is a great feature! I'm looking forward to making use of it! I should have tested it earlier, and I'm sorry these comments are coming so late.

1. Can the branch and commit be defaulted to master and the most recent commit, respectively? Those will be the most common choices and the interaction would be greatly streamlined with defaults there. (Branch is defaulted on Chrome but not on Safari...)
2. The branch and commit settings do not work on Safari/Mac. They *do* work on Chrome and Firefox, so I'm not sure what is going on there.
3. The notebook server does not seem to actually be at the requested commit. See the screenshot below. The URL indicated that we should be at commit c735d545, but `git log` shows that we are at a later commit.

image

rokroskar:
> The URL indicated that we should be at commit c735d545, but git log shows that we are at a later commit.

was there an autosave branch? That functionality may not take starting from arbitrary branches/commits into account properly.

src/notebooks/Notebooks.container.js
lorenzo-cavazzi: Ok, guess you meant without the `!`, so instead of having
```
if (!StatusHelper.isUpdating(commits)) {
    // code here
}
```
it's more readable to have this
```
if (StatusHelper.isUpdating(commits)) return;
// code here
```

ciyer: Yes, you are right: that's what I meant.

ciyer:
No, it was on master. I tried again with the latest version of the notebook component (renku/renku-notebooks:0.4.0-9d1a78b), but I still end up at the latest commit, even if I ask for an earlier one.

rokroskar:
yes, you started from master but was there an autosave branch present?

ciyer:
Yes, there was an autosave branch.

lorenzo-cavazzi:
I updated the preview in https://lorenzotest.dev.renku.ch/

1. I have added the "auto select" logic to automatically select "master" as the default branch and the latest commit.
Please mind that switching between branches has a different logic: it keeps the selected commit if it's available also on the other branch, otherwise it is reset (this doesn't happen if no commit is selected).
I prefer this way because it is more explicit. Since everything is now auto-selected, a user may first change the commit, then the branch, and it's confusing if the last action affects also the commit. I am open to discuss this if anyone disagrees.
2. I have changed the branch and commit selection logic a little bit, I hope this could help to address the problem on safari. Let me know
3. @rokroskar I totally forgot about this during our last discussion in Bern. In the user interaction (pseudo)flowchart, we should at some point check if there is an autosaved branch and ask the user what to do (start from there/discard). The reference issue is #429 .
4. That error popup error comes form the component `<UncontrolledTooltip>` in reactstrap, there is an issue on their repo to fix it reactstrap/reactstrap#1255 . It happens when clicking on the refresh icon for branches/commits. We could live with that until it's solved or switch to the standard `<Tooltip>` by manually implementing the logic.

P.S. At the moment there is a limitation: if a notebook server starts for the combination `(branch1, commit1)`, it's not possible to start another one for `(branch2, commit1)`. Will be solved by SwissDataScienceCenter/renku-notebooks#186.

rokroskar:
@lorenzo-cavazzi the autosave problem that @ciyer encountered is a bug - the script that recovers from the autosave I believe only checks that it was made for the branch but not also for the commit. So if you have an autosave branch for `master` for a different commit, it will always reset you there. 

Still, we should probably ask the user what they want to do with the autosave data...

PR #476, opened 2019-05-29
chart cleanup

ableuler:


ciyer:
This would still be nice to have, but the PR is hopelessly out-of-date now, so I will close and a new PR should be re-opened to address this.

PR #485, opened 2019-06-05
fix: the interval of the notebook server is lowered when servers are ready

vfried:
The intervals of the notebook server start at 3000,
when the servers are all ready the polling is set to 6000
if they are unready again for reason they are seat to 3000 again.

On top of that, if there are no servers all the polling stops.

### DIFF
--- a/src/project/Project.js
+++ b/src/project/Project.js
@@ -29,7 +29,7 @@ import { connect } from 'react-redux'
 import { StateKind } from '../model/Model';
 import Present from './Project.present'
 
-import { ProjectModel, GraphIndexingStatus } from './Project.state'
+import { ProjectModel, GraphIndexingStatus, PollingInterval } from './Project.state'
 import Ku from '../ku/Ku'
 import { FileLineage } from '../file'
 import { ACCESS_LEVELS } from '../api-client';
@@ -71,7 +71,7 @@ class View extends Component {
   async fetchBranches() { return this.projectState.fetchBranches(this.props.client, this.props.id); }
   async fetchCIJobs() { return this.projectState.fetchCIJobs(this.props.client, this.props.id); }
   async startNotebookServersPolling() {
-    return this.projectState.startNotebookServersPolling(this.props.client, this.props.id);
+    return this.projectState.startNotebookServersPolling(this.props.client, this.props.id,PollingInterval.START);
   }
   async stopNotebookServersPolling() { return this.projectState.stopNotebookServersPolling(); }
   async stopNotebookServer(serverName) { return this.projectState.stopNotebookServer(this.props.client, serverName); }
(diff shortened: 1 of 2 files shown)


rokroskar:
can you also prevent a new request from being sent if the previous one has not received a response?

src/project/Project.js
```diff
@@ -71,7 +71,7 @@ class View extends Component {
   async fetchBranches() { return this.projectState.fetchBranches(this.props.client, this.props.id); }
   async fetchCIJobs() { return this.projectState.fetchCIJobs(this.props.client, this.props.id); }
   async startNotebookServersPolling() {
-    return this.projectState.startNotebookServersPolling(this.props.client, this.props.id);
+    return this.projectState.startNotebookServersPolling(this.props.client, this.props.id,3000);
```
ciyer: Could we keep the polling durations as constants somewhere? E.g., 
```
const PollingDuration = {
  STARTING: 3000,
  RUNNING: 60000,
  ...
}
```
This would then be

```
this.projectState.startNotebookServersPolling(this.props.client, this.props.id, PollingDuration.STARTING);
```

src/project/Project.state.js
```diff
@@ -234,6 +252,19 @@ class ProjectModel extends StateModel {
     }
     return client.getNotebookServers(id)
       .then(resp => {
+        const serverNames = Object.keys(resp.data).sort();
+        if(serverNames && serverNames.length > 0){
+          const allServersReady = serverNames.map((k, i) =>
+            resp.data[k].ready
+          ).filter(ready => ready === false).length === 0;
+          if(allServersReady && this.get('notebooks.pollingInterval')!== 60000){
```
ciyer: Following note above about the magic numbers, this should be
```
const pollingInterval = this.get('notebooks.pollingInterval');
if (allServersReady &&  pollingInterval !== PollingDuration.RUNNING) {
  this.changeNotebookServerPollingInterval(client, id, PollingDuration.RUNNING);
}
```

src/project/Project.state.js
```diff
@@ -234,6 +252,19 @@ class ProjectModel extends StateModel {
...
+            this.changeNotebookServerPollingInterval(client, id, 60000);
+          } else if(!allServersReady && this.get('notebooks.pollingInterval')!== 3000){
```
ciyer:
```
else if (!allServersReady &&  pollingInterval !== PollingDuration.STARTING) {
  this.changeNotebookServerPollingInterval(client, id, PollingDuration. PollingDuration.STARTING);
}

src/project/Project.js
vfried: Done

src/project/Project.state.js
vfried: Done

ableuler:
I think a more effective fix for the problem we've seen would be to trigger the next poll with a set timeout in the promise resolution part of the previous call.

ciyer:
We want to switch to something event driven and move away from polling, but we wanted a quick fix to relieve the pressure off JupyterHub for the moment.

Issue #408, opened 2019-03-28
Project Search "Your Projects" not updated


ciyer:
Create a new project in Renku, then go to the Projects page. This new project is not listed under "Your Projects". It seems like the list of your projects is cached. It should be updated when a new project is created.

lorenzo-cavazzi:
Currently, the lists of "starred" and "yours" projects are stored in the redux `renku` store inside the `user` object, while the list of all projects is coming from `ProjectListModel` in `pages` object. This distinction is useful to keep track of projects that are important to the user and display information on the landing page, but at the moment it leads to inconsistency between data and multiple react components with slightly different code.

I suggest addressing the problem in 3 steps
* Add a `refreshPersonalProject` function in `renku.user` store to be invoked by any code that modifies started or personal projects (enough to close this issue).
* Move `user.starredProjects` and `user.memberProjects` to `projects.starred` and `projects.member`. Even if these data change when the user logs on and off, they are not user's properties and they don't usually refresh with user's data.
* Uniform the components in `/projects` page so that changing between tabs simply change the projects list that is fed to the react Component (longer, should be moved to a new issue)

emmjab:
I ran into this bug with a user; I think this important to solve because someone could think their project didn't get created, and try to create it again.

Is it possible to solve this by forcing a refresh of the page when the user hits the "create project" button?

lorenzo-cavazzi:
Refreshing the page wouldn't be the best approach to solve this kind of problem.
Basically, we get and store information about projects in 2 different places, but we refresh only one of them when the user performs a number of project-related actions.

Thanks for your comment, it's good to be aware that users feel particularly confused by this. We will try to solve the issue soon.

PR #486, opened 2019-06-07
feat: button to fork a project

vfried:
New button added to fork a project in the project page.

image

When the button is clicked a modal opens and gives the user the option to choose a namespace (like in the project creation) --> we could also give the user the option to set a new name for the forked project.

image

The submit action is still not finished, it throws some errors (i think the project is not forked properly or maybe we need to do a first commit and set some other things like in the project creation) 

==> I did this before going on holidays and a bit in a rush, I reused lot's of the code of the create project and it could happen that there are redundant things or things that are not right.

### DIFF
--- a/src/api-client/project.js
+++ b/src/api-client/project.js
@@ -164,6 +164,54 @@ function addProjectMethods(client) {
   }
 
 
+  client.forkProject = (projectMeta) => {
+    const gitlabProject = {
+      id: projectMeta.id
+    };
+    if (projectMeta.projectNamespace != null) gitlabProject.namespace_id = projectMeta.projectNamespace.id;
+    const headers = client.getBasicHeaders();
+    headers.append('Content-Type', 'application/json');
+
+    let createGraphWebhookPromise;
+    const newProjectPromise = client.clientFetch(`${client.baseUrl}/projects/${projectMeta.id}/fork`, {
+      method: 'POST',
+      headers: headers,
+      body: JSON.stringify(gitlabProject)
+    })
+      .then(resp => {
+        if (!projectMeta.optoutKg) {
+          createGraphWebhookPromise = client.createGraphWebhook(resp.data.id);
+        }
+        client.clientFetch(`${client.baseUrl}/projects/${resp.data.id}/triggers` , {
+          method: 'POST',
+          headers: headers,
+          body: JSON.stringify({
+            description: `Automatic Fork trigger`
+          })
+        }).then(trigger => { client.clientFetch(`${client.baseUrl}/projects/${resp.data.id}/trigger/pipeline`, {
+          method: 'POST',
+          headers: headers,
+          body: JSON.stringify({
+            ref:'master',
+            token: trigger.data.token
+          })
+        });
+        });
+        return resp;
+      });
+
+    let promises = [newProjectPromise];
+    if (createGraphWebhookPromise) {
+      promises = promises.concat(createGraphWebhookPromise);
+    }    
+    return Promise.all(promises)
+      .then((results) => {
+        if (results.errorData)
+          return Promise.reject(results);
+        return Promise.resolve(results).then(() => results[0].data);
+      });
+  }
+
   client.setTags = (projectId, name, tags) => {
     return client.putProjectField(projectId, name, 'tag_list', tags);
   }
--- a/src/project/Project.present.js
+++ b/src/project/Project.present.js
@@ -40,7 +40,7 @@ import { Card, CardBody, CardHeader } from 'reactstrap';
 
 import FontAwesomeIcon from '@fortawesome/react-fontawesome'
 import faStarRegular from '@fortawesome/fontawesome-free-regular/faStar'
-import { faStar as faStarSolid, faInfoCircle, faExternalLinkAlt } from '@fortawesome/fontawesome-free-solid'
+import { faStar as faStarSolid, faInfoCircle, faExternalLinkAlt, faCodeBranch } from '@fortawesome/fontawesome-free-solid'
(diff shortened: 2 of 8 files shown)


rokroskar:
This will be a nice addition - one thing to consider is to trigger the CI pipeline immediately after cloning, otherwise the user has to do it manually by going to gitlab.

src/project/Project.present.js
```diff
@@ -251,6 +253,17 @@ class ProjectViewHeaderOverview extends Component {
                     placeholder={system.star_count} aria-label="starCount" readOnly={true} />
                 </form>
               </div>
+              <div className={`fixed-width-100 pr-1`}>
```
lorenzo-cavazzi: adding this element result in a misalignment of the whole container when the window size is close to small (check the screenshot on the right)

Screenshot from 2019-06-24 16-11-19

It's probably just a matter of changing lines 237 - 243 (with both at `xs={12} md={6}` is already better)

src/api-client/project.js
```diff
@@ -164,6 +164,54 @@ function addProjectMethods(client) {
   }
 
 
+  client.forkProject = (renkuProject) => {
```
lorenzo-cavazzi: Passing the single values here instead of the whole object would probably make it more readable

src/project/Project.present.js
vfried: Good fix, i will apply it now. Thanks!

src/api-client/project.js
vfried: I decided to pass only the project meta, since it's what I use. But since there is the project that you will fork and then the new project i think that passing only the attributes might make it a bit more confusing.

lorenzo-cavazzi: On second thought, I still think it makes sense to separate the single parameters because the `api-clients` shouldn't be aware of the model.

The advantages would be
* being independent of the code in `ProjectFork.container.js` -- now we have to change the api call if we change the model there
* having a reusable call to the fork api

`forkProject` uses only 3 properties from the argument object `projectMeta`, by using meaningful names the readability should not become worse :)

src/project/Project.present.js
```diff
@@ -251,6 +253,17 @@ class ProjectViewHeaderOverview extends Component {
...
+                <form className="input-group input-group-sm">
+                  <div className="input-group-prepend">
+                    <button className="btn btn-outline-primary" onClick={this.props.toogleForkModal}>
```
lorenzo-cavazzi: I was wondering if it would make sense to create a `<ForkButton>` component and move all the ForkModal related logic in the "ProjectFork" pages instead of saving some data in the "Project" pages (like the `forkModalOpen` in the "Project.state.js"). Or is there a reason why it's there?

ciyer:
Great improvement! 

How difficult are the following changes to implement?

1. Show the full name of the project (namespace and slug) that will be created when the fork is done. Similar to how it is on the New Project page. 

Compare:

image

image

2. Changing the "Project Home" in the Fork Project dialog causes a flash and redraws the dialog. Is it possible to avoid this?

3. The fork counter does not update when a project is forked, though this problem affects GitLab as well, so it may not be something that can be fixed in the UI.

4. The problem described in #408 also affects this feature, but that is probably a separate PR to fix for all situations.

Overall, though, a very useful new feature.

ciyer:
Also -- are images built after a project is forked?

rokroskar:
kicking off the CI-pipeline: https://docs.gitlab.com/ee/api/pipelines.html#create-a-new-pipeline

src/project/fork/ProjectFork.container.js
```diff
@@ -0,0 +1,178 @@
...
+import { StateKind, StateModel } from '../../model/Model';
+import { newProjectSchema } from '../../model/RenkuModels';
+import ForkProjectModal from './ProjectFork.present'
+
+
+class Fork extends Component {
+  constructor(props) {
+    super(props);
+    this.forkProject = new StateModel(newProjectSchema, StateKind.REDUX);
+    this.state = { statuses: [], namespaces: [], namespaceGroup: null, namespacesFetched: false , id: props.projectId };
```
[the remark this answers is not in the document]
vfried: I did it like this because I reused(copied) the code of create project but you are right, it would be better to only have one state so I decided to use redux.

src/project/Project.present.js
vfried: I think the info of wether the fork modal is visible or not has to be and controlled in the project page since there is where the modal is shown.

vfried:
1. Show the full name of the project (namespace and slug) that will be created when the fork is done. Similar to how it is on the New Project page. ==> FIXED

2. Changing the "Project Home" in the Fork Project dialog causes a flash and redraws the dialog. Is it possible to avoid this? ==> FIXED

3. The fork counter does not update when a project is forked, though this problem affects GitLab as well, so it may not be something that can be fixed in the UI. ==> I think this is a gitlab problem (it looks like the fork count is cached) I found a closed bugfix for this but it could be that the problem is still there... https://gitlab.com/gitlab-org/gitlab-ee/issues/8993 (i will try to investigate a bit more on this though)

4. The problem described in #408 also affects this feature, but that is probably a separate PR to fix for all situations. ==> I AGREE, we should fix this in a new pr

src/project/Project.present.js
lorenzo-cavazzi: ok fine, no need to change this here.

In general, I find useful to use this rule of thumb: if the information is not needed anywhere else on the parent page, move it to the child. In this case, no other part of the project page benefits from knowing if the modal is shown, and keeping it there makes the component less re-usable -- if we want to add the fork button anywhere else, we would need to also copy some code from the project pages... but in this case it's unlikely we will use it somewhere else :)

vfried:
The build image thing it's already done in this PR.

PR #498, opened 2019-06-28
Update Notebooks components

lorenzo-cavazzi:
This is the UI adaptation to handle the API responses from SwissDataScienceCenter/renku-notebooks#177 (requires also SwissDataScienceCenter/renku-notebooks#187)

A preview is available here: https://lorenzotest.dev.renku.ch

At the moment we have lost the distinction between "starting" and "stopping" servers, we only mark them as "pending". This may be acceptable, but the UI shows that after the stop API is invoked, the server will still be considered "ready" for a while.
An idea to improve this could be using an extra annotation to save in the pod the information that the stop command has been invoked by the user, then returning it in the `/servers` endpoint in the status.

### DIFF
--- a/src/notebooks/Notebooks.present.js
+++ b/src/notebooks/Notebooks.present.js
@@ -27,7 +27,7 @@ import { UncontrolledTooltip, UncontrolledPopover, PopoverHeader, PopoverBody } 
 import FontAwesomeIcon from '@fortawesome/react-fontawesome';
 import { faStopCircle, faExternalLinkAlt, faInfoCircle, faSyncAlt, faCogs } from '@fortawesome/fontawesome-free-solid';
 
-import { SpecialPropVal, StatusHelper } from '../model/Model';
+import { StatusHelper } from '../model/Model';
 import { Loader, InfoAlert, ExternalLink } from '../utils/UIComponents';
 import Time from '../utils/Time';
 import Sizes from '../utils/Media';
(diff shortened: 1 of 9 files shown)


src/notebooks/Notebooks.present.js
```diff
@@ -585,24 +599,18 @@ class StartNotebookOptions extends Component {
...
-            <Label>A Jupyterlab server is already starting, please wait...</Label>
-          </FormGroup>
-        );
-      }
-      else if (status === "stop") {
+      else if (status === "pending") {
         content = (
           <FormGroup>
-            <Label>A Jupyterlab server is stopping, please wait...</Label>
+            <Label>A Jupyterlab server is starting or terminating, please wait...</Label>
```
ciyer: Reword to: `A JupyterLab server for this commit is starting or terminating, please wait...`

src/notebooks/Notebooks.present.js
```diff
@@ -585,24 +599,18 @@ class StartNotebookOptions extends Component {
...
           </FormGroup>
         );
       }
       else {
         content = (
           <FormGroup>
-            <Label>A Jupyterlab server is already running but it is currently not available. Please wait...</Label>
+            <Label>A Jupyterlab server is already running but it is currently not available.
```
ciyer: Missing a comma: "A Jupyterlab server is already running*,* but it is currently not available."

lorenzo-cavazzi:
I have updated the text in the 2 messages and aligned all `Jupyterlab` to the Pascal case version `JupyterLab`

lorenzo-cavazzi:
I realized I forgot to move the project id filtering logic from "Projects" pages to the "notebook-servers" API. Now it has been updated and re-deployed to https://lorenzotest.dev.renku.ch

This logic could be simplified in the future using the new notebook service API from SwissDataScienceCenter/renku-notebooks#188

PR #500, opened 2019-07-01
Prevent multiple simultaneous calls to the servers endpoint

lorenzo-cavazzi:
Avoid overlapping queries to the `/servers` endpoint by implementing a very basic semaphore logic.

The first commit comes from #498, merge after that.

### DIFF
--- a/src/notebooks/Notebooks.state.js
+++ b/src/notebooks/Notebooks.state.js
@@ -32,7 +32,8 @@ const notebooksSchema = new Schema({
       polling: {initial: null},
       all: {initial: {}},
       status: {initial: null},
-      url: {initial: null}
+      url: {initial: null},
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi:
Rebased and ready to test. Preview available here: https://lorenzotest.dev.renku.ch

To test it, go to any page where notebooks are visualized (e.g. https://lorenzotest.dev.renku.ch/notebooks), open the network tab and try to slow down the connection so that a single response from the `/servers` API will take more than 3 seconds to come (e.g. from the chrome debugger, select a custom device with network throttling very bad).

It's actually easier to test by setting the `POLLING_INTERVAL` variable down to a very low value (e.g. 300 instead of 3000), but this requires to checkout the branch locally and running it through telepresence.

PR #501, opened 2019-07-02
feat: dropdown in project search to order projects

vfried:
With dropdown in project search to order projects by name, created date, and updated date. 

With sorting included (Ascending - Descending) --> I'm not sure this button icons are the best, if you have better ones i can change them.

image

### DIFF
--- a/src/project/list/ProjectList.container.js
+++ b/src/project/list/ProjectList.container.js
@@ -22,6 +22,12 @@ import ProjectList from './ProjectList.present'
 import ProjectListModel from './ProjectList.state'
 import qs from 'query-string';
 
+const orderByValuesMap = {
+  NAME: 'name',
+  CREATIONDATE: 'created_at',
+  UPDATEDDATE: 'last_activity_at'
+};
+
 class List extends Component {
   constructor(props) {
     super(props);
@@ -31,23 +37,28 @@ class List extends Component {
     this.handlers = {
       onSearchQueryChange: this.onSearchQueryChange.bind(this),
       onSearchSubmit: this.onSearchSubmit.bind(this),
-      onPaginationPageChange: this.onPaginationPageChange.bind(this)
+      onPaginationPageChange: this.onPaginationPageChange.bind(this),
+      onOrderByDropdownToogle: this.onOrderByDropdownToogle.bind(this),
+      changeSearchDropdownOrder: this.changeSearchDropdownOrder.bind(this),
+      toogleSearchSorting: this.toogleSearchSorting.bind(this) 
     };
   }
 
   componentDidMount() {
     this.model.set('perPage', this.perPage);
-    const {query, pageNumber, pathName} = this.getUrlSearchParameters(this.props.location);
+    const {query, pageNumber, pathName, orderBy, orderSearchAsc} = this.getUrlSearchParameters(this.props.location);
     this.model.setQuery(query);
     this.model.setPathName(pathName);
     this.model.setPage(pageNumber);
-
+    this.model.setOrderDropdownOpen(false);
+    this.model.setOrderBy(orderBy);
+    this.model.setOrderSearchAsc(orderSearchAsc);
     // save listener to remove it when unmounting the component
     // TODO: this could be removed if onPaginationPageChange/this.props.history.push worked
     //    also when only the search part changed
     const listener = this.props.history.listen(location => {
-      const {query, pageNumber, pathName} = this.getUrlSearchParameters(location);
-      this.onUrlParametersChange(query, pageNumber, pathName);
+      const {query, pageNumber, pathName, orderBy, orderSearchAsc} = this.getUrlSearchParameters(location);
+      this.onUrlParametersChange(query, pageNumber, pathName, orderBy, orderSearchAsc);
     });
     this.setState({listener});
   }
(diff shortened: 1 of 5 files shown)


lorenzo-cavazzi:
I was thinking if it makes sense to change the "Order By" text on the button with the current chosen element. While using it, I noticed it is not instantly clear which is the ordering element. But I am not sure if this will make the button functionality less obvious

src/project/list/ProjectList.container.js
```diff
@@ -70,50 +75,79 @@ class List extends Component {
...
     this.props.history.push(newUrl);
   }
 
+  onOrderByDropdownToogle() {
+    this.model.set('orderByDropdownOpen', !this.model.get('orderByDropdownOpen'));
+  }
+
+  changeSearchDropdownOrder(e){
+    this.model.set('orderBy', e.target.value);
+    this.props.history.push(this.urlFromQueryAndPageNumber(this.model.get('query'), 1 , this.model.get('pathName'), this.model.get('orderBy'), this.model.get('orderSearchAsc') ))
```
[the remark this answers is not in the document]
vfried: Done

src/project/list/ProjectList.container.js
```diff
@@ -70,50 +75,79 @@ class List extends Component {
...
     const pathName = this.model.get('pathName');
-    const newUrl = this.urlFromQueryAndPageNumber(query, newPageNumber, pathName);
+    const orderBy= this.model.get('orderBy');
+    const orderSearchAsc= this.model.get('orderSearchAsc');
+    const newUrl = this.urlFromQueryAndPageNumber(query, newPageNumber, pathName, orderBy, orderSearchAsc);
     this.props.history.push(newUrl);
   }
 
+  onOrderByDropdownToogle() {
+    this.model.set('orderByDropdownOpen', !this.model.get('orderByDropdownOpen'));
```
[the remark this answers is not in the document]
vfried: Done

src/project/list/ProjectList.container.js
```diff
@@ -70,50 +75,79 @@ class List extends Component {
...
 
-  urlFromQueryAndPageNumber(query, pageNumber , pathName) {
-    return `${pathName}?q=${query}&page=${pageNumber}`
+  orderByValuesMap(){
+    return {
+      NAME: 'name',
+      CREATIONDATE: 'created_at',
+      UPDATEDDATE: 'last_activity_at'
+    }
+  }
```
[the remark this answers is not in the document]
vfried: I did it as a function to be consistent on the way we do things, since the url map is done the other way i thought it would be better to do it like that but i just changed it to a constant. In case you think i should revert it i can.

vfried:
> I was thinking if it makes sense to change the "Order By" text on the button with the current chosen element. While using it, I noticed it is not instantly clear which is the ordering element. But I am not sure if this will make the button functionality less obvious

I thought about this too and it could be done but I decided to do it this way because as you said it wouldn't be clear what the dropdown does otherwise. I also tryied putting order by and the selected action but it was too much text.

I think if in the future we have aditional search functionalities/filters we could improve the way it looks by transforming this into a full search bar line or something like what google does.

src/project/list/ProjectList.container.js
```diff
@@ -70,50 +81,95 @@ class List extends Component {
...
+    const newUrl = this.urlFromQueryAndPageNumber(query, newPageNumber, pathName, orderBy, orderSearchAsc);
     this.props.history.push(newUrl);
   }
 
+  onOrderByDropdownToogle() {
+    this.model.setOrderDropdownOpen(!this.model.get('orderByDropdownOpen'));
+  }
+
+  changeSearchDropdownOrder(e){
+    this.model.set('orderBy', e.target.value);
```
lorenzo-cavazzi: similarly to `this.model.setOrderDropdownOpen` at line 119, I would also change this to `this.model.setOrderBy`

src/project/list/ProjectList.container.js
```diff
@@ -70,50 +81,95 @@ class List extends Component {
...
+        1,
+        this.model.get('pathName'),
+        this.model.get('orderBy'),
+        this.model.get('orderSearchAsc')
+      )
+    )
+  }
+
+  toogleSearchSorting(){
+    this.model.set('orderSearchAsc', !this.model.get('orderSearchAsc'));
```
lorenzo-cavazzi: similarly to `this.model.setOrderDropdownOpen` at line 119, I would also change this to `this.model.setOrderSearchAsc`

src/project/list/ProjectList.present.js
```diff
@@ -160,7 +183,13 @@ class ProjectsSearch extends Component {
           <span></span>
       }
       <Col md={8}>
```
lorenzo-cavazzi: If you resize your browser (width), the `Search` button will go on a new line alone (see screenshot below). If you resize it more, it will move again to the right pace, and resizing further will move it to the new line again. We can skip this intermediate step my changing the `md={8}` to something like `xs={12}` (or even removing it)

Screenshot from 2019-07-03 14-25-02

lorenzo-cavazzi:
> I thought about this too and it could be done but I decided to do it this way because as you said it wouldn't be clear what the dropdown does otherwise. I also tryied putting order by and the selected action but it was too much text.
> 
> I think if in the future we have aditional search functionalities/filters we could improve the way it looks by transforming this into a full search bar line or something like what google does.

Right, it makes sense to leave it as it is now and improve it in the future when we need

src/project/list/ProjectList.present.js
vfried: Good point, fixed now.

src/project/list/ProjectList.container.js
vfried: Fixed

src/project/list/ProjectList.container.js
vfried: Fixed

ciyer:
Very nice implementation!

Would it be possible to sort "Your Projects" and "Starred Projects" by update date?

vfried:
> Very nice implementation!
> 
> Would it be possible to sort "Your Projects" and "Starred Projects" by update date?

Done!

BTW: i tested and the pipeline works for me. ==> I first have to create a trigger to the use the trigger token to run the pipeline. (I cannot start a pipeline without a token).

vfried:
> Working for me. It probably needs to be rebased before it can be merged.
> 
> Would it make sense (an be easy) to add this search also to starred and personal projects?

It makes sense but it wouldn't be that easy... since the project search and the starred/your projects are different components. So we would need to do this in a separate issue/pr.

PR #506, opened 2019-07-04
feat: added option to create empty project

### LINKED ISSUE #504: allow creation of an empty project
The "new project" page should include an option to create a project but skip the initialization step. There could be a check box at the bottom with text like "Do not initialize the renku project" or similar.

vfried:
A new option was added to the templates dropdown to create an empty project. 

image

Closes #504

### DIFF
--- a/src/api-client/project.js
+++ b/src/api-client/project.js
@@ -118,6 +118,8 @@ function addProjectMethods(client) {
     });
   }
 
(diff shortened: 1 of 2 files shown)


vfried:
Fixed the problem with the other templates. I tested and all the templates worked for me.

Issue #508, opened 2019-07-08
Fork project not forking to a group


vfried:
When you try to fork a project to a group it forks it to your username instead of the selected group.

PR #509, opened 2019-07-08
fix: fork to project works for groups

vfried:
There was an error on the JSON body sent on the POST request, it was set to "namespace_id" but it should have been "namespace".


rokroskar:
Shouldn't this close #508?

PR #512, opened 2019-07-08
feat: pipeline starts without creating a trigger

### LINKED ISSUE #511: Start pipeline without creating trigger (after project fork)
Run pipeline after fork without creating a trigger. 

As... https://docs.gitlab.com/ee/api/pipelines.html#create-a-new-pipeline

vfried:
The pipeline starts now without creating a trigger.

To test fork a project and go to CI/CD => pipelines in gitlab and you will see that after forking a project there is a pipeline running. 

For some reason it fails for old projects. I think in the future we can after redirect offer the option to restart failed pipelines (Like for Knowledge Graph).

Closes #511

### DIFF
--- a/src/api-client/project.js
+++ b/src/api-client/project.js
@@ -192,8 +192,79 @@ function addProjectMethods(client) {
     }
   }
 
+  client.getProjectStatus = (projectId) =>{
+    const headers = client.getBasicHeaders();
+    headers.append('Content-Type', 'application/json');
+    return client.clientFetch(`${client.baseUrl}/projects/${projectId}/import`, {
+      method: 'GET',
+      headers: headers
+    }).then(resp => {
+      return resp.data.import_status;
+    }).catch((error) => "error");
+  } 
 
-  client.forkProject = (projectMeta) => {
+  function runPipeline(projectId){
+    const headers = client.getBasicHeaders();
+    headers.append('Content-Type', 'application/json');
+    return  client.clientFetch(`${client.baseUrl}/projects/${projectId}/pipeline`, {
+      method: 'POST',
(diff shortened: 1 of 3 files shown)


rokroskar:
> For some reason it fails for old projects.

how old is old?

vfried:
> > For some reason it fails for old projects.
> 
> how old is old?

I'm not sure, I left the office and couldn't finish investigating on the issue. I was planning on trying to check tomorrow what's going on.

vfried:
> > > For some reason it fails for old projects.
> > 
> > 
> > how old is old?
> 
> I'm not sure, I left the office and couldn't finish investigating on the issue. I was planning on trying to check tomorrow what's going on.

I realised some forks where failing because I was trying to start the pipeline before the project was fully  forked. For this reason I now check project import status as stated here: https://docs.gitlab.com/ee/api/project_import_export.html#import-status 

I set up a timer and try 10 times to see if the pipeline is ready to be run or not (So far it was always successful on the 1st time).

I think the older projects where taking longer to being fetched or something and this is why they where failing.

ciyer:
I tried it out; the feature is very nice and the code looks good! I agree, though, that the feedback on the modal dialog should be more explicit: as you suggest, the fork and cancel buttons should be disabled and a short description of what's going on should be there.

vfried:
> I tried it out; the feature is very nice and the code looks good! I agree, though, that the feedback on the modal dialog should be more explicit: as you suggest, the fork and cancel buttons should be disabled and a short description of what's going on should be there.

Done... now buttons are disabled and there is a message as in the following image when the project is being forked.

image

Issue #256, opened 2018-07-03
Add interactivity to lineage graph


ciyer:
- [x] Clicking on a node should take you to the lineage view for that node
<img width="708" alt="image" src="https://user-images.githubusercontent.com/1196411/54537822-798fd400-4993-11e9-8c64-b3e86b052a1e.png">

( compare with current display where it is not clear what is current and what is past )
<img width="1175" alt="image" src="https://user-images.githubusercontent.com/1196411/55553659-12875480-56e1-11e9-85b4-4d333a35f931.png">

rokroskar:
should we close this and make a separate issue for seeing different commits? I think it's quite a different beast to tackle. @vfried @lorenzo-cavazzi ?

rokroskar:
or we can make an epic out of this since I think seeing and navigating commits might take several issues

vfried:
I think we can make a new epic with new issues for it

PR #519, opened 2019-07-15
feat: user is redirected on click to a node

vfried:
When a user clicks on a node of type "blob" (file) the user is redirected to the lineage of the clicked file.

This addresses the first point of the issue #256

### DIFF
--- a/src/file/Lineage.present.js
+++ b/src/file/Lineage.present.js
@@ -85,6 +85,11 @@ class FileLineageGraph extends Component {
   componentDidUpdate() {
     this.renderD3();
   }
+  
+  hasLink(nodeId, centralNode){
+    return this.props.graph.nodes
+      .filter(function(node) { return (node.id === nodeId && node.id !== centralNode && node.type === "blob") })[0];
+  }
 
   renderD3() {
     // Create the input graph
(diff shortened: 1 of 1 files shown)


src/file/Lineage.present.js
```diff
@@ -85,6 +85,16 @@ class FileLineageGraph extends Component {
   componentDidUpdate() {
     this.renderD3();
   }
+  
+  hasLink(nodeId, centralNode){
+    return this.props.graph.nodes
+      .filter(function(node) { return (node.id === nodeId && node.id !== centralNode) })
+      .filter((node)=> { return node.type === "blob" })[0];
```
lorenzo-cavazzi: It's probably not a big step forward, but performance-wise we could use a single filter by adding the second condition to the first function
```
.filter(function(node) { return (node.id === nodeId && node.id !== centralNode && node.type === "blob" ) })[0];
```

src/file/Lineage.present.js
```diff
@@ -85,6 +85,16 @@ class FileLineageGraph extends Component {
...
+  }
+
+  getNodeLabel(nodeId){
```
lorenzo-cavazzi: There is another `getNodeLabel` function outside the `FileLineageGraph` class, I believe that is the only one called. In that case, we could remove this

vfried: You are right, I used this on a previous version and forgot to delete it.

src/file/Lineage.present.js
vfried: Good point, fixed.

PR #532, opened 2019-07-17
fix: add elements to anonymous user navbar

### LINKED ISSUE #426: Anonymous NavBar
Anonymous users can't see the Projects link and the search input in the top NavBar

lorenzo-cavazzi:
Navbar for anonymous users shows a link to the project list and the search box, as it already happens for logged in users.

Screenshot from 2019-07-17 16-45-22

fix #426

### DIFF
--- a/src/landing/NavBar.js
+++ b/src/landing/NavBar.js
@@ -162,7 +162,12 @@ class AnonymousNavBar extends Component {
           </button>
 
           <div className="collapse navbar-collapse" id="navbarSupportedContent">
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi:
I have looked into that, but currently the feedback only searches through your own projects and the starred projects. These are saved in the user store, we may want to change it since it created other issues (like #408 , which is not as trivial as it appears since the search is currently split between all projects and starred/your projects)

ciyer:
Ok, let's go ahead and merge it.

PR #536, opened 2019-07-22
fix(lineage): "no lineage information" message wasn't readable

vfried:
The no lineage message was displayed like this: 

image

And now is readable: 

image


vfried:
This is irrelevant since the "no lineage message is no longer a graph with a single node". 
Now it's just text.

PR #548, opened 2019-07-24
feat(lineage): query graphql endpoint to get the lineage

### LINKED ISSUE #516: consume lineage data from the graphql endpoint
https://github.com/SwissDataScienceCenter/renku/pull/587 introduces a graphql endpoint for KG queries. The UI should use this endpoint for lineage queries.

lorenzo-cavazzi:
This PR introduces a minimal graphql client that is used to query the lineage data.

I have decided to use the [Apollo client library](https://www.apollographql.com/docs/react/) and to add it to the `APIclient` so that other `api-client` functions can be easily adapted.

Alternative solutions I have considered are:
* creating a separate `graphql-client` object --> it didn't add any real benefit but slightly complicated the UI (I have a PR ready for that in case it's useful)
* using no graphql library or lightweight solution like [nanographql](https://github.com/yoshuawuyts/nanographql) --> the only benefit is a small footprint

Preview available here: https://lorenzotest.dev.renku.ch

fix #516 

Requires SwissDataScienceCenter/renku-gateway#164
(along with SwissDataScienceCenter/renku#587 and SwissDataScienceCenter/renku-graph#88)


lorenzo-cavazzi:
I removed `react-apollo` from the dependencies since it's not needed in this PR

PR #550, opened 2019-07-24
fix(lineage): URL fixed and minor css improvements

### LINKED ISSUE #546: malformed URL when clicking on a node in the graph
If the text of a node is shortened, this can result in a path with `...` added, which is invalid:

image

vfried:
The url is correct now, and minor css improvements (workflow nodes with text in the center and buttons with new styling).

Fixes #546

### DIFF
--- a/src/file/Lineage.css
+++ b/src/file/Lineage.css
@@ -14,6 +14,10 @@ g.data rect {
   stroke: unset;
 }
 
+g.workflow {
+  line-height: 1;
+}
+
 g.workflow polygon {
   fill: #7fbf7bc9;
   stroke: unset;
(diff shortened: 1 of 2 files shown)


src/file/Lineage.css
```diff
@@ -30,11 +34,42 @@ g.workflow polygon {
...
   height: 65vh;
   width: 80vw;
   cursor:pointer;
+}
+
+.zoomButtons{
+  position: absolute;
+  right: 0;
+  background: #fff;
+  padding-right: 20px;
```
lorenzo-cavazzi: I would change this to `margin-right: 20px` so that it doesn't occupy space, otherwise it seems that nodes in that area (right to the buttons) are cut away.

Screenshot from 2019-07-24 15-53-50

We can also remove `background: #fff;` so that the buttons are rounded normally instead of having a squared shape due to the background.

vfried: Done

PR #553, opened 2019-07-30
feat(lineage): style data different from code

ciyer:
Style data and code in a way that is more visibly apparent.

### DIFF
--- a/src/file/Lineage.css
+++ b/src/file/Lineage.css
@@ -6,12 +6,12 @@ text {
 g.code rect {
   fill: #998ec399;
   stroke-width: 1.5px;
-  stroke: #333;
 }
 
 g.data rect {
-  fill: #998ec399;
-  stroke: unset;
+  fill: #ffffff;
+  stroke: #aaa;
+  stroke-width: 0.7px;
 }
 
 g.workflow {


vfried:
Could you add a screenshot of how it looks?

ciyer:
@vfried Please take a look at this change and provide some feedback. I think the result is easier to read and makes the distinction between files that are code and those that are data more clear.

Compare:
  (old)   https://dev.renku.ch/projects/249/files/lineage/src/clean_data.py
  (new) https://sekhar.dev.renku.ch/projects/249/files/lineage/src/clean_data.py

ciyer:
Sure!

### Old

image


### New

image

vfried:
I agree that it's easier to make a difference between nodes... but i'm not a fan of the not styling

Maybe we could add a light grey background

image

or add a thick grey border... 

image

(Personally i prefer the first)

ciyer:
I'm certainly a proponent of "less is more" and removing as much as much as possible, so I prefer the gray border to the gray background, but no background, no border is still my favorite. :) Maybe we can get another pair of eyes here. @lorenzo-cavazzi?

lorenzo-cavazzi:
I am also a minimalist but here I somehow prefer the solution with the borders (no background) to the one without.

ciyer:
How about this thin gray border?

image

rokroskar:
Hmm just looking at this now, styling data as just white with a black border de-emphasizes the data compared to the code. 

Take this view for example:

image

from https://dev.renku.ch/projects/86/files/lineage/img/pc_adelaide.png -- the data cells are barely noticeable.

PR #554, opened 2019-07-30
fix: edit project metadata bug

### LINKED ISSUE #427: Settings->Edit tags and description does not work
Editing of tags and description is broken.

ciyer:
Fix bug that prevents editing of project metadata.

Fix #427.

### DIFF
--- a/src/project/Project.present.js
+++ b/src/project/Project.present.js
@@ -361,9 +361,9 @@ class ProjectFilesNav extends Component {
       data={this.props.filesTree}
       lineageUrl={this.props.lineagesUrl}
       projectUrl={this.props.fileContentUrl}
-      setOpenFolder={this.props.setOpenFolder} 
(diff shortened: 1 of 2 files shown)


ciyer:
Can test on https://sekhar.dev.renku.ch. May need to reload the page, as the old logic may be cached.

lorenzo-cavazzi:
It works for me.
There is a minor issue: project tags are not updated in the projects list --  this is true only for "starred" and "your projects", but it is related to #408 and the logic behind those 2 tabs need to be updated.

P.S. It may be a good idea to add a margin under the `Update` button

Screenshot from 2019-08-05 13-45-43

I will add this suggestion in another commit, feel free to throw it away if you prefer another solution

ciyer:
Thanks for fixing the spacing. Looks better now.

Issue #525, opened 2019-07-15
Add a datasets tab inside a project page


vfried:
Inside the project overview page, there should be a dataset page.

![image](https://user-images.githubusercontent.com/1196411/61541039-214f0880-aa3f-11e9-8225-807a16b1ef01.png)

Clicking on a dataset takes you to the project dataset details page.

The following information should be displayed:
- Dataset name
- Visibility
- Creators
- Description
- Relevant Metadata (in: https://renku-python.readthedocs.io/en/latest/models/datasets.html)
- Files in the dataset (The files also have separate metadata, we should make sure that all relevant information regarding the files is displayed, either in a table, a popup or something similar Issue: ).

Mockup:
![image](https://user-images.githubusercontent.com/42647877/61228431-93191080-a726-11e9-8a1c-f9ab9fd0c157.png)

(if the user clicks on the title of a dataset she should be redirected to the dataset details page ( #522 )

Related to SwissDataScienceCenter/renku#594

ciyer:
It should also be possible to manage data sets (create, delete, rename) here.

rokroskar:
can this be closed?

vfried:
It has only been pushed to the feature branch, and we are still not fetching the datasets from the KG. But that can go in a new issue i guess.

rokroskar:
Right forgot about that - no lets keep it open until it's complete. Thanks!

PR #555, opened 2019-07-31
feat(datasets): tab for datasets inside project overview

vfried:
There is a tab for datasets inside project overview tab, there still needs to be a button that redirects the user to the detailed overview of the dataset. It still needs to be decided what information exactly we want to display.

image

Addresses #525


vfried:
The two dataset views inside project are now connected... this looks like this 
image

PR #562, opened 2019-08-12
WIP feat(datasets): with datasets explorer

vfried:
This addresses #525 .

There is now a dataset explorer that is similar to the file explorer but for datasets. 

This is "WIP" for two reasons:
1) We are getting the datasets info from the yml file in .renku instead of the KG.
2) This only works for datasets imported with the latest renku version. The older versions have the files in the dataset in a different way than the new ones so this makes everything break. Also, the older version for data import don't include so much data.

Test Deployment:
https://virginiatest.dev.renku.ch/projects/412/datasets/cc0c4e5f87d7411485c77577b0addca2/

Preview:
image

### DIFF
--- a/src/dataset/Dataset.present.js
+++ b/src/dataset/Dataset.present.js
@@ -1,0 +1,125 @@
+/*!
+ * Copyright 2018 - Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import { Col, Card, CardHeader, CardBody, Table } from 'reactstrap';
+import { Loader } from '../utils/UIComponents';
+import faDownload from '@fortawesome/fontawesome-free-solid/faDownload'
+import FontAwesomeIcon from '@fortawesome/react-fontawesome'
+
+function DatasetTableRow({dataset, label, prop, propFunc=null, type=null}) {
+  if (dataset[prop] == null) return null;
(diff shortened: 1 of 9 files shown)


rokroskar:
can we make the URL field clickable?

vfried:
> can we make the URL field clickable?

Yes, I will do it

src/dataset/Dataset.present.js
```diff
@@ -0,0 +1,168 @@
...
+
+function DisplayDetails(props){
+  return <Card key="datasetFiles">
+    <CardHeader className="align-items-baseline">
+      <span className="caption align-baseline">Dataset Info</span>
+    </CardHeader>
+    <CardBody>
+      <Table borderless>
+        <tbody>
+          {
```
ciyer: There is some code duplication here. To make it easier to adapt, there could be a helper function here:
```
function datasetTableRow(dataset, label, prop, propFunc=null) {
  if (dataset[prop] == null) return null;
  const propVal = (propFunc) ? propFunc(dataset[prop]).toString() : dataset[prop].toString();
  return (
              <tr>
                 <td><strong>{label}:</strong></td>
                 <td>{propVal}</td>
               </tr>
    )
}
```

Then, the lines below  can be:
```
datasetTableRow(props.dataset, 'Created', 'created');
datasetTableRow(props.dataset, 'Published', 'date_published', (val) => val["@value"]);
...
```

src/dataset/Dataset.present.js
```diff
@@ -0,0 +1,168 @@
...
+
+function DisplayFiles(props){
+  if (props.files === undefined) return null;
+
+  return <Card key="datasetDetails">
+    <CardHeader className="align-items-baseline">
+      <span className="caption align-baseline">Dataset Files</span>
+    </CardHeader>
+    <CardBody>
+      <Table borderless>
```
ciyer: Also add `size='sm` here as above.

src/project/Project.js
```diff
@@ -189,13 +195,32 @@ class View extends Component {
     }
   }
 
+  getSelectedDataset(datasets) {
+
+    if (datasets === undefined) return undefined;
+
+    const datasetPath = this.props.location.pathname.endsWith('/') 
+      ? this.props.location.pathname.substr(0,this.props.location.pathname.length-1)
```
ciyer: This would be clearer as `this.props.location.pathname.slice(0, -1)`

src/project/Project.js
```diff
@@ -189,13 +195,32 @@ class View extends Component {
...
+      : this.props.location.pathname;
+    const datasetId = datasetPath.split('/').pop();
+
+    let selectedDataset = datasets.filter((
```
ciyer: This can be written as ```
let selectedDataset datasets.filter( d => datasetId === d.identifier )[0]
```

vfried:
> Looks very nice! I have some minor code and display styling suggestions.

I did all the changes you requested, it looks better now, and the code is a bit cleaner. 
Thanks for your suggestions!

Issue #561, opened 2019-08-12
allow specifying refname to use for fetching the project templates


rokroskar:
# Description

We need a way of allowing the templates to evolve without breaking existing deployments. In the old templates mechanism, this was handled by using a folder for each version. With the new templates, we want to move to a cleaner solution and more flexible solution.

# Solution

The project-template fetching logic on the new project page should be adapted as follows:
- We need to introduce a Kubernetes chart configuration option called `project-template-refname` that specifies a `refname` to use for fetching project templates.
- If no refname is specified, use `master`.

This refname should be used when we fetch content from github. E.g., to get the templates manifest, use the url:

`https://raw.githubusercontent.com/SwissDataScienceCenter/renku-project-template/${refname}/manifest.yaml`

Similarly for any of the individual templates.

# Unresolved questions

How will this interact with user, group, and instance templates? It is conceivable that one gitlab is the backend for multiple renku instances (running different versions).

rokroskar:
Can we also have the git repo specified? The chart options could look something like:

```
templates-repository:
  url:
  ref:
```

rokroskar:
> How will this interact with user, group, and instance templates? It is conceivable that one gitlab is the backend for multiple renku instances (running different versions).

I think when there is ambiguity we need to just treat the templates as git repos - if it isn't obvious which version/tag/ref to use, we need to present the available options to the user on project creation.

rokroskar:
can we close this? cc @ciyer @vfried

PR #566, opened 2019-08-14
feat(templates): the  user can configure a url and ref to fetch templates from

vfried:
Addresses #561 

I created the branch 'test-branch-manifests' inside our templates project for testing reasons.

In case there is no branch and url configured it fails. Specifications say that default should be master but if there is no URL it will fail anyway unless we want to set the url from our templates project as default along with master.

I had to modify also my public/config.json to make it work, i'm not sure something is missing in the configuration I made. My config file looks like this now (i put it here because is an untracked file):

{
  "BASE_URL": "https://virginia.dev.renku.ch",
  "GATEWAY_URL": "https://virginia.dev.renku.ch/api",
  "WELCOME_PAGE": "IyMgV2VsY29tZSB0byBSZW5rdSB0aHJvdWdoIHRlbGVwcmVzZW5jZQpTb21lIGRlcGxveW1lbnQtc3BlY2lmaWMgaW5mb3JtYXRpb24gd2lsbCBiZSByZWFkIGZyb20gdGhlIHlvdXIgdmFsdWVzLnlhbWwgZmlsZSBhbmQgYmUgZGlzcGxheWVkIGFzIG1hcmtkb3duIGZpbGUuCg==",
  "RENKU_VERSION": "latest",
  "RENKU_TEMPLATES_URL": "https://github.com/repos/SwissDataScienceCenter/renku-project-template",
  "RENKU_TEMPLATES_REF": "test-branch-manifests"
}

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -21,6 +21,8 @@ echo " Configuration:"
 echo " GATEWAY_URL=${GATEWAY_URL:-http://gateway.renku.build}"
 echo " BASE_URL=${BASE_URL:-http://renku.build}"
 echo " RENKU_VERSION=${RENKU_VERSION}"
+echo " RENKU_TEMPLATES_URL=${RENKU_TEMPLATES_URL}"
+echo " RENKU_TEMPLATES_REF=${RENKU_TEMPLATES_REF}"
 echo "==================================================="
 
 tee > /usr/share/nginx/html/config.json << EOF
@@ -28,7 +30,9 @@ tee > /usr/share/nginx/html/config.json << EOF
   "BASE_URL": "${BASE_URL:-http://renku.build}",
   "GATEWAY_URL": "${GATEWAY_URL:-http://gateway.renku.build}",
   "WELCOME_PAGE": "${WELCOME_PAGE}",
-  "RENKU_VERSION": "${RENKU_VERSION}"
+  "RENKU_VERSION": "${RENKU_VERSION}",
+  "RENKU_TEMPLATES_URL": "${RENKU_TEMPLATES_URL}",
+  "RENKU_TEMPLATES_REF": "${RENKU_TEMPLATES_REF}"
 }
 EOF
 
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -20,7 +20,6 @@ echo "==================================================="
 echo " Configuration:"
 echo " GATEWAY_URL=${GATEWAY_URL:-http://gateway.renku.build}"
 echo " BASE_URL=${BASE_URL:-http://renku.build}"
-echo " RENKU_VERSION=${RENKU_VERSION}"
 echo " RENKU_TEMPLATES_URL=${RENKU_TEMPLATES_URL}"
 echo " RENKU_TEMPLATES_REF=${RENKU_TEMPLATES_REF}"
 echo "==================================================="
@@ -30,7 +29,6 @@ tee > /usr/share/nginx/html/config.json << EOF
   "BASE_URL": "${BASE_URL:-http://renku.build}",
   "GATEWAY_URL": "${GATEWAY_URL:-http://gateway.renku.build}",
   "WELCOME_PAGE": "${WELCOME_PAGE}",
-  "RENKU_VERSION": "${RENKU_VERSION}",
   "RENKU_TEMPLATES_URL": "${RENKU_TEMPLATES_URL}",
   "RENKU_TEMPLATES_REF": "${RENKU_TEMPLATES_REF}"
 }
--- a/helm-chart/renku-ui/values.yaml
+++ b/helm-chart/renku-ui/values.yaml
@@ -57,3 +57,7 @@ welcomePage:
     ## Welcome to Renku!
     Renku is software for collaborative data science. With Renku you can share code and data,
     discuss problems and solutions, and coordinate data-science projects.
+
+templatesRepository:
+  url: https://github.com/SwissDataScienceCenter/renku-project-template
+  ref: 0.1.1
--- a/src/api-client/project.js
+++ b/src/api-client/project.js
@@ -25,10 +25,14 @@ const FileCategories = {
   workflows: (path) => path.startsWith('.renku/workflow/'),
 };
 
-const TemplatesReposUrl = {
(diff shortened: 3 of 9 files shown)


rokroskar:
why use the `api.github.com` URL instead of just github.com?

vfried:
> why use the `api.github.com` URL instead of just github.com?

Because we are calling the github api, otherwise it fails. Unless we add the "api." string part ourselves.

From github api documentation:
image

rokroskar:
I think it would make more sense for whoever is deploying the chart to just put the URL of the repo - it's not that you usually put API URLs in configuration files.

vfried:
> I think it would make more sense for whoever is deploying the chart to just put the URL of the repo - it's not that you usually put API URLs in configuration files.

True, I will do the change so they only put the repo url.

vfried:
> I think it would make more sense for whoever is deploying the chart to just put the URL of the repo - it's not that you usually put API URLs in configuration files.

Done, now the url should be https://github.com/repos/SwissDataScienceCenter/renku-project-template or similar

rokroskar:
should be just https://github.com/SwissDataScienceCenter/renku-project-template - without the `repos` in the middle

vfried:
> should be just https://github.com/SwissDataScienceCenter/renku-project-template - without the `repos` in the middle

True, I changed it so now it works without the "/repo" part. It will work now assuming that the users put the proper URL.

helm-chart/renku-ui/values.yaml
```diff
@@ -57,3 +57,7 @@ welcomePage:
     ## Welcome to Renku!
     Renku is software for collaborative data science. With Renku you can share code and data,
     discuss problems and solutions, and coordinate data-science projects.
+
+templatesRepository:
+  url: https://github.com/SwissDataScienceCenter/renku-project-template
+  ref: test-branch-manifests
```
rokroskar:
```suggestion
  ref: master
```

vfried: It's with the test-branch-manifests for testing reasons, if you consider is ready to be merged i do this change.

rokroskar: this is the default value - for testing you can specify a different one during `helm upgrade` with `--set ui.templatesRepository.ref=test-branch-manifests` if you need to - there shouldn't be any need to keep it here

vfried: ok, good to know! I switched to master now

src/api-client/project.js
```diff
@@ -387,10 +395,11 @@ function addProjectMethods(client) {
     }, 'json', false)
   }
 
-  client.getProjectTemplates = () => {
-    return fetchJson(TemplatesReposUrl.TREES_MASTER)
+  client.getProjectTemplates = (renkuTemplatesUrl, renkuTemplatesRef) => {
+    const formatedApiURL = getApiURLfromRepoURL(renkuTemplatesUrl);
+    return fetchJson(`${formatedApiURL}/git/trees/${renkuTemplatesRef}`)
```
lorenzo-cavazzi: since you fetch `${<url>}/git/trees/${<template>}` multiple times, it may be a good idea to define a function and call that where you need it

src/api-client/project.js
```diff
@@ -449,12 +458,13 @@ function carveProject(projectJson) {
 //       make sense at some point to serve the project template from the GitLab
 //       instance we're working with.
 
-function getPayload(projectName, renkuVersion, projectTemplate) {
+function getPayload(projectName, renkuTemplatesUrl, renkuTemplatesRef, projectTemplate) {
   // Promise which will resolve into the repository sub-tree
   // which matches the desired version of the renku project template.
-  const subTreePromise = fetchJson(TemplatesReposUrl.TREES_MASTER)
+  const formatedApiURL = getApiURLfromRepoURL(renkuTemplatesUrl);
+  const subTreePromise = fetchJson(`${formatedApiURL}/git/trees/${renkuTemplatesRef}`)
```
lorenzo-cavazzi: `fetchJson(${<url>}/git/trees/${<template>})` --> function

src/api-client/project.js
```diff
@@ -449,12 +458,13 @@ function carveProject(projectJson) {
...
     .then(data => data.tree.filter(obj => obj.path === projectTemplate)[0]['sha'])
-    .then(treeSha => fetchJson(`${TemplatesReposUrl.TREES}${treeSha}?recursive=1`));
+    .then(treeSha => fetchJson(`${formatedApiURL}/git/trees/${treeSha}?recursive=1`));
```
lorenzo-cavazzi: `fetchJson(${<url>}/git/trees/${<template>})` --> function

src/api-client/project.js
vfried: I'm not sure about this. If i make a function to call the fetchJson function it will end up looking almost the same, and the exact same case only happens twice since the third case it's with the tree sha and it adds the ?recursive=1 at the end.

lorenzo-cavazzi: I thought it could make sense to have it only in one place so it's easier to understand that they are doing the same thing. The `?recursive=1` could easily be a parameter. Something like this
```
function getGitTrees(apiUrl, templateUrl, recursive=false) {
  let url = `${apiUrl}/git/trees/${templateUrl}`;
  if (recursive)
    url += "?recursive=1";
  return fetchJson(url);
}
```
Btw I am going to accept the PR as it is, feel free to use the function or not, it's not a big difference :slightly_smiling_face:

docker-entrypoint.sh
```diff
@@ -21,14 +21,18 @@ echo " Configuration:"
...
+echo " RENKU_TEMPLATES_REF=${RENKU_TEMPLATES_REF}"
 echo "==================================================="
 
 tee > /usr/share/nginx/html/config.json << EOF
 {
   "BASE_URL": "${BASE_URL:-http://renku.build}",
   "GATEWAY_URL": "${GATEWAY_URL:-http://gateway.renku.build}",
   "WELCOME_PAGE": "${WELCOME_PAGE}",
-  "RENKU_VERSION": "${RENKU_VERSION}"
+  "RENKU_VERSION": "${RENKU_VERSION}",
```
rokroskar: do we still need this?

ciyer: You are right: it is not used anywhere. We have removed this.

PR #573, opened 2019-08-20
chore(release): update version to 0.6.4

ciyer:
Update version to 0.6.4 after 0.6.3 release.


vfried:
Just a quick question... shall we keep on changing the package.json version in future releases?

ciyer:
Very observant! Yes, I think we should keep the package.json up-to-date in future releases. I have added this to the release procedure documentation.

PR #575, opened 2019-08-21
feat(notebooks): check docker image related pipeline before starting a new server

### LINKED ISSUE #495: Check notebook image build status
~~Requires #472~~ *merged*

Ensure the user can't try to start a new notebook server if the notebook image is not available -- i.e. not ready yet or in an error state.

- [ ] In the "Launch Notebook" page, when the user selects a branch/commit, check the notebook image status before allowing to select options and start the notebook
- [ ] Add the button to allow the user to (re)trigger the image creation when it's unavailable
- [ ] Display status through the logs until the image is building (poll to refresh) or it has failed

edit: remember to remove the badge on the top-right of the project page, and also the query to the API during the page loading -- if not used somewhere else
Screenshot from 2019-07-01 17-57-33

lorenzo-cavazzi:
When starting a new interactive environment, the UI now checks for the Docker image availability.
Behind the scenes, only the GitLab pipelines are effectively checked -- technically it is still possible that the pipeline containing the image build process has correctly terminated but the image is not available anymore in the Docker repository.

Preview available here: https://lorenzotest.dev.renku.ch/

How to test: open a project, create a new commit and open the "start new server" page just after. Verify that the "Docker Image" badge says something is ongoing and is updated later.
It's possible to play a bit with the relevant pipeline by opening it in GitLab and start/stop/re-run it. The information should be updated almost immediately.

fix #495

Screenshot from 2019-08-22 09-29-48

### DIFF
--- a/src/notebooks/Notebooks.present.js
+++ b/src/notebooks/Notebooks.present.js
@@ -23,8 +23,10 @@ import { Link } from 'react-router-dom';
 import { Form, FormGroup, FormText, Label, Input, Button, Row, Col, Table } from 'reactstrap';
 import { UncontrolledButtonDropdown, DropdownToggle, DropdownMenu, DropdownItem } from 'reactstrap';
 import { UncontrolledTooltip, UncontrolledPopover, PopoverHeader, PopoverBody } from 'reactstrap';
+import { Badge } from 'reactstrap';
 import FontAwesomeIcon from '@fortawesome/react-fontawesome';
-import { faStopCircle, faExternalLinkAlt, faInfoCircle, faSyncAlt, faCogs } from '@fortawesome/fontawesome-free-solid';
+import { faStopCircle, faExternalLinkAlt, faInfoCircle, faSyncAlt} from '@fortawesome/fontawesome-free-solid';
+import { faCogs, faCog, faExclamationTriangle, faRedo } from '@fortawesome/fontawesome-free-solid';
 import { Modal, ModalHeader, ModalBody, ModalFooter } from 'reactstrap';
 
 import { StatusHelper } from '../model/Model';
@@ -346,17 +348,32 @@ class NotebookServerRowAction extends Component {
 
 // * StartNotebookServer code * //
 class StartNotebookServer extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { ignorePipeline: null };
+  }
+
+  setIgnorePipeline(value) {
+    this.setState({ ignorePipeline: value });
+  }
+
   render() {
     const { branch, commit } = this.props.filters;
     const { branches } = this.props.data;
+    const { pipelines } = this.props;
     const fetching = {
       branches: StatusHelper.isUpdating(branches) ? true : false,
+      pipelines: pipelines.fetching,
       commits: this.props.data.fetching
     }
-    const show = {
-      commits: !fetching.branches && branch.name ? true : false,
-      options: !fetching.branches && branch.name && !fetching.commits && commit.id
-    }
+    let show = {};
+    show.commits = !fetching.branches && branch.name ? true : false;
+    show.pipelines = show.commits && !fetching.commits && commit.id;
+    show.options = show.pipelines && pipelines.fetched && (
+      pipelines.main.status === "success" || pipelines.main.status === undefined
+      || this.state.ignorePipeline
+      || this.props.justStarted
+    );
 
     return (
       <Row>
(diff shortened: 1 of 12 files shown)


rokroskar:
Works as advertised! One cosmetic thing - I would remove is the "Please wait..." as it's not really necessary (you have no other option) and it makes the line wrap unnecessarily. 

image

lorenzo-cavazzi:
The line wrap depends on a number of factors like your system font and the width of the browser window, I wouldn't bother too much about that.

Still, I agree we can remove that part but I am not sure if it would be clear that no other actions are required by the user apart from waiting (sometimes the image takes some time). Should we maybe re-phrase it completely? Something like `Please wait until the pipeline used to build the Docker image for the interactive environment has terminated.` -- any change/addition that makes it sound better is welcome

src/notebooks/Notebooks.present.js
```diff
@@ -474,6 +490,135 @@ class StartNotebookBranchesOptions extends Component {
...
+  }
+}
+
+class StartNotebookPipelinesContent extends Component {
+  componentDidMount() {
+    this.props.ignorePipeline(false);
+  }
+
+  render() {
+    const pipeline = this.props.pipelines.main;
```
ciyer: Can you render an empty div or something similar if ignorePipeline === true? Currently, after you ignore, you still see the buttons, even though an action has already been taken.

ciyer: More precisely -- after ignoring, the UI should somehow indicate the ignore was selected. After launching, none of the buttons should be shown anymore.

src/notebooks/Notebooks.present.js
```diff
@@ -474,6 +490,135 @@ class StartNotebookBranchesOptions extends Component {
...
+            <UncontrolledPopover trigger="hover" placement="top" target="image_ignore">
+              <PopoverBody>
+                The base image will be used instead.
+                <br /><FontAwesomeIcon icon={faExclamationTriangle} /> This may lead to unexpected errors.
+              </PopoverBody>
+            </UncontrolledPopover>
+            &nbsp;
+            <a className="btn btn-primary btn-sm mb-1" target="_blank" rel="noopener noreferrer"
+              href={pipeline.web_url} id="image_check_pipeline">
+              <FontAwesomeIcon icon={faExternalLinkAlt} /> Check pipeline
```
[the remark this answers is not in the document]
lorenzo-cavazzi: Since it's a button, I have added the sentence without the final point

src/notebooks/Notebooks.present.js
lorenzo-cavazzi: Not choosing the ignore button results in hiding the buttons. There will be a link to the GitLab pipeline (similar to what happens for running pipelines).
Starting the server will also preserve the choice and show the normal flow with the image warning still there -- without buttons

lorenzo-cavazzi:
All the points should be addressed now. I have redeployed  https://lorenzotest.dev.renku.ch

src/notebooks/Notebooks.present.js
```diff
@@ -474,6 +494,147 @@ class StartNotebookBranchesOptions extends Component {
...
+            </a>
+          </FormText>
+        </Label>
+      );
+    }
+    else if (pipeline.status === "failed" || pipeline.status === "canceled") {
+      let actions;
+      if (this.props.ignorePipeline || this.props.justStarted) {
+        actions = (
+          <FormText color="primary">
```
ciyer: Can you change this to be 
```Using base image instead.
            <a href={pipeline.web_url} target="_blank" rel="noreferrer noopener">
              <FontAwesomeIcon icon={faExternalLinkAlt} /> View pipeline in GitLab.
            </a>
```

lorenzo-cavazzi: guess you copied the suggestion wrong, the following 3 lines are already exacltly like your suggestion :slightly_smiling_face:

ciyer: Sorry, should be:
```
The base image will be used instead. This may work fine, but it may lead to unexpected errors.
            <a href={pipeline.web_url} target="_blank" rel="noreferrer noopener">
              <FontAwesomeIcon icon={faExternalLinkAlt} /> View pipeline in GitLab.
            </a>
```

lorenzo-cavazzi: Updated. This is the output now

Screenshot from 2019-08-23 15-47-20

PR #577, opened 2019-08-22
refactor: add local timezone aware date format

### LINKED ISSUE #571: notebooks: git commit timestamps use UTC
We should use the local time zone if possible.

lorenzo-cavazzi:
The commit datetime showed to the user while launching a new notebook is now local timezone aware.

fix #571

### DIFF
--- a/src/notebooks/Notebooks.present.js
+++ b/src/notebooks/Notebooks.present.js
@@ -652,7 +652,7 @@ class StartNotebookCommits extends Component {
       commits;
     const commitOptions = filteredCommits.map((commit) => {
       return <option key={commit.id} value={commit.id}>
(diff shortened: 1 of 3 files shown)


lorenzo-cavazzi:
It turns out that it's not that easy to create a proper unit test because JavaScript doesn't allow to create a new date for a specific timezone https://stackoverflow.com/questions/15141762/how-to-initialize-a-javascript-date-to-a-particular-time-zone

I added a pretty long piece of code to test on the current user timezone. Basically, it works on string manipulation. This is the best way that came to my mind to avoid writing a test that is simply a re-write of the function to be tested. The other way would be adding an external library for time management, which doesn't look like a good way to test it either.

lorenzo-cavazzi:
just rebased and squashed since it's a single topic

Issue #167, opened 2018-04-12
Renku UI Projects URL scheme


ciyer:
Currently, we have the following URL scheme for information around a project:

Entity | URL
------|-----
Project Home | /projects/[project-id]/
Project Kus | [project-home]/kus
Project Ku |  [project-home]/kus/[ku-id]
Project Settings |  [project-home]/settings
Project File | [project-home]/[path-to-file]

There are a several problems with this:
1. Projects should be in namespaces, and the namespaces should be visible
2. Project URLs should use the project name instead of the opaque id
3. References to files should be explicit, not implicit since this may result in namespace collisions

For these reasons, the UI URL scheme should be changed to the following:

Entity | URL|Status
------|-----|------
Project Home | /projects/[project-namespace]/[project-name]|
Project Kus | [project-home]/kus|&#10004;
Project Ku |  [project-home]/kus/[ku-id]|&#10004;
Project Settings |  [project-home]/settings|&#10004;
Project File Listing | [project-home]/files/|&#10004;
Project File Lineage| [project-home]/files/lineage/[path-to-file]|&#10004;
Project File Content| [project-home]/files/blob/[path-to-file]|&#10004;

rokroskar:
The title of this issue is "Renku URL scheme" but the description only refers to Projects - with #579 merged, can this be closed?

vfried:
I changed the name, we can close it now.

PR #579, opened 2019-08-22
refactor: project url uses namespace and name instead of id

vfried:
This addresses #167 

Old URLs that where using the project id will be rerouted to the corresponding namespace/project-name url is possible.

There could be some edge cases that overlap with the new routing, priority will be given to the new schema, i think this edge cases don't exist in reality in renku.

Test Deployment:
https://virginiatest.dev.renku.ch/

Since this traverses all of the projects routing I would appreciate you testing it, if you find some error let me know :)

### DIFF
--- a/src/project/Project.js
+++ b/src/project/Project.js
@@ -50,10 +50,30 @@ class View extends Component {
     this.projectState = new ProjectModel(StateKind.REDUX);
   }
 
+  componentWillMount(){
+    if(this.props.projectPathWithNamespace === undefined && this.props.projectId !== undefined){
+      this.redirectProjectWithNumericId();
+    }
+  }
+
   componentDidMount() {
-    // fetch only if user data are already loaded
-    if (this.props.user.available === true) {
-      this.fetchAll();
+    if(this.props.projectPathWithNamespace !== undefined){
+      // fetch only if user data are already loaded
+      if (this.props.user.available === true) {
+        this.fetchAll();
+      }
+
+      // in case the route fails it tests weather it could be a projectid route
+      const routes=['overview','kus','ku_new','files','lineage','notebooks',
+        'data','workflows','settings','pending','launchNotebook','notebookServers'];
+      const available = this.props.core ? this.props.core.available : null;
+      const potentialProjectId = this.props.projectPathWithNamespace.split('/')[0];
+      const potentialRoute = this.props.projectPathWithNamespace.split('/')[1];
+
+      if(!available && !isNaN(potentialProjectId) && routes.indexOf(potentialRoute) > 0){
+        this.redirectAfterFetchFails(this.props.projectPathWithNamespace,
+          this.props.location.pathname.replace("/projects/"+potentialProjectId,''));
+      }
     }
   }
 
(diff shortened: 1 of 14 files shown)


src/project/Project.js
```diff
@@ -50,11 +50,29 @@ class View extends Component {
...
     }
+
+    // in case the route fails it tests weather it could be a projectid route
+    const routes=['overview','kus','ku_new','files','lineage','notebooks',
+      'data','workflows','settings','pending','launchNotebook','notebookServers'];
+    const available = this.props.core ? this.props.core.available : null;
+    const potentialProjectId = this.props.projectSlug.split('/')[0];
+    const potentialRoute = this.props.projectSlug.split('/')[1];
+
+    if(!available && !isNaN(potentialProjectId) && routes.indexOf(potentialRoute) > 0){
```
lorenzo-cavazzi: I guess it's `>= 0` instead of `>`

lorenzo-cavazzi:
I am trying to navigate on all pages.

I get an error when I try to visualize a notebook file while in "lineage view", like in this one
https://virginiatest.dev.renku.ch/projects/lorenzo.cavazzi.tech/zurich-bikes-tutorial/files/lineage/notebooks/zhbikes-notebook.ipynb

lorenzo-cavazzi:
Another small issue. When I try to use an old project url, I get an error for the overview page.
E.G. https://virginiatest.dev.renku.ch/projects/95/ (with or without the final slash) doesn't work, but any sub url of that project works fine, like https://virginiatest.dev.renku.ch/projects/95/settings.

src/project/Project.js
vfried: Actually, there was a check missing, in case the projectSlug is undefined, meaning that is actually a projectId

vfried: But well spotted anyways

vfried:
> I am trying to navigate on different pages.
> 
> I get an error when I try to visualize a notebook file while in "lineage view", like in this one
> https://virginiatest.dev.renku.ch/projects/lorenzo.cavazzi.tech/zurich-bikes-tutorial/files/lineage/notebooks/zhbikes-notebook.ipynb

I just tried to navigate to the graph of a file with format .ipynb on de.renku.ch and it also fails. This is not a problem of the file path but i'm not sure where this problem comes from.

vfried:
> E.G. https://virginiatest.dev.renku.ch/projects/95/ (with or without the final slash) doesn't work, but any sub url of that project works fine, like https://virginiatest.dev.renku.ch/projects/95/settings.

Fixed

vfried:
Yes, sure... wouldn't it be feat(project)?

lorenzo-cavazzi:
This has already been approved, but I found a couple of edge cases that may be confusing for the users

* Endless spinning wheel when trying to access a non existing project by id -- or a private project as well (https://virginiatest.dev.renku.ch/projects/9500/)
Screenshot from 2019-08-27 13-29-48

* Empty project when trying to access a wrong url path (https://virginiatest.dev.renku.ch/projects/abcdefg)
Screenshot from 2019-08-27 13-30-12

vfried:
> This has already been approved, but I found a couple of edge cases that may be confusing for the users
> 
> * Endless spinning wheel when trying to access a non existing project by id -- or a private project as well (https://virginiatest.dev.renku.ch/projects/9500/)
>   Screenshot from 2019-08-27 13-29-48
> * Empty project when trying to access a wrong url path (https://virginiatest.dev.renku.ch/projects/abcdefg)
>   Screenshot from 2019-08-27 13-30-12

Ill fix this before we push it...

vfried:
I fixed the not found pages, there where some errors and also some new pages added. I think they could have been part of a new issue but i wasn't sure, since is a small change i added it here. 

If you navigate to a path that doesn't exist inside a project you see this:
image

If you navigate to a path under /projects that doesn't exist you see this: 
image

If you navigate to a project with a numeric id that doesn't exist:
image

rokroskar:
This doesn't seem like a huge problem but when I go to a URL for a blob directly, I see the 404 page flash for a split second... Then the correct page with the file content loads.

vfried:
> This doesn't seem like a huge problem but when I go to a URL for a blob directly, I see the 404 page flash for a split second... Then the correct page with the file content loads.

I see what you mean... I'll try to fix it.

lorenzo-cavazzi:
> I am trying to navigate on different pages.
> 
> I get an error when I try to visualize a notebook file while in "lineage view", like in this one
> https://virginiatest.dev.renku.ch/projects/lorenzo.cavazzi.tech/zurich-bikes-tutorial/files/lineage/notebooks/zhbikes-notebook.ipynb

info: this will be addressed by #582

vfried:
> * We have 2 different "Project not found" pages. Wouldn't it be better to have a single one?

I added a new not found page but is for when the project was found but the url inside the project wasn't found... this allows the user to still navigate inside the project. If we put a simple project not found page the user can't navigate inside the project and it would also be technically wrong since the project is there.

rokroskar:
Now that this is in place, we should discuss the larger topic of providing URIs for Renku entities. This should be in-sync with how entities are identified in the KG.

PR #582, opened 2019-08-28
feat(notebooks): add polling to Jupyter button in notebook preview

### LINKED ISSUE #559: Add polling capability to LaunchNotebookButton component
The functionality of the `<LaunchNotebookButton>` component has been restored by #556 but it should be further improved to always display the realtime status of the JupyterLab notebook.

The first iteration may be simply based on polling. It makes sense to wait for #518 to be closed and to re-use the logic in `Notebooks` components -- split them if needed.

lorenzo-cavazzi:
The `LaunchJupyter` component has been superseded by `JupyterButton`. It gives more detailed feedback on the current interactive environment status and it keeps the information updated by polling the `/servers` API.

This implied a few changes:
* move part of the logic from components in `File` to new ones in `Notebooks`
* get rid of all the code checking notebooks status only once at component load
* remove unused svg icons from css files
* create a new `JupyterIcon` component -- the svg icon may be handled differently (http://tiny.cc/i4p7bz) but that requires changing the SVG code to properly handle greyscale with the  `fill` property. The current approach is close to what we do with `react-fontawesome`, so it should look familiar.

Preview available here: https://lorenzotest.dev.renku.ch

**How to test**: open any notebook files from the `File` tab of a project and try to click on the Jupyter icon on the top right.
E.G. https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/zurich-bikes-tutorial/files/blob/notebooks/zhbikes-notebook.ipynb

fix #559


lorenzo-cavazzi:
As discussed, I'll handle this case in #597 by adopting a coherent solution (checking the existence of "master" and selecting another branch if not available) since it wouldn't make sense to ask the user to start a new environment and then not being able to detect it.

PR #583, opened 2019-08-28
chore(release): version 0.6.4

lorenzo-cavazzi:
Release version 0.6.4


lorenzo-cavazzi:
At the moment it's `0.6.4` and that is in line with the target release.
I believe we need to change it just after the release when we also bump the chart version
https://github.com/SwissDataScienceCenter/sdsc-dev-docs/tree/master/renku-release#bump-chart-version

vfried:
Sorry, i didn't see it was changed already.

PR #587, opened 2019-08-30
fix(lineage): fixed ui crash and hanging load for lineage display

ciyer:
For certain files, there seems to be problems with the request for a notebook server and the request for lineage that causes the UI to either hang when displaying lineage or crash entirely. 

Compare 

https://dev.renku.ch/projects/cramakri/renku-tutorial-flights/files/lineage/notebooks/00-FilterFlights.ipynb

With 

https://sekhar.dev.renku.ch/projects/cramakri/renku-tutorial-flights/files/lineage/notebooks/00-FilterFlights.ipynb

### DIFF
--- a/src/api-client/project.js
+++ b/src/api-client/project.js
@@ -354,7 +354,7 @@ function addProjectMethods(client) {
     const headers = client.getBasicHeaders();
     const url = `${client.baseUrl}/notebooks/${projectPath}/${commitSha}`;
     return client.clientFetch(url, { method: 'GET', headers})
-      .then(resp => resp.data)
+      .then(resp => resp.data, resp => ({url: null}))
       .then(server => server.url ? server.url : null)
   }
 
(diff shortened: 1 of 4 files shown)


src/api-client/project.js
```diff
@@ -354,7 +354,7 @@ function addProjectMethods(client) {
     const headers = client.getBasicHeaders();
     const url = `${client.baseUrl}/notebooks/${projectPath}/${commitSha}`;
     return client.clientFetch(url, { method: 'GET', headers})
-      .then(resp => resp.data)
+      .then(resp => resp.data, resp => ({url: null}))
       .then(server => server.url ? server.url : null)
```
vfried: Do we need the two lines of code here? (I ask because i'm not sure)

vfried: I mean the 357 and 358

ciyer: You are correct, data returns data, not a promise, so there is no need to wait on it. I think this code structure is a holdover from an earlier implementation where there was an extra promise there.

vfried: ok, we should link this PR to the one that lorenzo has in progress, so he knows about this changes.

PR #591, opened 2019-09-03
feat(notebooks): always select the latest commit when changing branch

### LINKED ISSUE #588: Automatically load commits after changing the branch in notebook server launch panel
A small issue that I just came across:

After changing the branch in the `Start new JupyterLab server` form, the commits dropdown remains empty until I manually refresh the commits (see screenshot). Changing the branch should automatically trigger loading the available commits. 

image

lorenzo-cavazzi:
Selecting a different branch now automatically select the last available commit -- after the commit list is refreshed.

~Preview available here: https://lorenzotest.dev.renku.ch~

fix #588

### DIFF
--- a/src/notebooks/Notebooks.container.js
+++ b/src/notebooks/Notebooks.container.js
@@ -138,31 +138,24 @@ class StartNotebookServer extends Component {
 
   async selectCommit(commitId) {
     if (this._isMounted) {
-      // get the useful commitId
-      if (!commitId) {
-        const oldCommit = this.model.get("filters.commit");
-        if (oldCommit && oldCommit.id)
-          commitId = oldCommit.id;
-        else
-          commitId = "latest";
-      }
-
-      // get the proper commit and set it
+      // filter the list of commits according to the constraints
       const maximum = this.model.get("filters.displayedCommits");
       const commits = maximum && parseInt(maximum) > 0 ?
         this.model.get("data.commits").slice(0, maximum) :
         this.model.get("data.commits");
+      let commit = commits[0];
 
-      let commit;
-      if (commitId === "latest")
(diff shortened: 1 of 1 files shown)


ciyer:
I feel like this new behavior is more intuitive to use. I found a corner case, though. If the project has no commits, then the UI is stuck in `updating branches` forever.

image

https://lorenzotest.dev.renku.ch/projects/cramakri/empty-project/launchNotebook

lorenzo-cavazzi:
The project has neither branches nor commits. How should we behave in this case? Should we just inform the user that an interactive environment requires at least one branch/commit? Or should we provide extra information/instructions?

rokroskar:
I suppose that we should still give them the option of launching a notebook server, but it will be with the instance-configured default image. This would be nice because they don't have to care about any local setup. On the other hand, it might be easier to simply insist that the user makes a commit first.

lorenzo-cavazzi:
At the moment the notebook service requires a `commit_sha` to start a new server. Omitting it results in a 400 error with the following message
> Invalid payload. 'namespace', 'project', and 'commit_sha' are mandatory.

This should be addressed, but it's probably not in the scope of this bugfix PR and is currently affecting every environment. I suggest discussing this in a separate issue -- or in the KG epic /SwissDataScienceCenter/renku#591 if it involves changes in the notebooks service

ciyer:
I think it is adequate to just show a message that a commit is necessary to start an interactive environment, no need to start a server on a project with no commits, since I think this is a rare case. If that is a small change, you can do it here, if it is a larger change then let us make a new issue.

lorenzo-cavazzi:
@ciyer I'll create a new issue for that and I'll work on it right away

PR #592, opened 2019-09-03
fix(notebooks): correctly update pipeline information

### LINKED ISSUE #590: Misleading image availability information during notebook launch
Merging #575 has introduced very useful information about the availability of a docker image when trying to launch a new notebook server. For most of the use-cases, this information is correct. However, as this information is gathered from the pipelines API and not from the registry API, this information is wrong for all cases where no pipeline has ever been run.

After forking a project, an image build is only triggered for the latest commit of the master branch. When choosing an older commit or a different branch, the UI will tell the user that the corresponding image is ready, but silently fall back on the default image when launching a notebook, leading to very confusing results.

A better solution would be to check the relatively recently introduced gitlab container registry API for the existence of the right image. I am not sure if this has been tried already? Otherwise, there must be a way to alert the user in cases where we don't know if the image is available or not.

lorenzo-cavazzi:
This addresses an annoying bug happening in the following situation

- the user tries to start an interactive environment from a commit where all the pipelines/jobs related to building the Docker image are available.
- later, the user selects another commit where the pipeline or the job is not found

In this case (as well as a couple of similar combinations), the information about the Docker images availability wasn't updated correctly.

**Testing**: the easiest way to reproduce the issue is to fork a project, then try to select an older commit -- the pipeline to build the Docker image exists only on the latest commit on the `master` branch

fix #590


rokroskar:
doesn't this close #590?

lorenzo-cavazzi:
> doesn't this close #590?

it does, there is a "re #590" in the description but I guess `re` is not a valid keyword in Github. Let me change it to `fix`

Issue #560, opened 2019-08-09
Allow search by username


emmjab:
The search interface currently only allows searching over project names. It would be useful to also be able to search for projects by username.

PR #593, opened 2019-09-03
feat(projects): search by username and group

vfried:
The user can search by user name and group name. 

Before reviewing the pr is important to understand that gitlab has big limitations on the search api, at some point this will have to be addressed differently.

It looks as follows:

image

image

Test deployment: https://virginiatest.dev.renku.ch/

Closes: #560

### DIFF
--- a/src/api-client/project.js
+++ b/src/api-client/project.js
@@ -120,6 +120,24 @@ function addProjectMethods(client) {
     });
   }
 
+  client.getProjectsBy = (searchIn, userOrGroupId, queryParams) => {
+    let headers = client.getBasicHeaders();
+    return client.clientFetch(`${client.baseUrl}/${searchIn}/${userOrGroupId}/projects`, {
+      method: 'GET',
+      headers,
+      queryParams
+    });
+  }
+
+  client.searchUsersOrGroups = (queryParams, searchIn) => {
+    let headers = client.getBasicHeaders();
+    return client.clientFetch(`${client.baseUrl}/${searchIn}`, {
+      method: 'GET',
+      headers,
+      queryParams
+    }).then(result => result.data);
+  }
+
   client.getProjectFiles = (projectId) => {
     return client.getRepositoryTree(projectId, { path: '', recursive: true }).then((tree) => {
       const files = tree
--- a/src/project/list/ProjectList.container.js
+++ b/src/project/list/ProjectList.container.js
@@ -28,6 +28,12 @@ const orderByValuesMap = {
   UPDATEDDATE: 'last_activity_at'
 };
 
+const searchInValuesMap = {
+  PROJECTNAME: 'projects',
+  USERNAME: 'users',
+  GROUPNAME: 'groups'
+};
+
 class List extends Component {
   constructor(props) {
     super(props);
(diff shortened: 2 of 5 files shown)


src/project/list/ProjectList.container.js
```diff
@@ -81,44 +94,97 @@ class List extends Component {
...
     this.model.setOrderDropdownOpen(!this.model.get('orderByDropdownOpen'));
   }
 
+  onSearchInDropdownToogle() {
+    this.model.setSearchInDropdownOpen(!this.model.get('searchInDropdownOpen'));
+  }
+
+  changeSelectedUserOrGroup(userId){
+    this.model.setSelectedUserOrGroup(userId);
+    this.props.history.push(
```
ciyer: Could you turn this into a method? Something like
```
pushNewSearchToHistory() {
    this.props.history.push(
       this.urlFromQueryAndPageNumber(
         this.model.get('query'),
         1,
         this.model.get('pathName'),
         this.model.get('orderBy'),
         this.model.get('orderSearchAsc'),
         this.model.get('searchIn'),
         this.model.get('selectedUserOrGroup')
       )
     )
}
```

This code appears in several places, and it would make the intention behind the code a little clearer.

ciyer:
Look very nice! This is extremely useful functionality, since I often know who created the project, but cannot remember how the project is called.

I have been testing and found the following:
- When not logged in, search by user doesn't work and has an endless spinning indicator
- We should indicate in the results what was searched for (e.g., Projects by User Name). After entering a search string, you no longer see the domain of the search.
- The first username/group name should be selected, and you should see the results for that one without an extra click 
- On the one hand, I like that you can search even with an empty username or group name search string, but we cannot display the results of the search in this way. There could be hundreds of users (think gitlab.com). This display is fine if there are a small number of users (<11), but does not work for a large number of users.

src/api-client/project.js
```diff
@@ -92,6 +92,45 @@ function addProjectMethods(client) {
...
+  }
+
+  client.searchUsers = (username, queryParams = {}) => {
+    let headers = client.getBasicHeaders();
+    let queryParamsUsername = {};
+
+    return client.clientFetch(`${client.baseUrl}/users?search=${username}`, {
+      method: 'GET',
+      headers,
+      queryParamsUsername
```
lorenzo-cavazzi: I believe the expected attribute name here is `queryParams`, you could change in `queryParams: queryParamsUsername`.

Even better moving all the query parameters together instead of having one in the url
```
let queryParamsUsername = { ...queryParams, search: username };

return client.clientFetch(`${client.baseUrl}/users}`, {
      method: 'GET',
      headers,
      queryParams: queryParamsUsername
```

src/api-client/project.js
```diff
@@ -92,6 +92,45 @@ function addProjectMethods(client) {
...
+    });
+  }
+
+  client.searchGroups = (groupname, queryParams = {}) => {
+    let headers = client.getBasicHeaders();
+    let queryParamsUsername = {};
+    return client.clientFetch(`${client.baseUrl}/groups?search=${groupname}`, {
+      method: 'GET',
+      headers,
+      queryParamsUsername
```
lorenzo-cavazzi: same here

```
let queryParamsUsername = { ...queryParams, search: groupname };

return client.clientFetch(`${client.baseUrl}/groups`, {
      method: 'GET',
      headers,
      queryParams: queryParamsUsername
```

src/project/list/ProjectList.container.js
vfried: Done

src/api-client/project.js
vfried: Fixed

src/project/list/ProjectList.container.js
vfried: > * We should indicate in the results what was searched for (e.g., Projects by User Name). After entering a search string, you no longer see the domain of the search.

I've addressed this but i'm not sure that it looks that good, if you have any suggestion on what we can do i'm happy to change it. For now it looks OK I guess.

ciyer: I think this is fine for the moment. We can improve it in a later issue.

PR #578, opened 2019-08-22
feat(notebooks): notebook servers for anonymous users

lorenzo-cavazzi:
Notebook servers tab is now displayed also to users without sufficient permissions for the project.
The page contains an alert with meaningful information about what the user can do.

Screenshot from 2019-08-22 14-44-39

fix #565

PR #597, opened 2019-09-05
feat(notebooks): handle non-standard branches projects

### LINKED ISSUE #595: handle interactive environments for projects without a commit
Since projects can be created from an empty template, we need to provide adequate feedback to users trying to start an interactive environment without having any commit.
Dropping this requirement requires some changes, but notifying the user that at least a branch/commit is required should be a straightforward change in the UI.

Extracted from #591

lorenzo-cavazzi:
The UI showed wrong feedback when trying to start a new interactive environment for projects without a `master` branch or without any branch/commit at all (note that 1 branch requires at least 1 commit, and also the other way around is also true).

This PR handles the 2 cases as follows:
* no branch named "master" --> another branch is selected
* no branch (and therefore no commit) --> the user is notified that at least 1 commit is necessary. Following the approach similar to #578, a few suggestions are provided to the user 
Screenshot from 2019-09-05 16-29-14

**How to test**:
1. create a project starting from the empty template, then go to the "Notebook servers" tab and click on "Start new server"
2. clone the project locally, add a branch with a name different from "master", create a commit and push to `origin`

fix #595

### DIFF
--- a/src/file/File.container.js
+++ b/src/file/File.container.js
@@ -22,7 +22,9 @@ import ReactDOM from 'react-dom';
 import hljs from 'highlight.js';
 
 import { atobUTF8 } from '../utils/Encoding';
-import { StyledNotebook } from './File.present';
+import { StyledNotebook, JupyterButtonPresent } from './File.present';
+import { ACCESS_LEVELS } from '../api-client';
+import { StatusHelper } from '../model/Model';
 
 const IMAGE_EXTENSIONS = ['jpg', 'jpeg', 'png', 'tiff', 'pdf', 'gif'];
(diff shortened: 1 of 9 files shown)


lorenzo-cavazzi:
P.S. tests will fail, I will fix them after #582 is merge since it should already address the problem

lorenzo-cavazzi:
I added the code to handle projects without "master" branch also for the "Launch Jupyter" button in the File preview notebooks. This completes the PR by handling also the new component introduced in #582.

An updated preview is available at https://lorenzotest.dev.renku.ch

PR #598, opened 2019-09-06
fix(projects): your project refreshes every time the user enters in the tab

vfried:
The user projects now are updated: 
1) when a project is created
2) when a project enters in the "your project" tab and the amount of the member project changed (I could do a comparison of the project ids for more accuracy if necessary. I did it this way because i am worried that comparing ids will take a bit more of time.

The starred projects are also updated in (2)


ciyer:
This fixes an annoying longstanding problem, thanks! When creating a new project, it works well for me on the "Your Projects" page as well. I couldn't see a difference with the "starred projects," though. Should that update when I star a project?

vfried:
> This fixes an annoying longstanding problem, thanks! When creating a new project, it works well for me on the "Your Projects" page as well. I couldn't see a difference with the "starred projects," though. Should that update when I star a project?

I think the only problem left to address is the landing page... it works for starred projects but not for your projects. I'm still not sure how to address this but i'm investigating.

ciyer:
- If I create a new project, I see this now on the Projects page and on the landing page.
- If I start a project, I do not this any page without a refresh.

vfried:
I just pushed new changes... it works for me. Now when a user is in the landing page the projects get refreshed if there are new starred or member projects. There are still some side cases that we could address fetching the projects from the client directly. In this solution i still have to fetch the projects to check if they where changes so we could consider only using the starred and member projects on the jump bar but just fetch projects everywhere else.

vfried:
I made a change inside this tab... The fix i had done was re fetching starred and user projects every time the user was entering the your projects/starred tab and comparing the length of the projects lists and then updating. I didn't liked this solution that much and since we are already fetching projects i decided the best thing was to just the latest list fetched.

I unified the search/starred/your projects component for this and they all use the same serach bar, meaning that you can now search, order, etc inside all tabs.

image

vfried:
> My suggestion: if it is easy, get the Your Projects>search by user to work the same as on the Starred Projects. If it is difficult, then just remove the option to search by user or group on the Your Projects tab. I do not think this is an essential feature.

I decided to hide from the search in the user and group options but leave the tab so the bar doesn't change so much when the user changes tabs... so basically they only have one option. This change has also been done for the starred projects since i don't finde much use on this but i can put it back for the starred.

PR #602, opened 2019-09-06
feat(fileview): the user can preview files with format rout, rproj, rst...

vfried:
Addresse #600


rokroskar:
does this close #600?

PR #606, opened 2019-09-11
Display commit and timestamp information about the file in the file view

### LINKED ISSUE #487: Commit hash and timestamp in file browser
Would it be add to add some more info here

image

or per file 

image

?

ciyer:
Commit hash, with link to commit in github, commit message, commit author, and commit date/time has been added to the display.

It looks like 

image

Running on https://sekhartest.dev.renku.ch/ for testing.

Fix #487.

### DIFF
--- a/src/file/File.present.js
+++ b/src/file/File.present.js
@@ -34,6 +34,36 @@ import { CheckNotebookStatus, CheckNotebookIcon } from '../notebooks'
 import { Loader } from '../utils/UIComponents';
 import { API_ERRORS } from '../api-client';
 
+
+/**
+ * Display the Card with file information. Has the following parameters:
+ *
+ * @param {string} filePath - Path of the file
+ * @param {string} commitHash - Commit hash for the file
+ * @param {Component} buttonGraph - Button to switch to graph view
+ * @param {Component} buttonGit - Button to switch to GitLab.
+ * @param {Component} buttonJupyter - Button to switch to Jupyter.
+ * @param {Object} body - Content to show as the body of the card
+ * @param {Component} lfsBadge - Badge to show for LFS (or null)
+ */
+class FileCard extends React.Component {
+  render() {
+    return <Card>
+      <CardHeader className="align-items-baseline">
+        {this.props.lfsBadge}
+        {this.props.filePath}
+        <span className="caption align-baseline">&nbsp;File view</span>
+        <div className="float-right" >
+          {this.props.buttonJupyter}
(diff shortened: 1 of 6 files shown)


ciyer:
@cchoirat If you get a chance to try this out, we would be interested in your feedback.

cchoirat:
is it on dev?

ciyer:
Not yet, but you can test on https://sekhartest.dev.renku.ch/

vfried:
I like it but i think for smaller screens it might be taking too much space... I think we could do something like this... 

image

For this i made a small change:
image

lorenzo-cavazzi:
@vfried I think floating right doesn't help in a small screen unless the commit message is very short.

Screenshot from 2019-09-13 14-41-06

I kind of prefer the 2 lines version. I would probably cut the commit message chars at a given number (60? 100?) to avoid long commit messages taking multiple lines.

lorenzo-cavazzi:
Ops I see only now that you mentioned cutting down long commit messages, guess the comment was edited but I didn't refresh the page :sweat_smile:

ciyer:
I'm using the "title" that comes with the commit information from gitlab. This is restricted to some number of characters (but I need to look at the code to see how many).

src/file/File.present.js
```diff
@@ -24,55 +24,86 @@ import NotebookPreview from '@nteract/notebook-render';
...
+    }
+    return <Card>
+      <CardHeader className="align-items-baseline">
+        {this.props.lfsBadge}
+        {this.props.filePath}
+        <span className="caption align-baseline">&nbsp;File view</span>
+        <div className="float-right">
+          {this.props.buttonJupyter}
+          {this.props.buttonGit}
+          {this.props.buttonGraph}
```
lorenzo-cavazzi: This is currently not rendered. I quickly tried to identify why but I didn't find the reason. Anyway, the problem is probably in `ShowFile` since the React debugger tells me that the property is set to `null`

Screenshot from 2019-09-13 15-12-18

ciyer: Nice catch! That slipped in when I wasn't careful enough with a refactoring. It was easy to fix.

ciyer:
I see. The `title` field took that commit message up to the first `\n`. I've limited it to 120 chars now.

ciyer:
With some flexbox magic fairy dust, I have it working as Virginia suggested, if there is space:

image

And otherwise, as before:

image

vfried:
Well done, looks good!

PR #608, opened 2019-09-13
fix(navbar): the key cloack user profile opens in a new tab

### LINKED ISSUE #401: Not back to current project after visiting profile page
If from a project page (for example https://renkulab.io/projects/385/)

image

I go to my profile

image

the `Back to renku-ui` button doesn't bring me back to my project, but to https://renkulab.io/.

vfried:
Closes #401 

I think the cleanest option for this is as @ciyer mentioned, opening a new tab.

The user might have to refresh after changing things but i don't see this as a major issue. Otherwise we might have to send key cloack a redirect link... i'm not sure that is possible.


vfried:
Ok, i will add it

vfried:
Done, well spotted!

PR #613, opened 2019-09-13
fix:description in project search is now limited to 2 lines

### LINKED ISSUE #609: Limit ammount of description chars inside the project search
This doesn't look good, we should limit the ammount of chars we display.

image
Also when logged in...
image

vfried:
Closes #609 

Note: The ...  at the end of the cropped text are missing because ellipsis only adds them when you are cropping one line. We could do it with javascript but it would be less responsive.

I think the result here is good...
image


vfried:
I changed the fix to be a javascript fix instead of a css fix...I decided to set up a limit of 250 chars and it looks good when you make the screen big and small, i think there is no need to make different limits in text for different screen sizes since this looks good.

Issue #482, opened 2019-06-05
Clean up interactive environment launch page


ciyer:
# Description

The page to launch interactive environments should be cleaned up. The form fields should have a more logical ordering. Any options that are not relevant should be hidden (e.g., if there are no GPUs available, do not show the GPU input; only show the Server URLs that the project has defined)

# Mock Up

<img width="612" alt="image" src="https://user-images.githubusercontent.com/1196411/58937620-3dd70000-8773-11e9-95e9-2f5339629e22.png">

lorenzo-cavazzi:
According to #568 we will use `Interactive environment` instead of `Activity` as the tab text. #599 also defines a new tab for collaboration, so "merge requests" will be in that tab.

This leaves 2 entries on the left menu, 1 for the list of running notebooks and the other for starting a new one. After speaking with @ciyer , we agree to define these 2 pages as `Running` and `New`  -- any other suggestion is welcome :slightly_frowning_face:

lorenzo-cavazzi:
The labeling and the order of the listed properties come from the API `notebooks/server_options`, that in turn reads them from a file stored in `/etc/renku-notebooks/server_options.json` on the server -- but I failed to identify who is loading the file.
Anyway, ordering and wording should be changed there. I also think it doesn't make sense to return useless parameters (e.g. 0/0 gpus), these should probably be removed from the file.

I like the multiple selection buttons in the mockup.
The whole page is changed quite a lot and now there are other dropdown menus (branches, commits). I think those should stay as the are.

![Screenshot from 2019-09-16 16-17-23](https://user-images.githubusercontent.com/43481553/64965809-f269dd00-d89d-11e9-8f7e-28cdc374f037.png)

PR #618, opened 2019-09-16
feat(environments): create environment route with sub-routes

lorenzo-cavazzi:
This PR integrates the "launch notebook" page into the new "Environments" tab, applying also minor changes to the layout according to #482. Please note that it doesn't filter/change the environment options (refer to the issue).

Screenshot from 2019-09-17 08-54-59

### DIFF
--- a/src/project/Project.present.js
+++ b/src/project/Project.present.js
@@ -607,8 +607,8 @@ function notebookLauncher(userId, accessLevel, notebookLauncher, fork, postLogin
 class ProjectEnvironments extends Component {
   render() {
     return [
-      <Col key="nav" xs={12} md={2} lg={2}>
-        <Nav pills className={'flex-column'}>
+      <Col key="nav" xs={12} md={2}>
+        <Nav pills className="flex-column mb-3">
           <NavItem>
             <RenkuNavLink to={this.props.notebookServersUrl} title="Running" />
           </NavItem>
@@ -618,7 +618,7 @@ class ProjectEnvironments extends Component {
         </Nav>
         {/* <ProjectViewOverviewNav {...this.props} /> */}
       </Col>,
-      <Col key="content" xs={12} md={10} lg={10}>
+      <Col key="content" xs={12} md={10}>
         <Switch>
           <Route exact path={this.props.notebookServersUrl}
             render={props => <ProjectNotebookServers {...this.props} />} />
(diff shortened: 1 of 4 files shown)


lorenzo-cavazzi:
It turns out that the ordering of the options returned by the server is different from the one defined in https://github.com/SwissDataScienceCenter/renku-notebooks/blob/e185002467c75954d749e29e3a7c5f55574257d6/helm-chart/renku-notebooks/values.yaml#L50

After digging a little bit into the details, it seems the responsible is the helm `toJson` function here  https://github.com/SwissDataScienceCenter/renku-notebooks/blob/e185002467c75954d749e29e3a7c5f55574257d6/helm-chart/renku-notebooks/templates/configmap.yaml#L13

We could add an `ordering` properties to the `serverOptions` properties to help the UI rendering options in the desired order.

lorenzo-cavazzi:
The first request is reasonable, but it's currently not easy to implement a clean solution. The main problem is that information about notebooks is available only inside a notebook container component: they instantiate a "private" react store and poll the resource independently, so adding a `<StartNotebookServer>` component inside `<Notebooks>` will end up in polling `/notebooks` api twice.
Changing it is not trivial, we could implement a not-so-elegant workaround but I believe we should create an issue (epic?) to address this problem in a coherent way for every redux store we are using in the UI.

We may go for another solution
* redirect to the `/environments/new` page after the notebooks' data are first loaded -- if this happens after the spinning wheel, it should be (almost) transparent to the user
* add a link to click to go to the `new` page -- this will still require an extra click but it will be less confusing
* ... any further idea?

lorenzo-cavazzi:
I implemented a solution for all the 3 topics and updated the deployment in https://lorenzotest.dev.renku.ch

Should I create a new issue out of the first request about no running environments? (display the "new" page instead of "running"?)

src/project/Project.present.js
```diff
@@ -601,21 +601,44 @@ function notebookLauncher(userId, accessLevel, notebookLauncher, fork, postLogin
...
+      <Col key="nav" xs={12} md={2}>
+        <Nav pills className="flex-column mb-3">
+          <NavItem>
+            <RenkuNavLink to={this.props.notebookServersUrl} title="Running" />
+          </NavItem>
+          <NavItem>
+            <RenkuNavLink to={this.props.launchNotebookUrl} title="New" />
+          </NavItem>
+        </Nav>
+        {/* <ProjectViewOverviewNav {...this.props} /> */}
```
ciyer: Was this intended to be removed? If it is supposed to be here, maybe add a comment explaining why.

lorenzo-cavazzi: I forgot it there, will remove it :slightly_smiling_face:

PR #619, opened 2019-09-16
fix(fileviewer): files with type commit have a preview now

vfried:
Context: there could be files of type "commit" inside the file tree of gitlab, we only have one example of this in the vom_natt project. 
image

Before when the user was clicking the file nothing was happening (because there was no link) and now the fileviewer tries to display the file: 
image


vfried:
> Would it make sense to write something like `Unable to preview commit files`? At the moment it tries to get the extension name ending up in a confusing message where the extension is `.<filename>`

You are right... this is confusing. 
Commit files can be previewed i think, but this has no content. I decided to preview the empty content anyways, since we can show it.... so it looks like this now.

image

PR #625, opened 2019-09-18
fix(project): html content is rendered properly in readme

### LINKED ISSUE #605: Render HTML in project README files correctly
HTML is valid markdown. However, when HTML is part of a README (or the entire README is an HTML document), Renku does not render the HTML correctly. See for example https://renkulab.io/projects/534/.

vfried:
In order to properly render HTML content inside the readme files I tried several approaches.

First I updated ReactMarkdown to the latest version and used this...

image

But it didn't worked. 

Then i found out that gitlab has an api that parses the react markdown to html 
https://docs.gitlab.com/ee/api/markdown.html
But this didn't worked for some cases (i.e: https://virginia.dev.renku.ch/projects/virginiafriedrich/test-html-2/)

Then I tried to change the call when fetching the file (so it looked the same as the one that gitlab does). This also didn't worked.

At the end i decided to download the "showdown" parser, that parses markdown to html.
Then purify the parsed HTML with "DOMPurify" and then inserting it into a div. (This did work!) 

**Test:** https://virginiatest.dev.renku.ch/projects/virginiafriedrich/test-html-description

Before:
image

After:
image

Closes #605


vfried:
I replaced all appearances of react markdown with renku markdown, a component that is in utils.

I moved sanitizedHTMLFromMarkdown to HelperFunctions and the tests for it to Utils.test.js

Now the Kus can also render html...
image

image

PR #626, opened 2019-09-18
feat(fork): the user can change the name of the forked project

vfried:
There is a new field inside the project fork modal where the user can change the name of the forked project.

image


vfried:
I've corrected the statuses thing, it's ready to be reviewed

Issue #499, opened 2019-07-01
Update notebook view


lorenzo-cavazzi:
Refactor the servers table used in `Project` and `Notebooks` pages to separate the `Action` button from the status indication.

This is the list of all the possible actions based on the current status:
    * running: connect, stop, get logs
    * failed: stop, get log
    * pending: get log
The "get log" functionality requires #414.

The new layout should
* [ ] Add an extra column to display the status (as a badge) and an info message when suitable (e.g. error state)
* [ ] Change the action button color back to the standard and add the proper action in the dropdown menu according to the list above.

![Screenshot from 2019-06-19 14-52-40](https://user-images.githubusercontent.com/43481553/59767169-fd62a080-92a1-11e9-92dd-ca498d430ce2.png)

EDIT: remember to edit the relevant documentation after this is closed since the layout will be completely different from the old one and possibly confusing for new users (`Environemnts` sub-pages, dropdown action buttons, ...)

pameladelgado:
This seems like a great functionality to have.
Does the `failed` +logs option include when k8s can't schedule a notebook because of lack of resources?
What does the `stop` action do in case of failure?

lorenzo-cavazzi:
1. Not sure if in that case the environment/pod data are returned at all, my bet is no but maybe @mohammad-sdsc has a different answer
2. The relevant code is here https://github.com/SwissDataScienceCenter/renku-notebooks/blob/0a6e3c484a04b0b217c74479cd24a7d2587fc8ff/renku_notebooks/api/notebooks.py#L192
In short, it invokes this `v1.delete_namespaced_pod( pod_name, kubernetes_namespace, grace_period_seconds=30)` where `v1` comes from the k8s python library
```
from kubernetes import client
v1 = client.CoreV1Api()
```

mohammad-sdsc:
K8s reports PodStatus which includes information if a pod is not schedule-able. See `Unschedulable` in https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions.
This information is passed to GUI as well.

lorenzo-cavazzi:
closed by #633
The logs feature will be addressed soon in #414

Issue #414, opened 2019-04-01
Show notebook server logs in UI


rokroskar:
- [ ] Add "show logs" option in the action column.
- [ ] Display them in a modal and refresh by polling

- [ ] _Further update: introduce a better visualization without modal, possibly with live updates (we need a SSE/WS endpoint?)_

PR #633, opened 2019-09-26
feat(environments): show more status data and allow force stop

### LINKED ISSUE #361: Cannot stop/restart a failed notebook server
How to reproduce:
- Create a docker image that fails on start, e.g. exit 1 in entrypoint/cmd
- Try to start the notebook server which fails, it cannot be stopped and cannot be restarted.

lorenzo-cavazzi:
This PR change the table layout of the environments table and adds a couple of extra options for non-running servers:

* show a popup containing kubernetes pod information
* allow force-stop

How to test: start one or more interactive environments, check the environment pages and try to force stop.
Projects that can't start correctly:
* https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/slow-build
* https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/failing-image

So far, I have not been able to re-produce a pure "error" state (any suggestion is welcome). Maybe the labeling should be tuned: I will try to identity "pending" states that should be labeled as "error".

fix #361, re #499

### DIFF
--- a/src/api-client/notebook-servers.js
+++ b/src/api-client/notebook-servers.js
@@ -35,13 +35,16 @@ function addNotebookServersMethods(client) {
     });
   }
 
-  client.stopNotebookServer = (serverName) => {
+  client.stopNotebookServer = (serverName, force = false) => {
     const headers = client.getBasicHeaders();
     const url = `${client.baseUrl}/notebooks/servers/${serverName}`;
+    let parameters = {};
+    if (force) parameters.force = true;
 
     return client.clientFetch(url, {
       method: 'DELETE',
-      headers
+      headers,
+      queryParams: parameters
     }, "text")
       .then(resp => {
         return true;
--- a/src/notebooks/Notebooks.container.js
+++ b/src/notebooks/Notebooks.container.js
@@ -63,8 +63,8 @@ class Notebooks extends Component {
     this.coordinator.stopNotebookPolling();
   }
 
-  stopNotebook(serverName) {
-    this.coordinator.stopNotebook(serverName);
+  stopNotebook(serverName, force) {
+    this.coordinator.stopNotebook(serverName, force);
   }
 
   mapStateToProps(state, ownProps) {
--- a/src/notebooks/Notebooks.present.js
+++ b/src/notebooks/Notebooks.present.js
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-import React, { Component } from 'react';
+import React, { Component, Fragment } from 'react';
 import Media from 'react-media';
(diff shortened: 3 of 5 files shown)


rokroskar:
You can make an image without `jupyterhub` installed and it will fail to start the container. That should give you a failing container (different from failing to pull an image).

lorenzo-cavazzi:
I force-psuhed bacause I forgot to include `Notebooks.state` :man_facepalming: . I also updated https://lorenzotest.dev.renku.ch/

@rokroskar thanks, I'll try that.

lorenzo-cavazzi:
@rokroskar it seems that the workflow is the same also in that case. The `notebook` pod state hangs in state `waiting`, therefore no "error" state

rokroskar:
ah true, I guess for an error it would have to be something like a missing volume or a secret/configmap etc.

ciyer:
Found an issue when testing:
- Start an environment on one of the project above (I used slow-build)
- Kill the environment
- Try to start a new environment for the same project -- this is not possible.

rokroskar:
This could be an issue with jupyterhub that we've seen before, where the server remains in the db even though it doesn't exist anymore.

lorenzo-cavazzi:
I believe this is related to the `force` delete introduced by this PR SwissDataScienceCenter/renku-notebooks#205 . When the delete is forced, only the k8s code is used to remove the pod and probably jupyterhub stase stays dirty. I'll try to fix that

rokroskar:
I believe the problem is only during spawning - if the server is already up, jupyterhub happily notices that it's been deleted and reacts accordingly. If we can pin down the behavior a bit better we can open an issue with jupyterhub.

lorenzo-cavazzi:
I did a couple of tests and the problem seems to be indeed related to spwaning pods. JupyterHub is sloppy in reacting also to running pods removal, but if eventually does -- if you remove a running pod from rancher and you try to start a new environment immediatly after, it doesn't work.

I guess it worth to update to `JupyterHub 1.0.0` before opening any issue, but I have not been able to succeed in that. I'll try again soon.

lorenzo-cavazzi:
I removed the code to force stop an environment (I left it commented out since it will be re-used soon for #414 ). #361 can't be addressed with the current force stop feature since it leaves JupyterHub in a dirty state and the user won't be able to start a new environment from the same branch/commit until the hub is rebooted (and he doesn't get feedback of this failure). Fortunately, updating JupyterHub to 1.0.0 seems to give us another way to cope with the hanging pods -- this will be addressed separately.

I squashed in 2 commits to keep the history clean.

PR #580, opened 2019-08-26
feat(datasets): datasets inside projects

vfried:
This feature PR, it groups smaller PRs inside a feature branch

PRs included:
https://github.com/SwissDataScienceCenter/renku-ui/pull/562
https://github.com/SwissDataScienceCenter/renku-ui/pull/555

Closes:
#525
#602

PR #646, opened 2019-10-10
feat(environments): show kubernetes log per server

### LINKED ISSUE #414: Show notebook server logs in UI

lorenzo-cavazzi:
This adds the possibility to fetch and show the kubernetes logs for each environment. In this first implementation, I used a simple modal to display logs. I know it's not one of @ciyer favorite solutions, we can discuss it and visualize them in a different way
Sadly, it's not possible to get logs during some spawn phases. This means that, usually, there aren't logs for hanging pods, making this addition less useful than expected.

Preview available at https://lorenzotest.dev.renku.ch/

How to test: start an environment, use the "get logs" button during any phase.

fix #414

### DIFF
--- a/src/notebooks/Notebooks.container.js
+++ b/src/notebooks/Notebooks.container.js
@@ -50,9 +50,15 @@ class Notebooks extends Component {
       this.coordinator.setNotebookFilters(props.scope, true);
     }
 
+    this.state = {
+      showingLogs: false
+    };
+
     this.handlers = {
-      stopNotebook: this.stopNotebook.bind(this)
-    }
+      stopNotebook: this.stopNotebook.bind(this),
+      fetchLogs: this.fetchLogs.bind(this),
+      toggleLogs: this.toggleLogs.bind(this)
+    };
   }
 
   componentDidMount() {
@@ -67,11 +73,24 @@ class Notebooks extends Component {
     this.coordinator.stopNotebook(serverName, force);
   }
 
+  fetchLogs(serverName) {
+    this.coordinator.fetchLogs(serverName);
+  }
+
+  toggleLogs(serverName) {
+    const nextState = !this.state.showingLogs
+    this.setState({ showingLogs: nextState });
+
+    if (nextState)
+      this.fetchLogs(serverName)
+  }
+
   mapStateToProps(state, ownProps) {
     return {
       handlers: this.handlers,
-      ...state.notebooks
-    }
+      ...state.notebooks,
+      logs: { ...state.notebooks.logs, show: this.state.showingLogs }
+    };
   }
 
   render() {
--- a/src/notebooks/Notebooks.container.js
+++ b/src/notebooks/Notebooks.container.js
@@ -78,7 +78,11 @@ class Notebooks extends Component {
   }
 
   toggleLogs(serverName) {
-    const nextState = !this.state.showingLogs
+    let nextState;
+    if (this.state.showingLogs !== serverName)
+      nextState = serverName;
+    else
+      nextState = false;
     this.setState({ showingLogs: nextState });
 
     if (nextState)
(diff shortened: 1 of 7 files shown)


lorenzo-cavazzi:
It seems there are problems with the version deployed on `lorenzotest` and it's currently not possible to deploy a new environment. I'll try to fix it

lorenzo-cavazzi:
I somehow managed to fix the problem. The PR is now ready for the review and a preview is available here https://lorenzotest.dev.renku.ch/

src/notebooks/Notebooks.container.js
```diff
@@ -50,9 +50,15 @@ class Notebooks extends Component {
       this.coordinator.setNotebookFilters(props.scope, true);
     }
 
+    this.state = {
+      showingLogs: false
+    };
+
```
vfried: I'm wondering if this should be in the store... i understand is internal state of the component, if you want to leave it here is ok.

src/notebooks/Notebooks.present.js
```diff
@@ -401,28 +425,22 @@ class NotebookServerRowAction extends Component {
...
-      */
+      defaultAction = (<Button {...classes} onClick={() => this.props.toggleLogs(name)}>Get logs</Button>);
     }
 
-    const alternateToggleStyle = {whiteSpace: "nowrap", borderRadius: "0.2rem",
-      paddingRight: "0.5625rem", paddingLeft: "0.5625rem"};
+    const alternateToggleStyle = {
+      whiteSpace: "nowrap", borderRadius: "0.2rem",
+      paddingRight: "0.5625rem", paddingLeft: "0.5625rem"
+    };
```
vfried: This looks a bit like css, shouldn't this be in the Project.css file or some css file

lorenzo-cavazzi: good point! It was not added in this PR (I guess it appears as changed because I run the auto-formatter on the page) but I can address this easily since we are already touching the code.

lorenzo-cavazzi:
I agree with the sizing concerns, I have made the modals bigger on wider screens and the text a bit smaller.

I am not sure about changing the letter spacing and background color. I think the font used on `<pre>` elements is very readable, and the terminal used in environments for new projects currently has a white background.

src/notebooks/Notebooks.container.js
vfried: You are right

lorenzo-cavazzi:
> I can't see the changes on size, i assume they haven't been deployed to lorenzotest, i will aprove the changes assuming that.

Right I didn't update the preview there :man_facepalming: 
Let me know if you prefer me to update it before merging the PR.

rokroskar:
Nice feature, I think this will still be useful for debugging issues with libraries and kernels. But - why don't I see all the logs? Seems like a very limited number of lines is shown.

lorenzo-cavazzi:
Currently, the last 100 lines are shown -- we can change this setting or allow the user to personalize it.
Apart from that, I see on the UI the same lines I see on rancher. Is it different for you?

lorenzo-cavazzi:
There was an issue reported by @ciyer when multiple pods were running.

I fixed that and added
* project's info on the Modal header
* Download button to save a log (mind that the file name in the preview is not up-to-date, the correct name is `Logs_{server_name}.txt`)

I updated the preview on https://lorenzotest.dev.renku.ch/

lorenzo-cavazzi:
> Very useful! I added a commit to make error handling more robust. Let us wait until #580 is merged and then rebase this against master, then I can approve.

ok great!
I added another commit to make the modal scrollable "in-place" and to improve the sizing on smaller windows.

rokroskar:
> Currently, the last 100 lines are shown -- we can change this setting or allow the user to personalize it.
> Apart from that, I see on the UI the same lines I see on rancher. Is it different for you?

It's probably the same, but I hate looking at logs in Rancher :) I guess this is a reasonable first implementation for now to avoid crashing the browser. 👍

PR #647, opened 2019-10-14
feat(dataset): datasets general search

### LINKED ISSUE #521: Search for datasets
The user should be able to search for datasets. 

For this we need to:
[ ] Create a new menu item next to the projects and notebooks in the top bar
[ ] Add a tab so the user can see their own datasets. 
       --> If the user is logged out a message under your project should be shown inviting the user to log in or create an account.
       --> If the user is logged in, a list of the datasets of the users should be displayed
[ ] Add a tab so the user can search for public datasets.
        --> When the user lands on this page, all datasets should be displayed.
        --> The user can search by text and order by relevant parameters.
        --> There should the same pagination system as in projects search.
        --> When the user clicks on a dataset it should be redirected to the details page of the dataset (On issue: #522)

**As a first step we will just search for datasets but not do the "your datasets" tab, and the search box will be just a free-text search box**

Mockups: 

image

image

Related to https://github.com/SwissDataScienceCenter/renku/issues/594

vfried:
Closes #521-datasets-search

Allow the users to search for datasets in the KG.

image

Testing virginiatest.dev.renku.ch  --> i will deploy again when collaboration tab is merged.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -30,12 +30,14 @@ import { BrowserRouter as Router, Route, Switch, Redirect }  from 'react-router-
 
 import Project from './project/Project'
 import Issue from './issue/Issue'
+import DatasetList from './dataset/list/DatasetList.container'
 import { Landing, LoggedInNavBar, AnonymousNavBar, FooterNavbar } from './landing'
 import { Notebooks } from './notebooks';
 import { Login } from './authentication'
 import Help from './help'
 import NotFound from './not-found'
 import { UserAvatar } from './utils/UIComponents'
+import ShowDataset from './dataset/Dataset.container'
 
 class RenkuNavBar extends Component {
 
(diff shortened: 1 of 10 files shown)


vfried:
Rebased and ready to be tested....
https://virginiatest.dev.renku.ch/datasets?q=dat*&orderBy=projectsCount&orderSearchAsc=true

lorenzo-cavazzi:
Great feature!

I would personally add by default a `*` to any search since that would be the standard behavior most of the users expect from a search feature. Maybe ticking an `advanced search` box could remove the `*` and show the help tips for using all the special chars.

I was playing a little bit on `virginiatest` and I noticed that the number of projects where the dataseta is used is sometimes bigger than the actual number I see listed once I click on it.
I initially thought this is due to some private projects that can't be listed due to privileges restriction but in some cases the difference seems to be really big. Is that number correct?

Screenshot_20191105_090540
Screenshot_20191105_090556

vfried:
I did a small fix in the layout, should be ready to be tested

PR #654, opened 2019-10-25
Add a copy button for project repository urls

### LINKED ISSUE #653: Copy button for project URLs and file paths
# Description

Project URLs should have a copy button

image

ciyer:
Close #653.

Preview available at `https://sekhar.dev.renku.ch/`

Added copy to clipboard button to:

**File View**
image

**Settings View**
image

### DIFF
--- a/src/file/File.present.js
+++ b/src/file/File.present.js
@@ -32,7 +32,7 @@ import faGitlab from '@fortawesome/fontawesome-free-brands/faGitlab';
 
 import { FilePreview } from './index';
 import { CheckNotebookStatus, CheckNotebookIcon } from '../notebooks'
-import { Loader } from '../utils/UIComponents';
+import { Clipboard, Loader } from '../utils/UIComponents';
 import { Time } from '../utils/Time';
 
 const commitMessageLengthLimit = 120;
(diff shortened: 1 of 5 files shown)


src/file/File.present.js
```diff
@@ -79,6 +79,9 @@ class FileCard extends React.Component {
       <CardHeader className="align-items-baseline">
         {this.props.lfsBadge}
         {this.props.filePath}
+        &nbsp;
+        <Clipboard clipboardText={this.props.filePath} />
```
lorenzo-cavazzi: This currently copies only the relative path (e.g. `notebooks/zhbikes-notebook.ipynb`). Wouldn't it be more useful if it copied the full file url?

rokroskar: We may need two buttons or some way of selecting which you want - the idea is to provide an easy way to get the path so it can be used with the `--source` flag to `renku dataset add`

rokroskar: but I agree that in another context you may want the fill URL

lorenzo-cavazzi: yes, it depends on what the user is doing. I am not sure if adding another button to copy the full url is a good idea, I fear it may be confusing. We can probably merge this PR as it is and re-consider the full URL if users will ask for that

PR #655, opened 2019-10-29
feat(state): allow using immutability-helper commands on model objects

### LINKED ISSUE #631: Add a force parameter to `Model.setObject`
The `Model.setObject()` function allows setting different properties at the same time, triggering a single refresh to the UI (a single `UPDATE` actions goes through the redux store).

Anyway, since it doesn't touch/overwrite properties of an object that are not explicitly added to the object, it's sometimes needed to split it one or multiple `Model.set()` calls.
E.G. reset `filter.commit` data
> `Model.get("filter.commit")`
> { id: 1234, author: xxx, active: true }
>`Model.setObject("filter.commit", {})`
> { id: 1234, author: xxx, active: true }
>`Model.setObject("filter.commit", { active: false })`
> { id: 1234, author: xxx, active: false }
>`Model.set("filter.commit", {})`
> { }

Since `Model.set()` works on a single variable, it's desirable to have a mechanism to force overwriting the involved objects when using `Model.setObject()`.
> `Model.get("filter.commit")`
> { id: 1234, author: xxx, active: true }
>`Model.setObject("filter.commit", { active: false }, true)`
> { active: false }
>`Model.setObject("filter.commit", { }, true)`
> { }

lorenzo-cavazzi:
Allow using any command available from the immutability-helper library https://github.com/kolodny/immutability-helper#available-commands

This enables to use `.setObject()` to perform atomic updates to the store even when trying to remove object attributes or array items, without forcing to invoke `.set()`. A few examples of ` notebooks` model changes are provided in the second commit.

Unexpected behavior may happen when trying to update attributes starting with `$`, but this doesn't look to be a big deal.

### How to test
A bit tricky, it requires to run the PR through telepresence
* Start 2 interactive environments
* Don't perform any operation on the first, do something on the second (open it and try a command in the command line)
* Click on the action "Get logs". With this PR in place, the logs are the same as visualized on rancher/kubectl. By changing the code on Notebooks.state.js from `data: { $set: lines }` to `data: lines`, the logs for the first notebook appear mixed with the second one's logs, because the `data` variables is not cleaned properly as defined by the`$set` command.

fix #631

### DIFF
--- a/src/model/Model.js
+++ b/src/model/Model.js
@@ -459,8 +459,6 @@ function nestedPropertyAccess(propAccessorString, obj) {
   return leaf;
 }
 
-window["callImmutableUpdate"] = immutableUpdate;
-
 // A redux reducer that will handle immutability-helper
 // updates.
 function modelUpdateReducer(state, action) {
--- a/src/model/Model.js
+++ b/src/model/Model.js
@@ -422,11 +422,14 @@ function updateObjectFromObject(obj, currentObject){
     if (obj[prop] instanceof Object && currentObject[prop]) {
       updateObj[prop] = updateObjectFromObject(obj[prop], currentObject[prop])
     }
+    else if (prop[0] === "$" && prop.length && prop.length > 1) {
+      updateObj[prop] = obj[prop];
+    }
     else {
-      updateObj[prop] = {$set: obj[prop]}
+      updateObj[prop] = { $set: obj[prop] };
     }
   });
-  return updateObj
+  return updateObj;
 }
 
 // Create a mongodb-style update object from a plain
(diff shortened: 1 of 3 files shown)


lorenzo-cavazzi:
Definitely!

I added a somehow verbose set of tests to act as documentation. The idea is to make it explicit that `.setObject` on existing objects and arrays doesn't overwrite them, therefore the `$set` command is needed in those cases.
Even if a similar result could be achieved by invoking multiple times `.set()`, an update operation like the one in the last test couldn't be performed in a single iteration before this PR, leading to possibly inconsistent states (e.g. `fetched` variable `true`, but `items` variable still empty).

PR #657, opened 2019-10-30
Update libraries to fix issues in audit

ciyer:
We have not updated the libraries in a while and there are some issues that have been discovered in the dependencies. This update should fix currently known critical issues.

An issue was found with mouse pointer flickering caused by tooltips on mouseover in Chrome (e.g., the launch, gitlab, and lineage icons in the file view). This was a problem before this update, and we have only upgraded some of the libraries to avoid having this problem. After this update, it occasionally occurs, but in a manageable amount. Hopefully, updating the libraries to the latest major version (#666) will resolve this problem.

The PR is available for testing at https://sekhar.dev.renku.ch/


vfried:
All works, as lorenzo said we do get some warning, also when runing tests (for componentWillMount and componentWillReceiveProps) we should make new issues for replacing this.

I also get an error for the notebook display... but the test passes and the notebook files can be displayed.
image

ciyer:
This has been superseded by #673, which is a superset of these changes.

PR #669, opened 2019-11-05
chore: prepare 0.7.1 release

ciyer:
Updates to Changelog for the 0.7.1 release.


vfried:
Isn't this missing ?
image

lorenzo-cavazzi:
@vfried I think they are already set at `0.7.1`, they should be updated to `0.7.2` after the current release as in #670

Issue #524, opened 2019-07-15
Create a dataset in project


vfried:
[restricted the issue to just creating a dataset within a project]

The dataset tab of a project should have a button for creating a dataset.

The page for creating a dataset should have a form with the following parameters:
- Name
- Description
- Files (there should be a metadata file and 0 or more data files)

![image](https://user-images.githubusercontent.com/1196411/67273750-5bf19280-f4bf-11e9-9091-9e016eea7c86.png)

![image](https://user-images.githubusercontent.com/1196411/67270457-01553800-f4b9-11e9-9f88-1ecbbd4a77b3.png)

Related to #594

ciyer:
I think the ability to create a dataset should also be in project, and creating a dataset from dataset search will require providing a project as well.

vfried:
We should also allow the user to upload files from a URL (like in the CLI)

rokroskar:
can this be closed?

PR #679, opened 2019-11-15
feat(dataset): create a dataset using the ui

vfried:
This PR addresses #524 

To create the form I implemented a form generator that will allow us to create a form from from a Schema. This can then be reused in case we need to create other forms.

In this case:

**Schema:**
image

**Call to the Form Generator (Form Panel)**
image

**Generated Form**
image

**Form with errors**
image

The text editor uses CKEditor
https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/frameworks/react.html

The file uploader uses FilePond
https://pqina.nl/filepond/

**### TO BE DONE:**
- Tests for the form generator
- Call to the backend to create the dataset and upload files (submitCallback)


ciyer:
The generic form framework is very cool!

vfried:
TO-DO:
- Improve upload of large files
- Progress bar for file uploads
- Fix redirection (a dataset is not visible until after you refresh the page), this could be because the KG is adding the dataset
- Retry button for when upload fails

src/project/datasets/new/DatasetNew.present.js
```diff
@@ -0,0 +1,105 @@
...
+      props.history.push({pathname: `/projects/${props.projectPathWithNamespace}/datasets`});
+     }
+  });
+  }
+
+  if(props.accessLevel < ACCESS_LEVELS.MAINTAINER){
+    return <Col sm={12} md={8} lg={10}>
+    <Alert timeout={0} color="primary">
+      Acces Denied. You don't have rights to create datasets for this project.<br /><br /> 
+      <FontAwesomeIcon icon={faInfoCircle} />  If you where recently added to this project try refreshing the page.
```
ciyer: Text: `If you were recently given access to this project, you might need to refresh the page first`.

Could we add a button that the user can click to refresh the page?

src/utils/formgenerator/fields/FileUploaderInput.js
```diff
@@ -0,0 +1,242 @@
...
+      if(response.status >=400) throw new Error();
+      response.json().then((body) => {
+        if (body.error) {
+          setFilesErrors(prevFilesErrors => [...prevFilesErrors, 
+            getFileObject(file.name, file.size, undefined, body.error.reason, undefined )]
+          )
+          return [];
+        } else {
+          let new_file = body.result.files[0];
+          if(body.result.files[0].file_name !== undefined && body.result.files[0].file_name !== file.name){
```
ciyer: I think this can be:
```if (new_file.file_name !== undefined && new_file.file_name !== file.name) { ```

File: src/utils/formgenerator/fields/FileUploaderInput.js [hunk 3]
```diff
@@ -0,0 +1,242 @@
...
+            new_file = getFileObject(file.name, new_file.file_size, new_file.file_id, undefined, new_file.file_name)
```
ciyer: Maybe call this variable `newFileObj` to avoid confusion

src/utils/formgenerator/fields/FileUploaderInput.js
```diff
@@ -0,0 +1,242 @@
...
+  let onDropOrChange = (e) => {
+    e.preventDefault();
+    e.persist();
+    if(!disabled){
+      let files = e.dataTransfer ? e.dataTransfer.files : e.target.files;
+      let droppedFiles = getFilteredFiles(Array.from(files))
+      droppedFiles.map(file => uploadFile(file));
+      setFiles([...files, ...droppedFiles]);
+      const newDisplayFiles = droppedFiles
+        .map(file => ( getFileObject(file.name, file.size, null, undefined, undefined)))
```
ciyer: I think you don't need one set of parents:
```.map(file => getFileObject(file.name, file.size, null, undefined, undefined))```

File: src/utils/formgenerator/fields/FileUploaderInput.js [hunk 5]
```diff
@@ -0,0 +1,242 @@
...
+              <th className="font-weight-light">Delete</th>
+            </tr>
+          </thead>
+          <tbody className={disabled? "disabled-input":""}>
+            { displayFiles.map((file, index) => (
+              <tr key={file.file_name + "file"} onClick={()=>{}}>
+                <td>{index + 1}</td>
+                <td>
+                  <span>{file.file_name}</span>
+                  {file.file_alias ? <small><br></br><span className="text-danger"> *File was renamed to:{file.file_alias}</span></small> : null}
```
ciyer: We should probably mention why the file was renames:
```File name, <i>{file.file_name}</i>, contains disallowed characters; it has been renamed to <i>file.file_alias}</I>```

**Comment by vfried** [comment]
I just pushed a fix for the requested changes, I also changed what happens after the form is submitted.

The function now:
- Waits until the dataset is created
- After that a timer is set where the function fetches the datasets of the project and checks if the dataset is there, if it is there the user is redirected the user to HTTPS://...datasets/new-dataset-id (Before this the timeout is cleared)

**Comment by vfried** [comment]
Changed the text editor so it doesn't use the CKEditor and added message in case the dataset isn't found and was just created.

File: src/project/datasets/DatasetsListView.js [hunk 6]
```diff
@@ -20,6 +21,23 @@ function DatasetListRow(props){
...
+function CreateDatasetButton(props){
+  if(props.visibility.accessLevel >= ACCESS_LEVELS.MAINTAINER)
+    return <span className="float-right throw-right-in-flex">
+      <UncontrolledButtonDropdown size="sm">
+        <DropdownToggle color="primary" className="alternateToggleStyle" caret>
+          <FontAwesomeIcon icon={faPlus} style={{ color: 'white', backgroundColor: "#5561A6" }} />
+        </DropdownToggle>
+        <DropdownMenu right={false}>
+          <DropdownItem>
+            <Link to={props.newDatasetUrl}>New Dataset</Link>
```
ciyer: Could you also add a link to the newDatasetUrl to the top "+" menu?

 image

src/model/RenkuModels.js
```diff
@@ -218,5 +219,45 @@ const notebooksSchema = new Schema({
   }
 });
 
+const datasetFormSchema = new Schema({
+  name: {
+    initial: "",
+    name: 'name',
+    label: 'Name',
+    type: FormGenerator.FieldTypes.TEXT,
+    parseFun: expression => FormGenerator.Parsers.slugFromTitle(expression),
```
ciyer: I like how you set the form up to automatically convert input into a slug. Very slick!

src/project/datasets/DatasetsListView.js
vfried: Done

vfried:
I added a retry button for when the upload fails.

Added help text for the description of the dataset.

Regarding the 404 on the dataset, i do wait to see if the dataset was created in the KG and then redirect the user, i think that even though the kg knows it has a new dataset sometimes it can't fetch it. We should address this in a new pr though. I did add a message so the users know that the dataset could be there even if they get a 404.

Testing available in: datasetstest.dev.renku.ch

ciyer:
This is a very nice new feature!

PR #708, opened 2019-12-10
fix: correctly handle nested groups in the URLs

### LINKED ISSUE #716: 404 Flash when navigating to a project by ID
# Description

When navigating to a project by project ID, there is a brief flash of a 404 page before the actual content for the project appears. This should be fixed.

ciyer:
Solve the display side of https://github.com/SwissDataScienceCenter/renku/issues/716

Full support will be in another issue.

Deployed on https://sekhar.dev.renku.ch/ for testing.

### DIFF
--- a/src/file/File.container.js
+++ b/src/file/File.container.js
@@ -209,7 +209,7 @@ class JupyterButton extends React.Component {
  * File content display container component
  *
  * @param {Object} client - api-client used to query the gateway
- * @param {string} filePath - path to the file
+ * @param {string} filePath - path to the file - for the JupyterNotebook button. See docs for JupyterButton
  * @param {string} branchName - optional branch name, defaults to master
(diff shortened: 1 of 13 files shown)


src/file/File.container.js
```diff
@@ -210,6 +210,7 @@ class JupyterButton extends React.Component {
  *
  * @param {Object} client - api-client used to query the gateway
  * @param {string} filePath - path to the file
+ * @param {string}  - for the JupyterNotebook button. See docs for JupyterButton
```
lorenzo-cavazzi: the second description line for the same parameter shouldn't start with `@param {string}`

```
 * @param {string} filePath - path to the file for the JupyterNotebook
 *    button. See docs for JupyterButton
```

ciyer: Oops! Fixed now.

PR #709, opened 2019-12-11
chore: fix npm dependencies

ciyer:
Fix dependencies so that linting works correctly in VS Code.


ciyer:
I have fixed that problem.

Issue #711, opened 2019-12-16
UI Crashes on Display a Notebook


ciyer:
# Description

When I try to display this notebook, the UI crashes:

  https://dev.renku.ch/projects/cramakri/biktes-tutorial-testing/files/blob/notebooks/zhbikes-notebook.ipynb

ciyer:
The problem seems to be in the notebook rendering. If I comment out the render method in the `StyledNotebook` component, the UI does not crash.

PR #712, opened 2019-12-17
fix: update a more minimal set of deps

ciyer:
Even though npm may ask you do install it, do not include fibers for the moment! It causes problems for notebook rendering
"fibers": "^4.0.2"


ciyer:
Fix #711

PR #717, opened 2020-01-10
feat(environments): read project level notebook options

### LINKED ISSUE #481: Fetch server options from the project config files
After SwissDataScienceCenter/renku-project-template#29 is finished, we should fetch the server options for running a server from the metadata.yaml.

After the user selects which server he wants to run we should show the right links/icons in the UI (in notebook server and in the file view).

lorenzo-cavazzi:
When starting a new interactive environment, the UI now checks for project options available in the `.renku/renku.ini` file, it tries to apply the settings and it displays popup warnings where needed.
More info in the docs here: SwissDataScienceCenter/renku#723

**HOW TO TEST**
A preview is available at https://lorenzotest.dev.renku.ch/
This is a project with some already defined custom options: https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/projectoptions

fix #481

Screenshot_20200110_172044

### DIFF
--- a/src/notebooks/Notebooks.state.js
+++ b/src/notebooks/Notebooks.state.js
@@ -99,7 +99,8 @@ class NotebooksCoordinator {
         namespace: null,
         project: null,
         branch: { $set: {} },
-        commit: { $set: {} }
+        commit: { $set: {} },
+        options: { $set: {} }
       },
       pipelines: {
         fetched: null,
@@ -258,10 +259,10 @@ class NotebooksCoordinator {
             if (parsedOption === "default_url")
               option = "defaultUrl";
             let value = parsedOptions[parsedOption];
-            if (value === "true") {
+            if (value && value.toLowerCase() === "true") {
               value = true;
             }
-            else if (value === "false") {
+            else if (value && value.toLowerCase() === "false") {
               value = false;
             }
             else if (!isNaN(value)) {
@@ -326,11 +327,10 @@ class NotebooksCoordinator {
     let warnings = [];
     Object.keys(projectOptions).forEach(option => {
       const optionValue = projectOptions[option];
-      if ((globalOptions[option]
-        && globalOptions[option].options
-        && globalOptions[option].options.length
-        && globalOptions[option].options.includes(optionValue))
-        || option === "defaultUrl"
+      if (
+        (globalOptions[option] && globalOptions[option].type !== "enum") ||
+        (globalOptions[option] && globalOptions[option].options && globalOptions[option].options.length && globalOptions[option].options.includes(optionValue)) ||
+        (option === "defaultUrl")
       ) {
         filterOptions[option] = optionValue;
       }
--- a/src/notebooks/Notebooks.state.js
+++ b/src/notebooks/Notebooks.state.js
@@ -335,7 +335,7 @@ class NotebooksCoordinator {
         filterOptions[option] = optionValue;
       }
       else {
-        warnings.push(option);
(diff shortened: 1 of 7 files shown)


src/notebooks/Notebooks.state.js
```diff
@@ -205,21 +214,133 @@ class NotebooksCoordinator {
...
+    const filters = this.getQueryFilters();
+    const requestFilters = JSON.stringify(filters); // keep these to verify answer validity
+    const projectId = `${encodeURIComponent(filters.namespace)}%2F${filters.project}`;
+    this.model.setObject({
+      options: {
+        fetching: true,
+        warnings: { $set: [] }
+      }
+    });
+    this.client.getRepositoryFile(projectId, RENKU_INI_PATH, filters.commit, 'raw')
```
ciyer: Could the function argument to `getRepositoryFile` be broken into a few functions? I think that would help readability.

src/notebooks/Notebooks.state.js
```diff
@@ -205,21 +214,133 @@ class NotebooksCoordinator {
...
+        filterOptions[option] = globalOptions[option].default;
+      }
+    })
+
+    // try to overwrite default options with project options
+    let warnings = [];
+    Object.keys(projectOptions).forEach(option => {
+      const optionValue = projectOptions[option];
+      if (
+        (globalOptions[option] && globalOptions[option].type !== "enum") ||
```
ciyer: Could the inside of this `if` be extracted to a function? Something like `isOptionSupported`?

src/utils/HelperFunctions.js
```diff
@@ -63,5 +63,35 @@ function simpleHash(str) {
   return ("0000000" + (hval >>> 0).toString(16)).substr(-8);
 }
 
+function parseINIString(data) {
```
ciyer: Could you write a few tests for this?

src/notebooks/Notebooks.state.js
lorenzo-cavazzi: Since most of the code was meant to convert the renku.ini raw content, I moved to its own function `NotebooksHelper.parseProjectOptions` with proper test coverage.
Also, the logic in the following `catch` function for handling missing files was somehow redundant. That case is now treated as an empty file

src/notebooks/Notebooks.state.js
lorenzo-cavazzi: Definitely. I moved those checks to its own function `NotebooksHelper.checkOptionValidity`, improved the logic to include type checking and added proper test coverage.

src/utils/HelperFunctions.js
lorenzo-cavazzi: Added

lorenzo-cavazzi:
I was a bit sloppy in reviewing the code before pushing.
I have now simplified the logic as suggested and added the corresponding tests
.
The preview on https://lorenzotest.dev.renku.ch/ has been updated.

ciyer:
Code looks good! I'm testing this feature, and I followed the documentation to request an amount of memory
```
* `mem_request [float]`: memory quota (in GBs)
```

and have created a renku.ini file
```
[renku "interactive"]
mem_request = 2
```
But it is not working as expected.

image

lorenzo-cavazzi:
Oh right, the `mem_request` field is not `2` but `2G`. I have to modify the docs to avoid confusion

lorenzo-cavazzi:
I have just pushed a fix for that SwissDataScienceCenter/renku#723

PR #730, opened 2020-01-23
Add link to discourse to help page

### LINKED ISSUE #710: Add UI Link to Discourse
# Description

We have a Discourse page set up: https://renku.discourse.group. The UI should include links to it.

ciyer:
Fix #710

See preview at https://sekhar.dev.renku.ch/help

### DIFF
--- a/src/help/Help.present.js
+++ b/src/help/Help.present.js
@@ -23,19 +23,36 @@
  *  Presentational components for help.
  */
 
+import React, { Component } from "react";
+import { Route } from "react-router-dom";
+import { RenkuNavLink } from "../utils/UIComponents";
 
-import React, { Component } from 'react';
-import { Route }  from 'react-router-dom';
-import { RenkuNavLink } from '../utils/UIComponents';
+import { Row, Col } from "reactstrap";
+import { Nav, NavItem } from "reactstrap";
 
-import { Row, Col } from 'reactstrap';
-import { Nav, NavItem } from 'reactstrap';
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faDiscourse, faGithub, faGitter } from "@fortawesome/free-brands-svg-icons";
 
+function ExternalDocsLink(props) {
+  return (
+    <a href={props.url} target="_blank" rel="noreferrer noopener">
+      {props.title}
+    </a>
+  );
+}
+
+function ExternalDocsIconLink(props) {
+  return (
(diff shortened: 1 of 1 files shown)


src/help/Help.present.js
```diff
@@ -57,19 +62,23 @@ class HelpGetting extends Component {
   render() {
     return (
       <div>
+        <h2>Discourse</h2>
+        <p>
+          We maintain a forum on <ExternalDocsLink url="https://renku.discourse.group" title="Discourse" /> for
+          discussion about Renku. This is a good place to ask questions and find answers.
```
vfried: *discussions? (I'm not sure this is correct or not, just asking)

rokroskar:
How easy would it be to somehow template these pages or otherwise disconnect them from the code? I'm thinking that non-sdsc deployments might want to put different information here.

src/help/Help.present.js
ciyer: Both _discussion_  and _discussions_ can be used here grammatically because discussion is both a countable and an uncountable noun: https://en.wiktionary.org/wiki/Appendix:Glossary#uncountable

There is a subtle difference in meaning, but not a fundamental one.

ciyer:
It would be simple to let deployments specify the content of the help page in Markdown and render that instead of our default page. 

That works for static content like this, but we are also thinking about a plug-in mechanism to make it possible to extend Renkulab and would allow introduction of dynamic content, and maybe it would be better to structure help as a plug-in.

lorenzo-cavazzi:
I like the new content of "Getting Help", it definitely looks better than before!

A minor layout consideration: when the page becomes larger, the content is not aligned anymore on the left and, depending on the browser width, there may be a big gap. Does it make sense to keep it aligned on the left as it's the case with other "tabbed" pages? (e.g. a project's page)

Screenshot_20200127_085752

Screenshot_20200127_085909

rokroskar:
> That works for static content like this, but we are also thinking about a plug-in mechanism to make it possible to extend Renkulab and would allow introduction of dynamic content, and maybe it would be better to structure help as a plug-in.

Seems like markdown support would get us most of the way there - why would you need dynamic content on the help pages?

ciyer:
Dynamic content on the help page is probably not necessary, but a plug-in that would display on the _Project_ page would need to be dynamic, and I think we should have just one plug-in mechanism, not different ones depending on whether content is static or dynamic.

ciyer:
@lorenzo-cavazzi I switch to a Flexbox-based layout. Try this one out.

Issue #733, opened 2020-01-27
Add support for group avatar


vfried:
Right now we only display avatars for users, we should also display the avatars of groups.

PR #741, opened 2020-01-29
refactor(state): use redux model to store user data

### LINKED ISSUE #632: Move UserState under the Schema/Model representation
The user's data are managed by `src/app-state/UserState.js` that is used to create the `renku` redux store in the standard redux way instead of using our Schema/Model representation.

This should be refactored to have a `user` sub-model as part of the general model, and a user coordinator to provide the required functions used by the UI components.
Tip: be careful when moving the `UserState.reducer` since it contains multiple actions and our representation contains only `UPDATE`

lorenzo-cavazzi:
User data have been migrated to the global model.
Most data are now part of the `user` sub-schema, while projects related data are in the `projects` sub-schema.

This required a small re-factory of multiple components. Please take some time to test it.

***Hot to test it***
A preview is available on https://lorenzotest.dev.renku.ch/
Try to navigate (keeping the javascript console open to notice hidden errors) and do some actions involving user-related content, e.g. :
* star/unstar project
* create new project
* search through personal/starred projects
* starting new interactive environment

fix #632


lorenzo-cavazzi:
A fix for the tests is coming, it doesn't have an impact on the test version

lorenzo-cavazzi:
Fixed and re-deployed to https://lorenzotest.dev.renku.ch/ :heavy_check_mark:

src/contribution/Contribution.present.js
```diff
@@ -17,32 +17,35 @@
...
-  Nav, NavLink, DropdownMenu, DropdownItem, DropdownToggle, Dropdown} from 'reactstrap';
+import {
+  Row, Col, Button, Input, TabContent, TabPane, NavItem,
+  Nav, NavLink, DropdownMenu, DropdownItem, DropdownToggle, Dropdown
+} from 'reactstrap';
 import classnames from 'classnames';
 import Collapse from 'react-collapse';
 
-import { Avatar, TimeCaption, RenkuMarkdown } from '../utils/UIComponents'
+import { UserAvatar, TimeCaption, RenkuMarkdown } from '../utils/UIComponents';
```
vfried: We might want to leave this as "Avatar" instead of "UserAvatar"... since we will need to add the avatar also to groups in #733

vfried: I think this is a small change request, we could just merge and change this when we address #733

lorenzo-cavazzi:
@vfried I agree on that. I think all the relevant rules are defined using eslint (the file with the rules is `.eslintrc.json`). On VSCode, I use the extension `ESLint` and, when I auto-format, it applies a bit of changes.
I think we could enforce requiring proper formatting when we submit a PR -- in renku-python they are already doing that (tests fail if code is not properly formatted)

PR #760, opened 2020-02-11
feat(collaboration): improvements to reach UI consistency inside issues

### LINKED ISSUE #690: UI Consistency - Collaboration / Issues
# Description

The display of the issues should be modified a bit to provide more information and be consistent to the rest of the UI.

image

image

Use [fa-comments](https://fontawesome.com/icons/comments?style=solid) and [fa-list-ul](https://fontawesome.com/icons/list-ul?style=solid) for the icons.

vfried:
Closes #690 

The ui looks like this with the change...
image

image


lorenzo-cavazzi:
Another detail that can be improved is the empty issue that is quickly displayed when opening an issue from the list.
If it's a one-liner fix (maybe just adding a `loader=true` variable while the API call is ongoing), it would be great to include it in this PR. Otherwise, I can create another issue for that

vfried:
Added loader for when issue is loading.

Added error message for when the issue is not found or there is some other error.

PR #761, opened 2020-02-11
Add support for sentry

### LINKED ISSUE #393: Integrate Sentry
We should use sentry to track exceptions.

  https://docs.sentry.io/error-reporting/quickstart/?platform=browsernpm

lorenzo-cavazzi:
This PR introduces the support for tracking expectations in Sentry.
It works, but any suggestions for improvements are welcome :slightly_smiling_face: 

**How to test**
The errors are reported here https://sentry.dev.renku.ch/swissdatasciencecenter/renku-ui/

* Using telepresence: execute the telepresence script with "SENTRY=1" `SENTRY=1 ./run-telepresence.sh`. Generate any error, e.g. by adding `throw Error("fake")` in `index.js` **after** `// configure Sentry`
* On a deployment: you can deploy it to any namespace after updating the value file as per SwissDataScienceCenter/sdsc-dev-docs#57 . Remember to add a fake error as pointed out above.
* It is temporarily possible to generate fake errors by visiting https://lorenzo.dev.renku.ch/
The errors take up to one or two minutes to show up.

fix #393

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -20,7 +20,8 @@ echo "==================================================="
 echo " Configuration:"
 echo " GATEWAY_URL=${GATEWAY_URL:-http://gateway.renku.build}"
 echo " BASE_URL=${BASE_URL:-http://renku.build}"
-echo " SENTRY_DNS=${SENTRY_DNS}"
+echo " SENTRY_URL=${SENTRY_URL}"
+echo " SENTRY_NAMESPACE=${SENTRY_NAMESPACE}"
 echo " RENKU_TEMPLATES_URL=${RENKU_TEMPLATES_URL}"
 echo " RENKU_TEMPLATES_REF=${RENKU_TEMPLATES_REF}"
 echo "==================================================="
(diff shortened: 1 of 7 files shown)


jachro:
Hi Lorenzo. One comment. Not so sure where would be the best place for that in the UI code but I found it very useful to add an `environment` query parameter to the Sentry DSN so then you know where the error occurred.

lorenzo-cavazzi:
Thanks for the suggestion!
By `environment` you mean the namespace where it comes from? In this case, for the UI it's easier because we can easily derive that from the `url` tag.

This is an exception example we get with the current configuration.

Screenshot_20200212_082258

jachro:
Cool! That looks really nice and indeed the env is maybe not needed.

lorenzo-cavazzi:
On second thought, it makes sense to add an `environment` variable in the same style of `renku-graph` -- as @rokroskar pointed out, it's useful to filter events cross-projects. Thanks again @jachro for the suggestion.

I also took the time to add user data to the error.

**TESTING**
* You can generate fake errors from https://lorenzotest.dev.renku.ch/ (just click on the link and refresh the page a few times) and verify that Sentry correctly collects them from https://sentry.dev.renku.ch/swissdatasciencecenter/renku-ui/issues/
* for telepresence/personal namespace deployments, please follow the instructions on the first post. Be sure to add data to the value file as per SwissDataScienceCenter/sdsc-dev-docs#58

lorenzo-cavazzi:
I need also a @SwissDataScienceCenter/kartografen review :slightly_smiling_face:

PR #773, opened 2020-02-19
feat(lineage): adapt to the lineage endpoint update

lorenzo-cavazzi:
The UI lineage needs to be adapted to prevent it from failing due to the new nodes id format.
SwissDataScienceCenter/renku-graph#248 adds extra properties to the lineage query we can use in order to stop guessing nodes properties from the id.

I mixed a few fixes in this PR
* Update lineage query to get more data and stop guessing nodes properties from the id
* Fix location issues that prevented switching correctly between files and lineage
* Allow viewing folders lineage
* Remove unnecessary scrollbar on svg elements

Please take some time to check the lineage of a few projects.
Preview available here:  https://lorenzotest.dev.renku.ch/

### DIFF
--- a/src/api-client/graph.js
+++ b/src/api-client/graph.js
@@ -74,7 +74,7 @@ function addGraphMethods(client) {
     const query = gql`
       query getLineage($projectPath: ProjectPath!, $filePath: FilePath!) {
         lineage(projectPath: $projectPath, filePath: $filePath) {
-          nodes { id, label, type },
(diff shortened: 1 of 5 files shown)


lorenzo-cavazzi:
Something looks a bit weird...
E.G. https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/flights-tutorial/files/lineage/data/output/2019-01-flights-filtered.csv

PR #774, opened 2020-02-20
feat(collaboration): added commit and discussion tabs and ui consistency improvements

### LINKED ISSUE #692: UI Consistency - Collaboration / Merge Requests
# Description

Make the feature set around merge requests a bit more complete and make the UI consistent with the rest of Renku, Issues in particular.

image

Like Issues, MRs should be expanded with a button to get back to the list.

image

vfried:
Closes #692 

The Merge Requests tab looks like this now...

image

image

image

image


src/merge-request/MergeRequest.present.js
```diff
@@ -17,65 +17,155 @@
...
-import { ExternalLink } from '../utils/UIComponents';
+function MergeRequestHeader(props) {
 
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { faLongArrowAltLeft as faLeftArrow }from '@fortawesome/free-solid-svg-icons';
+  const buttonGit = <ExternalIconLink tooltip="Open in GitLab" icon={faGitlab} to={props.externalMRUrl} />
 
+  const actionButton = props.showMergeButton ?
+    <TooltipToggleButton
+      onClick={props.onIssueStateChange} tooltip={`Merge`}
```
ciyer: This should be connected with the `onMergeClick` action. Right now, it is not possible to merge a merge request.

image

src/merge-request/MergeRequest.present.js
```diff
@@ -17,65 +17,155 @@
...
+      active={true}
+      activeIcon={faBoxOpen} inactiveIcon={faBox}
```
ciyer: Using an icon here is nice, but I do not think the box is a good choice for "merge". The Font-Awesome code-merge icons are all pro-only, so we cannot use those right now, but I think this one is not bad for merge:

  https://fontawesome.com/icons/project-diagram?style=solid

vfried: I agree with the fact that the box is bad... there is a problem with the project diagram, i think we use it for the graph view in the files explorer. I'll try to find some icon that works better for the merge requests.

vfried: I changed the iconf or a "merge icon", it looks like this now...
image

src/merge-request/MergeRequest.present.js
vfried: I just fixed that, it was a copy paste issue 🤦‍♂

PR #778, opened 2020-02-21
feat(datasets): user can import datasets though the UI

### LINKED ISSUE #528: Import datasets from different sources
The user should be able to import datasets from sources like Zenodo, etc.

vfried:
Closes #528 

There is a new button in the UI so the user can import datasets using the UI.

There are some things to be fixed in the backend for this to properly work. 

Button:
image

Form:
image

Deployment: https://virginiatest.dev.renku.ch/

### DIFF
--- a/src/api-client/dataset.js
+++ b/src/api-client/dataset.js
@@ -32,14 +32,14 @@ export default function addDatasetMethods(client) {
     if (controller) queryParams.signal = controller.signal;
 
     return fetch(`${client.baseUrl}/renku/cache.files_upload?override_existing=true`, queryParams)
-    .then(response => {
-       if (controller !== undefined)
+      .then(response => {
+        if (controller !== undefined)
           response.controller = controller;
         return response;
-    });
+      });
   };
 
-  client.addFilesToDataset = (projectUrl, datasetName, filesList, signal) => {
+  client.addFilesToDataset = (projectUrl, datasetName, filesList) => {
     let headers = client.getBasicHeaders();
     headers.append("Content-Type", "application/json");
     headers.append("X-Requested-With", "XMLHttpRequest");
@@ -54,7 +54,7 @@ export default function addDatasetMethods(client) {
     };
 
     return client.clientFetch(`${client.baseUrl}/renku/cache.project_clone`,
-    queryParams
+      queryParams
     ).then(response => {
       if (response.data.error) { return response; }
       else
@@ -91,6 +91,7 @@ export default function addDatasetMethods(client) {
         return response;
 
       project_id = response.data.result.project_id;
+
       return client.clientFetch(`${client.baseUrl}/renku/datasets.create`, {
         method: "POST",
         headers: headers,
@@ -100,7 +101,6 @@ export default function addDatasetMethods(client) {
           "project_id": project_id
         })
       });
-
     })
       .then(response => {
         if (response.data.error) { return response; }
@@ -117,5 +117,51 @@ export default function addDatasetMethods(client) {
             });
           } return response;
       });
+  };
+
+  client.datasetImport = (projectUrl, renkuDataset) => {
+    let headers = client.getBasicHeaders();
+    headers.append("Content-Type", "application/json");
+    headers.append("X-Requested-With", "XMLHttpRequest");
+
+    let project_id;
+
+    return client.clientFetch(`${client.baseUrl}/renku/cache.project_clone`, {
+      method: "POST",
+      headers: headers,
+      body: JSON.stringify({
+        depth: 1,
+        git_url: projectUrl
+      })
+    }).then(response => {
+      if (response.data.error !== undefined)
+        return response;
+
+      project_id = response.data.result.project_id;
+
+      return client.clientFetch(`${client.baseUrl}/renku/datasets.import`, {
+        method: "POST",
+        headers: headers,
+        body: JSON.stringify({
+          "dataset_uri": renkuDataset.uri,
+          "project_id": project_id
+        })
+      });
+
+    });
+  };
+
+
+  client.getJobStatus = (job_id) => {
+    let headers = client.getBasicHeaders();
+    headers.append("Content-Type", "application/json");
+    headers.append("X-Requested-With", "XMLHttpRequest");
+
+    return client.clientFetch(`${client.baseUrl}/renku/jobs/${job_id}`, {
+      method: "GET",
+      headers: headers
+    }).then(response => {
+      return response.data.result;
+    });
   };
 }
--- a/src/dataset/Dataset.present.js
+++ b/src/dataset/Dataset.present.js
@@ -132,14 +132,13 @@ export default function DatasetView(props) {
           }).catch(error => {
             if (fetchError === null) {
               if (!unmounted && error.case === API_ERRORS.notFoundError) {
-                setFetchError(
-                  "Error 404: The dataset that was selected does not exist or could not be accessed." +
-                  "If you just created the dataset try reloading the page."
-                );
-              }
+                setFetchError("Error 404: The dataset that was selected does" +
+                " not exist or could not be accessed. If you just created or" +
+                " imported the dataset try reloading the page.");
+            }
               else if (!unmounted && error.case === API_ERRORS.internalServerError) {
                 setFetchError("Error 500: The dataset that was selected couldn't be fetched.");
-              }
+            }
             }
           });
       }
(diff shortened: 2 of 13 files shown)


rokroskar:
preview?

vfried:
image

vfried:
Updated... and ready to be tested

image

I am creating a deployment...

rokroskar:
What is the URL that this lives on? What I mean is, can I do something like

```
https://renku/projects/myproject/datasets/import?url=doi://12345
```

This would make integration with third party pages like dataverse datasets much easier I guess?

vfried:
> What is the URL that this lives on? What I mean is, can I do something like
> 
> ```
> https://renku/projects/myproject/datasets/import?url=doi://12345
> ```
> 
> This would make integration with third party pages like dataverse datasets much easier I guess?

At the moment you can't do that but we can make a new issue to allow this to happen

rokroskar:
Yes! we should investigate what the best flow would be for this, but it would be super useful.

vfried:
> Yes! we should investigate what the best flow would be for this, but it would be super useful.

I edited my comment on top...

ciyer:
This is a nice feature. We should use it in the tutorial. I like the UI and the code looks great as well.

I have a few bugs that came up in testing:

1. "Enter" in the URI/DOI field causes the page to reload; I would expect it to submit the form

2. Put junk into the URI/DOI field, and you get "Please wait, dataset import will start soon" forever

image

3. If you try to import the same file twice, the server returns with an error, but the UI stays in "Please wait, dataset import will start soon"

image

image

vfried:
Fixed:

1. The enter was actually a problem on how the form was built, the button had the onClick event instead of the Form having the onSubmit event.

2. This is a backend problem, i also detected this but i forgot to say to sam. I actually stop the dataset import after a looong time of waiting (the problem is i don't want to just tell the user that dataset import failed too fast)

3. I also detected this problem and forgot to tell to sam, i am not getting anything in the UI that makes me realise that there is an error... all i get is that the job was ENQUEUED.

image

jsam:
I added tests for those cases and things seem to pass there correctly now.

2) (In tests) This case splits into 2 possible outcomes:
2.1) requested format is invalid in which case you should receive the following response:

```
{
   "result":{
      "updated_at":"2020-03-05T17:27:12.431676",
      "extras":{
         "error":"Invalid parameter value - Could not process junkjunkjunk.\nCannot parse URL."
      },
      "job_id":"f938c53eac7b43ed8b2462146cbfa7dd",
      "created_at":"2020-03-05T17:27:12.309677",
      "state":"FAILED"
   }
}
```
2.2) requested uri could not be resolve (for example invalid doi or url) in which case you should receive the following response:

```
{
   "result":{
      "created_at":"2020-03-05T17:32:12.233638",
      "updated_at":"2020-03-05T17:32:12.783466",
      "extras":{
         "error":"Invalid parameter value - Could not process 10.5281/zenodo.11111111111111111.\nURI not found."
      },
      "state":"FAILED",
      "job_id":"387d7e7970024b048be0fc272156f888"
   }
}

```

3) In this case you should receive response like this:

```
{
   "result":{
      "created_at":"2020-03-05T17:18:11.089451",
      "updated_at":"2020-03-05T17:18:30.178826",
      "extras":{
         "description":"pybossa-v3.1.2.zip",
         "total_size":714923,
         "progress_size":714923,
         "error":"Dataset exists: \"scifabricpybossa_v31_v312\"."
      },
      "job_id":"1cfc032c7cdd4791881cb8ae1c14ddcb",
      "state":"FAILED"
   }
}
```


Those error message are coming directly from the core module so they should match the same behavior as we have on the CLI. If you need messaging to be improved we should improve it there.

Let me know if you are not receiving expected results (after we merge the PR).

vfried:
@jsam thanks for the changes, i will test them ASAP!!!

src/api-client/dataset.js
```diff
@@ -90,32 +90,77 @@ export default function addDatasetMethods(client) {
...
+          return client.clientFetch(`${client.baseUrl}/renku/datasets.add`, {
+            method: "POST",
+            headers: headers,
+            body: JSON.stringify({
+              "dataset_name": renkuDataset.name,
+              "files": renkuDataset.files,
+              "project_id": project_id
+            })
+          });
+        } return response;
```
lorenzo-cavazzi: something is weird with the alignment here, this should be one tab to the right. I guess the linter shouldn't accept this :thinking: I'll check the setting.
Could you correct this manually? There are other misalignment in this page

src/dataset/Dataset.present.js
```diff
@@ -156,10 +155,8 @@ export default function DatasetView(props) {
           }).catch(error => {
             if (fetchError === null) {
               if (!unmounted && error.case === API_ERRORS.notFoundError) {
-                setFetchError(
-                  "Error 404: The dataset that was selected does not exist or could not be accessed." +
-                  "If you just created the dataset try reloading the page."
-                );
+                setFetchError("Error 404: The dataset that was selected does not exist or" +
+                +" could not be accessed. If you just created or imported the dataset try reloading the page.");
```
lorenzo-cavazzi: There are two `+` one after the other here, I believe this may generate an error

src/model/RenkuModels.js
```diff
@@ -314,6 +314,26 @@ const issueFormSchema = new Schema({
...
 });
 
+const datasetImportFormSchema = new Schema({
+  uri: {
+    initial: "",
+    name: "uri",
+    label: "URI/DOI",
+    edit: false,
+    help: "This will import the dataset with the DOI (Digital Object Identifier) 10.5281/zenodo.3352150" +
+      "and make it locally available. Dataverse and Zenodo are supported, with DOIs (e.g. 10.5281/zenodo.3352150" +
```
lorenzo-cavazzi: missing space at the beginning of the line

vfried:
@lorenzo-cavazzi i just fixed the lint problems and also the error when entering dataset pages, i think this was related to some other PR, should be fixed now after the rebasing.

src/model/RenkuModels.js
```diff
@@ -314,6 +314,26 @@ const issueFormSchema = new Schema({
...
 });
 
+const datasetImportFormSchema = new Schema({
+  uri: {
+    initial: "",
+    name: "uri",
+    label: "URI/DOI",
+    edit: false,
+    help: "This will import the dataset with the DOI (Digital Object Identifier) 10.5281/zenodo.3352150" +
+    "and make it locally available. Dataverse and Zenodo are supported, with DOIs (e.g. 10.5281/zenodo.3352150" +
```
lorenzo-cavazzi: missing space at the beginning of the string
`" and make it [...]`

src/project/Project.present.js
```diff
@@ -598,10 +601,11 @@ class ProjectViewDatasetsList extends Component {
       return <Col sm={12} md={8} lg={10}>
```
lorenzo-cavazzi: I noticed this by chance, but I guess also here the value `md` and `lg` may be inverted
`sm={12} md={10} lg={8}`

src/project/datasets/import/DatasetImport.present.js
```diff
@@ -0,0 +1,159 @@
...
+
+                  });
+              }, 10000);
+            }
+          });
+      });
+  };
+
+  if (props.accessLevel < ACCESS_LEVELS.MAINTAINER) {
+    return <Col sm={12} md={8} lg={10}>
```
lorenzo-cavazzi: Are the value `md` and `lg` inverted here? I believe it was intended to be
`sm={12} md={10} lg={8}`

src/project/datasets/import/DatasetImport.present.js
```diff
@@ -0,0 +1,159 @@
...
+          .then(response => {
+            if (response.data.error !== undefined) {
+              setSubmitLoader(false);
+              setServerErrors(response.data.error.reason);
+            }
+            else {
+              let cont = 0;
+              let monitorJob = setInterval(() => {
+                props.client.getJobStatus(response.data.result.job_id)
+                  // eslint-disable-next-line max-nested-callbacks
```
lorenzo-cavazzi: No need to change it this time, but in general the linting rule is intended to avoid deeply nested callbacks that are difficult to maintain. It would be nice to split this in multiple functions in the future.

Furthermore, I believe the code invoking `clint.<anyfunction()>` belongs to container/state components, not the present.

src/project/datasets/import/DatasetImport.present.js
vfried: Good catch, i also fixed this in other places where it was also wrong.

src/project/datasets/import/DatasetImport.present.js
vfried: Just fixed the things you pointed out, I also separated the code a bit in smaller functions so it's easier to maintain.

lorenzo-cavazzi: great! thanks for taking the time to improve this :+1:

Issue #771, opened 2020-02-19
change projects link on menubar


emmjab:
## Motivation

There's a mismatch between which projects you're expecting to be searching over. For instance, many students in a class that were looking for the course material to fork couldn't find the the project.

**currently**
https://renkulab.io/projects: YOUR projects - link in the menu bar after the search box and first tab in the submenu for a user's view; should be all projects
https://renkulab.io/projects/search: ALL projects - first search box in the menu bar and third tab in the submenu for a user's view

at the very least they should be consistent.

## Proposal

https://renkulab.io/projects/user=[username]: YOUR projects
https://renkulab.io/projects/: ALL projects

and search boxes that correspond

## Context

currently:
<img width="841" alt="Screenshot 2020-02-19 at 15 13 53" src="https://user-images.githubusercontent.com/1056472/74842410-a8909180-532a-11ea-801a-5c26d1169786.png">

emmjab:
The main thing I'm hoping for is that it's more clear which search box and which button correspond to searching over *your* projects vs searching over *all public* projects.

Things that could change: text (e.g. from "Search" to "Search over *all* Projects" v. "Search over *your* projects"), but also from a UX workflow standpoint, when you're just starting out and you don't have many projects, it's confusing to see a search bar. 

**Simplest solution, do the following three things**
1. Change "Search" to "All Projects":
```
Your Projects | Starred Projects | All Projects
```

2. Change "Projects" in the top menu bar to "Your Projects"
(But more likely, when we have "first class citizen" users, this will just be "Home")

3. In the search box in the menu bar, change the text:
From "Jump to or search..." to "Search all public projects..."

rokroskar:
I like this suggestion! I only have one comment:

> From "Jump to or search..." to "Search all public projects..."

you're technically searching over all projects which you can access if I'm not mistaken.

emmjab:
Ok maybe just “Search all projects” then, and the subtleties can be sorted out with some kind of “cant find what you’re looking for? Project may be private”

ciyer:
Good catch! The labels made sense when they were initially set, but as more features were added, the similarities increased, and now I agree that the situation is a bit confusing. This is my suggestion:

<img width="355" alt="image" src="https://user-images.githubusercontent.com/1196411/75434853-7991a580-5952-11ea-934e-d3f89ed5efa0.png">

[...]

PR #780, opened 2020-02-24
feat: add basic page for namespaces

### LINKED ISSUE #750: feat: show projects for a user
When I open a profile of some user and try to see which public projects that person has I'm greeted with this never ending loading circle and information never gets to be displayed.

image

Expected behavior would be to nicely preview that persons public projects.

lorenzo-cavazzi:
Users and groups are not first-class citizens yet. Since users may try to reach the `/projects/<namespace>` url anyway, we want to provide useful information when hitting that route.

With this PR, the existence of the `<namespace>` is tested and, according to the result (user, group or non-existing), actions are suggested to the user.

fix #750

**How to test**
Try to reach a few `/projects/<namespace>` URLs, at least one for each of the 3 following types
* User: E.G. https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech
* Group: E.G. https://lorenzotest.dev.renku.ch/projects/eurobioc2019
* Non-existing: E.G. https://lorenzotest.dev.renku.ch/projects/nonexistent

### DIFF
--- a/src/namespace/Namespace.present.js
+++ b/src/namespace/Namespace.present.js
@@ -1,0 +1,145 @@
+/*!
+ * Copyright 2020 - Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ *  renku-ui
+ *
+ *  Namespace.present.js
+ *  Namespace presentational components.
+ */
+
+import React from 'react';
+import { Link } from 'react-router-dom';
+import { Row, Col, Container } from 'reactstrap';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faInfoCircle } from '@fortawesome/free-solid-svg-icons';
+
+import { Loader, InfoAlert, ExternalLink } from '../utils/UIComponents';
+
+const NamespaceProjects = (props) => {
+  const { namespace } = props;
+  // TODO: I should get the URLs from the redux store: #779
+  const searchUrl = "/projects/search";
+  const searchProjectUrl = (project) => { return `${searchUrl}?q=${project}`; };
+  const searchUserUrl = (user) => { return `${searchUrl}?searchIn=users&q=${user}`; };
+  const searchGroupUrl = (group) => { return `${searchUrl}?searchIn=groups&q=${group}`; };
+
+  let checking = null;
+  if (props.user.fetching || props.group.fetching)
+    checking = (<div>
+      <p>Searching for {namespace}...</p>
+      <Loader />
+    </div>);
+
+  let outcome = null;
+  let userOrGroup = "";
+  if (checking == null) {
+    if (props.user.fetched && props.user.data.id) {
+      userOrGroup = "User";
+      const projectsUrl = searchUserUrl(props.user.data.username);
+      const gitlabUrl = props.user.data.web_url;
+      outcome = (
+        <NamespaceUserActions namespace={namespace} projectsUrl={projectsUrl} gitlabUrl={gitlabUrl} />
+      );
+    }
+    else if (props.group.fetched && props.group.data.id) {
+      userOrGroup = "Group";
+      const projectsUrl = searchGroupUrl(props.group.data.full_path);
+      const gitlabUrl = props.group.data.web_url;
+      outcome = (
+        <NamespaceGroupActions namespace={namespace} projectsUrl={projectsUrl} gitlabUrl={gitlabUrl} />
+      );
+    }
+    else {
+      const userUrl = searchUserUrl(namespace);
+      const groupUrl = searchGroupUrl(namespace);
+      const projectUrl = searchProjectUrl(namespace);
+      outcome = (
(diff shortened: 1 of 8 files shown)


rokroskar:
cc @emmjab @cchoirat

emmjab:
Thanks @lorenzo-cavazzi !

I see a typo in `group` on: https://lorenzotest.dev.renku.ch/projects/renku-qa
```
Currently, we don't provide grouü information but you can try one of the following:
```

Other than that, looks good to me!

lorenzo-cavazzi:
fixed, thanks! (I didn't update the deployment on `lorenzotest`)

src/namespace/Namespace.present.js
```diff
@@ -0,0 +1,156 @@
...
+  // TODO: I should get the URLs from the redux store: #779
+  const searchUrl = "/projects/search";
+  const searchProjectUrl = (project) => { return `${searchUrl}?q=${project}`; };
+  const searchUserUrl = (user) => { return `${searchUrl}?searchIn=users&q=${user}`; };
+  const searchGroupUrl = (group) => { return `${searchUrl}?searchIn=groups&q=${group}`; };
+
+  let checking = null;
+  if (props.user.fetching || props.group.fetching)
+    checking = (<div>
+      <p>We are checking if {namespace} is either a user or a group...</p>
```
ciyer: Suggestion -- I think this can be a little simpler.
```<p>Searching for {namespace}...</p>```

File: src/namespace/Namespace.present.js [hunk 2]
```diff
@@ -0,0 +1,156 @@
...
+        <NamespaceNotfoundActions namespace={namespace} userUrl={userUrl} groupUrl={groupUrl} projectUrl={projectUrl} />
+      );
+    }
+  }
+
+  return (
+    <Container fluid>
+      <Row>
+        <Col>
+          <h3>Projects for {props.namespace}</h3>
```
ciyer: Suggestion: Say here if the namespace is a user or group.

```<h3>{userOrGroup} <i>{props.namespace}</i></h3>```

File: src/namespace/Namespace.present.js [hunk 3]
```diff
@@ -0,0 +1,156 @@
...
+          {outcome}
+        </Col>
+      </Row>
+    </Container>
+  );
+}
+
+const NamespaceUserActions = (props) => {
+  return (<div>
+    <p>We have found a user with name {props.namespace}.</p>
```
ciyer: This line is unnecessary if you take the suggestion above.

src/namespace/Namespace.present.js
```diff
@@ -0,0 +1,156 @@
...
+
+    <InfoAlert timeout={0}>
```
ciyer: I don't really like the InfoAlert here. It makes the page feel more unfinished than it has to, and it makes no sense to close the InfoAlert, because then this page does not provide anything useful.

Couldn't we just do 
```
	      <ul className="mb-0">
        <li className="mb-1">
          Browse user&apos;s <Link
            className="btn btn-primary btn-sm" role="button" to={props.projectsUrl}>projects list
          </Link>
        </li>
        <li>
          Visit user&apos;s <ExternalLink url={props.gitlabUrl} size="sm" title="GitLab page" />
        </li>
      </ul>
```
And leave it at that (without the InfoAlert)?

src/namespace/Namespace.present.js
```diff
@@ -0,0 +1,156 @@
...
+          Visit user&apos;s <ExternalLink url={props.gitlabUrl} size="sm" title="GitLab page" />
+        </li>
+      </ul>
+    </InfoAlert>
+  </div>);
+}
+
+const NamespaceGroupActions = (props) => {
+  return (<div>
+    <p>We have found a group with name {props.namespace}.</p>
```
ciyer: See comments on NamespaceUserActions.

src/namespace/Namespace.present.js
```diff
@@ -0,0 +1,156 @@
...
+          Visit group&apos;s <ExternalLink url={props.gitlabUrl} size="sm" title="GitLab page" />
+        </li>
+      </ul>
+    </InfoAlert>
+  </div>);
+}
+
+const NamespaceNotfoundActions = (props) => {
+  return (<div>
+    <p>We couldn&apos;t find any user nor group with name {props.namespace}.</p>
```
ciyer: Wording:
```<p>We could not find a user or group with name {props.namespace}.</p>```

File: src/namespace/Namespace.present.js [hunk 7]
```diff
@@ -0,0 +1,156 @@
...
+
+    <InfoAlert timeout={0}>
```
ciyer: I think the InfoAlert makes sense in this case, so this can be left as is.

src/namespace/Namespace.present.js
lorenzo-cavazzi: I added `<h3>{userOrGroup} {props.namespace}</h3>` since I tried the italics for the project name but it looked weird

lorenzo-cavazzi:
@ciyer thanks for the suggestion, I implemented all of them apart from a minor difference (I commented on the change suggestion).

I am nor sure if I prefer the version with or without `<InfoAlert>`. Anyway, I updated the preview on `lorenzotest`

emmjab:
IMHO it doesn't change too much from a usability standpoint, still seems good to me. (I do have some usability things with https://github.com/SwissDataScienceCenter/renku-ui/issues/771, which is only tangentially related but probably will involve updating some links here too?)

src/api-client/instance.js
```diff
@@ -38,6 +38,16 @@ function addInstanceMethods(client) {
...
+  client.getUserByPath = (path) => {
+    const headers = client.getBasicHeaders();
+    const queryParams = { username: encodeURIComponent(path) };
+    return client.clientFetch(`${client.baseUrl}/users`, {
+      method: 'GET',
+      headers,
+      queryParams
+    })
+  }
 }
```
vfried: Shouldn't this be in user.js ?

lorenzo-cavazzi: this makes definitely sense, let me move it

lorenzo-cavazzi:
@vfried you are right, in the future it makes sense to treat users and groups as first entity citizens and this should go under a different path. We could open another issue to keep track of that (probably an epic would me more appropriate)

ciyer:
I agree with both vfried and lorenzo-cavazzi . I think the `projects/{namespace}` page url should focus on projects and there should be `user/{username}` and `group/{groupname}` pages for the other purposes, but, I think it is nice to provide a link to information about the user/group since you many want that.

Issue #756, opened 2020-02-07
Multiple calls to /datasets endpoint


lorenzo-cavazzi:
When you load/reload a dataset page in the browser, the `/dataset` endpoint is invoked 2 times. This doesn't happen when following a link to the same dataset with the UI already loaded.

E.G. Open this link and check the network monitor tab in the debug tools
https://dev.renku.ch/datasets/42467b81-2526-49b8-9669-2460631538a2/

lorenzo-cavazzi:
The route cause seems to be in the `src/index.js` file. Everything is re-rendered after `userCoordinator.fetchUser()` completes.
A possible solution would be showing a loading page while checking if the user is logged in. This would also drop the requirement to check for `user.fetched` in many components and will prevent many other useless API calls (e.g. in the project page)

PR #781, opened 2020-02-24
fix(avatars): small fixes for avatars

### LINKED ISSUE #767: Avatars
## Description

User avatars have a few minor glitches

* [ ] When loading, they flash a couple of times (reproduce: go on the Landing page)
* [ ] The size of the anonymous icon is smaller than the others (42x48 instead of 52x52) making the flashing more evident and misaligning the text (reproduce: look at project pages with at least an anonymous icon -- see image below)
* [ ] When adding a comment to an issue, the user appears to be anonymous (reproduce: add a comment to any existing issue -- see image below)

## Context

Tested on https://dev.renku.ch

***Different icon sizes***
Screenshot_20200217_110634

***Anonymous comment***
Screenshot_20200217_110617

vfried:
Closes #767 

Fixes:
- When loading, they flash a couple of times (reproduce: go on the Landing page)
- The size of the anonymous icon is smaller than the others (42x48 instead of 52x52) making the flashing more evident and misaligning the text (reproduce: look at project pages with at least an anonymous icon)
- When adding a comment to an issue, the user appears to be anonymous (reproduce: add a comment to any existing issue)

### DIFF
--- a/src/project/shared/Projects.state.js
+++ b/src/project/shared/Projects.state.js
@@ -42,6 +42,7 @@ class ProjectsCoordinator {
   }
 
   async getFeatured() {
+    if(this.model.get("featured.fetching"))return;
     // set status to fetching and invoke both APIs
     this.model.set("featured.fetching", true);
     const promiseStarred = this.client.getProjects({ starred: true, order_by: "last_activity_at" })
--- a/src/contribution/Contribution.container.js
+++ b/src/contribution/Contribution.container.js
@@ -126,6 +126,7 @@ class NewContribution extends React.Component {
       contribution: {
         body: ''
       },
+      submitting: false,
       files: [],
       loading: false,
       mentions: [],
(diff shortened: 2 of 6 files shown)


src/project/shared/Projects.state.js
```diff
@@ -42,6 +42,7 @@ class ProjectsCoordinator {
   }
 
   async getFeatured() {
+    if(this.model.get("featured.fetching"))return;
```
lorenzo-cavazzi: Does this case happen? If yes, could you add a comment/issue on how to reproduce it? (I would like to fix it)
Anyway, it won't hurt to test if we are not already fetching :+1:

vfried: Yes, this happens... i think it was one of the reasons why the avatars where blinking...

in dev:
image

in virginiatest:
image

lorenzo-cavazzi: oh wow, indeed it happens (a lot!)
I'll double check the reason

vfried: ok, shall i merge this or do you want to investigate?

lorenzo-cavazzi: Go on and merge this, your fix works well.
I think fixing the source problem may require a bit more time (or maybe not if it's related to #756) but I wouldn't keep this PR pending in the meanwhile :slightly_smiling_face:

PR #782, opened 2020-02-25
fix: prevent multiple api calls due to user temporarily being anonymous

### LINKED ISSUE #756: Multiple calls to /datasets endpoint

lorenzo-cavazzi:
It turns out that the whole application was rendered twice every time it was loaded on the browser because the user data were fetched and the submodel `user` was consequently updated.

This PR introduces a check on `user.fetched` and `user.fetching` that waits until the fetch completes before showing anything. Usually, this process takes (far) less than 1 second. If it takes longer, the user see a temporary loading page

Screenshot_20200225_110449

fix #756


ciyer:
@vfried @lorenzo-cavazzi 
A question about the bouncer/spinning wheel. Should we

1. Use just the bouncer everywhere
2. Use just the spinning wheel everywhere (maybe improve this)
3. Use the spinning wheel sometimes, use the bouncer sometimes

I think the spinning wheel looks great at small sizes like when it is the size of regular text. This is the case on the interactive environments page in several places. I do not think it looks so good when the size is bigger.  So I'm for:

- Use the spinner at small sizes
- Use something else (either the bouncer or one of these https://loading.io/css/) at larger sizes.

Opinions?

lorenzo-cavazzi:
I agree that the Loader doesn't look that great for big sizes (especially full page).
I like the bouncer in that situation, but it may be anything else.

P.S. I wanted to keep the "loader" component on this page different from the others since it's a special case -- no header nor footer are available, only a blank page before any content is loaded. I got inspired by [the Cloudflare "checking your browser" message](https://aws1.discourse-cdn.com/cloudflare/original/3X/b/6/b65b0b52307c977fbf98b7f06813b31e536ef0f9.png) displayed before the site is actually loaded.
Not sure if this was a great idea :slightly_smiling_face:

vfried:
This is a great fix, i think in this case it could be ok to use the bouncer, maybe it could be in purple to make it more similar to the loader.

Regarding keeping/removing the loader, i think if the bouncer is only used here it's ok that we have different things, I don't particularly like the loader but i'm not sure that the bouncer will look better, we could still try.

ciyer:
I like the bouncer too, but I would prefer to have just one component, Loader, to represent this. Let us make the bouncer a "mode" of the Loader, so we have one component. How about:
```<Loader mode="bouncer" />```. The default mode will be "spinner".

**Comment by lorenzo-cavazzi** [comment]
I have opened another issue for the bouncer #785 

A preview for this PR is now available at https://lorenzotest.dev.renku.ch/

File: src/App.js [hunk 1]
```diff
@@ -36,98 +35,110 @@ import { Login } from './authentication';
...
-              <Route exact path="/project_new"
-                render={(p) =>
-                  <Project.New key="project_new"
+    const { user } = this.props;
+    if (!user.fetched && user.fetching) {
+      return (
+        <Jumbotron className="delay1s bg-white">
+          <h3 className="text-center">Checking user data</h3>
+          <Bouncer />
+        </Jumbotron>
```
vfried: Maybe this should go inside the Bouncer, so if we want to display a message under the bouncer it looks the same every time it's used.

lorenzo-cavazzi: I wouldn't put the `<Jumbotron>` element inside the `<Bouncer>`, I think it's a too invasive element.
I would rather keep the bouncer a bit cleaner so that it's easier to re-use. Anyway, we may further refine the bouncer/loader component as part of #785

vfried: ok, i think we can discuss this inside 785, I see what is inside the jumbotron being copy-pasted everywhere where we decide to use the bouncer.... but we can talk about this later.

vfried: What about the text, do you think we should change it to a renku related thing or you want to leave it at Checking user data?

lorenzo-cavazzi: I didn't think much about it, but I am open to other alternatives :slightly_smiling_face: 
`Verifying Renku user permissions`?

vfried: Checkin user data in renku?

vfried: You can also leave it like this, i can approve the pr if you want...

lorenzo-cavazzi: Let's keep it like this for now, it's on a page that the users should almost never see

PR #783, opened 2020-02-25
improve and enforce linting

lorenzo-cavazzi:
As discussed, we would like to improve our linting strategy.

This PR does the following
* update linting libraries
* fix the broken rules
* add more restrictive linting rules
* lint add test files
* fail CI on linting errors/warning
* lint all the current files

The rules may be changed as it suits us, feel free to comment if you don't like some of them.
Some files look weird. I didn't manually changed all of them, I used `eslint --fix` and then I went through the remaining errors/warning. We could improve the formatting as we work on them (VScode with the `eslint` plugin does a great job on that)

**Info**
* To run the linter, use `npm run lint`
* Most of the changes are for linting the files. That is done all in the last commit.
* ESLint rules reference
** standard: https://eslint.org/docs/rules/
** react: https://github.com/yannickcr/eslint-plugin-react


lorenzo-cavazzi:
I will resolve the conflicts only before merging since it takes some time and every PR will cause a lot of conflicts

ciyer:
Thanks a lot for doing this! It will be great to have this.

I like the rules you set up with the one exception of quotation handling? Is it possible to allow both double and single quotes?

lorenzo-cavazzi:
Yes, we can allow both, but I would rather choose one and stick with that. There are many files with single and double quotes alternating every few lines.

Currently, this is the setting
`"quotes": ["warn", "double", { "allowTemplateLiterals": true, "avoidEscape": true }]`
The preference is on double quotes `"`, allowing the single quotes `'` only when using the double would require escaping. Template literals ` are always accepted.

We can change it https://eslint.org/docs/rules/quotes or, if you really prefer to allow both, just drop it.

lorenzo-cavazzi:
yes, I'll definitely keep a separated final commit for the linted files, and I'll squash only a few commits in between (the "rules adjustment" ones). Let's try to solve the conflicts :slightly_smiling_face:

PR #803, opened 2020-03-03
WIP fix(datasets): dataset endpoint is called once inside datasets explorer

### LINKED ISSUE #802: Multiple calls to dataset endpoint inside file explorer
## Motivation

When you are inside the file explorer the dataset endpoint is called multiple times.

image

for:
https://dev.renku.ch/projects/virginiafriedrich/test-datasets-test-datasets/datasets/244ac88a-180b-4a29-a0c3-f0c35f54cffc/

## Proposal

The reason for this is that the dependency array inside hooks has dependencies that make the component reload several times.

vfried:
In order to prevent several calls to the endpoint I added a new "useEffect" function that will only activate a readyToFetch when everything needed to fetch the dataset is there. 

Closes #802

### DIFF
--- a/src/dataset/Dataset.present.js
+++ b/src/dataset/Dataset.present.js
@@ -100,12 +100,12 @@ export default function DatasetView(props) {
 
   const [dataset, setDataset] = useState(undefined);
   const [fetchError, setFetchError] = useState(null);
+  const fetching = useRef(false);
   const [readyToFetch, setReadyToFetch] = useState(false);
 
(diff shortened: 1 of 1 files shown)


vfried:
Not really, without the fix you see this...

image

And with the fix there is a single call.

ciyer:
This is what I see when I run the code in this branch:

image

vfried:
Looks like some other change fixed this issue, therefore we can close the pull request.

PR #808, opened 2020-03-04
fix(datasets): search calls KG once and paging problem fixed

### LINKED ISSUE #757: Dataset search
The dataset search feature has a couple of issues to address.

* Clicking on `search` invokes the `/datasets` API twice
* Page number is not reset when searching for a different dataset. If the second search has fewer results, it may show the user a blank page.
E.G. Assume the user searches for a dataset and the result has >1 pages (like the generic `*`). If the user goes to page >=2, then searches for a different dataset with less results, the page number is too high and it shows no results even if there are some (manually change to url to `&page=1` to see them).

vfried:
Closes #757 

- When pressing search /datasets endpoint is invoked only 1 time.
- Fixed a small typo error under the search bar.
- Fix pagination problem (there are only 3 dataset's in renkulab so i can't test this) 

### DIFF
--- a/src/dataset/list/DatasetList.state.js
+++ b/src/dataset/list/DatasetList.state.js
@@ -92,6 +92,13 @@ class DatasetListModel extends StateModel {
     this.performSearch();
   }
 
+  resetBeforeNewSearch() {
+    this.setObject({
+      currentPage: 1,
+      pages: { $set: [] }
+    });
+  }
+
   getSorting() {
     const searchOrder = this.get("orderSearchAsc") === true ? "asc" : "desc";
     switch (this.get("orderBy")) {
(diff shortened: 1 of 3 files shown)


src/dataset/list/DatasetList.state.js
```diff
@@ -92,6 +92,11 @@ class DatasetListModel extends StateModel {
     this.performSearch();
   }
 
+  resetBeforeNewSearch() {
+    this.set("currentPage", 1);
+    this.set(`pages`, []);
```
lorenzo-cavazzi: It makes sense to merge these 2 operations to a single `.setObject` to update the redux store at once
```
this.setObject({
      currentPage: 1,
      pages: { $set: [] }
});
```

vfried: Totally

PR #810, opened 2020-03-05
Merge loader spinner and bouncer

### LINKED ISSUE #785: Rework loader element at full size
## Motivation

The current spinning wheel used on the `<Loader />` component looks nice as an inline element, but its use as a full page loader is not that nice.

## Proposal

We can polish and use the `<Bouncer />` element implemented in #782 as a full-size loader.

ciyer:
Fix #785 

Can test on https://sekhar.dev.renku.ch/

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -128,7 +128,7 @@ body {
 
 .bouncer {
   text-align: center;
-  margin: 50px auto 25px auto;  
+  margin: 50px auto 25px auto;
 }
 
 .bouncer span {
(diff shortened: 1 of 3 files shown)


ciyer:
I agree, I would prefer to have the CSS separate, but I decided to do a minimal refactoring. Don't you get a bouncer now when projects are loading?

vfried:
I get a bouncer but since the projects load fast it almost don't see it moving.

vfried:
As i said above i tried making it a bit faster and it doesn't look good so i think we can approve and see how this feels while we are using it, we can always try to tweak it a bit later.

PR #818, opened 2020-03-11
chore: enforce indent eslint rule

lorenzo-cavazzi:
The `react/jsx-indent` rule was not checking all the indentations.
This fixes that by adding also the eslint `indent` rule.
The second commit fixes the current indent warnings in all the files.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,6 +21,7 @@
     "react/prop-types": "off",
     "curly": ["warn", "multi-or-nest", "consistent"],
     "react/jsx-indent": ["warn", 2],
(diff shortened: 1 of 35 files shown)


vfried:
Hey, i'm currently testing this... i'm getting this warning on telepresence:

image

vfried:
> I'm just getting a weird warning that i left in a message above.
> 
> Besides that it looks good.

lorenzo-cavazzi:
oh right, I missed that. An `href` attribute is missing in the `<a>` element since I removed the linting disabler `/* eslint-disable-next-line */`. I have just added it back instead of changing the component -- we could address that with low priority the first time we need to modify it

lorenzo-cavazzi:
ok I rebased and fixed the missing pieces, it's ready

PR #825, opened 2020-03-13
fix(dataset): doi and source are display only when needed

### LINKED ISSUE #824: Improve DOIs and Source display
The UI displays DOIs and Sources that shouldn't be there:

image

In this case:
---> if the dataset is inside the project only the Source should be displayed
---> if the dataset is in the "global" view the source and the doi shouldn't be displayed (the link to external should also be hidden)

vfried:
Closes #824 
 (More info on the issue)

### DIFF
--- a/src/dataset/Dataset.present.js
+++ b/src/dataset/Dataset.present.js
@@ -205,7 +205,8 @@ export default function DatasetView(props) {
           </Link>
           : null
         }
-        {dataset.url ?
+        { dataset.url &&
+        (props.insideProject || dataset.sameAs.includes("doi.org")) ?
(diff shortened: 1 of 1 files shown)


src/dataset/Dataset.present.js
```diff
@@ -228,7 +228,8 @@ export default function DatasetView(props) {
           </Link>
           : null
         }
-        {dataset.url ?
+        { dataset.url &&
+        (props.insideProject || !dataset.sameAs.includes(props.client.baseUrl.replace("/api", ""))) ?
```
lorenzo-cavazzi: Are we sure we want to test this using `props.client.baseUrl.replace("/api", "")`? In some deployments, the `/api` link may be in a different base user than the UI

PR #840, opened 2020-03-17
fix(datasets): dataset displays folder information

vfried:
The datasets display folder information now...

image

I deleted the links to the file and graph display but we could potentially add them next to the file name or something similar...

Testing: https://virginiatest.dev.renku.ch/datasets/9b3b47b7-277e-4368-87bd-c66513701dc5/

### DIFF
--- a/src/dataset/Dataset.present.js
+++ b/src/dataset/Dataset.present.js
@@ -33,7 +33,7 @@ function DisplayFiles(props) {
 
   return <Card key="datasetDetails">
     <CardHeader className="align-items-baseline">
-      <span className="caption align-baseline">Dataset files</span>
+      <span className="caption align-baseline">Dataset files ({props.files.length})</span>
     </CardHeader>
     <CardBody>
       <FileExplorer
--- a/src/dataset/Dataset.present.js
+++ b/src/dataset/Dataset.present.js
@@ -19,11 +19,11 @@
 import React, { useState, useEffect } from "react";
 import { Row, Col, Card, CardHeader, CardBody, Table, Alert, Button } from "reactstrap";
 import { Link } from "react-router-dom";
-import { Loader } from "../utils/UIComponents";
+import { Loader, FileExplorer } from "../utils/UIComponents";
 import DOMPurify from "dompurify";
 import { API_ERRORS } from "../api-client";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import { faFile, faProjectDiagram, faExternalLinkAlt, faPen } from "@fortawesome/free-solid-svg-icons";
+import { faExternalLinkAlt, faPen } from "@fortawesome/free-solid-svg-icons";
 import { GraphIndexingStatus } from "../project/Project";
 import KnowledgeGraphStatus from "../file/KnowledgeGraphStatus.container";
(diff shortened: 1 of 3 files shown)


src/dataset/Dataset.present.js
```diff
@@ -36,34 +36,12 @@ function DisplayFiles(props) {
...
-                  <Link to={`${props.lineagesUrl}/${file.atLocation}`}>
-                    <FontAwesomeIcon icon={faProjectDiagram} />
-                  </Link>
-                </td> : null}
-            </tr>
-          )}
-        </tbody>
-      </Table>
+      <FileExplorer
+        files={props.files}
```
ciyer: Is there a reason for keeping this commented-out lines around? If not, let's delete them.

src/utils/FileExplorer.js
```diff
@@ -0,0 +1,180 @@
...
+          setOpenFolder={this.props.setOpenFolder}
+          hash={this.props.hash}
+        />;
+      })
+      : null;
+
+    let elementToRender;
+    if (this.props.node.jsonObj !== null) {
+      elementToRender =
+        <div className={order + " " + hidden}>
```
ciyer: Would it makes sense to declare a constant: `const eltClassName = `{order} {hidden}`? I see this used a few times.

src/utils/FileExplorer.js
```diff
@@ -0,0 +1,180 @@
...
+          <div className={"fs-element"}>
```
ciyer: I think this would be a bit nicer as `<div className="fs-element">`. It's kind of minor, but I think it's a little clearer, even though the semantics are exactly the same.

src/utils/FileExplorer.js
```diff
@@ -0,0 +1,180 @@
...
+            {icon} {this.props.node.name}
+          </div>
+        </div>
+      ;
+    }
+    else {
+      elementToRender = this.state.childrenOpen ?
+        <div className={order + " " + hidden} >
```
ciyer: See comment above.

src/utils/FileExplorer.js
```diff
@@ -0,0 +1,180 @@
...
+          <div className={"fs-element"} onClick={this.handleIconClick} >
```
ciyer: See comment above.

src/utils/FileExplorer.js
```diff
@@ -0,0 +1,180 @@
...
+            {icon} {this.props.node.name}
+          </div>
+          <div className="pl-3">
+            {children}
+          </div>
+        </div>
+        :
+        <div className={order + " " + hidden} >
```
ciyer: See comment above.

src/utils/FileExplorer.js
```diff
@@ -0,0 +1,180 @@
...
+          <div className={"fs-element"} onClick={this.handleIconClick}>
```
ciyer: See comment above.

src/dataset/Dataset.present.js
vfried: not really, i can add this change easily

vfried:
I did the changes requested and updated virginiatest

PR #846, opened 2020-03-19
feat: add options to properly convert markdown files

lorenzo-cavazzi:
Improve support for markdown files preview.

Here is an example: https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/new-markdown
Compare with https://dev.renku.ch/projects/lorenzo.cavazzi.tech/new-markdown


lorenzo-cavazzi:
Something is still wrong with the images but I couldn't find an easy fix. I will open another issue for that

Issue #796, opened 2020-03-02
Allow users to unzip files on dataset file upload


vfried:
We should allow the users to import a zip folder and unzip it in the backend.

This should also be properly displayed in the UI.

rokroskar:
There is still the limitation that uploading large files is not currently supported.

ciyer:
Closed by #849

PR #849, opened 2020-03-20
feat(dataset): users can unzip files in the backend #849 #796

vfried:
Closes #849 

When a user uploads a zip file we ask the user if they want to unzip the file:

image

After they choose all work as normal.

Possible modification: we can see if the user chose or not, at the moment all files that are not in ready state are dismissed on form submit.

Testing available in: virginiatest.dev.renku.ch

### DIFF
--- a/src/api-client/dataset.js
+++ b/src/api-client/dataset.js
@@ -11,7 +11,7 @@ export default function addDatasetMethods(client) {
     });
   };
 
-  client.uploadFile = (file, controller) => {
+  client.uploadFile = (file, controller, unpack_archive = false) => {
     const data = new FormData();
     data.append("file", file);
     data.append("file_name", file.name);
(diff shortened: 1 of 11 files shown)


ciyer:
This is super useful and works well. I added two commits to make minor improvements to how some information is presented to make it more visible and clearer to understand. Please check if those are ok, and you can squash one or both if you want.

ciyer:
In testing, I noticed a couple of problems that are not directly related to this feature, but if they are easy to fix, it would be nice to include the fixes in this pr.

1. If you do not provide all the information, creating a dataset fails. This is expected and good, but I would also expect an error near the "Create" button, not just by the field (see screenshot)

image

Would it be hard to add something like "Please fix problems with the following fields: _name_"

vfried:
I changed your changes a little bit... it looks like this now:

image

Issue #811, opened 2020-03-05
Dataset display does not include folder information


ciyer:
## Description

The display of a dataset does not include folder information

## Reproduce

See https://dev.renku.ch/projects/cramakri/covid-19-dashboard/datasets/df11229d-6b56-46ee-8ad2-38ecdfbd7c44/

If you look at the file, you will see that the gitignore files are in different folders.

## Context

![image](https://user-images.githubusercontent.com/1196411/75985766-2a68e900-5eed-11ea-82a3-1f39eea17f7d.png)

ciyer:
Closed by #860

PR #860, opened 2020-03-24
fix(datasets): datasets folders are displayed correctly and open

### LINKED ISSUE #859: Folders in datasets aren't displayed correctly
There is a bug on the directory structure display on the UI.

This...
image

Should look like this...
image

--> We should also display the "/data" folder and first two (or one in case "data" isn't there) should be open.

vfried:
Closes #859 

(Changes of this pr are described in the issue)


ciyer:
Closes #811

PR #863, opened 2020-03-24
Prepare 0.9.0 release of UI

ciyer:


lorenzo-cavazzi:
This should be included in the release to keep the template version up-to-date with renku-python #871

ciyer:
> All good!
> I think the file `.metals/metals.h2.db` accidentally slipped in

Oh, good catch! Thanks. It has been removed now.

PR #865, opened 2020-03-24
WIP feat(environments): allow unprivileged users to start interactive environments

### LINKED ISSUE #857: Allow unprivileged users to start an interactive session
The UI should allow a user to start an interactive session, even if that user does not have write access to the repo or is not logged in at all. The UI should warn the user that she or he might loose some data.

------------
_Update (2020-04-01):_

After some work on the backend and some discussion below we have a clearer picture about the necessary changes on the UI side to enable this feature:

- [x] Display dialogue for starting notebook servers also to unprivileged users
- [ ] Query the registry API rather than jobs API to determine image availability. In case of an unavailable image and only for privileged users the jobs API should still be queried to distinguish failed image build from ongoing ones.
- [ ] When receiving a 401 response from `/api/notebooks/...`, do not trigger the normal login cycle. Instead send the browser to `/api/auth/jupyterhub/login-tmp?redirect-url=<current-path-or-full-url (url-encoded)>`. This will trigger a login cycle with the Jupyterhub instance which manages the temporary user sessions and then redirect to the desired UI page.
- [ ] _updated_: The above point only applies when `.Values.global.anonymousSessions.enable: true`. Otherwise that login attempt will fail. `.Values.global.anonymousSessions.enable` should be translated into an env variable and written into `config.json` which then UI can then use to determine if anonymous sessions are activated or not.

lorenzo-cavazzi:
Temporarily beta for test purposes.

Preview: https://lorenzotest.dev.renku.ch

fix #857

### DIFF
--- a/src/api-client/notebook-servers.js
+++ b/src/api-client/notebook-servers.js
@@ -33,6 +33,7 @@ function addNotebookServersMethods(client) {
     }).then(resp => {
       return { "data": resp.data.servers };
     });
+    // return Promise.resolve({ "data": {} });
   };
 
   client.stopNotebookServer = (serverName, force = false) => {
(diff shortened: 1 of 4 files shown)


lorenzo-cavazzi:
superseded by #893

PR #870, opened 2020-03-25
Fix notebook display issues

### LINKED ISSUE #838: Hide input cells of notebooks by default
## Motivation

Many notebooks are designed to be viewed by a public that might not care about the code. It should be possible to show notebooks with input cells hidden.

## Proposal

All notebooks should have a radio-button-type interaction to select whether the input cells are shown or not. For the default behavior, there are two options:

* Option 1:  Have a setting a for each notebook: show or hide input cells as default
* Option 2: For notebooks in the "notebooks" folder, show input cells, for notebooks everywhere else, hide input cells

Option 1 is more powerful, but a bit more work to implement, and adds more cognitive load on the user
Option 2 is very easy to implement, but could seem a bit arbitrary unless it is clearly documented and explained.

ciyer:
Fix #838 (Allow hiding/showing of notebook cells)

Also:
* Use Source Code Pro for output display
* Show stdout output in black
* Style tables as bootstrap tables

## Testing (compare old/new)

See the difference in fonts and colors

* https://dev.renku.ch/projects/cramakri/notebook-rendering/files/blob/notebooks/ouput-test.ipynb
* https://sekhar.dev.renku.ch/projects/cramakri/notebook-rendering/files/blob/notebooks/ouput-test.ipynb

See how the input cells are hidden when papermill runs in report mode

* https://dev.renku.ch/projects/cramakri/notebook-rendering/files/blob/notebooks/tables-test-report.ran.ipynb
* https://sekhar.dev.renku.ch/projects/cramakri/notebook-rendering/files/blob/notebooks/tables-test-report.ran.ipynb

See how the tables look different

* https://dev.renku.ch/projects/cramakri/notebook-rendering/files/blob/notebooks/tables-test-report.ran.ipynb
* https://sekhar.dev.renku.ch/projects/cramakri/notebook-rendering/files/blob/notebooks/tables-test-report.ran.ipynb

See how you can select if code is to be shown

* https://sekhar.dev.renku.ch/projects/cramakri/notebook-rendering/files/blob/notebooks/input-hidden-test-report.ran.ipynb (default is code hidden, but can show as well)
* https://sekhar.dev.renku.ch/projects/cramakri/notebook-rendering/files/blob/notebooks/ouput-test.ipynb (default is code shown, but can hide as well)

### DIFF
--- a/src/file/File.present.js
+++ b/src/file/File.present.js
@@ -35,7 +35,6 @@ import { Time } from "../utils/Time";
 
 const commitMessageLengthLimit = 120;
 
-
 /**
  * Display the Card with file information. Has the following parameters:
  *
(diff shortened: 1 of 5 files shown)


rokroskar:
This looks really great! 

I would hide the "default" button for code visibility and maybe make that a less obtrusive element in general (a switch/toggle instead of two/three buttons?)

lorenzo-cavazzi:
Nice feature! Personally, I don't dislike the radio buttons :slightly_smiling_face: 

I would just uniform the output to the "Commit" section above

From this
Screenshot_20200415_112500

To this
Screenshot_20200415_112621

ciyer:
> This looks really great!
> 
> I would hide the "default" button for code visibility and maybe make that a less obtrusive element in general (a switch/toggle instead of two/three buttons?)

I think it would be nice to simplify the buttons, but I'm not sure how the Notebook author can communicate to the UI how the notebook should be rendered if we do not offer the "default" option. If we wanted to get rid of an option I think the best one to drop would be "Hidden."

rokroskar:
I don't follow. I can have a default set in the notebook and the switch is simply flipped to the default side when the notebook is rendered. The state is on/off, I shouldn't have more than two options.

ciyer:
> I don't follow. I can have a default set in the notebook and the switch is simply flipped to the default side when the notebook is initially rendered. The state is on/off, I shouldn't have more than two options.
> 
> This is the same behavior we have right now for the environment launch options btw.

Oh, now I see the confusion. The situation here is more complex. In addition to the global setting for the notebook, each cell in the notebook can have metadata that indicates whether it is visible or not. In "default" mode, the notebook metadata is respected. In "visible" mode, the notebook metadata is overridden, and every cell's input section is set to visible. In "hidden" mode, the notebook metadata is overridden, and every cell's input section is set to hidden.

rokroskar:
Oh ok I thought we were just talking about a global on/off like we've done in the covid-19 project, I didn't realize you were solving the harder problem already :)  

IMHO it would already be awesome to have a global on/off switch, unless you foresee this as a pedagogical tool? Otherwise you either always want to see the code because you want to copy/paste snippets or you don't want to see the code at all.

ciyer:
> Oh ok I thought we were just talking about a global on/off like we've done in the covid-19 project, I didn't realize you were solving the harder problem already :)
> 
> IMHO it would already be awesome to have a global on/off switch, unless you foresee this as a pedagogical tool? Otherwise you either always want to see the code because you want to copy/paste snippets or you don't want to see the code at all.

If we are going to interpret the notebook cell visibility metadata, then we might as well do it correctly (this is the "default" mode). I think we can keep it simple for overriding the metadata and just offer "visible"/"hidden".

I am open to a better name for default, but I cannot think of one that isn't even more confusing.

rokroskar:
So if there is the default, then there should just be the default without having to give it a name or a button. Can we make it a toggle that makes it clear that it's a global override? and if you flip it in either direction it's ignoring the cell-specific setting?

ciyer:
Here is how the styling looks now:
image

vfried:
Hey, it's really cool that we get the option of hiding the code... i don't really like the way the buttons look, i think i would actually put them in the top bar next to "File View" or on the right, maybe as a dropdown.

ciyer:
> So if there is the default, then there should just be the default without having to give it a name or a button. Can we make it a toggle that makes it clear that it's a global override? and if you flip it in either direction it's ignoring the cell-specific setting?

That will make it clearer. Let me think about how to implement that.

vfried:
Hey, if the option is between two things we can use something like in the ckeditor (Coming soon) 
image

`<CustomInput
        className="float-right"
        type="switch"
        id="exampleCustomSwitch"
        name="customSwitch"
        label={switchLabel}
        checked={codeview}
        onChange={() => { setCodeview(!codeview); }}
      />`

ciyer:
I like that switch. We could do something like

image

ciyer:
This is what it looks like now:

image

image

rokroskar:
Thanks @ciyer I think that's much clearer!

src/file/File.present.js
```diff
@@ -106,105 +109,229 @@ class FileCard extends React.Component {
...
+  }
+
+  function setLocalMode(mode) {
+    setOverrideMode(mode);
+    setDisplayMode(mode);
   }
+
+  const overrideControl = (displayMode === NotebookSourceDisplayMode.DEFAULT) ?
+    null :
+    <ButtonGroup key="controls" size="sm">
```
lorenzo-cavazzi, line 271: The Buttons look very close to the text above (see picture)

Screenshot_20200420_091024

I would add a distance, maybe `className="mt-1" ` (or even `mt-2`)

ciyer: I tried both and prefer mt-1. Here is what it looks like now.

image

rokroskar:
Thanks all, this is going to make the notebooks view so much better!

Issue #607, opened 2019-09-13
Improvements to help access & content


ciyer:
# Description

We want to make several improvements to the help page.

# To Do

### Display version information

It would make sense to show the version information somewhere in the UI. GitLab does this in the help page, which would make sense for Renku as well.

# Implemented:

* Rename "Discourse" to "Forum"
* Easier access to help in menu
* Easier access to Forum and GitHub from bottom menu

# Under Consideration

* Rename "Gitter" to "Chat"

ciyer:
From where can we acquire version information? I think it would make sense for the gateway to provide it.

ableuler:
Hmm, I'm not sure if it would make sense to display version information just for Renku or also for the components?

ciyer:
I think we would display information for all the components.

rokroskar:
That needs a service with access to the k8s API

emmjab:
My POV: 

maybe along the lines of adding "help" to the names for things: 

### rename "Discourse" to "Help Forum"

and I vote for adding `Help` to the top navbar where gitlab has it, but explicitly write "Help"

ciyer:
This is what is currently implemented:

![image](https://user-images.githubusercontent.com/1196411/78031539-f8b83600-7363-11ea-92cc-16478e9d8f4d.png)

ciyer:
Closing as out of date.

PR #872, opened 2020-03-25
Help improvements

ciyer:
* Rename 'Discourse' to 'Forum'
* Add link to Forum in FooterNavbar
* Add Help Menu

image

Work towards #607 

Test deployment on https://sekhar.dev.renku.ch/

### DIFF
--- a/src/landing/NavBar.js
+++ b/src/landing/NavBar.js
@@ -25,12 +25,12 @@
 
 import React, { Component } from "react";
 import { Link } from "react-router-dom";
-import { Navbar, NavLink, Nav } from "reactstrap";
+import { Navbar, Nav } from "reactstrap";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faPlus } from "@fortawesome/free-solid-svg-icons/";
 
 import logo from "./logo.svg";
-import { RenkuNavLink, Loader, UserAvatar } from "../utils/UIComponents";
+import { ExternalDocsLink, Loader, RenkuNavLink, UserAvatar } from "../utils/UIComponents";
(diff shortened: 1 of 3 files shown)


src/landing/NavBar.js
```diff
@@ -223,6 +223,8 @@ class FooterNavbar extends Component {
           </Nav>
           <Nav className="ml-auto">
             <RenkuNavLink to="/help" title="Help" />
+            { /* <ExternalDocsLink url="https://renku.discourse.group" title="Help Forum" /> */ }
```
lorenzo-cavazzi: should we remove this comment?

rokroskar:
I know we decided against this, but would it make sense to also make the gitter link lay-person-friendly (i.e. Gitter --> Chat)?

src/landing/NavBar.js
ciyer: Oops!

ciyer:
I worry that "Chat" alone might not be clear enough. But the majority of the team was for keeping it as "Gitter".

rokroskar:
"Gitter chat"? 😄

lorenzo-cavazzi:
I wouldn't go with both words since it's a "single word" link in the bottom of the page and multiple words may be confusing -- they may look like 2 separate links, plus there is very little space.

I voted for `Chat` so I may be biased on that direction :smile: 
In any case, I would keep either the generic definition for both links (`Forum` + `Chat`) or the specific one (`Concourse` + `Gitter`). I find the current solution a bit confusing.

ciyer:
Yes, you correctly guessed the purpose. We should probably be consistent with that, though. An alternative is to use a icon.
image

PR #873, opened 2020-03-25
feat(datasets): users can add datasets that are inside renku to other projects

### LINKED ISSUE #527: Select datasets to add to new or existing project
Either on the project search or on the create project page there should be an instance where the user selects datasets and is redirected to the create project page, so a project is created using the selected datasets.

(We should discuss about how we will do this, i don't think doing a basket that will be across the entire user interface is a good idea, I propose we either keep it in the datasets search or we implement a workflow like project creation page).

vfried:
Closes #527 

### **Needs testing!!!**

I added a button in the datasets (for users that are logged in) so users can add datasets to projects.

image
(I don't really like the way this buttons looks, maybe we could do something to improve them
---> I know i was the one creating them haha)

After the user presses on the button a modal opens, and they can select a project from the list of projects where they are MAINTAINER.
image

image

They press Add dataset, they wait and they are redirected to the project/new-dataset

I am not 100% sure it works on datasets with unzipped files. (If this is the case it would be a backend issue)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -129,6 +129,7 @@ class App extends Component {
                   projectsUrl="/projects"
                   selectedDataset={p.match.params.datasetId}
                   logged={this.props.user.logged}
+                  model={this.props.model}
                 />}
               />
               <Route path="/datasets" render={
--- a/src/App.js
+++ b/src/App.js
@@ -128,6 +128,7 @@ class App extends Component {
                   client={this.props.client}
                   projectsUrl="/projects"
                   selectedDataset={p.match.params.datasetId}
+                  logged={this.props.user.logged}
                 />}
               />
               <Route path="/datasets" render={
--- a/src/api-client/dataset.js
+++ b/src/api-client/dataset.js
@@ -120,7 +120,7 @@ export default function addDatasetMethods(client) {
       });
   };
 
-  client.datasetImport = (projectUrl, renkuDataset) => {
+  client.datasetImport = (projectUrl, datasetUrl) => {
     let headers = client.getBasicHeaders();
     headers.append("Content-Type", "application/json");
     headers.append("X-Requested-With", "XMLHttpRequest");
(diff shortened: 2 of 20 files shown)


vfried:
Maybe the dropdown should look different too....

rokroskar:
Would adding an option to _create_ a project with the dataset be within the scope here?

ciyer:
This is very cool, something we have been wanting for a while and now pretty close to being finished! 

The functionality works well, I have some cosmetic things that I think should be fixed before we merge.


### Layout

The layout of the "Modify", "Add to Project", and "Go to Source" buttons looks a little off.

image


### Display of projects

Could the list of projects be grouped by namespace and sorted alphabetically? It's a little hard to find the project you want if you have more than a few.

image

And I think it would be nicer to use the same widget we use in the fork dialog to find projects.

image

vfried:
Updated with autosuggest field instead of dropdown...

image

I agree that the buttons don't look great, i am not sure about what to do with the buttons, we could make them become icons and then on hover tell what they do, or make a dropdown with options (not sure if this will advertise well the actions) another option is to remove the go to source, since the source is already on the description and leave the modify and add to project.

vfried:
List now is sorted, i am not sure we need to have it grouped, at least for now.

I changed the buttons, and did a small change on the layout of the datasets inside project.
( <Col key="datasetcontent" sm={12} md={9} lg={8}> )

image

vfried:
Small updates, is ready to be tested in virginiatest

vfried:
Results are now grouped, i also removed the 100 projects limit, all projects are there now, the dropdown has a scroll bar for when the results are too many.

image

vfried:
Testing on https://virginiatest.dev.renku.ch

vfried:
I just merged the changes, but i can make a small issue for this.

PR #876, opened 2020-03-26
feat(datasets): label changed

### LINKED ISSUE #875: Change label of dataset import
Renku datasets can be imported now, so we should change the label of the dataset import feature.

vfried:
Closes #875 

Label changed from this: "Dataverse or Zenodo dataset URL or DOI:"

To:
image

### DIFF
--- a/src/model/RenkuModels.js
+++ b/src/model/RenkuModels.js
@@ -319,7 +319,7 @@ const datasetImportFormSchema = new Schema({
   uri: {
     initial: "",
     name: "uri",
-    label: "Dataverse or Zenodo dataset URL or DOI:",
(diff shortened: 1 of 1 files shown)


src/model/RenkuModels.js
```diff
@@ -319,7 +319,7 @@ const datasetImportFormSchema = new Schema({
   uri: {
     initial: "",
     name: "uri",
-    label: "Dataverse or Zenodo dataset URL or DOI:",
+    label: "Dataverse, Zenodo or Renku Dataset URL:",
```
ciyer: Could we make this to "Renku Dataset URL; Datavarse or Zenodo URL or DOI"?

vfried: yes

PR #890, opened 2020-04-01
Add links to GitLab

### LINKED ISSUE #867: link to gitlab from renkulab nav bar
## Motivation

It's hard for users to find the link to gitlab when they haven't created any projects yet, and this is a thing they have to do to be able to push existing projects to renkulab (because they have to set up an ssh key from gitlab).

## Proposal

put the link to gitlab in the nav bar (or somewhere else global)

ciyer:
Add links to the following GitLab URLs:

- User Profile
- User Settings
- GitLab Home

image

Can be tested at https://sekhar.dev.renku.ch/

Fix #867

### DIFF
--- a/src/landing/NavBar.js
+++ b/src/landing/NavBar.js
@@ -31,7 +31,7 @@ import { faPlus } from "@fortawesome/free-solid-svg-icons";
 import { faQuestionCircle } from "@fortawesome/free-regular-svg-icons";
 
 import logo from "./logo.svg";
-import { ExternalDocsLink, Loader, RenkuNavLink, UserAvatar } from "../utils/UIComponents";
+import { ExternalDocsLink, ExternalLink, Loader, RenkuNavLink, UserAvatar } from "../utils/UIComponents";
 import { getActiveProjectPathWithNamespace } from "../utils/HelperFunctions";
 import QuickNav from "../utils/quicknav";
 
@@ -68,7 +68,7 @@ class RenkuToolbarItemUser extends Component {
         {this.props.userAvatar}
       </a>
       <div key="menu" className="dropdown-menu dropdown-menu-right" aria-labelledby="profile-dropdown">
-        <ExternalDocsLink url={`${gatewayURL}/auth/user-profile`} title="Profile" className="dropdown-item" />
+        <ExternalLink url={`${gatewayURL}/auth/user-profile`} title="Profile" className="dropdown-item" role="link" />
         <DropdownItem divider />
         <a id="logout-link" className="dropdown-item"
           href={`${gatewayURL}/auth/logout?redirect_url=${redirect_url}`}>Logout</a>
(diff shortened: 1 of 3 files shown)


emmjab:
Thanks @ciyer !! I think this is good for now!

Slightly worried that the GitLab icon will be hard to describe for new-to-gitlab users, but we can include the icon in the docs I guess. It does do a good job saving space, and the first link once you click it does say GitLab.

Just noticed from testing that the help menu doesn't have a link to the docs~ but that's in another PR, will comment there.

vfried:
I think that having the link to profile and settings under the user and under the gitlab sign could be a bit redundant, especially since both links are right. next to each other... 

image

image

Shall we remove some of this links somewhere?

src/utils/UIComponents.js
```diff
@@ -568,24 +599,47 @@ function IconLink(props) {
   </span>;
 }
 
+function ExternalIconLinkWithTooltip(props) {
+  // eslint-disable-next-line no-unused-vars
```
vfried, line 603: You can remove the  //eslint-disable.... if you remove also the setter since you are not using it.

vfried:
> Thanks @ciyer !! I think this is good for now!
> 
> Slightly worried that the GitLab icon will be hard to describe for new-to-gitlab users, but we can include the icon in the docs I guess. It does do a good job saving space, and the first link once you click it does say GitLab.
> 
> Just noticed from testing that the help menu doesn't have a link to the docs~ but that's in another PR, will comment there.

I am all in for helping users have an easy beginning with renku, but i think we ned to draw a line about how much we "hide" gitlab inside renku, since for other users having gitlab is also seen as a benefit. I think that at some point someone using renku that has no idea about gitlab will have to make a quick search to see what it it.

src/utils/UIComponents.js
ciyer: Good suggestion! I have made this improvement.

emmjab:
> I am all in for helping users have an easy beginning with renku, but i think we ned to draw a line about how much we "hide" gitlab inside renku, since for other users having gitlab is also seen as a benefit. I think that at some point someone using renku that has no idea about gitlab will have to make a quick search to see what it it.

wait i think we're on the same side? 😅 
basically what i mean is that it's hard for someone to google for an icon -- my suggestion would be to replace the fox icon with the word "GitLab"

> I think that having the link to profile and settings under the user and under the gitlab sign could be a bit redundant, especially since both links are right. next to each other...

Also confused --> before the "Settings" and "Profile" were **renku** settings and profile in the "User" menu.

ciyer:
> I think that having the link to profile and settings under the user and under the gitlab sign could be a bit redundant, especially since both links are right. next to each other...
> 
> image
> image
> 
> Shall we remove some of this links somewhere?

You are right. In this case, since they are so close to each-other, it's confusing to have the links redundantly. I have removed them from the user menu and just left them in the GitLab menu.

ciyer:
> Also confused --> before the "Settings" and "Profile" were **renku** settings and profile in the "User" menu.

There is no difference ATM between Renku Settings and GitLab settings, so those links lead to the same place.

vfried:
Cool, i think it's better if we only have them in one place.

src/landing/NavBar.js
```diff
@@ -49,6 +50,20 @@ class RenkuNavBar extends Component {
   }
 }
 
+function gitLabSettingsUrlFromProfileUrl(webUrl) {
+  // Yes, the settings URL ends with 'profile'; the profile URL ends with the username
```
lorenzo-cavazzi, line 54: :man_facepalming:

PR #893, opened 2020-04-03
feat(environments): allow all users to start interactive environments

### LINKED ISSUE #857: Allow unprivileged users to start an interactive session

lorenzo-cavazzi:
Allow all users (both unprivileged and anonymous) to start environments on public projects.

* As an unprivileged user, you can start environments also in protected projects and check the pipelines/jobs status.
* As an anonymous user, only the image status will be checked without further details. You are not allowed to ignore that, so it shouldn't be possible to start environments from the wrong image.
A temporary login (not requiring name or any user action) will be triggered when accessing the environments page. This is needed to correctly associate environments to users.

Testing this PR requires an appropriate backend. A preview is available here:
https://andreas.dev.renku.ch

***How to test***
Try to access any public project and start an environment there.

supersede #865 
close #857

### DIFF
--- a/helm-chart/renku-ui/templates/deployment.yaml
+++ b/helm-chart/renku-ui/templates/deployment.yaml
@@ -48,6 +48,8 @@ spec:
             - name: MAINTENANCE
               value: {{ .Values.maintenance | default (printf "false") | quote }}
               {{- end }}
+            - name: ANONYMOUS_SESSIONS
+              value: {{ .Values.global.anonymousSessions.enabled | default (printf "false") | quote }}
           livenessProbe:
             httpGet:
               path: /
--- a/src/App.js
+++ b/src/App.js
@@ -52,6 +52,10 @@ class App extends Component {
       );
     }
 
+    // check anonymous sessions settings
+    const anonymousSessions = this.props.params["ANONYMOUS_SESSIONS"] === "true";
+    const blockAnonymous = !user.logged && !anonymousSessions;
+
     return (
       <Router>
         <div>
@@ -91,6 +95,7 @@ class App extends Component {
                   params={this.props.params}
                   model={this.props.model}
                   user={this.props.user}
+                  blockAnonymous={blockAnonymous}
                   {...p}
                 />}
               />
@@ -111,6 +116,7 @@ class App extends Component {
                   standalone={true}
                   client={this.props.client}
                   model={this.props.model}
+                  blockAnonymous={blockAnonymous}
                   {...p}
                 />}
               />
--- a/src/api-client/pipeline.js
+++ b/src/api-client/pipeline.js
@@ -16,6 +16,8 @@
  * limitations under the License.
  */
 
+import { API_ERRORS } from "./errors";
+
 function addPipelineMethods(client) {
   client.runPipeline = (projectId) => {
     const headers = client.getBasicHeaders();
@@ -80,6 +82,46 @@ function addPipelineMethods(client) {
       method: "POST",
       headers,
     }).then(response => response.data);
+  };
+
+  /**
+   * Get the list of available container registries. It should be 1 per project if there
+   * is at least a valid tagged image, otherwise 0.
+   *
+   * @param {number|string} projectId - project id or slug
+   */
+  client.getRegistries = (projectId) => {
+    const headers = client.getBasicHeaders();
+    headers.append("Content-Type", "application/json");
+    const url = `${client.baseUrl}/projects/${projectId}/registry/repositories`;
+
+    return client.clientFetch(url, {
+      method: "GET",
+      headers,
+    }).then(response => response.data);
+  };
+
+  /**
+   * Get the image data for a specific tag.
+   *
+   * @param {number|string} projectId - project id or slug
+   * @param {number} registryId - registry id
+   * @param {string} tag - tag name, our convention uses the first 7 chars from commit id
+   */
+  client.getRegistryTag = (projectId, registryId, tag) => {
+    const headers = client.getBasicHeaders();
+    headers.append("Content-Type", "application/json");
+    const url = `${client.baseUrl}/projects/${projectId}/registry/repositories` +
+      `/${registryId}/tags/${tag}`;
+
+    return client.clientFetch(url, { method: "GET", headers })
+      .then(response => response.data)
+      .catch((error) => {
+        // 404 is expected when nothing is available for the target tag
+        if (error.case === API_ERRORS.notFoundError)
+          return null;
+        throw error;
+      });
   };
 }
 
--- a/src/notebooks/Notebooks.present.js
+++ b/src/notebooks/Notebooks.present.js
@@ -50,7 +50,7 @@ class Notebooks extends Component {
 
     return <Row>
       <Col>
-        <NotebooksTitle standalone={this.props.standalone} />
+        <NotebooksTitle standalone={this.props.standalone} message={this.props.message} />
         <NotebookServers
           servers={this.props.notebooks.all}
           standalone={this.props.standalone}
@@ -75,7 +75,14 @@ class NotebooksTitle extends Component {
   render() {
     if (this.props.standalone)
       return (<h1>Interactive Environments</h1>);
-    return (<h3>Interactive Environments</h3>);
+
+    if (!this.props.message)
+      return (<h3>Interactive Environments</h3>);
+
+    return [
+      <h3 key="title">Interactive Environments</h3>,
+      <div key="message">{this.props.message}</div>
+    ];
   }
 }
 
@@ -591,7 +598,7 @@ class StartNotebookServer extends Component {
   render() {
     const { branch, commit } = this.props.filters;
     const { branches } = this.props.data;
-    const { pipelines } = this.props;
+    const { pipelines, message } = this.props;
     const fetching = {
       branches: StatusHelper.isUpdating(branches) ? true : false,
       pipelines: pipelines.fetching,
@@ -606,10 +613,15 @@ class StartNotebookServer extends Component {
       || this.props.justStarted
     );
 
+    const messageOutput = message ?
+      (<div key="message">{message}</div>) :
+      null;
+
     return (
       <Row>
         <Col sm={12} md={10} lg={8}>
           <h3>Start a new interactive environment</h3>
+          {messageOutput}
           <Form>
             <StartNotebookBranches {...this.props} />
             {show.commits ? <StartNotebookCommits {...this.props} /> : null}
--- a/src/notebooks/Notebooks.present.js
+++ b/src/notebooks/Notebooks.present.js
@@ -40,6 +40,30 @@ import Sizes from "../utils/Media";
 
 import "./Notebooks.css";
 
+class NotebooksDisabled extends Component {
+  render() {
+    const postLoginUrl = this.props.location ? this.props.location.pathname : null;
+    const to = { "pathname": "/login", "state": { previous: postLoginUrl } };
+    const info = postLoginUrl == null ?
+      null :
+      (
+        <InfoAlert timeout={0} key="login-info">
+          <p className="mb-0">
+            <Link className="btn btn-primary btn-sm" to={to} previous={postLoginUrl}>Log in</Link> to use
+            interactive environments.
+          </p>
+        </InfoAlert>
+      );
+
+    return (
+      <div>
+        <p>This Renkulab deployment doesn&apos;t allow unauthenticated users to start Interactive Environments.</p>
+        {info}
+      </div>
+    );
+  }
+}
+
 // * Notebooks code * //
 class Notebooks extends Component {
   render() {
@@ -47,10 +71,13 @@ class Notebooks extends Component {
     const loading = this.props.notebooks.fetched ?
       false :
       true;
+    const message = this.props.message ?
+      (<div>{this.props.message}</div>) :
+      null;
 
     return <Row>
       <Col>
-        <NotebooksTitle standalone={this.props.standalone} message={this.props.message} />
+        <NotebooksTitle standalone={this.props.standalone} />
         <NotebookServers
           servers={this.props.notebooks.all}
           standalone={this.props.standalone}
@@ -59,13 +86,15 @@ class Notebooks extends Component {
           fetchLogs={this.props.handlers.fetchLogs}
           toggleLogs={this.props.handlers.toggleLogs}
           logs={this.props.logs}
-          scope={this.props.scope} />
+          scope={this.props.scope}
+        />
         <NotebooksPopup
           servers={serverNumbers}
           standalone={this.props.standalone}
           loading={loading}
           urlNewEnvironment={this.props.urlNewEnvironment}
         />
+        {serverNumbers ? null : message}
       </Col>
     </Row>;
   }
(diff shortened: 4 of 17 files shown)


rokroskar:
what happens if I have a bunch of environments running as an anonymous user but then I follow the advice on the page and log in? I guess I can't find the other environments i started anymore, right?

ableuler:
> what happens if I have a bunch of environments running as an anonymous user but then I follow the advice on the page and log in? I guess I can't find the other environments i started anymore, right?

Yes, that ist correct. You'll have them back when you log out though... Would be cool if we could make this transition smoother for users.

src/api-client/pipeline.js
```diff
@@ -81,6 +84,44 @@ function addPipelineMethods(client) {
       headers,
     }).then(response => response.data);
   };
+
+  /**
+   * Get the array of pipeline from GitLab
```
ciyer, line 89: I think this comment is inaccurate (copy/paste?).

src/api-client/pipeline.js
```diff
@@ -81,6 +84,44 @@ function addPipelineMethods(client) {
...
+    const url = `${client.baseUrl}/projects/${projectId}/registry/repositories`;
+
+    return client.clientFetch(url, {
+      method: "GET",
+      headers,
+    }).then(response => response.data);
+  };
+
+  /**
+   * Get the array of pipeline from GitLab
```
ciyer, line 105: I think this comment is inaccurate (copy/paste?).

src/notebooks/Notebooks.present.js
```diff
@@ -801,6 +846,29 @@ class StartNotebookPipelinesBadge extends Component {
 class StartNotebookPipelinesContent extends Component {
   render() {
     const pipeline = this.props.pipelines.main;
+    const pipelineType = this.props.pipelines.type;
+    const { pipelineTypes } = NotebooksHelper;
+
+    // unexpected case
+    if (pipelineType !== pipelineTypes.anonymous && pipelineType !== pipelineTypes.logged)
+      return (<div><p>Unexpected state.</p></div>);
```
ciyer, line 854: Can we tell the user what to do in this situation? Reload? Contact an admin?

src/notebooks/Notebooks.state.js
```diff
@@ -508,6 +522,73 @@ class NotebooksCoordinator {
...
+          lastParameters: newParameters
+        }
+      });
+      const registries = await this.client.getRegistries(projectId)
+        .catch(error => {
+          this.model.set("pipelines.fetching", false);
+          throw error;
+        });
+
+      // TODO: check if filters are oupdated due to discrepancies
```
ciyer, line 561: oupdated -> outdated?

src/project/Project.present.js
```diff
@@ -876,46 +834,93 @@ class ProjectEnvironments extends Component {
...
 
+function notebookWarning(userLogged, accessLevel, fork, postLoginUrl, externalUrl) {
+  if (!userLogged) {
+    const to = { "pathname": "/login", "state": { previous: postLoginUrl } };
+    return (
+      <InfoAlert timeout={0} key="permissions-warning">
+        <p>
+          <FontAwesomeIcon icon={faExclamationTriangle} /> As
+          an anonymous user, you can start <ExternalLink role="text" title="Interactive Envirnonments"
+            url="https://renku.readthedocs.io/en/latest/developer/services/notebooks_service.html" /> but
```
ciyer, line 845: ` but` -> `, but`

src/project/Project.present.js
```diff
@@ -876,46 +834,93 @@ class ProjectEnvironments extends Component {
...
+        </p>
+      </InfoAlert>
+    );
+  }
+  else if (accessLevel < ACCESS_LEVELS.DEVELOPER) {
+    return (
+      <InfoAlert timeout={0} key="permissions-warning">
+        <p>
+          <FontAwesomeIcon icon={faExclamationTriangle} /> You
+          have limited permissions for this project.
```
ciyer, line 860: Suggestion: 
```You have limited permissions for this project. You can launch an interactive environment, but you will not be able to save any changes. If you want to save your work [...]```

**Comment by ciyer** [comment]
I think the information presented here is good, but I think the order should be swapped. Instead of 

### Current
image

### Suggestion

image

And also, instead of "Log in to get all the benefits", how about "Log in for full access."

**Comment by ciyer** [comment]
I think the banner here should be the same width as the table:

image

**Comment by ciyer** [comment]
When the image is not available, I see different information when anonymous or logged in.

When anonymous, it says the build is pending:

image

When logged in, it says there is no image available:

image

I realize that it might not be possible to find out if the image is building in anonymous mode, so I think it should always just say the image is not available. We should probably say something like:

"The image for this commit is not currently available. Please [contact a project owner](link to list of project members) for help."

File: src/notebooks/Notebooks.present.js (hunk 3 above)
Comment (lorenzo-cavazzi) [reply src/notebooks/Notebooks.present.js:854]: Good point. Actually, this case should never happen, I added it for extra precaution during the development/test phase but I could remove it completely now (furthermore, there is nothing useful the user can do in that case)

**Comment by lorenzo-cavazzi** [comment]
> I think the information presented here is good, but I think the order should be swapped. Instead of
> 
> ### Suggestion
> 
> image

Do you think the message should be at the end only when there is no running environment or should I always put the message at the end of page? (It will go after the table if there are running environments).

File: src/notebooks/Notebooks.state.js (hunk 4 above)
Comment (lorenzo-cavazzi) [reply src/notebooks/Notebooks.state.js:561]: I can remove also this `TODO` since I added somewhere else a line of code to re-trigger the function when something relevant changes (e.g. the commit)

**Comment by ciyer** [comment]
> Do you think the message should be at the end only when there is no running environment or should I always put the message at the end of page? (It will go after the table if there are running environments).

Good point. Only if there are no environments running!

**Comment by lorenzo-cavazzi** [comment]
Thanks for the thorough review!

I addressed all points, but I ended up with a different (more verbose) explanation for unavailable Docker images. Feel free to suggest something different.

I add the screenshot, I'll update the deployment after I have implemented also the check on `.Values.global.anonymousSessions.enabled`

Screenshot_20200407_172222

**Comment by lorenzo-cavazzi** [comment]
@ableuler 

I have updated the PR to check for `.Values.global.anonymousSessions.enabled`. When `false`, I assume the Interactive Environments are disabled for anonymous users, therefore I display a message like this

Screenshot_20200407_194938

Tomorrow I'll try to deploy all the components for supporting unprivileged environments to a new deployment. If you want to update https://andreas.dev.renku.ch, this PR now requires the `tmp-login` endpoint to support the `redirect_url` variable (instead of `redirect-url`)

**Comment by ableuler** [comment]
Thanks @lorenzo-cavazzi for the updates to the UI. I have added two deployments for testing (including the latest changes to the UI).
https://tmp-on.dev.renku.ch/ (anonymous sessions enabled)
https://tmp-off.dev.renku.ch/ (anonymous sessions disabled)

**Comment by lorenzo-cavazzi** [comment]
Additional feedbacks from @ciyer 
* We may remove the warning message on the Environments list page when at least one environment is running -- no need to keep bothering the user with the same warning again and again
* There are cases when the `/registry/repositories` endpoint return an empty array (e.g. https://tmp-on.dev.renku.ch/projects/andreas/covid-19-dashboard/environments/new ) and the UI keep saying `Checking Docker image status...`
This seems to happen when the repo has no valid Docker mages at all (double check that) -- we can provide the same feedback as when the target image is missing from the registry

**Comment by rokroskar** [comment]
I see some repeated requests when accessing the notebook endpoints as a non-authenticated user: 

image (2)

**Comment by lorenzo-cavazzi** [comment]
> I see some repeated requests when accessing the notebook endpoints as a non-authenticated user [...]

Checking the URL, it seems that there is one repetition. It doesn't seem to come from the UI since it's in `/jupyterhub/.../login-tmp`, while the UI goes to `/api/auth/jupyterhub/.../login-tmp` only once

Screenshot_20200409_153525

**Comment by lorenzo-cavazzi** [comment]
I fixed the issues found by @ciyer and deployed the new version to https://lorenzotest.dev.renku.ch/

**Comment by ableuler** [comment]
> > I see some repeated requests when accessing the notebook endpoints as a non-authenticated user [...]
> 
> Checking the URL, it seems that there is one repetition. It doesn't seem to come from the UI since it's in `/jupyterhub/.../login-tmp`, while the UI goes to `/api/auth/jupyterhub/.../login-tmp` only once
> 
> Screenshot_20200409_153525

Yes, I can confirm what @lorenzo-cavazzi wrote: `/jupyterhub-tmp/hub/api/oauth2/authorize` is hit twice as a part of the regular login flow of the tmp-authenitcator. Also note that `/api/auth/jupyterhub/login-tmp` and `/jupyterhub-tmp/services/notebooks/login-tmp` are two different endpoints.

File: src/App.js [hunk 7]
```diff
@@ -52,6 +52,10 @@ class App extends Component {
       );
     }
 
+    // check anonymous sessions settings
+    const anonymousSessions = this.props.params["ANONYMOUS_SESSIONS"] === "true" ? true : false;
```
ableuler, line 56: Is this notation any different from just writing `this.props.params["ANONYMOUS_SESSIONS"] === "true"`?

src/App.js
```diff
@@ -52,6 +52,10 @@ class App extends Component {
...
+    const blockAnonymous = !user.logged && !anonymousSessions ? true : false;
```
ableuler, line 57: Same comment as above.

src/api-client/pipeline.js
```diff
@@ -63,6 +65,7 @@ function addPipelineMethods(client) {
       method: "GET",
       headers,
     }).then(response => response.data);
+    //return Promise.resolve([{ id: projectId, "status": "success" }]);
```
ableuler, line 68: This was useful for me, guess you can remove it now :)

src/api-client/pipeline.js
```diff
@@ -81,6 +84,46 @@ function addPipelineMethods(client) {
...
+
+  /**
+   * Get the list of available container registries. It should be 1 per project if there
+   * is at least a valid tagged image, otherwise 0.
+   *
+   * @param {number|string} projectId - project id or slug
+   */
+  client.getRegistries = (projectId) => {
+    const headers = client.getBasicHeaders();
+    headers.append("Content-Type", "application/json");
```
ableuler, line 96: Not necessarily for this PR, but I guess the commonly used headers should be defined as named constants to avoid hunting a typo for half a day at some point.

src/notebooks/Notebooks.present.js
```diff
@@ -801,6 +868,32 @@ class StartNotebookPipelinesBadge extends Component {
...
+    // anonymous
+    if (pipelineType === pipelineTypes.anonymous) {
+      if (pipeline && pipeline.path)
+        return null;
+
+      return (
+        <div>
+          <Label>
+            <p>
+              <FontAwesomeIcon icon={faExclamationTriangle} /> The image for this commit is not currently available.
```
ableuler, line 883:
```suggestion
              <FontAwesomeIcon icon={faExclamationTriangle} /> The image for this commit is currently not available.
```

src/notebooks/Notebooks.present.js
```diff
@@ -801,6 +868,32 @@ class StartNotebookPipelinesBadge extends Component {
...
+            </p>
+            <p>
+              Since building it takes a while, consider waiting a few minutes if the commit is very recent.
+              <br />Otherwise, you can either select another commit or <ExternalLink role="text" size="sm"
```
ableuler, line 887:
```suggestion
              <br />Otherwise, you can either select a previous commit or <ExternalLink role="text" size="sm"
```

helm-chart/renku-ui/templates/deployment.yaml
```diff
@@ -44,6 +44,8 @@ spec:
               value: {{ required "templateRepository.url must be specified, e.g., https://github.com/repos/SwissDataScienceCenter/renku-project-template" .Values.templatesRepository.url | quote }}
             - name: RENKU_TEMPLATES_REF
               value: {{ required "templateRepository.ref must be specified, e.g., master" .Values.templatesRepository.ref | quote }}
+            - name: ANONYMOUS_SESSIONS
+              value: {{ .Values.global.anonymousSessions.enabled | quote }}
```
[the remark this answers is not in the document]
lorenzo-cavazzi: I thought about putting a default in the `values` file but, since it's not a UI only setting, I would prefer to put the default value here -- it's not necessary, but I agree it may be easier to read

src/App.js
lorenzo-cavazzi: right, I got used to always specify the 2 values when using the ternary operator to avoid react inline code errors, but it's definitely not necessary here :+1:

src/api-client/pipeline.js
lorenzo-cavazzi: ops! Good catch!

src/api-client/pipeline.js
lorenzo-cavazzi: This makes sense, especially since there are only a couple of headers that are set most of the time. I guess having `ContentType.JSON`, maybe in addition to `client.setContentType()`, would be neat.
I'll create another issue for this.

src/notebooks/Notebooks.present.js
ciyer: I prefer the original wording: "The image for this commit is not currently available."

src/notebooks/Notebooks.present.js
lorenzo-cavazzi: I personally think `another` is more precise here. It's possible you tried to select a previous commit that has no valid image, but a more recent one does.

ciyer: I think the original wording is better. It is also possible that the image build on an older commit failed. In that case, the user might want to try out a more recent commit.

ableuler: ok, no problem - fine for me 👍

src/notebooks/Notebooks.present.js
ableuler: 👍

src/api-client/pipeline.js
ableuler: 👍

PR #898, opened 2020-04-08
fix(datasets): dataset has all folders from unziped file

### LINKED ISSUE #896: Wrong creation of dataset
I am trying to create a dataset in Renkulab by importing the zip file `ABdataset.zip` with the following hierarchy:

├── ABdataset
│   ├── 1900
│   │   └── 06_collectedinformation.pickle
│   ├── 1901
│   │   └── 06_collectedinformation.pickle
│   ├── 1902
│   │   └── 06_collectedinformation.pickle
│   ├── 1903
│   │   └── 06_collectedinformation.pickle
│   ├── 1904
│   │   └── 06_collectedinformation.pickle
│   ├── 1905
│   │   └── 06_collectedinformation.pickle

After uploading the zip file in the UI of Renku, it prompted that there was an error with the naming, and it was changed:
Screenshot 2020-04-07 at 15 08 53

But it did not provide any more information. However, when finally clicking on Create Dataset, this was the final result:
Screenshot 2020-04-07 at 15 13 16

I replicated this error in the following repo:
https://renkulab.io/projects/luis.salamanca/test_importdataset

And here I am in case I can help further 😄 

Thanks!
Cheers

vfried:
Fixes #896

(Luis made a good explanation of the problem on the issue)

Now the folder looks like this: 
image
There is still a small problem since there is an extra file there that wasn't on the folder structure.

Also, the redirect after the operation is finished isn't working well now.

(We weren't sending all the id's for unzipped files and now we are)

### DIFF
--- a/src/utils/FileExplorer.js
+++ b/src/utils/FileExplorer.js
@@ -19,9 +19,9 @@ function buildTree(parts, treeNode, jsonObj, hash, currentPath, foldersOpenOnLoa
 
   let newNode;
   if (parts[0] === jsonObj.name)
-    newNode = { "name": parts[0], "children": [], "jsonObj": jsonObj, "path": currentPath };
+    newNode = { "name": parts[0], "children": [], "jsonObj": jsonObj, "path": currentPath, "id": jsonObj.id };
   else
-    newNode = { "name": parts[0], "children": [], "jsonObj": null, "path": currentPath };
+    newNode = { "name": parts[0], "children": [], "jsonObj": null, "path": currentPath, "id": jsonObj.id };
 
   const currentNode = treeNode.filter(node => node.name === newNode.name);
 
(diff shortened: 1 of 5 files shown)


vfried:
I am also adding a preview of the uploaded folder so the user can see what will be there:

image


image

vfried:
Update... this is ready to be tested, i decided to do one further improvement, I decided to remove the already uploaded files from the uploader area and instead put them on top with the folder structure. 

It looks like this now... 
image

src/utils/FileExplorer.js
```diff
@@ -53,7 +55,7 @@ function getFilesTree(files, foldersOpenOnLoad) {
     const dir = list[i].atLocation.split("/");
     buildTree(dir, tree, list[i], hash, "", foldersOpenOnLoad);
   }
-  const treeObj = { tree: tree, hash: hash };
+  const treeObj = { tree: tree, hash: hash, leaves: Object.values(hash).filter(file => file.isLeave) };
```
ciyer, line 58: `isLeave` => `isLeaf`

src/utils/formgenerator/fields/FileUploaderInput.js
```diff
@@ -302,9 +341,48 @@ function FileuploaderInput({ name, label, alert, value, setInputs, help, disable
...
+    if (file.file_uncompress === FILE_COMPRESSED.UNCOMPRESS_NO || file.file_uncompress === undefined ) {
+      return initialFilesTree.hash[file.file_path] ? <small><br></br>
+        <span className="text-info">
+          &nbsp;*This file will be skipped because
+          &nbsp;there is a file with the same name inside the dataset.
+        </span>
+      </small> : null;
+    }
+    if (file.folder_structure !== undefined) {
+      const repeatedFile = file.folder_structure.leaves
```
ciyer, line 357: I think it would be nice to show a better error message here. Something like:
```
const repeatedFiles = file.folder_structure.leaves
        .filter(file => {
          let suspectedRep = initialFilesTree.hash[partialFilesPath + file.path];
          return (suspectedRep !== undefined) ? suspectedRep.isLeaf : false;
        });
      return repeatedFiles.length > 0 ? 
        <small><br></br>
            <span className="text-info">
            &nbsp;*{repeatedFiles.length > 1 ?
                `{repeatedFiles.length} files are already in the dataset and will be skipped.`:
                `{repeatedFiles[0]} is already in the dataset and will be skipped.`
        </span></small> : null;
```

src/utils/FileExplorer.js
vfried: I googled the word, but i guess i got it wrong anyways  haha

src/utils/formgenerator/fields/FileUploaderInput.js
vfried: I agree, i wasn't sure what message would be good there.

rokroskar:
super minor comment - "un-ziped" should be "unzipped" 👍

vfried:
Thanks for your review, especially for the english corrections :) 

Requested changes where addressed https://viginia.dev.renku.ch has the new changes

PR #900, opened 2020-04-09
feat: add maintenace page

### LINKED ISSUE #894: Maintenace page
## Motivation

We are planning to do regular updates on GitLab (~1/month, expected duration <30min each) during which renkulab won't be usable. The first one will probably happen on 15.04.2020 and will last longer (a few hours).
We may want to put a maintenance page during that time.

## Proposal

A simple solution would be creating the page and set it on/off through a variable in `/helm-chart/renku-ui/values.yaml` that we read in `src/index.js`. Renku team members would be able to easily change that value to turn on/off the page.

A more complex solution would use an array of maintenances with a `start` and a `end` timestamp. The UI can show the maintenance page during these hours, as well as a warning message in the Landing page for a while (2 days?) before the event.

We could instead get the maintenance information from a backend API, maybe the same one that will provide version's information (related to #607).
Not sure if this is a useful addition unless we plan to add a minimal user interface for the Renku team members to easily add/browse the maintenances.

lorenzo-cavazzi:
Provide a maintenance page to be used during renkulab maintenances when important components are not available (e.g. Keycloack, GitLab, ...). This prevents any API call.

The maintenance page is piloted by `ui.maintenance` in the values files or the `MAINTENANCE` environment variable in the pod (the latter is created from the former).

By default `ui.maintenance: false`. It can be set to `true` or to any other string that will be used as a message to display on the maintenance page.

**TEST**
Preview available here: https://lorenzotest.dev.renku.ch/
Feel free to modify the values while testing (the easiest is modifying the `MAINTENANCE` UI pod env variable from rancher).

This is the maintenance page with the standard message:

Screenshot_20200414_150734

fix #894

### DIFF
--- a/helm-chart/renku-ui/templates/deployment.yaml
+++ b/helm-chart/renku-ui/templates/deployment.yaml
@@ -44,6 +44,10 @@ spec:
               value: {{ required "templateRepository.url must be specified, e.g., https://github.com/repos/SwissDataScienceCenter/renku-project-template" .Values.templatesRepository.url | quote }}
             - name: RENKU_TEMPLATES_REF
               value: {{ required "templateRepository.ref must be specified, e.g., master" .Values.templatesRepository.ref | quote }}
+              {{- if .Values.maintenance }}
+            - name: MAINTENANCE
+              value: "1"
+              {{- end }}
           livenessProbe:
             httpGet:
               path: /
--- a/helm-chart/renku-ui/templates/deployment.yaml
+++ b/helm-chart/renku-ui/templates/deployment.yaml
@@ -46,7 +46,7 @@ spec:
               value: {{ required "templateRepository.ref must be specified, e.g., master" .Values.templatesRepository.ref | quote }}
               {{- if .Values.maintenance }}
             - name: MAINTENANCE
-              value: "1"
+              value: {{ .Values.maintenance | default (printf "true") | quote }}
               {{- end }}
           livenessProbe:
             httpGet:
--- a/helm-chart/renku-ui/templates/deployment.yaml
+++ b/helm-chart/renku-ui/templates/deployment.yaml
@@ -46,7 +46,7 @@ spec:
               value: {{ required "templateRepository.ref must be specified, e.g., master" .Values.templatesRepository.ref | quote }}
               {{- if .Values.maintenance }}
             - name: MAINTENANCE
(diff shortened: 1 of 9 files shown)


helm-chart/renku-ui/values.yaml
```diff
@@ -62,6 +62,9 @@ templatesRepository:
   url: https://github.com/SwissDataScienceCenter/renku-project-template
   ref: 0.1.9
 
+# any string here will enable the maintenance page and be added on it as an info
```
ciyer, line 65: Maybe clarify: `Any string here, other than 'false', will ...`

run-telepresence.sh
```diff
@@ -75,7 +75,8 @@ tee > ./public/config.json << EOF
   "SENTRY_URL": "${SENTRY_URL}",
   "SENTRY_NAMESPACE": "${SENTRY_NAMESPACE}",
   "RENKU_TEMPLATES_URL": "https://github.com/SwissDataScienceCenter/renku-project-template",
-  "RENKU_TEMPLATES_REF": "master"
+  "RENKU_TEMPLATES_REF": "master",
+  "MAINTENANCE": "true"
```
ciyer, line 79: This is probably just for debugging this PR and should be removed now.

helm-chart/renku-ui/templates/deployment.yaml
```diff
@@ -44,6 +44,10 @@ spec:
               value: {{ required "templateRepository.url must be specified, e.g., https://github.com/repos/SwissDataScienceCenter/renku-project-template" .Values.templatesRepository.url | quote }}
             - name: RENKU_TEMPLATES_REF
               value: {{ required "templateRepository.ref must be specified, e.g., master" .Values.templatesRepository.ref | quote }}
+              {{- if .Values.maintenance }}
+            - name: MAINTENANCE
+              value: {{ .Values.maintenance | default (printf "true") | quote }}
```
ciyer, line 49: Shouldn't the default be `false`?

helm-chart/renku-ui/values.yaml
lorenzo-cavazzi: good point! I added also an extra line about `true`

run-telepresence.sh
lorenzo-cavazzi: oh right, that accidentally slipped in!

PR #902, opened 2020-04-14
build: add helm-chart dir to .dockerignore

ableuler:
The build context can get pretty big if one has multiple versions of the UI chart lying around.


ableuler:
Yeah, my locally built charts ended up there, amounting to >500MB of build context... 🐌

PR #904, opened 2020-04-15
feat(datasets): dataset creation and import unified

### LINKED ISSUE #862: Merge the New and Import Dataset pages
## Motivation
The *New* and *Import* dataset pages should be merged.

* I sometimes pick the wrong option and it takes several clicks to change
* It would make links in the UI easier if these were merged (e.g., the topnav '+' only needs one option for dataset)

## Proposal

Have just one url, `[project]/datasets/new` and there should be a radio button with two options, create or import.

image

vfried:
Closes #862 

Dataset creation and import are now unified:

image

image

image

### DIFF
--- a/src/api-client/project.js
+++ b/src/api-client/project.js
@@ -448,6 +448,20 @@ function addProjectMethods(client) {
     const headers = client.getBasicHeaders();
     return client.clientFetch(url, { method: "GET", headers }).then((resp) => {
       return resp.data;
+    }).then(resp => {
+      let fullDatasets = resp.map(dataset =>
+        client.fetchDatasetFromKG(dataset._links[0].href));
+      return Promise.all(fullDatasets);
+    });
+  };
+
+  //in the future we will gett all the info we need for the dataset list from this call...
+  client.getProjectDatasetsFromKG_short = (projectPath) => {
+    let url = `${client.baseUrl}/knowledge-graph/projects/${projectPath}/datasets`;
+    url = url.replace("/api", "");//The url should change in the backend so we don't have to do this
+    const headers = client.getBasicHeaders();
+    return client.clientFetch(url, { method: "GET", headers }).then((resp) => {
+      return resp.data;
     });
   };
 
--- a/src/dataset/Dataset.present.js
+++ b/src/dataset/Dataset.present.js
@@ -24,8 +24,6 @@ import DOMPurify from "dompurify";
 import { API_ERRORS } from "../api-client";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faExternalLinkAlt, faPen } from "@fortawesome/free-solid-svg-icons";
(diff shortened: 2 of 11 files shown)


rokroskar:
This is looking really nice! There is one thing that is confusing to me: when I click on the project Datasets tab, I see the "Dataset List" but then if I select "Add dataset" from the drop-down I lose that context and now I have just "Create" and "Import" with no heading. Furthermore, the page looks like "Create" and "Import" are the only things available under "Datasets":

image

So I could see "List" "Create" and "Import" all being in that left column at the same level... otherwise the hierarchy is I think a bit confusing.

vfried:
I see what you mean, i will do something to improve this.

vfried:
New set of changes, still testing needed and fixing an issue with the KG... 

Create
image

Import
image

Modify
image

vfried:
I pushed changes to adapt thing to new interactions... the datasets list will look like this now:

image

rokroskar:
This is starting to look very nice - I have a few points that could use some improvement imho: 

1. the button says "New Dataset" but then the sub-page says "Add dataset" - should these use the same terminology? I would vote for "add" since when you import it's not really new. 

2. I'm not sure it makes sense to have a toggle to go between "create" and "import" but I also don't have a better solution in mind. Somehow a toggle in my mind is to switch something on or off not to choose between two options. That said, since it is a toggle, I'm naturally tempted to click it in which case it's immediately obvious what it's doing. So maybe it's ok? 

3. there seems to be a bug that doesn't allow me to create a new dataset if there are no datasets in the project already. I have this screen: 

image

And if I click on the "new dataset" button nothing happens.

vfried:
@rokroskar thanks for your comments.... you are right in all the things you pointed out.

I changed the New Dataset to Add a Dataset. 

I also fixed the bug that for empty projects.

Regarding the switch, it's not the right component for this... I wasn't 100% sure when using it, what about this... ?

image

rokroskar:
Cool! In this case I think I like the radio buttons 😄 

I have one final nit, but I think this is a larger issue because it's the same on the issues page (no pun intended): 

image

I've got the "Add dataset" button floating sort of in the middle of nowhere - I don't have a ton of visual cues here that this is within a boundary of some sort. I realize the point is to keep things minimalist (which I generally like) but an outline of some sort would help imo. What makes it even more confusing is that some banners and elements (I think for the environments) span a totally different range.

vfried:
> I have one final nit, but I think this is a larger issue because it's the same on the issues page (no pun intended):

The option here would be to do something like this.... In case of the collaboration tab we will add a message saying that there are no issues/merge requests i think, so things are't floating so much in the middle of nowhere

image

rokroskar:
> We should use this as the model/template for other entities like issues and environments

Yes, I was going to point out that the different components feel pretty disjointed now in terms of presentation, but I do really like the dataset implementation here so hopefully we can standardize things a bit.

PR #909, opened 2020-04-20
fix: update the jupyter button used in file view

lorenzo-cavazzi:
This fixes the `CheckNotebookStatus` component by correctly setting the `"user"` subModel in the constructor and showing the normal commands also to unprivileged users.

It goes further by adding basic support for being re-directed back to the notebook page when starting an environment from there. That part needs polishing but it will be addressed in the proper issue.

***How to test***
Go to a project containing a notebook, click on the "Start environment" Jupyter icon. This should now work again and bring you back to the original page when done, even if the feedback you get there is still poor.
https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/zurich-bikes-tutorial/files/blob/notebooks/zhbikes-notebook.ipynb


rokroskar:
Thanks @lorenzo-cavazzi this is going in the right direction I think!

PR #913, opened 2020-04-22
fix(datasets): dataset_name changed to short_name

vfried:
"dataset_name" inside the core service changed to "short_name" this will allow us to use in the future for "dataset_name" a proper title with spaces and short name would be the title that we have now.

For testing this in your own env you need to have the latest version of renku-core(0.10.3-cf08d03)


vfried:
> Code makes sense. Could you just add a reference to the PR and issue in the commit message? And could you add that there is a breaking change here, since this needs a new version of a backend component?
> https://www.conventionalcommits.org/en/v1.0.0/#summary

Done!

PR #917, opened 2020-04-24
feat: show commit list

### LINKED ISSUE #623: Refactor (project): component state cleanup
We have too many redux stores; we should have just one redux store for the entire application. If a component needs local state, it should use react hooks.

While doing this, we should also reduce to a minimum the information passed to sub-components of the project page. Currently, they get the full project state, but they should only receive what they need.

Here are the locations where stores are created:

*createStore*
- [X] index.js: user --> #632 
- [x] Landing.js: Home --> #1180

*StateKind.REDUX*
- [X] Notebooks.state.js: NotebooksModel --> #627 
- [ ] DatasetList.state.js: DatasetListModel --> #1182
- [x] ProjectFork.container.js: Fork --> #1118
- [x] ProjectList.state.js: ProjectListModel --> #1181
- [ ] Project.js: View --> #1183
- [x] ProjectNew.container.js: New --> closed by #963

*mix*
- [x] Issue.js / CollaborationList --> #624 

*StateKind.REACT*
- [ ] QuickNav.container.js: QuickNavContainerWithRouter --> this doesn't necessarily need to be moved since data are stored in a local react state. We may still want to move it to the global redux store, but I can't see many synergies now (that may change with a search bar using a KG query)

### LINKED ISSUE #852: Show commit messages in renkulab
## Motivation

Seeing the commit log is helpful to understand what has been happening in a project. I find myself often going to GitLab to see this, so we should integrate it into Renkulab directly.

## Proposal

Add a tab "Commits" to the overview page and show the commit log there. They should be grouped by date -- there should be a header with the date and all commits in that date should be there. See mockup below.

### Interactions

Initially, 

- It should be possible to diff a diff of the commit to its immediate parent
- It should be possible to navigate to the commit in GitLab
- Link to start environment at commit

Other interactions can be done later.

## Context

image

## Notes

* There is already code to get commits in the Notebook component (Notebook.state.js probably). This should be moved up to a shared location.
* May want to consider adding this a project submodel of the main redux store

lorenzo-cavazzi:
Adds all the necessary logic to handle projects commits.

It includes a `ProjectCoordinator` component to handle storing project data to the global Redux store (#623). This created extra difficulties since we can't easily migrate all the `ProjectModel` logic at once. To simplify the temporary co-existence, I created a higher order function `withProjectMapped` to map project data from the global store to single components whenever needed.

The first commit closes #852 by adding a new "Commits" section in the project overview tab, while the second commit re-uses the `CommitsView` component in the merge request section.
The third commit injects all the commits-related logic in the environments components. We may want to clean up a little bit `Notebooks.container.js` when `branches` will be managed by `ProjectCoordinator`.

***Preview***: https://lorenzotest.dev.renku.ch
***How to test***: browse to any project, go to the new "Commits" section on the "Overview" page. Please check also the "Merge request" section and check a notebook in the "File" tab, starting also a new environment from there. This should cover all the relevant changes.

Screenshot_20200424_194059

### DIFF
--- a/src/utils/Commits.js
+++ b/src/utils/Commits.js
@@ -1,0 +1,180 @@
+/*!
+ * Copyright 2020 - Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ *  renku-ui
+ *
+ *  Commits.js
+ *  Commits code and presentation.
+ */
+
+import React from "react";
+import { Row, Col, ListGroup, ListGroupItem, ButtonGroup, Button, UncontrolledTooltip } from "reactstrap";
+
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faFolderOpen } from "@fortawesome/free-regular-svg-icons";
+
+import { ExternalLink, TimeCaption, Clipboard, Loader } from "./UIComponents";
+import Time from "./Time";
+
+import "./Commits.css";
+
+// Constants
+const CommitElement = {
+  DATE: "date",
+  COMMIT: "commit"
+};
+
+
+// Helper functions
+function createDateObject(commit) {
+  return {
+    type: CommitElement.DATE,
+    date: Time.parseDate(commit.committed_date),
+    readableDate: Time.getReadableDate(commit.committed_date)
+  };
+}
+
+function createCommitsObjects(commits) {
+  const enhancedCommits = commits.reduce(
+    (data, commit) => {
+      if (!data.lastDate || !Time.isSameDay(data.lastDate, commit.committed_date)) {
+        data.lastDate = commit.committed_date;
+        data.list.push(createDateObject(commit));
+      }
+      data.list.push({ ...commit, type: CommitElement.COMMIT });
+
+      return data;
+    },
+    { lastDate: null, list: [] }
+  );
+
+  return enhancedCommits.list;
+}
+
+
+// React components
+
+/**
+ * Display a list of commits with links to GilLab pages
+ *
+ * @param {Object[]} props.commits - list of commits as returned by GitLab api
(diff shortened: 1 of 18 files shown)


rokroskar:
Is there an issue/epic for this?

lorenzo-cavazzi:
@rokroskar yes, it's #852 . I accidentally wrote twice #623 in the issue description (fixed) but it was in the `linked issues`

lorenzo-cavazzi:
@ciyer good point! We may want to handle pagination better in the future, but in the meanwhile that message would make this limitation clear.
I added the info inside a `<InfoAlert>` Component, let me know if you prefer the plain text.

Here is a comparison (first is the current implementation)

Screenshot_20200428_190617

Screenshot_20200428_190647

lorenzo-cavazzi:
@vfried thanks for the suggestions!

I think the current spacing on the top is in line with the other tabs (see Description and Stats). Furthermore, making the top padding `0` would make the top border look thicker (it's visible in your image).

Im not sure about moving the buttons to the right. I see that it would make the page more similar to the `Cards` in the Files tab, but it would be awkward to have the number and the refresh button far from the `Commits` text they qualify.

ciyer:
> @ciyer good point! We may want to handle pagination better in the future, but in the meanwhile that message would make this limitation clear.
> I added the info inside a `<InfoAlert>` Component, let me know if you prefer the plain text.
> 
> Here is a comparison (first is the current implementation). The final implementation has `100` instead of `10`.
> 
> Screenshot_20200428_190617
> 
> Screenshot_20200428_190647

Nice! I prefer the second one, the plain text option. The InfoAlert looks a bit too aggressive to my eyes.

lorenzo-cavazzi:
@ciyer done
I updated https://lorenzotest.dev.renku.ch

src/utils/Commits.js
```diff
@@ -91,9 +91,9 @@ function CommitsView(props) {
     return (<Loader />);
 
   const enhancedCommits = createCommitsObjects(props.commits);
-  const commitsView = enhancedCommits.map(commit =>
+  const commitsView = enhancedCommits.map((commit, i) =>
     <SingleCommit
-      key={commit.id ? commit.id : commit.readableDate}
+      key={commit.id ? commit.id : `${i}-${commit.readableDate}`}
```
lorenzo-cavazzi, line 96: the same `readableDate` shouldn't appear twice. If that happens, it may be a bug in how the date is computed

lorenzo-cavazzi: Using `committed_date` instead of `authored_date` seems to be the right thing to do to avoid out of order dates. It's also in line with what GitLab shows

lorenzo-cavazzi:
* Switched `authored_date` for `committed_date` to keep dates consistency with GitLab and avoid duplicates
* Squashed commits and added a reference to the PR to keep the history clean

vfried:
> Im not sure about moving the buttons to the right. I see that it would make the page more similar to the `Cards` in the Files tab, but it would be awkward to have the number and the refresh button far from the `Commits` text they qualify.

The new changes look really good, I agree with you with the buttons and the top padding thing is a small detail. I see inside stats is a similar thing so it's ok if you leave it as it is.

lorenzo-cavazzi:
I had to rebase, could someone please re-approve?

PR #920, opened 2020-04-29
fix(environments): skip autosaves for unprivileged users

### LINKED ISSUE #906: Do not show autosave data recovery dialog for unprivileged sessions
## Description

When starting an environment which has autosaved data, UI displays a dialog to notify users that some data has been recovered. An anonymous user or a logged-in user with no developer access, 
should not see this dialog because there is nothing to recover for them. However, this dialog is still displayed for them.

## Reproduce

1. Use a dev env with has anonymous sessions enabled
1. Find a project which has autosaved data
2. Try to start an environment from that project/branch

## Context
`renku-notebooks` disables autosaving for those users (implemented in https://github.com/SwissDataScienceCenter/renku-notebooks/issues/280).

lorenzo-cavazzi:
This filters autosaved branches based on the username.
Preview temporarily available on https://lorenzo.dev.renku.ch/

***How to test***
* Generate an autosaved branch: go to any public project, start an environment, do any change and stop the environment without committing/pushing
* Verify that trying to start an environment from the same project will show you the autosaved branch popup -- don't start it!
* Log out and try to start an environment as an anonymous user or using another user without developer privileges on the project. This time you shouldn't see any popup.

fix #906

### DIFF
--- a/src/notebooks/Notebooks.container.js
+++ b/src/notebooks/Notebooks.container.js
@@ -335,13 +335,23 @@ class StartNotebookServer extends Component {
     this.selectCommit();
   }
 
+  filterAutosavedBranches(branches, username) {
+    if (!username || !branches)
+      return [];
+    return branches.filter(branch => branch.autosave.username === username);
+  }
+
   mapStateToProps(state, ownProps) {
+    const username = state.user.logged ?
+      state.user.data.username :
+      null;
+    const ownAutosaved = this.filterAutosavedBranches([...ownProps.inherited.autosaved], username);
     const augmentedState = {
       ...state.notebooks,
       data: {
         ...state.notebooks.data,
         branches: ownProps.inherited.branches,
-        autosaved: ownProps.inherited.autosaved
+        autosaved: ownAutosaved
       },
       externalUrl: ownProps.inherited.externalUrl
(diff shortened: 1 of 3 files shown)


lorenzo-cavazzi:
@ableuler thanks for pointing out that the username is already in the autosaved branch name. I updated the PR to filter autosaved branches per-user -- and changed a bad variable name that made me believe it was the namespace there instead of the username :sweat_smile: 

Updated https://lorenzo.dev.renku.ch

src/notebooks/Notebooks.container.js
```diff
@@ -335,13 +335,23 @@ class StartNotebookServer extends Component {
...
   }
 
+  filterAutosavedBranches(branches, username) {
+    if (!username || !branches)
+      return [];
+    return branches.filter(branch => branch.autosave.username === username);
+  }
+
   mapStateToProps(state, ownProps) {
+    const username = state.user.logged ?
```
ableuler, line 345: This is always the gitlab username, right?

lorenzo-cavazzi: yes, it should be that one, that is returned by the API `/user` as `data.username`

PR #931, opened 2020-05-15
feat(markdown): code is highlighted a bit more

vfried:
Right now code gets a bit "lost" inside markdown....
image

I think we should highlight it a bit better.. 
image

This is just a proposal, we could also do it a bit darker with lighter words or something similar.

We could also think about going all the way and using something like this https://github.com/sindresorhus/github-markdown-css 
(This makes all our markdown look like the gitlab one) ---> NOT SURE THIS IS WHAT WE WANT

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -184,3 +184,30 @@ body {
     opacity: 1;
   }
 }
+
+.renku-markdown code {
+  padding: .2em .4em;
+  font-size: 85%;
+  background-color: rgba(0, 0, 0, 0.04);
+  border-radius: 3px;
+}
+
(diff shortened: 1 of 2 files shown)


ciyer:
This looks nice! I think we should go ahead and deploy the gray code background, since it is already an improvement. We can still think about adopting the GitHub stylesheet. but that is separate for me.

vfried:
> This looks nice! I think we should go ahead and deploy the gray code background, since it is already an improvement. We can still think about adopting the GitHub stylesheet. but that is separate for me.

Ok, sounds good to me.

PR #933, opened 2020-05-15
fix(datasets): if a dataset fails we display non-failed datasets

### LINKED ISSUE #924: When one dataset fetch fails dataset's aren't displayed
## Description

When one of the datasets inside a project fails instead of displaying the "non-failed" datasets everything fails and it looks like the project has no datasets.

image

## Reproduce

It happens here... https://renkulab.io/projects/covid-19/covid-19-public-data/datasets

vfried:
Closes #924 

In case a dataset fetch fails all dataset fetch was failing.

This was causing a "no dataset inside this project" message.

Now in case a dataset can't be fetched we display all the datasets that where fetched with no problem.

I also replaced some calls to get dataset for simplified ones when we don't need to get all the dataset info (i.e redirect after dataset add).

### DIFF
--- a/src/api-client/project.js
+++ b/src/api-client/project.js
@@ -468,7 +468,9 @@ function addProjectMethods(client) {
   client.fetchDatasetFromKG = (datasetLink) => {
     const headers = client.getBasicHeaders();
     const datasetPromise = client.clientFetch(datasetLink, { method: "GET", headers });
-    return Promise.resolve(datasetPromise).then(dataset => dataset.data);
+    return Promise.resolve(datasetPromise)
+      .then(dataset => dataset.data)
+      .catch((error) => "error");
   };
 
   client.getProjectDatasetsFromKG = (projectPath) => {
@@ -480,7 +482,10 @@ function addProjectMethods(client) {
     }).then(resp => {
       let fullDatasets = resp.map(dataset =>
         client.fetchDatasetFromKG(dataset._links[0].href));
-      return Promise.all(fullDatasets);
+      return Promise.all(fullDatasets).then(datasets => {
+        //in case one of the dataset fetch fails we return the ones that didn't fail
+        return datasets.filter(dataset => dataset !== "error");
+      });
     });
   };
 
--- a/src/project/datasets/import/DatasetImport.container.js
+++ b/src/project/datasets/import/DatasetImport.container.js
@@ -54,7 +54,7 @@ function ImportDataset(props) {
 
   const findDatasetInKgAnRedirect = (oldDatasetsList) => {
     let waitForDatasetInKG = setInterval(() => {
-      props.client.getProjectDatasetsFromKG(props.projectPathWithNamespace)
+      props.client.getProjectDatasetsFromKG_short(props.projectPathWithNamespace)
         .then(datasets => {
           if (datasets.length !== oldDatasetsList.length)
             redirectUserAndClearInterval(datasets, oldDatasetsList, waitForDatasetInKG);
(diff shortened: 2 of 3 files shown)


vfried:
To make a dataset fail you can do this... 

Inside /api-client/project.js 

```
 client.getProjectDatasetsFromKG = (projectPath) => {
    let url = `${client.baseUrl}/knowledge-graph/projects/${projectPath}/datasets`;
    url = url.replace("/api", "");//The url should change in the backend so we don't have to do this
    const headers = client.getBasicHeaders();
    return client.clientFetch(url, { method: "GET", headers }).then((resp) => {
      return resp.data;
    }).then(resp => {
      let fullDatasets = resp.map(dataset => {
        if (dataset.name === "NAME-OF-DATASET-THAT-WILL-FAIL")
          return client.fetchDatasetFromKG(dataset._links[0].href + "123");
        return client.fetchDatasetFromKG(dataset._links[0].href);
      });
      return Promise.all(fullDatasets);
    });
  };
```

src/project/datasets/import/DatasetImport.container.js
```diff
@@ -54,7 +54,7 @@ function ImportDataset(props) {
 
   const findDatasetInKgAnRedirect = (oldDatasetsList) => {
     let waitForDatasetInKG = setInterval(() => {
-      props.client.getProjectDatasetsFromKG(props.projectPathWithNamespace)
+      props.client.getProjectDatasetsFromKG_short(props.projectPathWithNamespace)
```
ciyer, line 57: Can you explain why you switched from `getProjectDatasetsFromKG` to `getProjectDatasetsFromKG_short` here?

src/api-client/project.js
```diff
@@ -480,7 +482,10 @@ function addProjectMethods(client) {
     }).then(resp => {
       let fullDatasets = resp.map(dataset =>
         client.fetchDatasetFromKG(dataset._links[0].href));
-      return Promise.all(fullDatasets);
+      return Promise.all(fullDatasets).then(datasets => {
+        //in case one of the dataset fetch fails we return the ones that didn't fail
+        return datasets.filter(dataset => dataset !== "error");
```
ciyer, line 487: Would it make sense to put in a placeholder for the dataset that failed, showing the information we have and a message that the dataset could not be retrieved?

vfried:
What happens here is that i get from the KG a list of datasets that has an error, then i try to fetch details from this list and this is when the error makes things fail. This should not happen... i shouldn't be getting from the backend a list that has an error. This is why i decided to drop the error instead of displaying something.

After kuba finishes this https://github.com/SwissDataScienceCenter/renku-graph/issues/298 we will stop fetching dataset per dataset and will have to change this anyways... hopefully this should be solved also so i wouldn't put to many effort on fixing something that will change soon.

src/project/datasets/import/DatasetImport.container.js
vfried: The get projects from KG fetches the datasets list and also the details of the dataset (this means one call per dataset plus one call to get the datasets list). The dataset import fetches de dataset list to see if there was a dataset added and if there is it redirects the user. We don't need all the dataset details that the getProjectDatasetFromKG gives. Maybe i whould rename the functions to "getProjectDatasetsFromKgEnriched" and "getProjectDatasetFromKG"

src/api-client/project.js
vfried: I made a comment about this bellow... i think that we shouldn't overthink this because it will change soon since kuba will return datasets enriched from the KG.

src/project/datasets/import/DatasetImport.container.js
ciyer: 👍 I like the name `getProjectDatasetsFromKgEnriched`

Issue #921, opened 2020-04-30
Improve message when dataset add takes too long


ciyer:
## Description

If dataset add takes too long, it's not clear whether you should push "Cancel" or "Add dataset". Could we make it just an "OK" button?

![image](https://user-images.githubusercontent.com/1196411/80679980-412b5680-8abe-11ea-8297-e95438090935.png)

## Reproduce

1. Go to a dataset
2. Click on 'Add to project'
3. If this operation takes a long time, you will get the message above.

PR #934, opened 2020-05-18
fix(datasets): when dataset takes too long we display an ok button instead of cancel and add dataset

vfried:
When adding a dataset takes too long we display display OK button when import fails (This button will close the modal) instead of "Add Dataset" and "Cancel" this should be less confusing for the user. 

image

To test this pull the branch and inside /dataset/addtoproject/datasetadd.container change line 125 

from:
`if ((cont > 30 && job.state !== "IN_PROGRESS") || (cont > 50 && job.state === "IN_PROGRESS")) {`

to:
`if ((cont > 1 && job.state !== "IN_PROGRESS") || (cont > 1 && job.state === "IN_PROGRESS")) {`

this will make the wait too short so the dataset add should display an error message and the new button.

### DIFF
--- a/src/dataset/addtoproject/DatasetAdd.container.js
+++ b/src/dataset/addtoproject/DatasetAdd.container.js
@@ -37,11 +37,14 @@ function AddDataset(props) {
   const [serverErrors, setServerErrors] = useState(undefined);
   const [submitLoader, setSubmitLoader] = useState(false);
   const [submitLoaderText, setSubmitLoaderText] = useState(ImportStateMessage.ENQUEUED);
+  const [takingTooLong, setTakingTooLong] = useState(false);
 
   const closeModal = () =>{
     if (!submitLoader) {
       addDatasetToProjectSchema.project.value = "";
       addDatasetToProjectSchema.project.options = [];
+      setServerErrors(undefined);
+      setTakingTooLong(false);
       props.setModalOpen(false);
     }
   };
@@ -85,6 +88,7 @@ function AddDataset(props) {
                 addDatasetToProjectSchema.project.options = [];
                 clearInterval(waitForDatasetInKG);
                 setServerErrors(ImportStateMessage.KG_TOO_LONG);
+                setTakingTooLong(true);
               }
             }
           });
@@ -92,7 +96,7 @@ function AddDataset(props) {
     }
   };
 
-  function handleJobResponse(job, monitorJob, cont, oldDatasetsList, projectPath, kgProgressCompleted) {
+  function handleJobResponse(job, monitorJob, waitedSeconds, oldDatasetsList, projectPath, kgProgressCompleted) {
     if (job) {
       switch (job.state) {
         case "ENQUEUED":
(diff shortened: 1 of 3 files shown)


rokroskar:
How is "too long" defined?

vfried:
> How is "too long" defined?

We discussed it with sekhar... i think it's 3 minutes if it didn't start the import process and 4 or 5 if we started it... I can check to be sure if you want

src/dataset/addtoproject/DatasetAdd.container.js
```diff
@@ -121,6 +125,7 @@ function AddDataset(props) {
     if ((cont > 30 && job.state !== "IN_PROGRESS") || (cont > 50 && job.state === "IN_PROGRESS")) {
```
ciyer, line 125: Because the duration is in counts, it is hard to see how long has elapsed. (30 is 5 min, 50 is 8min 20s). I know this wasn't changed in this PR, but could we switch the code to convert that into seconds?

rokroskar:
so once the user clicks "ok" what happens? does the operation get cancelled? or it keeps running? If it keeps running, why wait in the first place?

I think it would make more sense to not wait for the final OK from the backend but instead show a "dataset import in progress" information somewhere so the user isn't locked in for the duration of this timeout.

vfried:
That's a good point @rokroskar but i think we might need to think on how to implement this change... i think it could be out of the scope of this PR. I'd have to think about how to implement something like that. I also think that we should change all the redirects after dataset add/creat/import since they all wait until the KG finished building to redirect the user.Maybe we could do something like comparing the datasets in the KG vs the datasets that the core service returns and guess that an import process is in progress.

rokroskar:
Right, because there are several stages to the import, you may need to check different places for the information. 

1. the import is triggered and the backend responds that the import is in progress - here we don't need to keep the user waiting, we can instead poll the status endpoint and provide that information in the dataset view
2. import is done - the repo has been pushed - this kicks off the triples generation. Here as well we don't really need to keep the user waiting - I _think_ the graph API should return the information about which commits are being processed, but that doesn't necessarily tell you which dataset that is. Maybe the dataset endpoint should return the commit so you could check for that on the graph service.

src/dataset/addtoproject/DatasetAdd.container.js
vfried: I just pushed the changes for this, i agree with rok regarding the fact that we shouldn't wait for datasets to be 100% there but i think if we want to change the redirect strategy we should do this in a separate issue/pr... If you think we should include it in this PR we i can also do it but it would take longer.

ciyer:
Our original thinking was that importing a dataset would not take very long, so it is ok to block the user while this is going on, but this assumption seems to be incorrect.

Just like we have a top-level environments page, we could add a top-level "Operations" page that would list any asynchronous operations that are running and allow you to get the status of them. Doing this is a larger task, though.

ciyer:
Fix #921

PR #959, opened 2020-06-04
refactor(collaboration): issues and merge requests have now small visual improvements

### LINKED ISSUE #854: Improvements to presentation of merge request
## Motivation

The merge request display is pretty useful, but a few small improvements would make it better.

## Proposal

* [x] Show the tabs in this order:  'Changes', 'Commits', 'Discussion'
* [x] Show notebooks and markdown files first, then the opaque files second
* [x] Like issues, we should show Open and closed in tabs
* [ ] Show markdown files side-by-side, similar to how notebooks are shown
* [ ] The card header for the sections should have an icon/link to the the diff tool in GitLab
* [ ] Don't prompt user to create a new MR for every branch. Potential MRs should be in another tab, besides Open and Closed
* [ ] Make actions obviously buttons.

(See marked-up screenshot below.)

changes desc
buttons

## Additions

The _back-to-list_ button in Issues and Merge Requests should look the same as in the datasets. Instead of
image

It should look like:
image

And, there should be a back-to-list button in the "New Issue" and "New Merge Request" pages, just like there is in the "New Dataset" page.

### LINKED ISSUE #891: Visually separate the description of an issue from a comment
An issue which looks like this 

image 

in GitLab is displayed in an unexpected way in our UI:
image.

It would be nice to display the issue description differently. Right now it looks like an out-of-order comment.

Maybe this can still make it into the epic #850?

vfried:
Closes #891.
Partial work on #854.

I wonder if we should change the buttons for going to gitlab and closing the issue, they could be a bit bigger and have text so users know a bit better what the button does, i don't thing they are so self - explanatory.

image

### DIFF
--- a/src/collaboration/issue/Issue.js
+++ b/src/collaboration/issue/Issue.js
@@ -25,14 +25,14 @@
 
 import React, { Component, useState } from "react";
 import { Provider, connect } from "react-redux";
-import { Row, Col, Button, Card, CardHeader, CardBody, Alert, } from "reactstrap";
+import { Row, Col, Button, Alert } from "reactstrap";
 import { faGitlab } from "@fortawesome/free-brands-svg-icons";
-import { faBoxOpen, faBox, faListUl } from "@fortawesome/free-solid-svg-icons";
+import { faBoxOpen, faBox } from "@fortawesome/free-solid-svg-icons";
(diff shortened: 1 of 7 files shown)


vfried:
I just checked the new changes and it looks better, thanks for doing this!

PR #960, opened 2020-06-04
feat(dataset): description display has ben shortened

vfried:
The description is now cropped... This is just a proposal, when the user hovers over the description he can see a bit more, we can do changes to this by just cropping and not showing more.

We could also crop the creator's list, in most cases is not too long but in other it could be.

image

TEST: https://virginiatest.dev.renku.ch/projects/virginiafriedrich/project-11/datasets
https://virginiatest.dev.renku.ch/projects/virginiafriedrich/markdown-images/datasets

### DIFF
--- a/client/src/project/Project.css
+++ b/client/src/project/Project.css
@@ -32,33 +32,10 @@ Button.alert-link {
   font-size: 14px;
   margin-block-start: 0.5em;
   margin-block-end: 0.5em;
-  overflow: hidden;
-  text-overflow: ellipsis;
-  display: -webkit-box;
-  -webkit-line-clamp: 3;
+  overflow: hidden; 
   -webkit-box-orient: vertical;
 }
 
-
-.list-group-item:hover .datasetDescriptionText{
-  -webkit-line-clamp: 8;
-}
-
-
-.list-group-item:hover .datasetDescriptionText::after{
-  background: unset;
-}
-
-.datasetDescriptionText:after{
-  content: "";
-  text-align: right;
-  position: absolute;
-  bottom: 0;
-  right: 0;
-  width: 100%;
-  height: 2em;
-  background: linear-gradient(to bottom, rgba(255, 255, 255, 0), rgba(255, 255, 255, 1) 75%);
-}
 
 .datasetAuthors{
   display: block;
--- a/client/src/project/Project.css
+++ b/client/src/project/Project.css
@@ -32,6 +32,40 @@ Button.alert-link {
   font-size: 14px;
   margin-block-start: 0.5em;
   margin-block-end: 0.5em;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  display: -webkit-box;
+  -webkit-line-clamp: 3;
+  -webkit-box-orient: vertical;
+}
+
+
+.list-group-item:hover .datasetDescriptionText{
+  -webkit-line-clamp: 8;
+}
+
+
+.list-group-item:hover .datasetDescriptionText::after{
+  background: unset;
+}
+
+.datasetDescriptionText:after{
+  content: "";
+  text-align: right;
+  position: absolute;
+  bottom: 0;
+  right: 0;
+  width: 100%;
+  height: 2em;
+  background: linear-gradient(to bottom, rgba(255, 255, 255, 0), rgba(255, 255, 255, 1) 75%);
+}
+
+.datasetAuthors{
+  display: block;
+  white-space: nowrap;
+  overflow: hidden;
+  width: 98%;
+  text-overflow: ellipsis;
 }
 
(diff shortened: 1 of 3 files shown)


vfried:
After this is merged I would like to apply these changes to https://dev.renku.ch/datasets as well so that all datasets lists look simmilar.

ciyer (dismissed):
This looks really nice! When you do a squash merge, just remember to include a reference to the issue and PR.

vfried:
@ciyer before i merge, what do you think about the long list of authors? should we display all the authors or shall i just crop it and leave only the first line of authors?

ciyer:
> @ciyer before i merge, what do you think about the long list of authors? should we display all the authors or shall i just crop it and leave only the first line of authors?

Good point, if it is easy to shorten the list of authors to, say a max of 3, in the list view, that would be a good improvement.

vfried:
I shortened the author's list so that it has only one line and made the bottom of the dataset element smaller because it was too big.

Result...
image

ciyer (dismissed):
👍

rokroskar:
The usual convention when an author list is shortened is to use e.g. "Bardi, Alessia, et al." or list the first three with last name, first name + et al. if you want to keep three. But I wouldn't just cut off one of the names in the middle like it is now. 

I'm not sure how I feel about the automatic expansion - I can see this getting pretty annoying if the description is particularly long and/or there are many datasets in the list with long descriptions. Did you consider having a clickable "expand" or "show more" or something like that?

vfried:
@rokroskar 
Authors ---> not sure i follow...  I will crop the list to 3 and add ", et al." at the end.

Yes, that is another option... we can also just crop the description to a few lines and if they want to see more they need to access the dataset. What do we want to do?

vfried:
Result... 
image

rokroskar:
> Yes, that is another option... we can also just crop the description to a few lines and if they want to see more they need to access the dataset. What do we want to do?

I think being able to see the description in the list is nice, but I wouldn't make it automatic. Still though, is there an overall limit to how much we should expand?

vfried:
Regarding the file size... for git lfs files i need to fetch the file before i can display the size of the file... do we want this?

ciyer:
I think the `et al.` if there are more than 3 authors is a good solution. I do not want a button to expand, though. Having a button there will disrupt the visual flow of the page and require an extra click to see the description. I think we should either leave as is (truncated description that can be expanded on mouse over), or just remove the description from the listing entirely.

This current change only applies to the project/dataset listing, but there is a plan to introduce it to the general dataset listing as well.

vfried:
I would truncate the description to one line only maybe...

rokroskar:
I agree, I think just the titles are a little too limiting. I can see the auto expand being especially problematic on the general search/dataset listing page where you just want to scroll through titles and read the first few sentences of a description. Now you would have the text dropping down all the time, potentially. I just checked how it's done on Zenodo and Dataverse and they both just do what we're currently doing to keep the elements all the same size, more or less. So I'd vote for that.

vfried:
Do we like this...
image

ciyer:
> I just checked how it's done on Zenodo and Dataverse and they both just crop to keep the elements all the same size, more or less. So I'd vote for that.

We looked at the Zenodo and Dataverse solutions, but there is a fundamental difference. Zenodo permits stylized content, but doesn't render it in the listing: it just takes the first N words from the description. Dataverse just contains text in its descriptions.

We thought the datasets could look more attractive in listings if we display stylized content in the listing.

See:  https://virginiatest.dev.renku.ch/projects/virginiafriedrich/project-11/datasets / https://virginiatest.dev.renku.ch/projects/virginiafriedrich/project-11/datasets/test-long-description/

rokroskar:
I'm all for seeing stylized content, but you were concerned about an "expand" button breaking the flow of the page - doesn't dropping down potentially many many lines break the flow of the page even more? When what I'm looking at is a _listing_. Just wondering what the thinking is.

ciyer:
> I'm all for seeing stylized content, but you were concerned about an "expand" button breaking the flow of the page - doesn't dropping down potentially many many lines break the flow of the page even more? When what I'm looking at is a _listing_. Just wondering what the thinking is.

This is only on mouse over. If you aren't interested in seeing more information, you can avoid this by not mousing over the description.

rokroskar (approved):
🙌

ciyer:
Just here to document the options considered. Here is the first (truncated/expand on mouse over):

(expand on mouse: truncated)
image

(expand on mouse: expanded)
image

And here is the second:
(always truncated)
image

ciyer (approved):
Took a bit of a detour to get here, but I think in the end, the solution looks very good. 🙌

Issue #703, opened 2019-12-02
Use renku service to create project


ciyer:
# Description

Projects are currently autonomously created in the UI. We should use the renku service to create projects to remove code duplication between the UI and backend and ensure uniform semantics

lorenzo-cavazzi:
The PR for creating the new endpoint in the core service is almost ready SwissDataScienceCenter/renku-python#1287. That requires a small re-work of the `new project` page since we can't use the endpoint for GitLab specific settings that are not stored in the repository (e.g. visibility and description)

lorenzo-cavazzi:
Here is my proposal

For the `description` field, I suggest to drop the current one and use the `description` [implemented in the templates](https://github.com/SwissDataScienceCenter/renku-project-template/tree/master/python-minimal#-name-) -- adjusted if necessary. This requires to change the projects list to avoid showing the `No description available` label to something similar to GitLab.

<kbd><img src="https://user-images.githubusercontent.com/43481553/83043267-f06a3780-a042-11ea-8b79-fd61ad922236.png" /></kbd>

For the visibility, I would leave the dropdown as it is now and invoke the GitLab API once the project is created, along with the KG API for indexing the project (transparent to the user).

We have a lot of extra flexibility on how to chose the repository/repositories to fetch the templates from, and we want to support more than 1 repo (see #641).
I would allow the admins to personalize this page from the [value file](https://github.com/SwissDataScienceCenter/renku-ui/blob/cb074bd8617e61a20f8352f07758b2dcaafddcb7/helm-chart/renku-ui/values.yaml#L61) by changing the `templateRepository` to allow:
1. providing more repositories
2. deciding if users can get templates from custom repositories

This is a stub of the values

```
templates:
  custom: true
  repositories:
   - url: https://github.com/SwissDataScienceCenter/renku-project-template
     ref: 0.1.11
     prefix: Renku
   - url: https://github.com/lorenzo.cavazzi.tech/renku-project-template
     ref: master
     prefix: OpenLab
```

In the UI, I suggest to split the project creation into 2 steps:
1. Select template
![Screenshot_20200527_174459](https://user-images.githubusercontent.com/43481553/83043639-65d60800-a043-11ea-9f19-d9fcf5f70e86.png)
2. Specify project data
![Screenshot_20200527_174433](https://user-images.githubusercontent.com/43481553/83043599-5a82dc80-a043-11ea-874b-1a8907e3f7cf.png)

[...]

PR #961, opened 2020-06-05
feat: improve project search and project list layout

lorenzo-cavazzi:
In #703, we suggest to slightly modify the project list layout, hiding the description when not available.

This PR introduces that change and it fixes a number of other issues related to project list and project search. The idea is to keep the layout consistent at different window sizes while adapting to the content (e.g. missing description, date positioning, ... -- see the screen below or the preview).

Preview: https://lorenzotest.dev.renku.ch/projects

Please feel free to suggest any further improvement.
The final result is closer to the current GitLab project list layout: https://dev.renku.ch/gitlab/

Screenshot_20200608_091701

Screenshot_20200608_091732

Screenshot_20200608_092950

### DIFF
--- a/src/landing/NavBar.js
+++ b/src/landing/NavBar.js
@@ -204,11 +204,10 @@ class LoggedInNavBar extends Component {
     if (null != nextRoute) this.props.history.push(nextRoute);
   }
   render() {
-    // TODO If there is is an active project, show it in the navbar
     return (
       <header>
         <nav className="navbar navbar-expand-sm navbar-light bg-light justify-content-between">
-          <span className="navbar-brand">
+          <span className="navbar-brand mr-2">
             <Link to="/"><img src={logo} alt="Renku" height="24" /></Link>
           </span>
           <button className="navbar-toggler" type="button" data-toggle="collapse" data-target="#navbarSupportedContent"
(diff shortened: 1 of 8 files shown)


src/landing/NavBar.js
```diff
@@ -204,33 +204,34 @@ class LoggedInNavBar extends Component {
...
+          <span className="navbar-brand mr-2">
             <Link to="/"><img src={logo} alt="Renku" height="24" /></Link>
           </span>
           <button className="navbar-toggler" type="button" data-toggle="collapse" data-target="#navbarSupportedContent"
             aria-controls="navbarSupportedContent" aria-expanded="false" aria-label="Toggle navigation">
             <span className="navbar-toggler-icon"></span>
           </button>
 
-          <div className="collapse navbar-collapse" id="navbarSupportedContent">
+          <div className="collapse navbar-collapse flex-wrap" id="navbarSupportedContent">
```
ciyer, line 218: This is very nice!

src/project/list/ProjectList.container.js
```diff
@@ -46,7 +46,7 @@ const urlMap = {
   projectsSearchUrl: "/projects/all",
   projectNewUrl: "/project_new",
   starred: "/projects/starred",
-  yourProjects: "/projects/your_projects"
+  yourProjects: "/projects"
```
ciyer, line 49: Good fix. That has probably been lingering for a while.

lorenzo-cavazzi:
Oh, that doesn't happen with other browsers (Firefox and Chrome).
I believe I have [identified the issue](https://stackoverflow.com/questions/57516373/image-stretching-in-flexbox-in-safari), fixing it should be a matter of adding `flex-direction: colum` to the image parent container.

I updated the preview. Could you check if it looks better?
https://lorenzotest.dev.renku.ch/projects/all?q=cramakri&page=1&orderBy=last_activity_at&orderSearchAsc=false&searchIn=users

lorenzo-cavazzi:
I made a couple of force pushes to squash a few commits and to add the links to the PR in the single commits.

Issue #641, opened 2019-10-04
UI Support for Instance-Level Templates


ciyer:
# Description

When creating a new project, the UI displays a list of templates that come from a helm-configurable repository. We should add a second such configuration for instance-level templates. This configuration would contain the URL to a repository that contains templates for the instance. This project should have the same structure https://github.com/SwissDataScienceCenter/renku-project-template with a manifest.yml and folders for each template. If this value is configured, we should show the instance-level templates together with the global templates in the UI.

# Documentation

We should document how to create instance-level templates. Otherwise, this feature isn't really useable.

rokroskar:
Should this be a list of potential template sources rather than a single repo?

lorenzo-cavazzi:
As discussed, it makes sense to wait for the service endopint for project creation to be ready

PR #963, opened 2020-06-12
feat: use core service to fetch templates and create projects

### LINKED ISSUE #703: Use renku service to create project

lorenzo-cavazzi:
Use the core service backend to read from a remote Renku template repository and create a new project.

It's still a draft, a few parts are still missing:
- fetching a custom user template (partially implemented)
- invoking the `templates.create_project` API
- changing the visibility after project creation

A preview is currently available here: https://lorenzotest.dev.renku.ch/projects/new

Closes #703 and #641 
Requires SwissDataScienceCenter/renku-python#1287

Reminders:
* Create PR to change docs/charts in "renku" and "sdsc-dev-docs"
* Create PR for the ops repositories to update the templates values
* Update the integration tests
* [Done] ~While we are working on this, we should flag projects that have the following names in the client validation because they will not make it to GitLab: https://docs.gitlab.com/ee/user/reserved_names.html~


rokroskar:
This is starting to look really great, thanks @lorenzo-cavazzi! I'm wondering if it's necessary to implement all of the bullet points in the first iteration - personally, I would be happy if the custom templates were left off in the first pass if it means this can be done quicker because it will already make a lot of other things much easier (resolving circular dependencies once and for all, for instance).

src/project/new/ProjectNew.state.js
```diff
@@ -0,0 +1,236 @@
...
+   *   the list in model.config.repositories will be used instead.
+   */
+  async getTemplates(sources = null) {
+    // use deployment repositories if nothing else is provided
+    if (!sources || !sources.length)
+      sources = this.model.get("config.repositories");
+
+    // verify sources and set fetching status
+    if (!sources || !sources.length) {
+      const errorText = "no template sources available in this Renkulab deployment.";
```
ciyer, line 117: I suggest:
```
const errorText = "No project templates are available in this RenkuLab deployment. Please notify a RenkuLab administrator and ask them to configure a project template repository.";
```

src/landing/NavBar.js
```diff
@@ -107,8 +107,11 @@ class RenkuToolbarItemPlus extends Component {
         Dataset
       </Link>
       : null;
-    const projectDropdown = <Link className="dropdown-item" id="navbar-project-new" to="/project_new">
-      Project
+    const projectDropdown = <Link className="dropdown-item" id="navbar-project-new" to="/projects/new">
+      Project (service)
+    </Link>;
+    const projectDropdownOld = <Link className="dropdown-item" id="navbar-project-new-old" to="/project_new">
```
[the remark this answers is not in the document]
lorenzo-cavazzi: yes, this will be removed in the final implementation

src/model/RenkuModels.js
```diff
@@ -172,6 +179,66 @@ const projectSchema = new Schema({
...
+      namespace: {
+        [Prop.SCHEMA]: new Schema({
+          id: { [Prop.INITIAL]: null, [Prop.MANDATORY]: true },
+          fetched: { [Prop.INITIAL]: null, [Prop.MANDATORY]: true },
+          fetching: { [Prop.INITIAL]: false, [Prop.MANDATORY]: true },
+          visibility: { [Prop.INITIAL]: null, [Prop.MANDATORY]: true },
+          visibilities: { [Prop.INITIAL]: [], [Prop.MANDATORY]: true }
+        })
+      },
+      userTemplate: {
```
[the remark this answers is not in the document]
lorenzo-cavazzi: Yes it's not implemented yet. I'll change to plural and leave it commented out

lorenzo-cavazzi:
I have addressed the points in the review and left the custom templates out for the moment being. I'll create another issue to keep track of that -- the implementation is quick but better testing this a bit before.

A few things to underline:
* @ciyer I have implemented a real-time validation that disables the "Create project" button and adds a popup with feedback when something is not correct. We can adjust it as needed, but I believe the Model validation is not required anymore since it tests fewer details. If you agree, I'll remove it.
* I still need to double-check the namespace dropdown menu, it doesn't work as expected.
* The backend needs to be fixed since it changes all the spaces with hyphens in the project title

Preview available here: https://lorenzo.dev.renku.ch
I'll wait for a review before creating the PR in the main Renku repository with updated docs, integration tests and values.

P.S: Sorry for force-pushing and losing the user's contributions, I couldn't find another easy way.

rokroskar:
Can templates provide the hint text? E.g. to have something other than "Insert a value"

lorenzo-cavazzi:
@ciyer Right, it doesn't make sense to remember the user input once the project has been created, but keeping other data (namespaces, templates, ...) is useful.
I fixed by re-setting the values on the main component constructor.

lorenzo-cavazzi:
@rokroskar 
I agree that "Initializing project" sounds better :smile: 

About the first point, do you mean that the hint text should also go in the placeholder of the input value instead of "Insert a value..."? Currently, it's on the bottom, similar to the template description. I kept it that way for coherence with the other components, but also because longer texts get cut in the placeholder area. The result on small screens (or also on big ones if the descriptions are too long) would be bad.

Screenshot_20200706_174540

Would removing the placeholder completely be helpful?

lorenzo-cavazzi:
The previous bugs are fixed, the updated preview is available at https://lorenzotest.dev.renku.ch

rokroskar:
> Would removing the placeholder completely be helpful?

Maybe? I guess that this field will be generated automatically for every value that the template requires, right? In that case there might be a lot of "Insert a value..." If it's required and not entered I'm guessing the validation will fail and you will show an error to the user anyway. So yes, from my pov it would be better to just leave it blank :)

rokroskar:
Ah so actually, even though the value is "required" the initialization can proceed just fine without it - so maybe we don't need to say it's required? We should probably make the variables in the templates manifest be a list and each item can have a `required: true/false` value to make this less ambiguous.

lorenzo-cavazzi:
In fact the popup says "Required for this template. Can be an empty string". Probably a bit confusing :grin: 
Adding the `required` field seems definitely a nice solution. It requires a bit of support from the service and it should work also from the command line. Should we move that to a second step and try to wrap up this first implementation?

I suggest to:
* Remove the confusing `i` popup
* Change the placeholder to "Template variable" (or "Template parameter")
* Leave the main description below the input

Screenshot_20200707_090813

ciyer:
I prefer **Template parameter**. Can we put initial capitalization on the template-parameter-description text? E.g., "Short description added..." rather than "short description added..."?

rokroskar:
We should leave it blank - the user won't know what we mean by "Template parameter/variable". +1 on the capitalization (this should (also) be fixed in the templates, but you may want to ensure capitalization here anyway since users will be supplying their own templates).

rokroskar:
> Should we move that to a second step and try to wrap up this first implementation?

yes, definitely - we should make an issue for it so we can follow up

vfried:
Hey, i took a quick look to the form and i wonder if we can make some small visual changes.

- Align Create project button to the right
- Add a cancel button that takes you back to where you where before
- Delete the info icons from the info (when you hover over them the pointer changes and it seems like this is a button)
- Make the title a bit smaller ("New Project")
- Add a bit of space on the left side of the form

We are generating other forms from the model with the form generator that makes all forms look the same and it would be good if we make this also look like that. 

To compare...

image

image

Sorry for asking this changes a bit late...

lorenzo-cavazzi:
Thanks all for the review and suggestions!

I updated the preview with the latest version of the UI and the core service. Please try it again and let me know.
https://lorenzotest.dev.renku.ch/

@rokroskar / @ciyer: descriptions and parameters capitalized, placeholder blank.

@ciyer: I overhauled the real-time validation. Now it's saved permanently in the store and accessible everywhere. The feedback is immediate but delayed for untouched inputs (until the first click on "Create Project"). We may further improve the title validation to avoid illegal chars.

@vfried: Thanks for the tips! I deleted the extra info icon and adjusted the pointer. For the title sizing, I left it unchanged because it's similar in the other main pages like Projects or Datasets, while the "New dataset" is a sub-page of a project page.
For the button changes, we can move it to another issue as discussed offline.
For the spacing on the left, I think it should be removed in the "Add dataset" page instead, since it's the only page where I find it and it makes content misaligned on the left (on Firefox, a scrollbar also appears on the bottom, maybe due to a `col` element inside another `col` instead of a `row`).

helm-chart/renku-ui/values.yaml
```diff
@@ -16,8 +16,8 @@ global:
 replicaCount: 1
 
 image:
-  repository: renku/renku-ui
-  tag: '0.7.4'
+  repository: lorenzocavazzitech/renku-ui
```
ciyer, line 19: This is probably a development value.

run-telepresence.sh
```diff
@@ -22,6 +22,11 @@ set -e
 CURRENT_CONTEXT=`kubectl config current-context`
 WELCOME_MESSAGE="## Welcome to Renku through telepresence
 Some deployment-specific information will be read from the your values.yaml file and be displayed as markdown file."
+TEMPLATES='{"custom":true,"repositories":
+[{"name":"Renku","ref":"master",
+"url":"https://github.com/SwissDataScienceCenter/renku-project-template"},
+{"name":"Point11","ref":"0.1.11",
```
ciyer, line 28: I think it makes sense to leave this in, since we might want to test this in the future, but I think the name should be "Telepresence" or something, so it will be obvious to the developer where this template came from.

helm-chart/renku-ui/values.yaml
lorenzo-cavazzi: ouch, that sneaked in. Thanks for spotting it!

lorenzo-cavazzi:
@ciyer changes done.
When merging to master, I will squash all the commits into a single one since it's a single big feature with breaking changes.

src/project/new/ProjectNew.present.js
```diff
@@ -20,126 +20,255 @@
...
+class Title extends Component {
+  render() {
+    const { handlers, meta, input } = this.props;
+    const error = meta.validation.errors["title"];
+    const url = "https://docs.gitlab.com/ce/user/reserved_names.html#reserved-project-names";
+
+    const help = (
+      <span>
+        <FontAwesomeIcon className="no-pointer" icon={faInfoCircle} /> There are a
+        few <ExternalLink url={url} title="reserverd names" role="link" /> you cannot use.
```
ciyer, line 120: Oops -- typo! Should be `reserved`

src/project/new/ProjectNew.state.js
```diff
@@ -0,0 +1,427 @@
...
+      projects.featured.member.map(project => project.path_with_namespace.toLowerCase()) :
+      [];
+
+    if (newInput)
+      input = Object.assign({}, input, newInput);
+    if (newTemplates)
+      templates = Object.assign({}, templates, newTemplates);
+
+    // ? reference https://docs.gitlab.com/ce/user/reserved_names.html#reserved-project-names
+    const reserverdNames = ["badges", "blame", "blob", "builds", "commits", "create", "create_dir",
```
ciyer, line 386: Typo -- should be `reservedNames`

src/project/new/ProjectNew.state.js
```diff
@@ -0,0 +1,427 @@
...
+    if (meta.namespace.fetching)
+      warnings["visibility"] = "Verifying visibility constraints.";
+    if (templates.fetching)
+      warnings["template"] = "Fetching templates.";
+
+    // check errors: require user intervention. Skip if there is a warning
+    if (!input.title)
+      errors["title"] = "Title is missing.";
+    else if (reserverdNames.includes(input.title))
+      errors["title"] = "Reserverd title name.";
```
ciyer, line 403: Typo: should be `Reserved title name.`

lorenzo-cavazzi:
Thanks @ciyer for double checking.
In the meanwhile, I did another round of tests and I spotted and fixed a couple of corner cases making the validation fail. The reason was the validation function not using the up-to-date list of user's projects (or missing it completely).

I deployed the latest version to https://lorenzotest.dev.renku.ch

helm-chart/renku-ui/values.yaml
```diff
@@ -58,9 +58,12 @@ welcomePage:
     Renku is software for collaborative data science. With Renku you can share code and data,
     discuss problems and solutions, and coordinate data-science projects.
 
-templatesRepository:
-  url: https://github.com/SwissDataScienceCenter/renku-project-template
-  ref: 0.1.11
+templates:
+  custom: true
```
rokroskar, line 62: what is this for?

rokroskar:
Is there still a blocker before this PR goes in? It's blocking the resolution of all the outstanding templating problems so it would be great to have it asap.

helm-chart/renku-ui/values.yaml
lorenzo-cavazzi: I started implementing #976 as part of this PR and then stopped, I didn't comment this line out

lorenzo-cavazzi:
@rokroskar We were waiting on renku-python -- integration tests were failing, it seems they are succeeding now so I can go on and merge all the template related PRs

lorenzo-cavazzi:
All up-to-date and ready to be merged
@rokroskar could you please re-approve? I had to solve conflicts when merging master
@ciyer do you want to re-test? I updated the preview in https://lorenzo.dev.renku.ch

helm-chart/renku-ui/values.yaml
rokroskar: ok that's what I figured, just wanted to double check.

rokroskar:
Thanks @lorenzo-cavazzi, great to see this complete!

PR #965, opened 2020-06-21
feat(datasets): new fields to dataset creation and edit of fields

### LINKED ISSUE #526: Edit basic dataset information
The user should be able to edit basic dataset information. for this on the dataset details page #522 we should add an edit button, that redirects the user to a form where the user can edit the dataset.

On this page, the user should also be able to add and delete files from a dataset.

Related to https://github.com/SwissDataScienceCenter/renku/issues/594

### LINKED ISSUE #790: Add new fields to dataset creation
Dataset Creation should have the following fields:

|field|type|autogenerated| already there | comments |
|-----------| ----------- | ----------- | ----------- | ----------- |
|id| - | x | x | |
|date_published| dateTime | x | | we should create a publish action, not sure if there shuld be a publish action from within the create |
|description| free text (HTML) | |x | |
|identifier| | x | x  | this should be a DOI or something similar |
|keywords| array of keywords/strings | | | |
|basedOn| | | | not sure about this field |
|license| | | | not sure about this field |
|name| text | | | right now is there but there is a bug when sending names with spaces so the name is the sort name |
|url| url text |x |x | on the create this is the renku url of the dataset |
|version|  | | | |
|created| dateTime | x |x | |
|short_name| |could be autogenerated | | is the same as name and shouldnt |
|_project| renku project | x | x | when a dataset is created it is created from inside a project |
|creator| array of creators | | | we should allow the user to input the full array of creators/authors of the dataset, possibly a creator could be an organization |
|in_language|  | | | |
|files| | | x | |
|tags| ? | ?  || not sure what this means|
|same_as| | | |  not sure we need this in creation |

### LINKED ISSUE #791: Add new fields to dataset edit
Dataset Edit should have the following fields:

|field|type|autogenerated| already there | comments |
|-----------| ----------- | ----------- | ----------- | ----------- |
|description| free text (HTML) | | | |
|keywords| array of keywords/strings | | | |
|basedOn| | | | not sure about this field |
|license| | | | not sure about this field |
|name| text | | | right now is there but there is a bug when sending names with spaces so the name is the sort name |
|version|  | maybe | | should there be a new version when the dataset is edited??? |
|short_name| |could be autogenerated | | |
|creator| array of creators | | | we should allow the user to input the full array of creators/authors of the dataset, possibly a creator could be an organisation |
|in_language|  | | | |
|files| | | x | |
|tags| ? | ?  || not sure what this means|
|same_as| | | |  not sure we need this in edit |

vfried:
Closes #790 
Closes #526 
Closes #791 

Looks like this PR is FINALLY ready to be tested, is not 100% reliable since sam is still fixing some issues with the EDIT. Since its not a small PR and the changes in the service will probably not cause changes in the UI code i think we can review this.

To review, create a dataset, edit the dataset, etc... 

Added the title and creators field

image

image

Let me know what you think... It's deployed on https://virginiatest.dev.renku.ch


lorenzo-cavazzi (changes requested):
Wow, this PR makes adding and editing datasets from the UI a smooth experience! Very well done!

I tried to test it thoroughly and I spotted a couple of details to be polished.
* The keywords contain a json object instead of the string value (see screenshot 1).
* The "Dataset files" card occasionally shows `Error fetching dataset files: internal error`, even if the API fetching the files seems to contain the proper values (see screenshot 1).
* I tried to edit an existing dataset (https://virginiatest.dev.renku.ch/projects/lorenzo.cavazzi.tech/test-datasets/datasets/sarscov2_detection_dog/) and I get a blank page with this error `(creator.email is null -- InputValidator.js@31)`
Not sure if only old projects can have authors without an email or if it's still possible (maybe from the cli?) but I would either handle this case or disable the edit button for projects not migrated to the latest version.
* The "Dataset" tab occasionally shows an error when trying to load the dataset page directly in the browser. The error seems to come from the core service (see screenshot 2) but there is no error logged in the core pods (I checked the pod and the 4 containers).
(e.g. https://virginiatest.dev.renku.ch/projects/lorenzo.cavazzi.tech/test-datasets/datasets/sarscov2_detection_dog/).
I believe this should be forwarded to the renku-core team, but since I managed to reproduce it only when loading the page directly and not when navigating to the dataset, I thought maybe some information is missing when invoking the API.

Screenshot_20201013_142634

Screenshot_20201013_143530

vfried:
Thanks for your review...
- The keywords issue is a problem that comes from the backend (provably something was imported wrongly). This @value that you see actually comes as a string from the core service. 
I see that in general this works well...
Imported dataset:
https://virginiatest.dev.renku.ch/projects/virginiafriedrich/more-dataset-testing/datasets/bip4covid19_impact_metr_17/
Local dataset:
https://virginiatest.dev.renku.ch/projects/virginiafriedrich/what-is-going-on/datasets/test-dataset-creation-all-fields/
(this are just examples) 

- Dataset Files card error ---> i think it's fixed but i couldn't find a reliable way of reproducing the error

- Creator email editing problem fixed ---> this has nothing to do with the migrations, regarding migrations i don't know if we want to not allow migrations, it could happen that two versions that are different don't affect dataset creation/editing and if they do they will get a message from the backend.

- As you said, this is a problem with the core service, i think it doesn't happen so much on the more sable releases but we should make sure that this is not constantly happening.

lorenzo-cavazzi:
I tried it again and I still get errors.
It's pretty easy to replicate by re-loading this dataset a couple of times in a row https://virginiatest.dev.renku.ch/projects/lorenzo.cavazzi.tech/test-datasets/datasets/sarscov2_detection_dog/

* The most promintent error is this one, apperently coming from the service, specifically from `GET /datasets.list` returning an error with code `32603` (pinging @jsam for some help with the debug). Trying to reload the page linked above a few times in a row within a short timespan should trigger it. I can't find any error in the core pods logs but it's easy to reproduce.
Screenshot_20201019_151653

* The other one is the `Error fetching dataset files: internal error`. This seems to be a UI error since the API response seems fine. I initially thought it happens sporadically, but I believe it's more complicated. I can't replicate it when the user is not the project's owner, but it happens _almost_ always when the user is the owner of the project. You can try by visiting the link above (should see the files) and then forking the project. This time, there should be an error (if not, try refreshing the page or going back to the list and again to the dataset details).
The first screen shows the (apparently valid) response from the service. The second shows the difference between the dataset view for project owner (on the right) and a third user (on the left)
Screenshot_20201019_151620

Screenshot_20201019_152145

lorenzo-cavazzi:
I forgot to mention that trying to browse the same dataset on dev works well with any user, so both points seem to be a regression
https://dev.renku.ch/projects/lorenzo.cavazzi.tech/test-datasets/datasets/sarscov2_detection_dog/

vfried:
I just fixed the datset files error, it's in a separate commit, a tiny mistake was causing a recall of the useEffect (this is a one line commit btw).

jsam:
@lorenzo-cavazzi @vfried Thanks for reporting the problem! However, I'm not sure if the problem is in the service itself. The version @vfried is using in this deployment is exactly the same to the one which is provisioned to the dev environment and things seems to be working fine there. I've manage to reproduce it successfully the first time I tried but since I did not find anything which could be problematic per say. Sadly, the core was provisioned without the `SENTRY_DSN` so any exception which was caught with internal error report, its stacktrace was lost. I asked @vfried to put the same sentry dsn value so if we get this error again it should be visible in the sentry (for now), which might help us troubleshoot this issue. However, change of environment variable triggered re-provisioning of the core containers and now the issue went away completely and I cannot reproduce it anymore. 

Let's try to use the deployment a bit and see if we can reproduce the behavior and catch the little bugger on the sentry.

:rocket:

src/api-client/dataset.js
```diff
@@ -132,14 +132,29 @@ export default function addDatasetMethods(client) {
 
         project_id = response;
 
-        return client.clientFetch(`${client.baseUrl}/renku/datasets.create`, {
-          method: "POST",
-          headers: headers,
-          body: JSON.stringify({
+        let postUrl = edit ? `${client.baseUrl}/renku/datasets.edit` : `${client.baseUrl}/renku/datasets.create`;
+        let body = edit ?
```
lorenzo-cavazzi, line 136: The 2 objects you are assigning to `body` seem to be the same. Is that correct? Should we remove the ternary operator?

lorenzo-cavazzi:
Thank you @jsam for helping with this! And great to hear that it was just a temporary problem :smile:  One minor comment: would it make sense to log the errors also in the pod logs to simplify debugging and be coherent with the other renku components?

lorenzo-cavazzi (changes requested):
Great improvement for the Dataset UX! Very nice work! :raised_hands: 
I have just a minor change request before merging this PR (see inline comment).

Here are 2 points we could improve in a follow-up issue:
1. I've noticed there are no dataset related tests. It would be great to add unit tests for the functions in `DatasetFunctions.js` and for the rendering of the main components.
2. I think there are still a few cases when there are duplicate calls to the APIs. It would be great to double-check when this is still happening (e.g. when clicking on the "modify" button, the `datasets.files` endpoint is called 3 times. Furthermore, I am not sure if it's necessary to invoke again the `datasets` and `status` endpoints but I may very well be wrong on this)
Screenshot_20201021_143717

src/api-client/dataset.js
vfried: Thanks for the review, you are right about the unit tests, I will create issues for that and for the duplicated call... I thought i solved it yesterday but maybe there is something I didn't catch that makes the page load twice/repeat calls.

Regarding the body thing, you are right... it's the same, it wasn't but with the latest changes, it became the same object

lorenzo-cavazzi: :+1:

lorenzo-cavazzi (approved):

PR #980, opened 2020-07-15
Integrate statuspage.io in the UI

### LINKED ISSUE #938: Maintenance
## Motivation

#894 created a page for showing maintenance information, but this is very rigid. We would like to be able to dynamically show information to the user.

## Proposal

Use renku.statuspage.io/api to get the information to show here. See the comments below.

### UI Change -- Help

Add a tab to the help page showing the status page information.

image

### UI Change -- Landing/Welcome

Add any scheduled maintenance information to the welcome screen on the landing page.

image

image

ciyer:
The is PR implements the design specified in #938.
The configuration for this feature is documented in https://github.com/SwissDataScienceCenter/renku/pull/1315.

The dev-values file template is updated in this PR: https://github.com/SwissDataScienceCenter/sdsc-dev-docs/pull/65

# Testing
- if the statuspage.id value is not configured, then you should not see the help/status page
- if the statuspage.id value is set to junk, then you should see a banner explaining that the statuspage is not correctly configured in the help/status page
- if you add a scheduled maintenance, you should a notification banner and details in the help/status page
- if a component is not stable, you should a notification banner and details in the help/status page

### Create a statuspage

You can create a free account at [statuspage.io](http://statuspage.io). 

### Setting the statuspage id

By default, no statuspage is configured. You can configure one by filling out the configuration.
```
ui:
  statuspage:
    id: [statuspage id]
```

The `run-telepresence.sh` script sets the statuspage to the dev statuspage id by default. You can override this by setting a STATUSPAGE_ID environment variable.

Finally, you can modify `src/index.js:81` and set the id explicitly. Just change this line to contain your statuspage id:
```const statuspageId = params["STATUSPAGE_ID"];```

The last option is nice for testing since you can change the statuspage id without having to redeploy.

# Screenshots

## Scheduled Maintenance

image

## Problems with Service (details)

image

## Problems with Service (landing page)
image

## Incorrectly configured statuspage

image

Fix #938

### DIFF
--- a/helm-chart/renku-ui/templates/deployment.yaml
+++ b/helm-chart/renku-ui/templates/deployment.yaml
@@ -48,6 +48,10 @@ spec:
               value: {{ .Values.gatewayUrl | default (printf "%s://%s/api" (include "ui.protocol" .) .Values.global.renku.domain) | quote }}
             - name: WELCOME_PAGE
               value: {{ .Values.welcomePage.text | b64enc | quote }}
+            {{- if .Values.statuspage }}
+            - name: STATUSPAGE_ID
+              value: {{ .Values.statuspage.id | quote }}
+            {{- end }}
               {{- if .Values.sentry.enabled }}
             - name: SENTRY_URL
               value: {{ .Values.sentry.url | quote }}
--- a/helm-chart/renku-ui/values.yaml
+++ b/helm-chart/renku-ui/values.yaml
@@ -95,3 +95,6 @@ privacy:
       contentClasses: text-white small
       buttonClasses: btn btn-sm btn-light mr-2
       buttonWrapperClasses: mt-2
+## If you maintain a statuspage.io page for the status of your RenkuLab instance, put the id for it here.
+statuspage:
+  id:
--- a/run-telepresence.sh
+++ b/run-telepresence.sh
@@ -38,6 +38,8 @@ else
   echo "Warning! your OS has not been tested yet"
 fi
 
+if [ -z "$STATUSPAGE_ID" ]; then STATUSPAGE_ID="r3j2c84ftq49"; else echo "STATUSPAGE_ID is set to '$STATUSPAGE_ID'"; fi
+
 if [[ $CURRENT_CONTEXT == 'minikube' ]]
 then
   echo "Exchanging k8s deployments using the following context: ${CURRENT_CONTEXT}"
@@ -93,7 +95,8 @@ tee > ./public/config.json << EOF
   "SENTRY_NAMESPACE": "${SENTRY_NAMESPACE}",
   "ANONYMOUS_SESSIONS": "true",
   "PRIVACY_ENABLED": "false",
-  "TEMPLATES": ${TEMPLATES}
+  "TEMPLATES": ${TEMPLATES},
+  "STATUSPAGE_ID": "${STATUSPAGE_ID}"
 }
 EOF
 
--- a/src/index.js
+++ b/src/index.js
@@ -40,9 +40,9 @@ Promise.all([configFetch, privacyFetch]).then(valuesRead => {
       params["PRIVACY_STATEMENT"] = privacy;
 
     // show maintenance page when necessary
-    const maintenace = params["MAINTENANCE"];
-    if (maintenace) {
-      ReactDOM.render(<Maintenance info={maintenace} />, document.getElementById("root"));
+    const maintenance = params["MAINTENANCE"];
+    if (maintenance) {
+      ReactDOM.render(<Maintenance info={maintenance} />, document.getElementById("root"));
       return;
     }
 
@@ -77,9 +77,13 @@ Promise.all([configFetch, privacyFetch]).then(valuesRead => {
     function mapStateToProps(state, ownProps) {
       return { user: state.user, ...ownProps };
     }
+
+    const statuspageId = params["STATUSPAGE_ID"];
+
     const VisibleApp = connect(mapStateToProps)(App);
     ReactDOM.render(
-      <VisibleApp client={client} params={params} store={model.reduxStore} model={model} />,
+      <VisibleApp client={client} params={params} store={model.reduxStore} model={model}
+        statuspageId={statuspageId} />,
       document.getElementById("root")
     );
   });
--- a/src/statuspage/Statuspage.present.js
+++ b/src/statuspage/Statuspage.present.js
@@ -1,0 +1,275 @@
+/*!
+ * Copyright 2020 - Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ *  renku-ui
+ *
+ *  Statuspage.present
+ *  Components for the displaying information from statuspage.io
+ */
+
+import React, { Fragment } from "react";
+import { Link } from "react-router-dom";
+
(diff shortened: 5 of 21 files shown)

### CODE REVIEW

**Comment by rokroskar** [comment]
This looks really nice! Does the user get a popup if there is a maintenance announcement that they haven't seen yet? Or do they have to go to "Status" to see that there is a planned or ongoing maintenance/outage? One option would be to have a small "indicator light" somewhere that would quickly give a clue that something is going on.

**Comment by ciyer** [comment]
> This looks really nice! Does the user get a popup if there is a maintenance announcement that they haven't seen yet? Or do they have to go to "Status" to see that there is a planned or ongoing maintenance/outage? One option would be to have a small "indicator light" somewhere that would quickly give a clue that something is going on.

This is in progress. Just added a screenshot for the outage announcement that is shown on the landing page. Adding the same for scheduled maintenance.

Once those are done, the only outstanding TODO is to poll periodically for changes.

**Comment by rokroskar** [comment]
Would it be useful to have a sample statuspage set up? I _think_ we can have one that is not published but should still allow you to use the API.

**Comment by ciyer** [comment]
> Would it be useful to have a sample statuspage set up? I _think_ we can have one that is not published but should still allow you to use the API.

I just set one up myself for testing. The id for the statuspage is a configuration parameter.

**Comment by rokroskar** [comment]
Ok perfect. I have a temp one that we set up for limited, but if you have it already then all good.

File: helm-chart/renku-ui/values.yaml [hunk 1]
```diff
@@ -73,3 +73,7 @@ sentry:
   enabled: false
   url: ''
   namespace: ''
+
+# If you maintain a statuspage.io page for the status of your RenkuLab instance, put the id for it here.
+statuspage:
+  id: ''
```
rokroskar, line 79:
```suggestion
##  If you maintain a statuspage.io page for the status of your RenkuLab instance, put the id for it here.
# statuspage:
#  id: 
```

helm-chart/renku-ui/templates/deployment.yaml
```diff
@@ -34,6 +34,8 @@ spec:
               value: {{ .Values.gatewayUrl | default (printf "%s://%s/api" (include "ui.protocol" .) .Values.global.renku.domain) | quote }}
             - name: WELCOME_PAGE
               value: {{ .Values.welcomePage.text | b64enc | quote }}
+            - name: STATUSPAGE_ID
```
rokroskar, line 37: would be better to put this into a conditional and not set it at all if the statuspage id is missing

run-telepresence.sh
```diff
@@ -38,6 +38,8 @@ else
   echo "Warning! your OS has not been tested yet"
 fi
 
+if [ -z "$STATUSPAGE_ID" ]; then STATUSPAGE_ID="5bcn9bqff4qt"; else echo "STATUSPAGE_ID is set to '$STATUSPAGE_ID'"; fi
```
rokroskar, line 41: don't put the renkulab id as a default

ciyer: Switched to renku-dev.

helm-chart/renku-ui/values.yaml
ciyer: Done.

helm-chart/renku-ui/templates/deployment.yaml
ciyer: Done.

helm-chart/renku-ui/values.yaml
rokroskar: Actually I think my suggestion will break the chart rendering with default values 🙈  

It should be

```
statuspage:
  id:
```

i.e. the elements should be there but not set to anything.

ciyer: I tested it, and I didn't have any problems with chart rendering in either case. Both "not specified at all" and "specified to empty" work. So it is your call how you want the chart to look.

rokroskar: I think it's better to just leave it null since that's what we have elsewhere and is usually what is done in helm charts.

ciyer: It's the way you requested now.

rokroskar (approved):

run-telepresence.sh
```diff
@@ -93,7 +95,9 @@ tee > ./public/config.json << EOF
   "SENTRY_NAMESPACE": "${SENTRY_NAMESPACE}",
   "ANONYMOUS_SESSIONS": "true",
   "PRIVACY_ENABLED": "false",
-  "TEMPLATES": ${TEMPLATES}
+  "TEMPLATES": ${TEMPLATES},
+  "ANONYMOUS_SESSIONS": "true",
```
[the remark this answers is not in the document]
ciyer: Good catch!

src/index.js
```diff
@@ -40,9 +40,9 @@ Promise.all([configFetch, privacyFetch]).then(valuesRead => {
       params["PRIVACY_STATEMENT"] = privacy;
 
     // show maintenance page when necessary
-    const maintenace = params["MAINTENANCE"];
-    if (maintenace) {
-      ReactDOM.render(<Maintenance info={maintenace} />, document.getElementById("root"));
+    const maintenance = params["MAINTENANCE"];
+    if (maintenance && maintenance !== "false" && maintenance !== "0") {
```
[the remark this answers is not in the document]
ciyer: 👍

src/statuspage/StatuspageAPI.js
```diff
@@ -0,0 +1,56 @@
...
+ *  @param {string} pageId - the pageId for this application's status page
+ */
+class StatuspageAPI {
+
+  constructor(pageId) {
+    this.pageId = pageId;
+    this.controller = null;
+  }
+
+  async summary() {
```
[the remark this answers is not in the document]
ciyer: I wasn't sure what to do about this. It is true that our other API code is in that folder, but those all calls to the gateway; the statuspage API goes directly to statuspage.io, so it made sense to me to keep that code together with the rest of the statuspage support.

If you want to argue for moving it, I am open to it.

src/statuspage/Statuspage.present.js
```diff
@@ -0,0 +1,275 @@
...
+      <br />
+      <h4>Scheduled Maintenance Details</h4>
+      <ScheduledMaintenance maintenances={summary.statuspage.scheduled_maintenances} />
+      <br />
+      <h4>Component Details</h4>
+      <ComponentStatusDetails components={summary.statuspage.components} />
+      <br />
+      <p>
+        For further information, see <a href={summary.statuspage.page.url}>{summary.statuspage.page.url}</a>.<br />
+        <TimeCaption caption="Status retrieved" time={summary.retrieved_at} endPunctuation=";" />
```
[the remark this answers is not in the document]
ciyer: Yes, I was also unsure about how best to handle this. I removed the "Status retrieved" part, but then I found it confusing because it wasn't clear if the information being displayed was retrieved, say two days ago, or just now, so I decided to leave it in.

ciyer:
> The code looks great, I left just a few minor comments.

Good catches there, thanks. I think I have addressed all your comments.

> I have 2 considerations (it's not necessary to address them in the PR):
> 
> * When I load the home page, I get the warning on top almost immediately, but after a few seconds it goes away for a moment and then it re-appears. It seems that data are fetched again when the user's projects are loaded.

Yes, this is very ugly, and I spent some time trying to figure out how to fix it, but I was not able to find a solution. I will make a follow-up issue. 

> * It would be nice to have the same info also on the landing page for anonymous users.

Good point. The UX should probably be a bit different for anonymous users. I will make a follow-up for this as well.

src/statuspage/Statuspage.present.js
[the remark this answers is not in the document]
lorenzo-cavazzi: An idea would be to re-compute the human-readable text every `n` seconds if the provided time hasn't changed in the meanwhile.
That's probably an improvement we can add in the `TimeCaption` component. We can open an issue if we think it's useful, but the impact here is rather limited.

src/statuspage/StatuspageAPI.js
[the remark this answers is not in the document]
lorenzo-cavazzi: I generally like the idea of keeping all the APIs in one place. It makes it easier to find them if something needs to be changed. Otherwise, in this case, I can't imagine re-using the API anywhere else, therefore it's probably not much important where the code is

lorenzo-cavazzi (approved):
All good! :tada: 

Let me know if it makes sense to create an issue for auto-updating the sentence for the `TimeCaption` component.

ciyer:
Follow-up issues:
- #1016: fix flashing
- #1017: added status pages for anonymous users

PR #981, opened 2020-07-15
feat: add privacy page and cookie banner

### LINKED ISSUE #949: renku-ui should give users a cookie opt-in/out dialog
## Motivation

renkulab should comply with the Cookie Law and let users know that we are using cookies to make the site function. This should be part of the privacy policy text. Ideally, this text should be modifiable by an admin so that other renku deployments can provide their own text easily. 

## Proposal

A standard "we use cookies, click OK" dialog somewhere and a link to the privacy policy.

### UI Changes -- Footer

image

## Context

https://gdpr.eu/cookies/

lorenzo-cavazzi:
This PR adds the cookie consent banner and a privacy page.
They should be easy to customize, so that a RenkuLab deployment can be adjusted without touching the UI code. I tried with `ConfigMaps` but I couldn get a working version (I left a stub in the last commit), so I ended up adding everything to the `values.yaml` file.
The long text for the Privacy page is expected to be a valid Markdown text, while the cookie banner has a standard layout/content, but both can be adjusted providing a custom html code and a json object -- here is the reference for the adjustable properties: https://github.com/Mastermindzh/react-cookie-consent

Preview: https://lorenzotest.dev.renku.ch

fix #949


helm-chart/renku-ui/templates/configmap.yaml
```diff
@@ -0,0 +1,40 @@
+# ---
```
ciyer, line 1: Is it possible to add a comment here that this is a WIP and not actually used?

helm-chart/renku-ui/values.yaml
```diff
@@ -70,3 +70,46 @@ sentry:
...
+    interdum diam commodo, non suscipit augue commodo. Praesent sapien arcu, ullamcorper eu sollicitudin in,
+    suscipit eu felis. Donec id nisl tincidunt, maximus odio a, eleifend mauris. Orci varius natoque penatibus
+    et magnis dis parturient montes, nascetur ridiculus mus. Nunc placerat massa rhoncus justo auctor, et pulvinar
+    velit cursus. Donec vitae nunc purus. Nunc libero dolor, sollicitudin at nibh ut, gravida dignissim dui. Sed
+    sit amet mauris porttitor, fringilla eros et, pulvinar eros.
+  banner:
+    content: |
+      This website uses cookies to enhance the user experience. By clicking
+      or navigating the site, you consent to the use of cookies in accordance with
+      our <a class="text-white" href="/privacy">Privacy Policy</a>.
```
ciyer, line 108: Could the link be styled so that is is clearly a link? Currently, it is not obvious that "Privacy Policy" can be clicked on.

helm-chart/renku-ui/templates/configmap.yaml
lorenzo-cavazzi: the idea is to completely drop this file if not useful (it's in the last commit along with another comment about `comfigMaps`)

helm-chart/renku-ui/values.yaml
lorenzo-cavazzi: Good point. I opted for underlining it. Feel free to suggest a different effect if you find something more intuitive.

Screenshot_20200720_233135

helm-chart/renku-ui/templates/configmap.yaml
rokroskar: I'm probably not understanding something, but why is it commented? If it's in the template it could be "excluded" with a conditional in the template

lorenzo-cavazzi: @rokroskar I'm currently not using the `configMap` at all cause I had problems getting the proper values in the UI, so I ended up storing everything in the `values.yaml`. If you think there are advantages in storing the content in a `configMap`, I'll try again.

rokroskar: I see - my proposal to use the configmap was motivated by the fact that these privacy policy texts will probably get quite long. Having the whole thing in the values file might make that file difficult to read.

lorenzo-cavazzi: Yeah that makes sense. I guess I should be able to get the value from the configMap with `valueFrom: configMapKeyRef: ` and somehow apply `b64enc` to prevent any problem when passing data to the UI.
I'll try again later.

lorenzo-cavazzi: ok, after many trials I found a solution. The content is now stored in a configmap in plain text (E.G: `lorenzo-renku-ui-privacy` https://rancher.renku.ch/p/c-l6jt4:p-bqms9/config-maps/lorenzo:lorenzo-renku-ui-privacy) and it's base64 encoded in the bash script that sets the env variables in the pod.

helm-chart/renku-ui/templates/deployment.yaml
```diff
@@ -50,6 +50,15 @@ spec:
               {{- end }}
             - name: ANONYMOUS_SESSIONS
               value: {{ .Values.global.anonymousSessions.enabled | default (printf "false") | quote }}
+            - name: PRIVACY_STATEMENT
```
rokroskar, line 53: I'm not sure it matters but you could mount the configmap as a file instead of putting all the text into an environment variable?

helm-chart/renku-ui/values.yaml
```diff
@@ -70,3 +70,46 @@ sentry:
   enabled: false
   url: ''
   namespace: ''
+
+privacy:
+  statement: |
```
rokroskar, line 75: This is not needed anymore if you're using the configmap?

helm-chart/renku-ui/values.yaml
```diff
@@ -70,3 +70,46 @@ sentry:
...
+    sollicitudin vel mauris ac, efficitur placerat nibh. Sed commodo ut urna eget rhoncus. Proin dignissim
+    sollicitudin turpis, in faucibus libero tristique eget. Vestibulum ac laoreet libero. Fusce laoreet odio
+    interdum diam commodo, non suscipit augue commodo. Praesent sapien arcu, ullamcorper eu sollicitudin in,
+    suscipit eu felis. Donec id nisl tincidunt, maximus odio a, eleifend mauris. Orci varius natoque penatibus
+    et magnis dis parturient montes, nascetur ridiculus mus. Nunc placerat massa rhoncus justo auctor, et pulvinar
+    velit cursus. Donec vitae nunc purus. Nunc libero dolor, sollicitudin at nibh ut, gravida dignissim dui. Sed
+    sit amet mauris porttitor, fringilla eros et, pulvinar eros.
+  banner:
+    content: |
+      This website uses cookies to enhance the user experience. By clicking
```
rokroskar, line 106: I would say that in our case the cookies we use all fall into the "required" category, i.e. we don't do any sort of tracking. So we could say here that "This website uses cookies in order to ensure basic functionality." or something similar.

helm-chart/renku-ui/templates/deployment.yaml
lorenzo-cavazzi: Yes, that's an alternative. I checked the limits on env variables and it's in the order of MBs, so we should have >1 million chars at disposal, enough for *most* Privacy Statement pages.
If you think a file is more elegant, I can change it, but I like that all the data passed from the deployment file are imported together in one place in the UI index page.

helm-chart/renku-ui/values.yaml
lorenzo-cavazzi: ooops, I forgot to remove it... thanks for point that out

helm-chart/renku-ui/values.yaml
lorenzo-cavazzi: I used *requires* instead of *uses*
> This website requires cookies in order to ensure basic functionality

helm-chart/renku-ui/templates/deployment.yaml
rokroskar: Right, that's why I said I'm not sure it matters :) But from the pov of someone, lets say, who is debugging the service and inspecting the environment of the pod, I guess it would be annoying to have 20 pages of a privacy policy spill out. It also seems a bit unnatural to have this information in an environment variable - if the service was running stand-alone you would expect this in a file. But if you think this is fine then keep it, if it causes issues we can change it.

lorenzo-cavazzi: > I guess it would be annoying to have 20 pages of a privacy policy spill out.
> It also seems a bit unnatural to have this information in an environment variable

On a second thought, I believe you are right, a file seems the proper place to put it. I mapped the configmap to a volume.

I am not very familiar with configmaps, how is this supposed to be modified by the admin of a specific environment? Should we expect them to change it after it's deployed, or should they provide it during `heml upgrade`?
If it's relevant, we can include configuration information in the dummy privacy statement instead of a `Lorem ipsum` fake text.

rokroskar: Yeah they will need to change the configmap by hand. This will require some documentation and an additional step after the installation. I like the idea of putting the documentation into this text though - maybe a short loremipsum followed by instructions on how to change it.

rokroskar: oh, but to answer your question - you would do it like this:

```
kubectl -n <namespace> edit configmap <configmap>
```

Since you will add the text to the configmap template you should be able to use the templating to give you more or less the exact command someone needs to run.

docker-entrypoint.sh
```diff
@@ -16,8 +16,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+export NGIX_PATH=/usr/share/nginx/html
```
rokroskar, line 19: typo in the variable name?

```suggestion
export NGINX_PATH=/usr/share/nginx/html
```

helm-chart/renku-ui/templates/deployment.yaml
```diff
@@ -19,6 +19,13 @@ spec:
         app: {{ template "ui.name" . }}
         release: {{ .Release.Name }}
     spec:
+      volumes:
+        - name: privacy
+          configMap:
+            name: {{ template "ui.fullname" . }}-privacy
+            items:
+            - key: privacy_statement
+              path: statement.md
```
rokroskar, line 28: Why not put the cookie banner configuration into files as well?

lorenzo-cavazzi: well, we can... I would say those config are more tweaks from the admin and I see them in the values instead of the privacy policy file (i.e. the banner config mainly change the layout and graphics)

ciyer:
I just made a commit to harmonize the banner text in the values.yaml to the default value in the Privacy.container.js code.

helm-chart/renku-ui/templates/deployment.yaml
ciyer: I have this version deployed in my namespace and I just tried

```kubectl --context switch-dev -n $DEV_NAMESPACE edit configmap $DEV_NAMESPACE-renku-ui-privacy```

After doing this, the UI no longer works. The call to the endpoint `/privacy` return the HTML for the UI, and the UI dies with

```
InvalidCharacterError: Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded.
```

File: helm-chart/renku-ui/templates/configmap.yaml [hunk 8]
```diff
@@ -0,0 +1,39 @@
...
+  name: {{ template "ui.fullname" . }}-privacy
+  labels:
+    app: {{ template "ui.name" . }}
+    chart: {{ template "ui.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  privacy_statement: |
+    # Privacy statement
+    ## Personal data we collect
```
ciyer, line 13: Do we want to change this text to include instructions for how to change it? Also see my comment about what happens if the configmap of a running instance gets edited.

helm-chart/renku-ui/templates/deployment.yaml
rokroskar: I don't _think_ the pod gets recreated automatically if you change the config map - you may need to redeploy it. Does it need to be base64 encoded?

lorenzo-cavazzi: The configMap content is not encoded, it's plain-text markdown.
I tried to manually edit the content in rancher and it works -- after manually re-deploying the UI (I think there is a way to automate this).
@ciyer We can try this step together to see what is going wrong

helm-chart/renku-ui/templates/configmap.yaml
lorenzo-cavazzi: I added very basic documentation to the default configmap.
Should we also add detailed information in the main renku documentation? Not sure where, probably here https://renku.readthedocs.io/en/latest/admin/index.html
We may want to discuss with the dev-ops people where to put this type of admin-level documentation

rokroskar: could you make a follow-up issue in the main renku repo?

lorenzo-cavazzi: Documentation PR here SwissDataScienceCenter/renku#1314

lorenzo-cavazzi:
@rokroskar  An empty value for the ConfigMap is now handled correctly and links to the privacy page are removed in that case. If the user manually enters the privacy page url, he will see a warning.
@ciyer I fear the wrong indentation of the `data` property in the ConfigMap caused a number of errors when trying to update it. It should be fixed now

Preview updated in https://lorenzotest.dev.renku.ch -- feel free to modify the ConfigMap there for testing.

lorenzo-cavazzi:
I made a few changes according to the comments in SwissDataScienceCenter/renku#1314.

An updated preview is available here: https://lorenzotest.dev.renku.ch

Since the default deployment doesn't enable the feature, I made these 2 operations for the preview:
- I manually created a ConfigMap named `privacy-page` starting from the provided sample.
- I customized the values file for the deployment with these

```
  privacy:
    enabled: true
    page:
      map: true
      configMapName: privacy-page
      configMapKey: privacy_statement
```

helm-chart/renku-ui/values.yaml
```diff
@@ -73,3 +73,22 @@ sentry:
   enabled: false
   url: ''
   namespace: ''
+
+privacy:
+  enabled: false
+  page:
+    map: false
```
rokroskar, line 80: `map` just means `enabled` right? Can we call it enabled?

helm-chart/renku-ui/templates/deployment.yaml
```diff
@@ -19,6 +19,15 @@ spec:
         app: {{ template "ui.name" . }}
         release: {{ .Release.Name }}
     spec:
+      {{- if and .Values.privacy.enabled .Values.privacy.page.map }}
+      volumes:
+        - name: privacy
+          configMap:
+            name: {{ .Values.privacy.page.configMapName | default (printf "%s-privacy-sample" (include "ui.fullname" .)) | quote }}
```
rokroskar, line 26: your default here is different from the default in the values file - this could get pretty confusing. I would suggest removing the default from the values file and including a comment there that briefly explains what that means (i.e. if it is left unset, the default privacy page will be generated).

helm-chart/renku-ui/values.yaml
lorenzo-cavazzi: It defines if the configmap is mapped to a local volume, so even if `true` it may map it but still be disabled (if no content is provided for the page).
But yeah, `enabled` is more descriptive and intuitive, I changed it :grin:

helm-chart/renku-ui/templates/deployment.yaml
lorenzo-cavazzi: Good point, I commented them out and added a short explanation

rokroskar:
🎉 I guess someone needs to make the privacy policy now... 🤔

lorenzo-cavazzi:
@rokroskar Yeah, that would be useful :smile: 
Didn't you get something from ETH?

rokroskar:
yep, it's in progress

PR #982, opened 2020-07-16
Stop using deprecated k8s beta APIs

ableuler:


ableuler:
> Changes look fine. I built and deployed this image, and everything seems to work, so 👍

Thanks! 🎉

Issue #971, opened 2020-07-01
Improvements to dataset create


emmjab:
## Motivation

I was helping a new user upload data, and it would have been nice to be able to specify the URL where the data was coming from in the dataset creation tab (instead of having to first download the data and reupload it into renkulab), which is functionality that exists in renku python. It was also unclear to the user without explanation the difference between create and import.

## Proposal

(1) It might be nice to include a short blurb about what a dataset is on the dataset add/import page with a link to the docs. E.g.:

A dataset is defined on RenkuLab as a named collection of zero or more files. The name, description, creator, creation date of the dataset, as well as the paths to all of the files that are added to the dataset, are tracked in a metadata file. If you are creating a dataset from a collection of files (either from your local machine or somewhere in the cloud), use "create". If you are importing a published dataset or a dataset from another Renku project (e.g. a set of files already with metadata, from Zenodo, Dataverse, or a Renku project), use "import".

create = "build a new dataset by filling out some metadata and uploading/importing individual data files"
import = "import a published dataset (already contains metadata) from zenodo or dataverse, or from another renku project."

(2) The `renku dataset add` command lets users specify URLs to pull data from. It would be useful to add this as an additional input box for upload.

ciyer:
Good points. I would prefer to use different words for the action of reusing a dataset, and the action of grabbing data from a URL, rather than using "import" for both cases, since I think that is confusing.

Here is my proposal for the changes to the UI.

In both cases the "More about datasets" should be a link to the documentation.

# Changes to Dataset Create

![image](https://user-images.githubusercontent.com/1196411/86453337-e4e2ee00-bd1d-11ea-905a-a9c7e2475ede.png)

The "URL" tab should have just one field to specify a URL.

# Changes to Dataset Import

![image](https://user-images.githubusercontent.com/1196411/86453359-ee6c5600-bd1d-11ea-9948-66637a57758d.png)

PR #985, opened 2020-07-20
feat(dataset): user can add files to datasets with urls

### LINKED ISSUE #800: Allow users to add a file to a dataset through a URL
We should allow users to add files to a dataset from a remote endpoint.

vfried:
The users can upload files using urls...
image

Closes #800 
Solves part of #971 
The other part of #971 is in #988

**HEADS UP!!!** There is a lot of duplicated code but this will be unified in #965 ---> given the fact that this pr allows user to edit most of the dataset metadata the create and edit calls to the backend are really similar and i prefer to unify this code at that point.


src/project/datasets/new/DatasetNew.present.js
```diff
@@ -34,6 +34,47 @@ import { faInfoCircle } from "@fortawesome/free-solid-svg-icons";
...
+        : null}
+      {props.jobsStats.inProgress.length > 0 ?
+        <div>
+          <strong>Uploads in progress:</strong>
+          <br />
+          {progress}
+        </div>
+        : null}
+      <br /><br />
+      <strong>If more files where added they should be in the dataset, it could take some time until
```
ciyer, line 73: I do not understand how/when this message is relevant. Can you explain that situation?

vfried: This is relevant in case you are uploading url files and files from your computer. It could be that there is some error in files from urls and files from your computer are still not in the KG, in that case you might see the dataset but not those files...

src/project/datasets/edit/DatasetEdit.present.js
```diff
@@ -24,104 +24,59 @@
...
-          });
-      }
-      else {
-        props.datasetFormSchema.name.value = props.dataset.name;
-        props.datasetFormSchema.description.value = props.dataset.description;
-        props.datasetFormSchema.files.value = props.dataset.hasPart;
-        setInitialFiles(props.dataset.hasPart);
-      }
-      setInitialized(true);
+    const tooLongText = "Knowledge graph build is taking too long," +
```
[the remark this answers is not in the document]
vfried: Done

src/project/datasets/edit/DatasetEdit.container.js
```diff
@@ -23,23 +23,193 @@
...
+
+  const submitCallback = e => {
+    setServerErrors(undefined);
+    setSubmitLoader(true);
+    const dataset = {};
+    dataset.name = datasetFormSchema.name.value;
+    dataset.description = datasetFormSchema.description.value;
+
+    const pendingFiles = datasetFormSchema.files.value
+      .filter(f => f.file_status === "pending").map(f => ({ "file_url": f.file_name }));
```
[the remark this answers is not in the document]
vfried: We do have constants, i just missed to replace this one... Fixed now.

vfried:
Changes look like this now...
image

src/project/datasets/new/DatasetNew.present.js
ciyer: Oh, I see. Can we make the message the following, then:
```The dataset has been created, but some files are still being processed. They will become visible when processing completes.```

PR #989, opened 2020-07-22
perf: adjust code fetching all projects

### LINKED ISSUE #986: Refactor getAllProjects
The function `getAllProjects` could be refactored to:
* avoid repeating code already in `getProjects`
* make the parameters explicit (improve readability)

from #983

lorenzo-cavazzi:
This improves the `client.getAllProjects` function and limits its use to whenever it's necessary.

* Prevent fetching user projects when browsing as an anonymous user.
* Re-use `client.getProjects` function in `client.getAllProjects` and make it more readable.
* Invoke `client.getAllProjects` also for starred projects to be sure we are not missing projects.

fix #986

EDIT: sorry for not making a preview available, I'm running out of test deployments :disappointed:

### DIFF
--- a/src/api-client/project.js
+++ b/src/api-client/project.js
@@ -96,9 +96,9 @@ function getFilesTreeLazy(client, files, projectId, openFilePath, lfsFiles) {
 
 }
 
-function addProjectMethods(client) {
+async function addProjectMethods(client) {
 
-  client.getProjects = (queryParams = {}) => {
+  client.getProjects = async (queryParams = {}) => {
     let headers = client.getBasicHeaders();
     return client.clientFetch(`${client.baseUrl}/projects`, {
       method: "GET",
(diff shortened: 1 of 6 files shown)


lorenzo-cavazzi:
While working on this, I came up 2 possible follow up -- I'll create the issues if they are relevant.

### 1.
We could further improve the projects fetching logic by:
* separating the `fetching` variable for starred and member projects.
* taking advantage of paging by fetching a few projects (5-10, enough to fill the preview in the Landing page), then lazy loading all the remaining. This would require a couple of extra control variables and minor adjustments to the current code.

The speedup is considerable only with a lot of member/starred projects (100+). We could probably address with lower priority.

### 2.
An interesting idea would be adding a global notification system so that we can popup info/ warning messages when something not too critical goes wrong.
E.G: errors in fetching starred projects currently result in `.catch(error => [])` without further actions. It's not easy to add visual feedback since it's invoked in many different places, and throwing an error is not desirable since the impact is limited. I would like to notify the user with a popup, possibly with a re-execute button. Ideally, I would add the warning to a global queue with a function like this: `this.model.addNotification("cannot get the starred projects", "Type.WARNING", functionToReInvoke)`.
If this makes sense, I'll create an issue. We probably need anyway an extensive notification system when we have a web socket backend, this could be the first step.

src/api-client/project.js
```diff
@@ -96,9 +96,9 @@ function getFilesTreeLazy(client, files, projectId, openFilePath, lfsFiles) {
 
 }
 
-function addProjectMethods(client) {
+async function addProjectMethods(client) {
```
ciyer: I think the addProjectMethods function should be synchronous.

ciyer:
> While working on this, I came up 2 possible follow up -- I'll create the issues if they are relevant.

I created #991 for the second suggestion. I think the first is also good, could you make an issue for that?

lorenzo-cavazzi:
Great!
The second issue is #993

PR #996, opened 2020-07-29
fix(datasets): new message for file upload limits

### LINKED ISSUE #973: data upload size limitations
## Motivation

Working with a user, took longer than we had patience for to upload 1GB so found an alternate way.

## Proposal

in small text in the upload box, it would be nice to specify the maximum size of the data upload. if there's no maximum, we could write that as well. we should also test to make sure a few large sizes are possible, and indicate about how long it might take to upload a few different sizes.

## Context

Some users had issues uploading larger (1GB+ data)

vfried:
I added a new message to give users a hint about what happens with each file size...

image

image

Closes #973


rokroskar:
"probably" is spelled with an Argentine accent :)

ciyer:
I made some changes, here is what it looks like now:

image

vfried:
> This is clearer now!

Much better!!! Thanks for rewording it!

PR #1000, opened 2020-08-03
feat(dataset): add dataset to project active and with migration check

### LINKED ISSUE #964: Uncomment the "Add to project" button
We commented the add to project button (inside datasets) and we should uncomment this when things are ready.

vfried:
When importing a dataset if the project you want to import the dataset to needs to be migrated you will see a warning message...

image

Testing available: https://virginiatest.dev.renku.ch/

In order to test this, you should try to import a dataset into a project that needs to be migrated using the add to project button. 

Closes #964

### DIFF
--- a/src/dataset/Dataset.present.js
+++ b/src/dataset/Dataset.present.js
@@ -22,7 +22,7 @@ import { Link } from "react-router-dom";
 import { Loader, FileExplorer, RenkuMarkdown } from "../utils/UIComponents";
 import { API_ERRORS } from "../api-client";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import { faExternalLinkAlt, faPen } from "@fortawesome/free-solid-svg-icons";
+import { faExternalLinkAlt, faPen, faPlus } from "@fortawesome/free-solid-svg-icons";
 import Time from "../utils/Time";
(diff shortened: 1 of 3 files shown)


rokroskar:
@lorenzo-cavazzi the import will work in that case though it will not show up in the UI... maybe it's ok for now until we switch to checking project-local dataset data with the core service instead of the KG?

lorenzo-cavazzi:
@rokroskar for me it's ok but this may be confusing for the users. If I try to re-import the dataset again after the KG integration is activated (something I assume one would try) I would get this error

Screenshot_20200810_163111

The information I get from the UI are conflicting.
I believe checking the webhook once the target project is selected is pretty easy, just a matter of invoking `GET /projects/<id>/graph/webhooks/validation` for the target project.

rokroskar:
Sure, I'm not saying the current situation is not confusing - but I'm wondering if it makes sense working around it too much since a much better fix will be available soon.

lorenzo-cavazzi:
Yeah I see your point.
I would rather put the feature on hold since the same thing can already be achieved by importing the dataset from the target project instead of "pushing" it from the source project (and that's error-free), but if you prefer to go on we can move forward with this PR.

PR #1008, opened 2020-08-13
feat: display of relative paths in markdown and improvement in file preview in markdown

### LINKED ISSUE #667: Display of images specified with relative path in readme files
There is no image preview for markdown files like: 

https://renkulab.io/gitlab/bst262-2019/problem1

Image from gitlab:
image

### LINKED ISSUE #941: URLs to project files
## Description

URLs linking to project files in markdown content are inconsistent with GitLab. In GitLab, one writes `path-to-file`, it Renkulab, it has to be `files/blob/[path-to-file]`.

We need to figure out if there is a way we can make this more consistent.

vfried:
I created a markdown component that fetches images from relative paths and also fixes relative links in markdown files.

I also improved/restyled the preview of files...

We only preview files with "!" in the front... since I think it's nice to also leave the option to not preview a file in case this is not desired.

image

Closes #667 
Closes #941

### DIFF
--- a/src/model/RenkuModels.test.js
+++ b/src/model/RenkuModels.test.js
@@ -2,7 +2,7 @@
 
 import { testClient as client } from "../api-client";
 import { StateKind, StateModel } from "../model/Model";
-import { Project, projectSchema } from "./RenkuModels";
+// import { Project, projectSchema } from "./RenkuModels";
 import { ProjectModel } from "../project/Project.state";
 
 describe("fetch project", () => {
--- a/src/utils/Markdown.js
+++ b/src/utils/Markdown.js
@@ -1,0 +1,186 @@
+import React, { useEffect, useState, useRef } from "react";
+import { sanitizedHTMLFromMarkdown } from "./HelperFunctions";
+import { CardBody, Card } from "reactstrap";
+import ReactDOMServer from "react-dom/server";
+import { FilePreview } from "../file";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faFile } from "@fortawesome/free-solid-svg-icons";
+
+const patterns = {
+  fileRefFull: /!\[(.*?)\]\((.*?)\)/g, //with !
+  fileRefTrigger: /(!?\[[^\])]+])\(([^)]*$)/,
+  urlFilesRef: /https?: \/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\+.~#?&//=]*)/,
+  urlRef: /(http|ftp|https):\/\/[\w-]+(\.[\w-]+)+([\w.,@?^=%&amp;:\/~+#-]*[\w@?^=%&amp;\/~+#-])?/
+};
+
+const IMAGE_EXTENSIONS = ["jpg", "jpeg", "png", "tiff", "pdf", "gif"];
+
+const STIFFNESS = 290;
+const DAMPING = 20;
+
+const REF_TYPES = {
+  LINK: "link",
+  IMAGE_PREV: "image",
+  FILE_PREV: "file", // with ! in the front
+};
+
+const getFileExtension = (file_path) => {
+  if (!file_path)
+    return null;
+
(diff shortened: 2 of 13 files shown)


vfried:
Preview available:
https://virginiatest.dev.renku.ch/projects/virginiafriedrich/overviewR

ciyer (commented):
Been testing this, and it is very nice, a major improvement to how it worked before! 

I noticed a couple of discrepancies to how images are handled in GitLab, and I think we should decide what to do about them.

1. In GitLab, relative paths work (e.g., I can do `../images/foo.png` that it will be resolved correctly), but this does not work in RenkuLab
2. In GitLab, images are set to `max-width: 100%`, so they never overflow, but they are shown full size in RenkuLab
3. In GitLab, images that are in LFS are rendered in markdown files, but this does not work in RenkuLab
4. Images are not rendered correctly in issue previews.

To see illustrations of these, compare the following pairs:

**Point 1**
[RenkuLab: something.md](https://virginiatest.dev.renku.ch/projects/cramakri/markdown-images/files/blob/afolder/something.md) and [GitLab: something.md](https://dev.renku.ch/gitlab/cramakri/markdown-images/-/blob/test/afolder/something.md)

**Points 2, 3**
[RenkuLab: README.md](https://virginiatest.dev.renku.ch/projects/cramakri/markdown-images/files/blob/README.md) and [GitLab: README.md](https://dev.renku.ch/gitlab/cramakri/markdown-images/-/blob/test/README.md)

**Point 4**
https://virginiatest.dev.renku.ch/projects/cramakri/markdown-images/collaboration/issues

I think we should make relative paths work in RenkuLab (1). For the max-width (2), I think the way it is handled in GitLab is a bit nicer, but I don't think it is a dealbreaker either. If handling LFS files is easy (3), I think it would be nice to have, but if it is difficult, it can be done later. I think we should show the alt-text for an image in an issue preview (4).

src/model/RenkuModels.test.js
```diff
@@ -2,7 +2,7 @@
 
 import { testClient as client } from "../api-client";
 import { StateKind, StateModel } from "../model/Model";
-import { Project, projectSchema } from "./RenkuModels";
+// import { Project, projectSchema } from "./RenkuModels";
```
ciyer, line 5: This was probably for debugging.

src/utils/Markdown.js
```diff
@@ -0,0 +1,194 @@
...
+  </div>;
+}
+
+/**
+ * Safely render markdown.
+ * @param {string} markdownText the markdown text to display
+ * @param {boolean} singleLine if true, render the output as a single line without line breaks
+ * @param {object} style any styles to apply
+ */
+function RenkuMarkdownDeep(props) {
```
ciyer, line 96: Why is this called "RenkuMarkdownDeep" -- what makes it "Deep"?

ciyer (changes requested):
Code looks good! I have two small things to fix: I think I found a leftover debugging comment, and there is a name I do not understand, but otherwise looks great.

src/utils/Markdown.js
vfried: Deep as in deep fetching of stuff, I guess the name is confusing. Maybe I can rename it to RenkuMarkdownWithFiles?

ciyer: How about `RenkuMarkdownWithPathTranslation`?

ciyer:
The fixes for the relative paths handle the most common case, which is great! Two change requests:

- `Rename RenkuMarkdownWithFiles` -> `RenkuMarkdownWithPathTranslation` (I think this is a clearer name)
- Could you write a couple of tests for the fixRelativePath function?

ciyer (changes requested):
The fixes for the relative paths handle the most common case, which is great! Two change requests:

- `Rename RenkuMarkdownWithFiles` -> `RenkuMarkdownWithPathTranslation` (I think this is a clearer name)
- Could you write a couple of tests for the fixRelativePath function?

vfried:
I addressed the requested changes and fixed a small bug I found with ids for file previews in issues.

ciyer (approved):
Looks good! If you squash, don't forget to add a reference to the issue #667 and the PR #1008 in the commit message.

PR #1010, opened 2020-08-17
fix: dataset name replaced by title where needed

### LINKED ISSUE #1009: Use dataset name and dataset title accordingly after changes in the KG
There were some changes in the dataset metadata in the kg and we need to make a quick fix to adapt our code to these changes.

Before in KG
dataset.name = "This is a dataset"
dataset.short_name --> wasn't there

Now in KG
dataset.name = this-is-a-dataset
dataset.title = "This is a dataset"

Before in CORE
dataset.short_name= this-is-a-dataset
dataset.name = "This is a dataset"

Now in CORE = Now in KG
dataset.name = this-is-a-dataset
dataset.title = "This is a dataset"

vfried:
Name was replaced by title for dataset display purposes...

Before:
image

After:
image

Right now the difference is only visible with datasets that have name and title that are not the same (like the flights dataset or zenodo imported datasets)

If you want to see the "before" you might have to deploy the renku-graph version: 1.5.0

Closes #1009

### DIFF
--- a/src/dataset/list/DatasetList.present.js
+++ b/src/dataset/list/DatasetList.present.js
@@ -41,7 +41,7 @@ class DatasetListRow extends Component {
           <Badge color="success" className="font-weight-light">{projectsCountLabel}</Badge>
         </div>
         <Link to={`${datasetsUrl}/${encodeURIComponent(dataset.identifier)}`}>
-          {dataset.name || "no title"}
+          {dataset.title || dataset.name}
         </Link>
         {
           dataset.published !== undefined && dataset.published.creator !== undefined ?
--- a/src/project/Project.present.js
+++ b/src/project/Project.present.js
@@ -1139,7 +1139,7 @@ class ProjectViewDatasetsOverview extends Component {
     let datasets = this.props.datasets.map((dataset) =>
       <OverviewDatasetRow
         key={dataset.identifier}
-        name={dataset.name}
+        name={dataset.title || dataset.name}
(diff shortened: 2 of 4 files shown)


src/project/Project.present.js
```diff
@@ -1139,7 +1139,7 @@ class ProjectViewDatasetsOverview extends Component {
     let datasets = this.props.datasets.map((dataset) =>
       <OverviewDatasetRow
         key={dataset.identifier}
-        name={dataset.name}
+        name={dataset.title}
```
lorenzo-cavazzi, line 1142: Shouldn't it be `dataset.title || dataset.name` also here?

vfried: This is true, i just fixed it...

src/dataset/list/DatasetList.present.js
```diff
@@ -41,7 +41,7 @@ class DatasetListRow extends Component {
           <Badge color="success" className="font-weight-light">{projectsCountLabel}</Badge>
         </div>
         <Link to={`${datasetsUrl}/${encodeURIComponent(dataset.identifier)}`}>
-          {dataset.name || "no title"}
+          {dataset.title || dataset.name}
```
ciyer, line 44: Do you still need a fallback as before? I.e.,
```{dataset.title || dataset.name || "no title"}```

vfried: No, since name is required for creating a dataset.

vfried: Title is not required so we do need dataset.name there

PR #1018, opened 2020-08-25
feat: remove superfluous server setup documentation

ciyer:
The server setup documentation isn't very helpful for the normal RenkuLab user. We can remove it.

The help page now looks like this:

image

### DIFF
--- a/src/help/Help.present.js
+++ b/src/help/Help.present.js
@@ -52,9 +52,6 @@ class HelpNav extends Component {
         <NavItem>
           <RenkuNavLink to={this.props.url.features} title="Features" />
         </NavItem>
-        <NavItem>
-          <RenkuNavLink to={this.props.url.setup} title="Set Up / Admin" />
-        </NavItem>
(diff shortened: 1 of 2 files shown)


src/help/Help.present.js
```diff
@@ -52,9 +52,6 @@ class HelpNav extends Component {
         <NavItem>
           <RenkuNavLink to={this.props.url.features} title="Features" />
         </NavItem>
-        <NavItem>
-          <RenkuNavLink to={this.props.url.setup} title="Set Up / Admin" />
```
lorenzo-cavazzi, line 56: Should we also remove `url.setup` from the `urlMap` function in ["src/help/Help.container.js"](https://github.com/SwissDataScienceCenter/renku-ui/blob/master/src/help/Help.container.js#L37) ?

ciyer: You are right. Done!

lorenzo-cavazzi (approved):

PR #1020, opened 2020-08-25
ci: fix test-chart action precedence

lorenzo-cavazzi:
The helm chart should be linted only for master/tagged version, not on the PRs.
This fixes a wrong precedence in the ci action preventing them from publishing the chart


ciyer (approved):
👍

PR #1023, opened 2020-08-31
fix: ckeditor toolbar works as expected now

### LINKED ISSUE #1022: Fix CKEditor multiple toolbar issue
There are two problems that should be fixed as part of this issue. Either by virtue of the upgrade or, otherwise, directly.

## Cannot edit content unless switched to "Markdown" mode.

image

## Switching markdown on and off leads to an accumulation of toolbars

image

_Originally posted by @ciyer in https://github.com/SwissDataScienceCenter/renku-ui/issues/997#issuecomment-683706981_

vfried:
I found this solution that fixes the ckeditor problem:

https://github.com/ckeditor/ckeditor5-react/issues/142#issuecomment-659060608

I implemented the solution but I am still not sure what are the consequences of this, besides the fact that the CKEditor works well now.

I ran a command found [here](https://github.com/browserslist/browserslist#query-composition) to compare the difference between the configurations, there is only one difference:

image

If we can live with this config change I would push this fix.

I saw that create react app also removed the "not ie 11" from their config here:
https://github.com/facebook/create-react-app/commit/f4f20a20d095c7ea97191292e9d17753ff24596a

Can be tested in https://virginia.dev.renku.ch/

Closes #1022

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4437,7 +4437,7 @@
     },
     "babel-plugin-syntax-jsx": {
       "version": "6.18.0",
-      "resolved": "http://registry.npmjs.org/babel-plugin-syntax-jsx/-/babel-plugin-syntax-jsx-6.18.0.tgz",
+      "resolved": "https://registry.npmjs.org/babel-plugin-syntax-jsx/-/babel-plugin-syntax-jsx-6.18.0.tgz",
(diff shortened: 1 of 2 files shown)


ciyer:
That is very weird, but it does indeed look good. Let's go with this!

ciyer (approved):
👍

PR #1029, opened 2020-09-05
refactor: dataset info partially retrieved from the core service

### LINKED ISSUE #997: Update CKEditor
We have a new TextEditor(CKEditor), we should continue improving the way it works and add it to where it's missing.

vfried:
Dataset info for projects is partially retrieved from the core service.

--> Dataset List is almost the same, I only took out the number of projects a dataset is in and replaced it for information about whether the dataset is or not in the KG:
image

--> Dataset list for projects that are not in the KG is now visible:
image

--> Inside datasets that are not in the KG: (this can be because the entire project doesn't have the kg activated or because the kg is building)
image 

--> Datasets that are not in the KG have the "add to project" button disabled.

There are no major changes that are visible in dataset preview, add and edit. This operations run much faster because they don't wait for the KG. Messages about KG errors were removed. 

When the planets are aligned all works much smoother, I have experienced some problems with the core but I think this is because things are running in production. Sam also mentioned some problems with the KG.
 
This is ready to be tested.

Preview: will be available soon, i built an image but there is something wrong with it.

To test in your own env you need to use virfried/renku-core:0.11.1-2b4d54e4 in the core service... This is a build from https://github.com/SwissDataScienceCenter/renku-python/pull/1508

Closes #997

### DIFF
--- a/src/api-client/dataset.js
+++ b/src/api-client/dataset.js
@@ -73,6 +73,43 @@ export default function addDatasetMethods(client) {
     });
   };
 
+  //we should use this instead of clone, this checks if the project was cloned already...
+  client.getProjectIdFromSvc = (projectUrl) => {
+    let headers = client.getBasicHeaders();
+    headers.append("Content-Type", "application/json");
+    headers.append("X-Requested-With", "XMLHttpRequest");
+
+    return client.clientFetch(`${client.baseUrl}/renku/cache.project_list`, {
+      method: "GET",
+      headers: headers
+    }).then(response => {
+      if (response.data.error !== undefined)
+        return response;
+
+      return response.data.result.projects.find(project => project.git_url === projectUrl);
+    }).then(cloned_project => {
+      if (cloned_project !== undefined)
+        return cloned_project.project_id;
+
+      return client.clientFetch(`${client.baseUrl}/renku/cache.project_clone`, {
+        method: "POST",
+        headers: headers,
+        body: JSON.stringify({
+          depth: 1,
+          git_url: projectUrl
+        })
+      }).then(response => {
+        if (response.data.error !== undefined)
+          return response;
+
+        return response.data.result.project_id;
+
+      }).then(project_id => {
+        return Promise.resolve(project_id);
+      });
+    });
+  };
+
   client.postDataset = (projectUrl, renkuDataset) => {
     let headers = client.getBasicHeaders();
     headers.append("Content-Type", "application/json");
@@ -80,29 +117,24 @@ export default function addDatasetMethods(client) {
 
     let project_id;
 
-    return client.clientFetch(`${client.baseUrl}/renku/cache.project_clone`, {
-      method: "POST",
-      headers: headers,
-      body: JSON.stringify({
-        depth: 1,
-        git_url: projectUrl
+    return client.getProjectIdFromSvc(projectUrl)
+      .then(response => {
+
+        if (response.data !== undefined && response.data.error !== undefined)
+          return response;
+
+        project_id = response;
+
+        return client.clientFetch(`${client.baseUrl}/renku/datasets.create`, {
+          method: "POST",
+          headers: headers,
+          body: JSON.stringify({
+            "name": renkuDataset.name,
+            "description": renkuDataset.description,
+            "project_id": project_id
+          })
+        });
       })
-    }).then(response => {
-      if (response.data.error !== undefined)
-        return response;
-
-      project_id = response.data.result.project_id;
-
-      return client.clientFetch(`${client.baseUrl}/renku/datasets.create`, {
-        method: "POST",
-        headers: headers,
-        body: JSON.stringify({
-          "name": renkuDataset.name,
-          "description": renkuDataset.description,
-          "project_id": project_id
-        })
-      });
-    })
       .then(response => {
         if (response.data.error) { return response; }
         else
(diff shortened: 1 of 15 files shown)


lorenzo-cavazzi (changes requested):
Works well! I've found a couple of minor fixes (see inline comments).
Other 3 things I noticed:
* From Overview --> Datasets, clicking on a dataset brings you to a (probably wrong) link where an error is generated and no further action is possible because the page goes blank (e.g. https://virginia.dev.renku.ch/projects/renku-qa/core-integration-test/overview/datasets)
* When you click on the Datasets tab, the first time it queries the `datasets.list` endpoint but the next times it doesn't. The information there doesn't refresh. I would query it every time the Datasets tab is loaded (it's already the case for Overview --> Datasets)
* The Datasets page alignment is wrong unless the page is small. I'll upload the screenshot for different widths to give an idea

lorenzo-cavazzi:
* Small size --> ok
  Screenshot_20200909_084934

* Medium size --> compressed on the left
  Screenshot_20200909_084947

* Large size --> ok-ish but right side could avoid wrapping word (as in small size)
  Screenshot_20200909_085006

ciyer (changes requested):
Have a some code/naming requests.

src/api-client/dataset.js
```diff
@@ -127,28 +159,53 @@ export default function addDatasetMethods(client) {
...
-        headers: headers,
-        body: JSON.stringify({
-          "dataset_uri": datasetUrl,
-          "project_id": project_id
-        })
       });
+  };
 
-    });
+  client.listProjectDatasetsFromCore = (git_url) => {
```
ciyer, line 182: I suggested `fooFromCoreService` above `fooFromCore` is also ok for me, but the naming format should be the same everywhere.

src/dataset/Dataset.present.js
```diff
@@ -170,16 +127,16 @@ export default function DatasetView(props) {
     <Row>
       <Col md={8} sm={12}>
         {
-          dataset.published !== undefined && dataset.published.datePublished !== undefined ?
+          props.dataset.published !== undefined && props.dataset.published.datePublished !== undefined ?
```
ciyer, line 130: Maybe extract `const dataset = props.dataset` to a local variable to avoid that duplication?

src/dataset/Dataset.present.js
```diff
@@ -223,47 +179,59 @@ export default function DatasetView(props) {
...
     <br />
     <DisplayProjects
-      projects={dataset.isPartOf}
+      projects={props.dataset.isPartOf}
       projectsUrl={props.projectsUrl}
     />
+    {
+      props.dataset.url === undefined ?
+        <Alert color="primary">
+          <strong>This dataset is not in the Knowledge Graph</strong>,
```
ciyer, line 215: Wording:

`<strong>This dataset is not in the Knowledge Graph</strong>;
this means that some operations on it are not possible.<br /><br />
If the dataset was created recently, and the Knowledge Graph integration for the project is active, the dataset should be added to the Knowledge Graph soon. Otherwise, you need to activate the Knowlede Graph to be able to use the full set of dataset features.
```

File: src/project/datasets/new/DatasetNew.present.js [hunk 4]
```diff
@@ -35,23 +36,17 @@ import { faInfoCircle } from "@fortawesome/free-solid-svg-icons";
...
-    }
     const failed = props.jobsStats.failed
       .map(job => <div key={"warn-" + job.file_url} className="pl-2">- {job.file_url}<br /></div>);
     const progress = props.jobsStats.inProgress
       .map( job => <div key={"warn-" + job.file_url} className="pl-2">- {job.file_url}<br /></div>);
     return <div>
       {props.jobsStats.tooLong ?
         <div>
-          {tooLongText}
+          This operation is taking too long and it will continue being processed in the background.<br />
```
ciyer, line 46: Wording:

```This operation is taking longer than expected; it will continue being processed in the background.<br />
           Please check the datasets list later to make sure that the new dataset is available. <br />
           You can also check the <Link to={props.overviewCommitsUrl}>commits list
           </Link> in the project to see if commits for the new dataset appear there.
```

ciyer (changes requested):
Some additional changes to wording.

src/project/Project.js
```diff
@@ -445,20 +448,20 @@ class View extends Component {
 
       datasetView: (p) => <ShowDataset
         key="datasetpreview" {...subProps}
-        progress={graphProgress}
         maintainer={maintainer}
-        forked={forked}
         insideProject={true}
         datasets={datasets}
+        datasetId={p.match.params.datasetId}
+        projectPath={projectPathWithNamespace}
```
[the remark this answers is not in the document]
vfried: Fixed

vfried:
Thanks for both of your reviews... All changes were addressed and reviewed, this is ready to be merged whenever renku-python and kg are ready to make a release.

vfried:
Things have been fixed and deployed to virginia.dev.renku.ch

src/project/Project.state.js
```diff
@@ -268,7 +268,7 @@ class ProjectModel extends StateModel {
       })
       .catch(err => {
         const datasets = [];
-        const updatedState = { datasets_kg: datasets, transient: { requests: { datasets_kg: false } } };
+        const updatedState = { datasets_kg: { $set: datasets }, transient: { requests: { datasets_kg: false } } };
```
lorenzo-cavazzi, line 271: I think the same `datasets_kg: { $set: datasets }` should be applied also a few lines above (264)

src/project/Project.state.js
```diff
@@ -268,7 +268,7 @@ class ProjectModel extends StateModel {
...
         this.set("core.datasets_kg", datasets);
         this.setObject(updatedState);
```
lorenzo-cavazzi, line 273: This duplicates the `datasets_kg` values in the redux store, so that I see the same value in both `core.datasets_kg` and `datasets_kg`. Is this intended or should this be `this.setObject( {core: updatedState} );`?
The same applies a few lines above (266)

lorenzo-cavazzi (commented):
Nice!
All good on my side, just a couple of comments to be sure the code is correct when setting the variables in the store (I believe either the `this.set()` or `this.setObject()` is not setting the proper variable, but maybe I misinterpreted -- see inline comments)

src/project/Project.state.js
vfried: I fixed this, i removed the this.set since we had the datasets lists duplicated, only one set is needed now. The other value was used by mistake in the overview of datasets but this is fixed now.

lorenzo-cavazzi (approved):
:+1:

src/dataset/Dataset.container.js
```diff
@@ -49,6 +49,9 @@ function mapDataset(dataset_core, dataset_kg, core_files) {
       dataset.sameAs = dataset_kg.sameAs;
       dataset.isPartOf = dataset_kg.isPartOf;
```
ciyer, line 50: So you want to add `dataset.insideKg = true` here?

ciyer (changes requested):
One question about the code in the latest commit.

src/dataset/Dataset.container.js
vfried: No, I would prefer to leave this empty since in the generic view this is not set (the datasets come from the KG and are not pre-processed)

ciyer: Could we just put a comment in there. It would be good to explain why there is an asymmetry.

vfried: Actually, while I was explaining myself better I realized that datasets are always going through this mapDataset function, it doesn't matter if they are inside a project or not.

It makes sense to have this property set properly... I just pushed the change

ciyer (approved):
🎉  Very nice! It's great to be able to look at datasets and create even on private projects that aren't in the KG.

rokroskar:
coming late to the party here, but nicely done! One (minor) question: have you thought about using some sort of icon or something to indicate whether the dataset is in the KG? The text seems a bit excessive.

vfried:
Yes, i thought about this... we have the icon that we are using for the KG inside files but i don't know how well this icon represents a graph

PR #1034, opened 2020-09-11
Get all code resources from webpack

### LINKED ISSUE #1027: Get all code resources from webpack
# Description

We currently reference `popper`, `bootstrap`, and `jquery` in the index.html and retrieve them from CDNs. We should include the following dependencies in `package.json`:

- https://www.npmjs.com/package/react-popper
- https://www.npmjs.com/package/jquery

and ensure they are available for bootstrap (bootstrap.js should already be included in `package.json`).

Then we can remove these references from index.html.

ciyer:
All code resources are retrieved from the renkulab server.

Before, `jquery`, `popper`, and `bootstrap` were retrieved from the web.

Compare before change:
image

With after:

Compare (new):
image

Now, only the Source Code Pro font is retrieved from the web, but if this is missing, it shouldn't cause a major problem for the UI, since it will default to another monospaced font.

The new version can be tested on https://sekhar.dev.renku.ch/

Fix #1027


lorenzo-cavazzi (approved):
:+1:

PR #1036, opened 2020-09-15
feat: client side notification system (core)

lorenzo-cavazzi:
Client-side notification system with toast notifications and dropdown menu in the top navbar.
Currently, only a couple of notebooks action trigger notifications.

For convenience, on the test deploment it's possible to create fake notifications from the notifications menu

Screenshot_20200915_124252

### DIFF
--- a/src/notifications/Notifications.container.js
+++ b/src/notifications/Notifications.container.js
@@ -27,7 +27,7 @@ import React, { Component } from "react";
 import { toast } from "react-toastify";
 import { connect } from "react-redux";
 
-import { NotificationsCoordinator, Notifications as NotificationsInfo } from "./Notifications.state";
+import { NotificationsCoordinator, NotificationsInfo } from "./Notifications.state";
 import { Notification, CloseToast, NotificationsMenu as NotificationsMenuPresent } from "./Notifications.present";
 
 
@@ -37,13 +37,16 @@ import { Notification, CloseToast, NotificationsMenu as NotificationsMenuPresent
  * @param {Object} client - api-client used to query the gateway
  * @param {Object} model - global model for the ui
  */
-const Notifications = (props) => {
-  const model = props.model.subModel("notifications");
-  const coordinator = new NotificationsCoordinator(props.client, model);
-  const { Levels, Topics } = NotificationsInfo;
+class NotificationsManager {
+  constructor(model, client) {
+    this.model = model.subModel("notifications");
+    this.coordinator = new NotificationsCoordinator(client, this.model);
+    this.Levels = NotificationsInfo.Levels;
+    this.Topics = NotificationsInfo.Topics;
 
-  // get toast settings -- can't be static anymore once users will be able to change settings from the UI
-  const settings = coordinator.getToastSettings();
+    // can't be static anymore once users will be able to change settings from the UI
+    this.settings = this.coordinator.getToastSettings();
+  }
 
   /**
    * Add a notification to the list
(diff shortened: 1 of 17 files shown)


src/notifications/Notifications.container.js
```diff
@@ -0,0 +1,196 @@
...
+import { Notification, CloseToast, NotificationsMenu as NotificationsMenuPresent } from "./Notifications.present";
+
+
+/**
+ * Notifications object - it's not a React component.
+ *
+ * @param {Object} client - api-client used to query the gateway
+ * @param {Object} model - global model for the ui
+ */
+const Notifications = (props) => {
```
ciyer, line 40: It probably doesn't make a difference semantically, but looking at this code, I expect a React component. Could we make Notifications just a class and the properties methods? I think that would make it easier to follow.

ciyer:
## General

This looks really good! So the idea is that any place where notifications were needed (e.g., in a React component, or in the API layer), a `notifications` object will be passed in, and the methods `add{Info, Success, etc}` could be called to create notifications? I think that will work nicely, and it should be easy to integrate notifications from the server, once we start getting them.

## Extensions 

One thing that I identified that we will probably need is the ability to greater customize display of notifications. The code in the `Notifications.present.js:Notification` component expects a particular format to the notifications. It might be nice to let the code that posts the notification also provide a component for displaying the notification. Doesn't need to be implemented from the start, but I think we should code with the assumption that this extension will be necessary later.

## Naming

It is potentially confusing that the object that is the interface to the notification system is called `Notifications`, and the Rect component that renders notifications is called `Notification`; I think we should come up with more distinct names to avoid confusion.

src/notifications/Notifications.container.js
lorenzo-cavazzi: Good suggestion! It will make this much less confusing when we will look at this again a few months in the future.

lorenzo-cavazzi:
Thanks for the early review!

> So the idea is that any place where notifications were needed [...] a notifications object will be passed in, and the methods add{Info, Success, etc} could be called to create notifications?

That's correct. This would make it easier to add notifications from non-react objects (like the coordinators). It may not be the most elegant way, since this requires to pass the `notifications` object around in the react components, but it's easy to use and flexible.

I agree with all points you mentioned. I have converted the `Notifications` object to a class and adjusted the naming (please feel free to suggest any appropriate name if you think it's still not intuitive).
I will tackle the suggestion in the `Extensions` section while implementing the notifications recap page since it should be immediately useful going in that direction.

ciyer:
One more comment -- we already discussed this, so it's here just as a _aide-mémoire_. If the user is already on the page that provides details about a notification, the notification should be suppressed.

E.g., if the user launches an interactive environment and is on the environments list page, we do not need to show the notification that the environment has started. In that case, we can just mark it as read. The tricky thing about this case is that there are two pages where the user might see this information: the project environments page or the global environments page. Both should be handled.

lorenzo-cavazzi:
I have updated the preview in https://lorenzotest.dev.renku.ch with the requested changes and fixed the interactive environments.
The `TEST - add multiple` button in the dropdown simulate adding notificcations linked to the help and environment pages. Trying to generate them while visiting the `/environments` and `/help` pages result in the skipping the respective toast notification and the +1. They will still appear in the (not yet ready) notifications list page and marked as read.

lorenzo-cavazzi:
Preview updated in https://lorenzotest.dev.renku.ch and ready for the final review.

P.S. the `TEST -` button (and functions) will be removed after the review. The functions have already been integrated in the unit tests.

ciyer (approved):
This this very nice! I think the core here is great for integrating notifications into the UI. Let us develop this a bit further in the notifications branch, and it will soon be ready to merge to master.

PR #1040, opened 2020-09-22
fix: change name ordering with title ordering

### LINKED ISSUE #1039: Can't order dataset by name
## Description

Ordering datasets by name doesn't work

## Reproduce

* Go to the `Dataset` page on dev: https://dev.renku.ch/datasets
* Try to order by `name`

## Context

From @jachro
I noticed that we introduced some regression bug recently when we modified the dataset’s name related fields (you remember, the `name` and `title`). So what happens now is that we cannot order by `name` on the free-text dataset search page anymore as we still use the `name` field while it should be `title`.

lorenzo-cavazzi:
The dataset search uses the property `title` instead of `name`.

Screenshot_20200922_155217

I'm not totally sure if in the UI we want to use the word `title` or if this should stay behind the scene and we keep using `name` (I see `name` in other parts as well)

fix #1039

### DIFF
--- a/src/dataset/list/DatasetList.state.js
+++ b/src/dataset/list/DatasetList.state.js
@@ -26,7 +26,7 @@
 import { Schema, StateKind, StateModel } from "../../model/Model";
 
 const orderByValuesMap = {
-  NAME: "name",
+  NAME: "title",
   DATE_PUBLISHED: "datePublished",
   PROJECTSCOUNT: "projectsCount"
 };
@@ -103,7 +103,7 @@ class DatasetListModel extends StateModel {
     const searchOrder = this.get("orderSearchAsc") === true ? "asc" : "desc";
     switch (this.get("orderBy")) {
       case orderByValuesMap.NAME:
-        return "name:" + searchOrder;
+        return "title:" + searchOrder;
       case orderByValuesMap.DATE_PUBLISHED:
         return "datePublished:" + searchOrder;
       case orderByValuesMap.PROJECTSCOUNT:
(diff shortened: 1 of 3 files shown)


ciyer (commented):
Code looks good, and the change works as advertised. Virginia should say whether we want to use "name" or "title" -- we should pick one and be consistent.

vfried:
The thing here is that a dataset has two properties, name and title... name is the short name of a dataset ("this-is-a-short-name") and the title is like "This is a title". Since the CLI has name as name and title as title I think we should also do that. In the long-standing or for dataset field changes this is better reflected since you can choose name and title when you create a dataset.

vfried:
After having a think about this I think that what happens in the KG is that if a dataset has no title the title of the dataset will be the name this is different in the CLI (i think also in the core service) in this case if a dataset has no title then title is empty/undefined/null (not sure which value), this is an inconsistency between services.

I think if this code works we can merge this, I don't think this inconsistency will affect our users.

src/dataset/list/DatasetList.state.js
```diff
@@ -26,7 +26,7 @@
 import { Schema, StateKind, StateModel } from "../../model/Model";
 
 const orderByValuesMap = {
-  NAME: "name",
+  NAME: "title",
```
vfried, line 29: I would change the name of this variable to TITLE and also in other places where it is used since there is also a name property and they are not the same.

vfried (changes requested):
Besides the variable name I think things are OK, if you change that I will approve. 👍

src/dataset/list/DatasetList.state.js
vfried: I can also do this fix if you are busy with something else...

vfried (approved):
👍

PR #1041, opened 2020-09-22
fix: only merge server options for the default_url (#1025)

### LINKED ISSUE #1025: Incorrect options for interactive environment
## Description

#950 introduced code to show options in the `renku.ini` file, even if they are not available in the list of server-provided options. This is the correct behavior for the `default_url` option, but is wrong for `cpu_request`, `gpu_request`, and `mem_request`. It looks like I can pick 4G of memory here, but the only supported options are actually 1G and 2G.

image

## Reproduce

Try to start an interactive environment on https://dev.renku.ch/projects/cramakri/ROS-Examples/environments/new.

The [renku.ini](https://dev.renku.ch/projects/cramakri/ROS-Examples/files/blob/.renku/renku.ini) specified 4G, but only 1 and 2 G are supported dev.

ciyer:
# Description

#950 introduced a change to prevent project options from disappearing, but this was applied to things like cpu_request and mem_request, which the user cannot actually control. With this change, only default_url can be extended by the project options in the `renku.ini` file, other options will be ignored and show a warning if the project default is not available in the server environment. 

image

The fix is deployed to https://sekhar.dev.renku.ch, to test, try out this project: https://sekhar.dev.renku.ch/projects/cramakri/project-defaults-test/environments/new

Fix #1025.

### DIFF
--- a/src/notebooks/Notebooks.present.js
+++ b/src/notebooks/Notebooks.present.js
@@ -1129,6 +1129,20 @@ class StartNotebookOptionsRunning extends Component {
   }
 }
 
+/**
+ * Combine the globalOptions and projectOptions to cover all valid options.
+ */
+function mergeEnumOptions(globalOptions, projectOptions, key) {
+  let options = globalOptions[key].options;
+  // defaultUrl can extend the existing options, but not the other ones
+  if (key === "defaultUrl"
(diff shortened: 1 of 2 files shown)


lorenzo-cavazzi (approved):
This works well and it's a very needed fix!

PR #1044, opened 2020-09-24
fix: anonymous interactive environments

### LINKED ISSUE #1030: Link to new interactive environment does not work in anonymous mode
## Description

I would like to provide anonymous users a link directly to the new environment launch page, so they can directly get to the feature they want, but this link brings up a login form.

  https://renkulab.io/projects/ROS/ROS-Examples/environments/new

Instead, it should take you the environment launch page. (It works correctly if you first go to https://renkulab.io/projects/ROS/ROS-Examples/environments and click new, but that is one extra step that is unnecessary).

lorenzo-cavazzi:
Fix an API call that mistakenly triggered the login workflow when trying to start an anonymous session. That happened only in specific situations.
This includes a minor workaround to prevent the UI from showing for a moment an error page when performing the "anonymous login".

***How to test it***
The problem happens almost always if you
* Use a chrome-based browser
* Goes directly to the `/environments/new` page

A good test can be done by comparing the 2 following pages (advice: use chrome/chromium in a Private Window)
https://lorenzotest.dev.renku.ch/projects/renku-test/test-2020-09-24-13-27-17/environments/new
https://dev.renku.ch/projects/renku-test/test-2020-09-24-13-27-17/environments/new

fix #1030

### DIFF
--- a/src/api-client/index.js
+++ b/src/api-client/index.js
@@ -115,6 +115,10 @@ class APIClient {
         }
       })
       .then(response => {
+        // This avoids showing errors for a second while doing the anonymous log-in.
+        // It should be solved in a more elegant way once we support interruptable fetch #776
+        if (!response && anonymousLogin)
+          return returnType === RETURN_TYPES.json ? { data: {} } : "";
         switch (returnType) {
           case RETURN_TYPES.json:
             return response.json().then(data => {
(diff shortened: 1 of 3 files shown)


ciyer (approved):
Nice!

PR #1049, opened 2020-09-28
Add button to clone renku project

### LINKED ISSUE #929: Add a button for (renku) cloning a project
Currently it's not too clear what url to use with the `renku clone` command and many users might not even be aware that this is an option that allows them to copy a project locally (e.g. for when they go somewhere with no internet access).

For this reason it'd be nice to have a button in the UI that streamlines this process, similar to the `Clone or Download` button on github
image

This could be next to the "Take me to gitlab" button we already have on the project page.

With content like:

> **Clone with Renku:**
> Install renku locally (link) then run
> `renku clone ...`

I'm not sure if we also want instructions to just clone with `git` or if we want to discourage that

lorenzo-cavazzi:
Adds a button with the commands to clone the project locally, similar to what we find in both the GitLab and GitHub interfaces.

Screenshot_20200928_205620

fix #929

### DIFF
--- a/src/project/Project.present.js
+++ b/src/project/Project.present.js
@@ -1327,7 +1327,7 @@ function RepositoryUrlRow(props) {
     <tr>
       <th scope="row">{props.urlType}</th>
       <td>{props.url}</td>
-      <td><Clipboard clipboardText={props.url} /></td>
+      <td style={{ width: 1 }}><Clipboard clipboardText={props.url} /></td>
     </tr>
   );
 }
@@ -1355,7 +1355,7 @@ function CommandRow(props) {
       <td>
         <code>{props.command}</code>
       </td>
-      <td><Clipboard clipboardText={props.command} /></td>
+      <td style={{ width: 1 }}><Clipboard clipboardText={props.command} /></td>
     </tr>
   );
 }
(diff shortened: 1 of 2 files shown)


ciyer (changes requested):
Nice solution, but could we move it to the settings page, like this:

image


I think we want it out of the way because most (though not all) of our users would use the interactive  environments rather than cloning the project. We will want the screen real-estate where "clone" is for a "launch" button.

vfried:
A small comment regarding the change, if we put this in that tab it should look similar to the ssh and HTTP that are already there...

ciyer:
> A small comment regarding the change, if we put this in that tab it should look similar to the ssh and HTTP that are already there...

Definitely. I was just rushed for time with the mock-up. I will fix it.

lorenzo-cavazzi:
Sorry for the force push, I messed up a bit my local repo.
I removed the button and moved the clone info to the setting page. This is how it looks now

Screenshot_20200930_143814

ciyer (approved):
Very nice!

ciyer (changes requested):
Wait spelling!

ciyer (approved):
Nice!

Issue #880, opened 2020-03-30
Show commit messages in enviroments pages


lorenzo-cavazzi:
## Motivation

Users asked for extra information about commits in the Environments pages.

## Proposal

Show the commit message on mouse hover. Maybe it's not the best interaction since commit is now a link, I would rather show the text when hovering/clicking on a :information_source:  icon.

## Context

Until now, all of the information required for the UI is sent by the `GET /servers` API. In this case, I would rather query the GitLab commit API to get the info only when needed because 1) it may be more bytes than all the other info combined and 2) it's not always visible, chances are that it's not needed in most of the cases

lorenzo-cavazzi:
closed by #1052

PR #1052, opened 2020-09-29
feat: show commit details on-request in environments pages

### LINKED ISSUE #853: Show date/time of commit in interactive-environments page
## Motivation

When collaborating, I sometimes launch multiple environments on the same branch, but I cannot remember which version is which anymore.

## Proposal

Show the date/time next to the commit hash in the interactive environment display. Alternative: show it on mouse-over.

## Context

See image here: which is the newer commit? 

image

lorenzo-cavazzi:
There are a couple of issues requiring to show commit data in the environments list. The proposed solutions were slightly different from the implemented one, but it turns out that other ways would require to implement operations in the backend that are possibly (probably) resource/time intensive for a marginal benefit.

In this PR, commit details are fetched directly through the GitLab API only when required by the user. The implemented interaction requires clicking on an info icon in the "commit" column (mind that the click on the commit-sha already redirects the user).
If the interaction is confusing, we can change it (e.g. remove the icon and change the popover trigger to "on-hover")

Screenshot_20200929_170347

I didn't make a preview because my test namespaces are already used, but this can be run through telepresence (no requirements on other components)

fix #853, #880

### DIFF
--- a/src/api-client/repository.js
+++ b/src/api-client/repository.js
@@ -98,7 +98,7 @@ function addRepositoryMethods(client) {
     );
   };
 
-  client.getRepositoryCommit = (projectId, commitSHA) => {
+  client.getRepositoryCommit = async (projectId, commitSHA) => {
     let headers = client.getBasicHeaders();
     return client.clientFetch(
       `${client.baseUrl}/projects/${projectId}/repository/commits/${commitSHA}`, {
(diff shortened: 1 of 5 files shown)


ciyer (approved):
Great improvement to the notebooks. Makes it much easier to remember where I was when I started the environment.

ciyer:
I couldn't do this on my deployment, but we should test this for anonymous users as well. I'm assuming it will work fine, but if we find something, we can fix it in a later issue.

lorenzo-cavazzi:
I verified that it works also for anonymous users

Screenshot_20201001_093358

PR #1058, opened 2020-10-02
feat: better explain that renku clone is preferred

### LINKED ISSUE #1057: Make choice of cloning method clearer
## Proposal

Right now, both options Git and Renku have equal weight. We should show the Git option in small text and add a note to emphasize that, though it is a possibility, the `renku clone` is the better choice.

## Context

Screenshot_20200930_143814

ciyer:
# Appearance

Here is the new appearance. Initially, only the renku command is shown. Clicking on the text opens the git command.

image

image

Fix #1057

### DIFF
--- a/src/project/Project.present.js
+++ b/src/project/Project.present.js
@@ -1361,28 +1361,62 @@ function CommandRow(props) {
   );
 }
 
+function GitCloneCmd(props) {
+  const [cmdOpen, setCmdOpen] = useState(false);
+  const { externalUrl, projectPath } = props;
+  const gitClone = `git clone ${externalUrl}.git && cd ${projectPath} && git lfs install --local --force`;
+  const gitHooksInstall = "renku githooks install";
+  return (cmdOpen) ?
+    <div style={{ fontSize: "smaller" }} className="mt-3">
+      <p className="font-italic">
+        If the <b>renku</b> command is not available, you can clone a project using Git.
+      </p>
+      <Table size="sm" className="mb-0" borderless={true}>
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi (dismissed):
Definitely more explanatory than before :+1:

src/project/Project.present.js
```diff
@@ -1361,28 +1361,61 @@ function CommandRow(props) {
...
+      <p className="font-italic">
+        If the <b>renku</b> command is not available, you can clone a project using Git.
+      </p>
+      <Table size="sm" className="mb-0" borderless={true}>
+        <tbody>
+          <tr>
+            <th scope="row">Git<sup>*</sup></th>
+            <td>
+              <code>{gitClone}</code>
+              <p className="mt-2">
```
lorenzo-cavazzi, line 1380: This `<p>` adds extra margin in the bottom, so that the second command `renku githooks install` is very distanced (see image).
I would add also the class `mb-0`

Screenshot_20201007_175948

lorenzo-cavazzi (changes requested):
I have a minor comment on the extra margin introduced by the `<p>` element. Feel free to discard it.

There is another detail to adjust: linting is failing on a long line (src/project/Project.present.js @ 1381).

src/project/Project.present.js
ciyer: Good point about the spacing. I have improved this and pushed.

lorenzo-cavazzi (approved):
Very neat solution with the show/hide link!

Issue #1060, opened 2020-10-05
Datasets display "Not in the Knowled Graph" when they are in the Knowledge Graph


vfried:
Some datasets inside projects are in the KG but display "not in the knowledge graph"

Example: 
https://dev.renku.ch/projects/renku-testing/project-9/datasets

PR #1061, opened 2020-10-05
fix: dataset displays kg message correctly

vfried:
Datasets are now displaying the correct message.

They display: 
"in the knowledge graph" if they are in the knowledge graph and the project has the knowledge graph activated
"not in the knowledge graph" in case one of the previous conditions isn't reached

---> This bug was introduced because there was a missing check to see if a project had the kg activated or not, in some cases we where also not fetching this information.

Testing:
--> check datasets for a project that you don't have rights in. I.E: https://dev.renku.ch/projects/virginiafriedrich/more-dataset-testing/datasets this project has the kg activated and the kg finished building so the datasets should display "in the knowledge graph"
--> this is fixed here... https://virginiatest.dev.renku.ch/projects/virginiafriedrich/more-dataset-testing/datasets

### DIFF
--- a/src/project/Project.js
+++ b/src/project/Project.js
@@ -294,6 +294,12 @@ class View extends Component {
 
   }
 
+  isGraphReady() {
+    const webhookStatus = this.projectState.get("webhook");
+    return webhookStatus.status || (webhookStatus.created && webhookStatus.stop)
+      || webhookStatus.progress === GraphIndexingStatus.MAX_VALUE;
+  }
+
   getStarred() {
     const featured = this.props.model.get("projects.featured");
     // return false until data are available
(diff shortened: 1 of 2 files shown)


src/project/Project.js
```diff
@@ -373,7 +373,8 @@ class View extends Component {
     const datasets = this.projectState.get("core.datasets");
     const graphProgress = this.projectState.get("webhook.progress");
     const webhookStatus = this.projectState.get("webhook");
-    const graphStatus = webhookStatus.status || (webhookStatus.created && webhookStatus.stop);
+    const graphStatus = webhookStatus.status || (webhookStatus.created && webhookStatus.stop)
```
ciyer, line 376: Looks good. Could you extract this to a function, since the same code is used in two places.

ciyer (changes requested):
I tested and it works now. I have a refactoring request. Also, could you add a reference to the issue and PR in the commit msg (or were you planning to do a squash merge)?

ciyer (approved):
👍 Remember to add a reference to #1060 and #1061 in the commit msg.

PR #1067, opened 2020-10-12
keycloak library for single sign out

lorenzo-cavazzi:
testing the keycloak library for single sign out


lorenzo-cavazzi:
This can be closed since we chose another solution #1163

PR #1069, opened 2020-10-14
build: update highlight library

lorenzo-cavazzi:
The highlight library we are using is outdated and the support for the R language has a few bugs.
This PR updates the library and adds it to the npm dependencies (previously, it was imported as a requirement for another dependency).

OLD (some comments not highlighted)
Screenshot_20201014_091812

NEW
Screenshot_20201014_090740

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10680,9 +10680,9 @@
       "dev": true
     },
     "highlight.js": {
-      "version": "9.12.0",
-      "resolved": "https://registry.npmjs.org/highlight.js/-/highlight.js-9.12.0.tgz",
-      "integrity": "sha1-5tnb5Xy+/mB1HwKvM2GVhwyQwB4="
+      "version": "10.2.1",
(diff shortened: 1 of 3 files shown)


vfried (approved):
Works much better!!!

There are still issues with:
.txt files https://virginia.dev.renku.ch/projects/lorenzo.cavazzi.tech/many_files_new/files/blob/data/file1.txt

I think we can still update this since it's a library update, I guess we just need to keep an eye open for when they do new releases and improvements.

lorenzo-cavazzi:
This is actually a good point. We weren't handling free text at all.
I added a commit for the files we want to preview without a specific syntax highlight (I couldn't find the docs but I found [this examples in the library's tests](https://github.com/highlightjs/highlight.js/blob/976afa1d0ce0cfcd62dea81ce676ef838f3e1025/test/special/noHighlight.js#L24)).

This is the result for text files

Screenshot_20201016_110301

ciyer (approved):
👍

Issue #1021, opened 2020-08-29
File list should scroll independently of full page


ciyer:
## Description

In the file browser, there is only one scroll bar for the entire page. The file view should have a maximum height and scroll on overflow.

## Reproduce

1. Go to https://dev.renku.ch/projects/cramakri/ros-examples/files
2. Scroll to the file "Dockerfile"
3. Click on Dockerfile
4. Where is it? It is at the top of the screen.

## Solution

We may need to query the browser for the container size and dynamically set max heights for the file listing. A library like this could be useful: https://github.com/react-container-query/react-container-query

PR #1070, opened 2020-10-14
file tree scroll independent from page scroll

lorenzo-cavazzi:
This PR is giving me headaches. The only way I found to address #1021 is to dinamically change the max height of the file nav element. This is hard because:
* Elements get re-created when the user interact with the interface (e.g. the `<FilesTreeView>` element is mounted multiple times, firing `componentDidMount` at every user click on the left file nav bar, even invoking the `constructor` again on some user inteactions).
* The browser size can change.
* When new content is retrieved from APIs, re-drawing the html element takes a non-zero time.

The current solution addresses the first 2 points (with some glitches). Before I try to tackle the 3rd one and to polish the code, I'd like to get early feedback and/or pointer to different solutions. I feel the current solution is getting too complicated.

Here is a preview: https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/many_files/files
The `data` folder contains a lot of files.

P.S. The reference issues suggest to user https://github.com/react-container-query/react-container-query . We are already using a similar library (react-media) but my understanding is that it can help to handle media queries programmatically, while we need something to dinamycally compute HTML elements sizes.


ciyer:
This is a step in the right direction! I wonder if this could be simplified by using two libraries:

- https://github.com/renatorib/react-sizes
- https://github.com/souporserious/react-measure

The first one provides the overall viewport size, the second one gives the sizes of components. Here is a thought -- could we just restrict the height of the file list and the file contents columns to be the same as the viewport height, and scroll on overflow? Then we would only need `react-sizes`.

lorenzo-cavazzi:
Thanks @ciyer for the suggestion! I tried with `react-measure` and I got good results when the file content on the right takes less space (in height) than the file browser, but it's complicated to handle the other way around.
One of the reasons for this is that the components get re-created every time anything on the file tree changes (e.g. clicking a file or a folder), making it very difficult to save any state.

An alternative solution I'm trying now is [react-sticky.](https://www.npmjs.com/package/react-sticky). It allows to make the file tree sticky when scolling down.
I still need to polish the solution, but here is a **preview**: https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/many_files/files/

Since that library is not maintained anymore, I will try a similar alternative (probably [react-stickynode](https://www.npmjs.com/package/react-stickynode))

vfried:
I think it will be great to push this change.

As I said in the standup, it would be good to consider making sticky both the file explorer and the file view so that both work in the same way... 
image

ciyer:
I like the code changes -- they are clear and understandable. The solutions for the most part works well, but there are two things to improve if possible.

1. The header looks weird on overflow, on *Safari*. Did not see this problem in Chrome.

image

2. Selecting a file scrolls the file list back to the top instead of staying at the current location. See 

image

In this, I selected `file73.txt`, but the list scrolled back to to the top.

src/project/filestreeview/FilesTreeView.js
```diff
@@ -105,11 +107,76 @@ class FilesTreeView extends Component {
...
     };
     this.toggle = this.toggle.bind(this);
   }
 
+  componentDidMount() {
+    this.updateTreeMaxHeight();
+    if (!this.state.listener) {
+      const listener = () => this.updateTreeMaxHeight();
+      this.setState({ listener });
+      window.addEventListener("resize", listener);
```
ciyer, line 120: Is it better to implement this as here, using the lower-level functions, or a higher-level library like `react-size` which would manage the listener and re-render the component on a size change?

ciyer (commented):
Code generally looks good, I'm just wondering if it wouldn't be better to use a library like `react-sizes` to handle the window resize events. Also, see my comment about the problem in Safari and the reset of the scroll position when something is selected.

src/project/filestreeview/FilesTreeView.js
lorenzo-cavazzi: Ah yes definitely.
I just realized I didn't push the final version of the code, so the one running the preview is actually (very) different from the one in this PR :facepalm:  I'm not adding listener anymore to `window`.

I will polish the code and push it soon, sorry for the confusion.

lorenzo-cavazzi:
@vfried thanks for the review. Your solution translates into setting a max height in both components and making them scroll independently. That was my first attempt but it turned out to be more difficult than expected because I couldn't compute the appropriate height

@ciyer since I messed up with the code I pushed vs the code i published on the preview, can I ask you if your feedback is valid for the preview here? https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/many_files/files/blob/data/file1.txt

ciyer:
> 
> @ciyer since I messed up with the code I pushed vs the code i published on the preview, can I ask you if your feedback is valid for the preview here? https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/many_files/files/blob/data/file1.txt

I am still having those problems I mentioned (Safari rendering problem, not scrolling to the highlighted file on Safari and Chrome).

lorenzo-cavazzi:
Target issue already closed by #1102

PR #1071, opened 2020-10-14
fix: jsonify structured error messages when fecthing custom templates

### LINKED ISSUE #1056: error on erroneous template URL
Requesting to fetch templates from a wrong URL causes the app to crash. This only seems to happen when the URL is formatted in a specific way, e.g.

```
https://<host>/<repo>.git
```

If instead it is formatted as 

```
https://<host>/<owner>/<repo>.git
```

the request makes it to the backend which then responds with "repository not found"

More details:

Sentry Issue: [RENKU-UI-2M](https://sentry.dev.renku.ch/organizations/swiss-datascience-center/issues/349/?referrer=github_integration)

```
Error: Objects are not valid as a React child (found: object with keys {owner}). If you meant to render a collection of children, use an array instead.
  at immutableUpdate (model/Model.js:149:21)
  at immutableUpdate (model/Model.js:237:22)
  at setObject (model/Model.js:223:10)
  at setObject (model/Model.js:279:20)
  at call (project/new/ProjectNew.state.js:199:16)
...
(45 additional frame(s) were not displayed)
```

lorenzo-cavazzi:
The error message we get when trying to fetch custom templates can be structured. Invoking `JSON.stringify` when that happens is the easiest solution to prevent the UI from crashing.

In the specific case mentioned in the reference issue, I *think* the message is changed in the core service and it becomes more cryptic compared to the original git command error.
If it turns out that users get confused, we could either try to intercept and parse specific errors in the UI or change the message we emit from the core service.

Preview: https://lorenzotest.dev.renku.ch/projects/new

Screenshot_20201014_165457

fix #1056

### DIFF
--- a/src/project/new/ProjectNew.present.js
+++ b/src/project/new/ProjectNew.present.js
@@ -623,12 +623,16 @@ class Create extends Component {
         content = <pre className="text-wrap">{error}</pre>;
       }
       else {
-        Object.keys(error).map(v => {
+        const errors = Object.keys(error).map(v => {
           const text = typeof error[v] == "string" ?
             `${v}: ${error[v]}` :
             `Error message: ${JSON.stringify(error[v])}`;
-          content = (<pre key={v} className="text-wrap">{text}</pre>);
+          return (<pre key={v} className="text-wrap">{text}</pre>);
         });
+        if (errors.length === 1)
+          content = (errors[0]);
+        else
+          content = error[0];
       }
       const fatal = templates.all && templates.all.length ? false : true;
       const description = fatal ?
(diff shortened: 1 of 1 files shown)


vfried (approved):
I will approve but I think we need to change the error message that comes from dev-core... otherwise, we will have to implement something in the UI to "fix" this error messages. But for that we would need to know what can come from the backend and what this means so we can translate it to an error that makes sense. Here it's asking the user to fill an "owner" field that is not displayed in the form for example, this makes no sense at all (in the case of users that are not familiar with git at least).

What i write here about error messages applies also to dastet/migration stuff, i am getting terrible error messages and maybe we could do some generic "error message translator" in the UI if the core doesn't return something better.

lorenzo-cavazzi:
I agree with your comment, what we get in the error response should be standardized to some extent to make it manageable and avoid confusing messages.
I opened this SwissDataScienceCenter/renku-python/issues/1616 on the core service repo.

PR #1081, opened 2020-10-18
Update of the ckeditor and added features to the editor

vfried:
I updated the ckeditor to the latest version.

https://ckeditor.com/blog/

They did several improvements, the markdown features are working MUCH better now and they have also some new features.

- Some text has autoformatting on writing like headings, bold, italics, code (you can type ** in the editor and they will recognize that you are writing with format) --> this doesn't work for image inserts for now, maybe we can do some configuration somewhere for this to work
- They improved the markdown processor that they are using --> before I had to do some fixes in their editor because some things where broken and I can now delete this fixes I did
- Same as on top for the markdown code features (they are still missing some stuff but they are better)
- We can insert images by copy/pasting and they are developing a feature that allows us to put an image from a link with a widget in the top bar --> it would be VERY cool if we could take this and do something so we can navigate the folders from a git project to select a file/image to paste
- I still can't manage to configure well the tables features so you can align content from a column (like the :----:) but i guess we can live with that
- They support todo list for markdown

image

image

I wonder if we should expand this to be used everywhere or if we should wait until all things are suported (images/files...) 
From what i see new issues have this editor but not the comments for the issues ---> we might be able to add a little comment that says that we will resolve links when we display them.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1907,12 +1907,167 @@
         "tslib": "^1.9.0"
       }
     },
+    "@ckeditor/ckeditor5-basic-styles": {
+      "version": "23.0.0",
+      "resolved": "https://registry.npmjs.org/@ckeditor/ckeditor5-basic-styles/-/ckeditor5-basic-styles-23.0.0.tgz",
+      "integrity": "sha512-3JqOfB3vWGlXm6CK5SvWgaubyYp/E4R1L4HvT1jzvwa6+QW9S4aqLK3O75XukfDleFDOodDcXDdRMQpI0OBk5g==",
+      "requires": {
+        "@ckeditor/ckeditor5-core": "^23.0.0",
+        "@ckeditor/ckeditor5-typing": "^23.0.0",
+        "@ckeditor/ckeditor5-ui": "^23.0.0"
+      }
+    },
+    "@ckeditor/ckeditor5-code-block": {
+      "version": "23.0.0",
+      "resolved": "https://registry.npmjs.org/@ckeditor/ckeditor5-code-block/-/ckeditor5-code-block-23.0.0.tgz",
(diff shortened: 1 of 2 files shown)


lorenzo-cavazzi (approved):
I tested a bit and all works fine :+1:
Since the actual code change is in the other repo https://github.com/SwissDataScienceCenter/ckeditor5-build-renku , I would suggest for the next time to open a PR there and review the actual code.
I leave my review here: I have seen `src/markdowneditor.js` contains a mix of single/double tabs and spaces. Could we re-format it to be consistent? (same in `htmleditor.js`). No need to updated the version here since the change doesn't affect the code.

lorenzo-cavazzi:
A general consideration: since the current `ckeditor5-build-renku` release has only a couple of config files in the `/src` folder, I was wondering if we prefer to keep that repo separate from the ui for some specific reason (different licensing? reusability?) or if it would be easier to merge it here to avoid this overhead.

ciyer:
> A general consideration: since the current `ckeditor5-build-renku` release has only a couple of config files in the `/src` folder, I was wondering if we prefer to keep that repo separate from the ui for some specific reason (different licensing? reusability?) or if it would be easier to merge it here to avoid this overhead.

This makes sense to me. Once #1043 is merged, it will be easy to add a subfolder for the ckeditor code.

vfried:
I agree with you, I think I did the CKEditor as a separate repo because in the previous versions there was much more code there and I thought we didn't need to have all that code in our repo but now that it's much smaller I will find a way of putting it inside the renku-ui repo.

I would do this in a separate issue though, I think we should see how other projects have done this since there are also tests inside.

vfried:
Another issue is to see if we can change the css in the editor so that the preview of the markdown text looks closer to how it will look in renku after published.

PR #1088, opened 2020-10-22
refactor: use a switch instead of a checkbox for boolean server options

### LINKED ISSUE #1038: use a switch instead of check box for pulling data
The "automatically fetch LFS data" check box would probably be better as a toggle switch: 

image

Like the one used here: 

image

lorenzo-cavazzi:
fix #1038

Top: newer version; bottom: older
Screenshot_20201022_202215


ciyer (approved):
👍 Works as advertised!

PR #1102, opened 2020-10-30
Stick file tree component and set max height

### LINKED ISSUE #1021: File list should scroll independently of full page

lorenzo-cavazzi:
Here is an alternative implementation of the file tree component with independent scroll from the main page (the other was #1070)

Following the reviews, I tried to

* remove any custom listener
* fix the visualization on Safari
* adjust the scroll

These required a few workarounds because the components in the project page are re-drawn every time the user clicks a file node on the tree. Changing this behavior would require a big re-design.
The only library I ended up using is [react-sticky](https://www.npmjs.com/package/react-sticky). I tried other solutions with [react-stickynode](https://www.npmjs.com/package/react-stickynode), [react-sizes](https://github.com/renatorib/react-sizes) and [react-measure](https://github.com/souporserious/react-measure), but the final result was either more verbose or more unstable.

PREVIEW: https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/many_files/files

P.S.  The current solution has a few workarounds that I don't really like, but I have already spent a lot of time on this and I couldn't find a better solution. Any alternative implementation is very welcome.

fix #1021

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -17323,6 +17323,15 @@
         }
       }
     },
+    "react-sticky": {
+      "version": "6.0.3",
+      "resolved": "https://registry.npmjs.org/react-sticky/-/react-sticky-6.0.3.tgz",
+      "integrity": "sha512-LNH4UJlRatOqo29/VHxDZOf6fwbgfgcHO4mkEFvrie5FuaZCSTGtug5R8NGqJ0kSnX8gHw8qZN37FcvnFBJpTQ==",
+      "requires": {
+        "prop-types": "^15.5.8",
+        "raf": "^3.3.0"
+      }
+    },
     "react-syntax-highlighter": {
       "version": "9.0.1",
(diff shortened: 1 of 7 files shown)


ciyer (approved):
There remain a few minor problems, but this is **much** better than the _status quo_, and the code is very understandable. I say, let's merge, and we can do the remaining improvements later -- no rush.

PR #1103, opened 2020-11-02
fix: dataset redirects well and doesn't wait for the kg

### LINKED ISSUE #1075: No redirection happens after adding dataset to a project
## Reproduce

1. Create a new project
2. Go to a dataset: https://dev.renku.ch/projects/mohammad.alisafaee/lego/datasets/sets/
3. Click Add to a project
4. The dataset gets imported and the dialog shows `Dataset was imported, you will be redirected soon...` with a spinning wheel forever.

## Note

It's OK to not to do any redirect and just show a success message.

## Context

- macOS Catalina
- Safari 14.0

vfried:
The redirect after dataset add was broken... it was trying to redirect a user to /project/datasets/identifier (this was the identifier like c3177e09-0582-494d-8be3-a558ded9ea63) the projects now use the dataset-name. This pr redirects the user to the right place.

On top of that, it doesn't wait for the dataset to be in the KG to redirect since it's not needed anymore. 

To test: try to add a dataset to a project in dev using the "add to project" button
To see the fix: do the same in virginia.dev.renku.ch 

Closes #1075

### DIFF
--- a/client/src/dataset/Dataset.container.js
+++ b/client/src/dataset/Dataset.container.js
@@ -36,7 +36,7 @@ export default function ShowDataset(props) {
 
   useEffect(()=>{
     let unmounted = false;
-    if (props.insideProject && datasetFiles === undefined && dataset.name) {
+    if (props.insideProject && datasetFiles === undefined && dataset && dataset.name) {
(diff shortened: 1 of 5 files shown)


ciyer (approved):
Nice, this works well now. Thanks!

PR #1108, opened 2020-11-04
feat: kg activation and kg in datasets improvements

vfried:
Unified status sign for /projects/datasets

image

In case the general sign is present there is no sign when entering a dataset view /projects/datasets/dataset-name 
(i deleted the big purple sign here)

image

In case the general sign is not present, there is a sign inside the dataset (This can happen if the dataset was recently added or the kg recently activated)

image

This alerts redirect the user to the status page

image

When the users activate the kg ---> here we use the same component as we use inside files (We could think about removing this in a future issue)
image

Testing: virginia.dev.renku.ch

### DIFF
--- a/client/src/dataset/Dataset.container.js
+++ b/client/src/dataset/Dataset.container.js
@@ -105,5 +105,7 @@ export default function ShowDataset(props) {
     projectPathWithNamespace={props.projectPathWithNamespace}
     dataset={dataset}
     fetchError={fetchError}
+    overviewStatusUrl={props.overviewStatusUrl}
+    projectInsideKg={props.projectInsideKg}
   />;
 }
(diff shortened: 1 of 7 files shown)


ciyer (commented):
Very nice change! I tested, and it works well in a variety of situations, and code looks great. 

Just one question before I approve. There is something strange going on with the project `core-integration-test`: It says, `Please wait, KG integration recently triggered` and shows an endless wait bounce. Is it something worth fixing in this PR, or is it unrelated enough that we should fix in another PR?

https://virginia.dev.renku.ch/projects/renku-qa/core-integration-test/overview/status

image

vfried:
I am not sure what is going on actually, I see we get this on the /status call...

image

@lorenzo-cavazzi do you know what that means?

lorenzo-cavazzi:
I think that is expected when the indexing starts and that project may very well be a special case since it's used from renku-python to run integration tests. It's possible it's not indexed at all. We can ping @SwissDataScienceCenter/renku-graph-maintainers for confirmation.

ciyer (approved):
👍

PR #1109, opened 2020-11-04
custom images for interactive environments

### LINKED ISSUE #1105: Support custom registry images when starting a new interactive environment
## Motivation

SwissDataScienceCenter/renku-notebooks#304 will allow users to provide a custom registry image for interactive environments and the `GET servers` endpoint will return more information about the registry image that has been used. We can enhance the UI to support these features.

## Proposal

* [x] add an extra field for `image` before the list notebook options in the new environment page. The value should come from the `renku.ini` file. We don't need to display it when it's not available.
* [ ] adapt the code that checks the image availability. When the custom `image` is specified, the check should be similar to the one for non-owned repositories (i.e. don't check the job but only the image availability in the registry)
* [x] update the server list visualization to include the information about the custom image and the fallback to a default image
* [ ] [bonus] allow the user to manually overwrite the default image in the "start environment" page on selected renkulab environments (i.e. specify if it's possible on the values.yaml file). Not sure if this is useful/desirable at all...

As a follow up, it would be great to allow users to change the ini file settings in the UI (should we have an endpoint in `renku-core` for that?)

## Context

I can add a mockup for the 3rd point. Very likely, the information will be included in the commit's popover whenever relevant. When the environment is started using a fallback image, we may want to display a different icon (e.g. a warning sign) to make explicit to the user that things didn't go as expected.

lorenzo-cavazzi:
Allows users to start interactive environments with a custom image.
A user can check the image used for an environment by clicking on the ready icon. Not sure if this is the best way to show this information (despite it shouldn't be very important). Any alternative suggestions are welcome.

**Preview**: https://lorenzotest.dev.renku.ch

P.S. Trying to start a new environment when an image is not available (both custom image or commit image) will result in getting a `404` response. The UI needs to show proper feedback in that case -- this is still missing.

fix #1105 

Screenshot_20201104_144705
Screenshot_20201109_165748
Screenshot_20201117_183412


lorenzo-cavazzi:
Now it verifies whether a default image has been used for an interactive environment, giving feedback to the user.

It's not deployed yet, but it can be tested using the image built from this PR https://github.com/SwissDataScienceCenter/renku-notebooks/pull/452
--> `lorenzocavazzitech/renku-notebooks:0.8.3-5e1faef`


Screenshot_20201117_183412

lorenzo-cavazzi:
Preview added here: https://lorenzotest.dev.renku.ch/
* This is a sample of project with a custom image: https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/custom-image
* This is a project with an unavailable image. Switch the `other` branch and ignore the Docker image error to start the environment with the default image: https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/faulty-image/environments/new

client/src/notebooks/Notebooks.present.js
```diff
@@ -472,12 +482,16 @@ class NotebookServerRowCompact extends Component {
   }
 }
 
-function getStatusObject(status) {
+function getStatusObject(status, defaulImage) {
```
ciyer, line 485: `defaulImage` -> `defaultIamge`

client/src/notebooks/Notebooks.present.js
```diff
@@ -1272,14 +1323,25 @@ class StartNotebookServerOptions extends Component {
...
+        { verb: "", plural: "s", aux: "are", article: "" } :
+        { verb: "s", plural: "", aux: "is", article: "a " };
+      const wrongVariables = unmatchedWarnings.map((w, i) => (
+        <span key={i}><i>{w}</i>: <code>{projectOptions[w].toString()}</code><br /></span>
+      ));
+
+      globalWarning = (
+        <Warning key="globalWarning">
+          The project configuration for interactive environments
+          contains {language.article}variable{language.plural} that {language.aux} either
```
ciyer, line 1338: 👍

ciyer (changes requested):
This looks very good, and the changes are easy to follow. I would like a different terminology for the "custom" images. I do not think that word is very descriptive. I suggest "pinned." I also think most users do not care which image is being used, so I suggest changing the new  environment form to display the information about the pinned image like this:

image

The "more info/pinned image" link should take the user to documentation about this feature.

@rokroskar what do you think?

lorenzo-cavazzi:
I addressed the 2 points from the review, but I used slightly different words to avoid the disputed `pinned` word. We can adjust that as we prefer before merging. Let me know if the sentence is clear.

Redeploying to https://lorenzotest.dev.renku.ch is taking longer due to the docker pull rate limit. Running through telepresence works when paired with the latest version of renku-notebook currently on `master`.

This is the result:
Screenshot_20201202_170254

client/src/notebooks/Notebooks.present.js
```diff
@@ -939,6 +994,28 @@ class StartNotebookPipelinesContent extends Component {
...
+
+      // this style trick makes it appear as the other Label + Input components
+      const style = { marginTop: -8 };
+      const url = "https://renku.readthedocs.io/en/latest/user/templates.html?highlight=.dockerignore#renku";
+      return (
+        <Fragment>
+          <Input type="input" disabled={true} id="customImage" style={style} value={projectOptions.image}></Input>
+          <FormText>
+            <FontAwesomeIcon className="no-pointer" icon={faInfoCircle} /> This project specifies
+            an <ExternalLink role="text" iconSup={true} iconAfter={true} url={url} title="image in the settings" />. A
```
ciyer, line 1011: Now that https://github.com/SwissDataScienceCenter/renku/pull/1737 has been approved, I think we can go with
```
 This project specifies a <ExternalLink role="text" iconSup={true} iconAfter={true} url={url} title="pinned image" />.
```

ciyer (changes requested):
With the agreement on terminology in https://github.com/SwissDataScienceCenter/renku/pull/1737, I think we can just use "pinned image."

client/src/notebooks/Notebooks.present.js
```diff
@@ -924,12 +975,16 @@ class StartNotebookPipelinesBadge extends Component {
         text = "not available";
       }
     }
+    else if (pipelineType === NotebooksHelper.pipelineTypes.customImage) {
+      color = "primary";
+      text = "custom";
```
ciyer, line 980: This should be changed to `text = "pinned";`

lorenzo-cavazzi:
I changed the texts. I didn't re-deploy the preview. Here is a screenshot showing the relevant part of the "New environment" page.

Screenshot_20201204_150925

ciyer (approved):
Fantastic! 🎉

PR #1115, opened 2020-11-05
Improve title validation on new project creation

### LINKED ISSUE #1107: Improve the title validation
## Description

When creating new projects, we don't validate the title with a proper set of rules (e.g. some chars are not allowed, see below). We should improve this to provide a better user experience and avoid project creation to fail with unclear errors.

The solution would be to
- [x] report an error as soon as the user tries to use an invalid char (REF: to be added)
- [x] in the slug preview, strip any char valid for the title but not for the slug (REF: [GitLab path regex code](https://gitlab.com/gitlab-org/gitlab/blob/master/lib/gitlab/path_regex.rb#L13)).
  Important! This must be coordinated with `renku-core` since we don't pass the slug to the backend but it's computer there with the same rules (REF: to be added).
- [x] when cheking for the project to be unique, verify both the project name and the project full slug

## Reproduce

Try to create a new project containing
- Invalid chars like `$` or `/`
- Chars valid for the name but not for the project slug like `é` or `ü`

## Context

Sentry errors: [RENKU-PYTHON-4B](https://sentry.dev.renku.ch/organizations/swiss-datascience-center/issues/715/?referrer=github_integration)

Error message on GitLab explaining the allowed set of chars
Screenshot_20201104_115401

lorenzo-cavazzi:
The "New Project" page already contains a number of checks to prevent the user from clicking on "Create Project" when the provided data are wrong. Both the title and the project slug were not tested deeply enough and users were getting confusing errors. We are now closer to [the GitLab slugify function](https://github.com/gitlabhq/gitlabhq/blob/7942fe679107b5e73e0b359f000946dbbf2feb35/app/assets/javascripts/lib/utils/text_utility.js#L48-L65), and we test unicode chars in the title. Sadly, we still don't allow emoji... yet :smile:

This PR solves the issue on the UI side, and it should be deployed together with SwissDataScienceCenter/renku-python#1691 that replicates similar functionalities in the renku-core backend APIs.

**Preview**: https://lorenzotest.dev.renku.ch/projects/new
**How to test**: create projects with creative names -- try to insert invalid chars like `/` or `$` as well as other Unicode letters like `ê` or `æ`

fix #1107


client/src/project/new/ProjectNew.state.js
```diff
@@ -450,12 +450,17 @@ class NewProjectCoordinator {
...
       errors["title"] = "Title is missing.";
     else if (reservedNames.includes(input.title))
       errors["title"] = "Reserved title name.";
+    else if (!verifyTitleValidity(input.title))
+      errors["title"] = "Title can contain only letters, digits, '_', '.', '-' or spaces.";
+    else if (input.title && !slugFromTitle(input.title, true))
+      errors["title"] = "Title must contain at least a standard ascii char or a number.";
     else if (projects && projectsPaths.includes(`${input.namespace}/${slugFromTitle(input.title, true)}`))
-      errors["title"] = "Title already in use in current namespace.";
+      errors["title"] = `Similar title (${slugFromTitle(input.title, true)}) ${slugExplanation}`;
```
ciyer, line 463: Change to 

`Title produces a project identifier (${slugFromTitle(input.title, true)}) that is already taken in the selected namespace. Please select a different title or namespace.`

This message is shown when I just enter the slug as a title, in which case no characters were changed, so I don't think we need to mention the transformation.

client/src/project/new/ProjectNew.state.js
```diff
@@ -450,12 +450,17 @@ class NewProjectCoordinator {
       warnings["template"] = "Must get the templates first.";
 
     // check errors: require user intervention. Skip if there is a warning
+    const slugExplanation = "already in use in current namespace. Non ascii chars were converted or stripped out.";
```
ciyer, line 453: Change to `already taken in the selected namespace.`

This message is shown when I just enter the slug as a title, in which case no characters were changed, so I don't think we need to mention that part.

client/src/project/new/ProjectNew.state.js
```diff
@@ -450,12 +450,17 @@ class NewProjectCoordinator {
...
     if (!input.title)
       errors["title"] = "Title is missing.";
     else if (reservedNames.includes(input.title))
       errors["title"] = "Reserved title name.";
+    else if (!verifyTitleValidity(input.title))
+      errors["title"] = "Title can contain only letters, digits, '_', '.', '-' or spaces.";
+    else if (input.title && !slugFromTitle(input.title, true))
+      errors["title"] = "Title must contain at least a standard ascii char or a number.";
```
ciyer, line 461: `Title must contain at least one letter (without any accents) or a number.`

client/src/utils/HelperFunctions.js
```diff
@@ -20,12 +20,36 @@
...
+ */
+function slugFromTitle(title, lower = false, separator = "-") {
+  // ? REF: https://github.com/gitlabhq/gitlabhq/blob/7942fe679107b5e73e0b359f000946dbbf2feb35
+  // ?        /app/assets/javascripts/lib/utils/text_utility.js#L48-L65
+  const rawProjectName = lower ?
+    title.trim().toLowerCase() :
+    title.trim();
+  const slug = rawProjectName
+    .replace(/[^a-zA-Z0-9_.-]+/g, separator) // remove invalid chars
+    .split(separator).filter(Boolean).join(separator); // remove separators duplicates
```
ciyer, line 42: Very clever!

client/src/utils/Utils.test.js
```diff
@@ -320,10 +320,53 @@ describe("Helper functions", () => {
     expect(splittedBranches.autosaved[0].autosave.commit).toEqual(commit);
     expect(splittedBranches.autosaved[0].autosave.finalCommit).toEqual(finalCommit);
   });
+});
 
-  it("function slugFromTitle", () => {
+describe("title related functions", () => {
```
ciyer, line 325: These tests are great, very helpful to understand the behavior.

client/src/utils/Utils.test.js
```diff
@@ -320,10 +320,53 @@ describe("Helper functions", () => {
...
+  });
+  it("function slugFromTitle lowercase with custom separator", () => {
+    expect(slugFromTitle("This is my Project", true, "+")).toEqual("this+is+my+project");
+  });
+
+  // verifyTitleValidity
+  it("function verifyTitleValidity - valid strings", () => {
+    expect(verifyTitleValidity("João-Mario")).toBeTruthy();
+    expect(verifyTitleValidity("здрасти_.и")).toBeTruthy();
+    expect(verifyTitleValidity("")).toBeTruthy();
```
ciyer, line 362: Why is this one a valid title?

ciyer (changes requested):
This is looking good, and the code is great. I have a few requests for changes.

More generally, would it be difficult to just strip accents rather than convert those letters into hyphens? E.g., turn `é` into `e` rather than `-`?

client/src/project/new/ProjectNew.state.js
lorenzo-cavazzi: We substitute all the non `[a-zA-Z]` letters with `-` _and_ we strip the extra `-`. This leads to possibly confusing cases.
E.G: If you have `namespace/test`, there are a number of title combinations that will produce the same slug, like `testü`, `test-ä`, `testé`, `-test-`, ...
I will change the message with what you suggested.

client/src/utils/HelperFunctions.js
lorenzo-cavazzi: Blatantly copied from here :grin:  https://github.com/gitlabhq/gitlabhq/blob/7942fe679107b5e73e0b359f000946dbbf2feb35/app/assets/javascripts/lib/utils/text_utility.js#L48-L65

client/src/utils/Utils.test.js
lorenzo-cavazzi: Good point, the name `verifyTitleValidity` is probably slightly misleading...
It actually checks only that valid chars are used, while other cases are tested separately in the coordinator component to give better feedback to the user (i.e. first letter wrong, title empty, ...).
Should I change the function name to `verifyTitleCharacters`?

lorenzo-cavazzi:
I also forgot to check for the first letter to be a number or a character. Added now.

> More generally, would it be difficult to just strip accents rather than convert those letters into hyphens? E.g., turn é into e rather than -?

It's not particularly difficult, and we could borrow [from this convenient function](https://github.com/gitlabhq/gitlabhq/blob/7942fe679107b5e73e0b359f000946dbbf2feb35/app/assets/javascripts/lib/utils/text_utility.js#L278-L351), but we tried to stick to the same rules GitLab is applying on its new project page.
The backend has to implement the very same logic (we don't pass the project part of the namespace slug as a separate argument to the backend) and adopting GitLab standard was preferred over deciding for our own flavor -- even if what you suggested was the first solution we considered :smile: 

Preview updated in https://lorenzotest.dev.renku.ch/projects/new

client/src/utils/Utils.test.js
ciyer: Good idea -- `verifyTitleCharacters` would make it clearer.

ciyer (changes requested):
I think the behavior is good with the improvements. It is very minor, but while we are here, we might as well make the improvements -- could you rename the function to `verifyTitleCharacters`?

Regarding the conversion of accents. I don't think the GitLab behavior is the best guide, because in GitLab the slug is user editable. I think our solution is better and easier to use, where the slug is just generated from the title, but it need the conversion from title to slug to be good. 

For example, if I wanted to create a project with the title `Caffè`, it would have the slug `caff`, which seems weird. I think the slug `caffe` would be better. In GitLab, I could fix the slug. In RenkuLab, I would need to create the project as `Caffe` and then rename it to `Caffè`. This seems needlessly complicated.

But, it's a minor thing. I think if you agree it is worth fixing, we should do it. If not, I will approve after the function rename.

lorenzo-cavazzi:
Well, your argumentation makes sense.
Plus, we have to deviate anyway from the GitLab conversion pattern because it seems to suggest [invalid patterns]( https://github.com/docker/distribution/blob/master/reference/regexp.go#L6-L20), leading to invalid references format for images in the GitLab registry ([example 1](https://dev.renku.ch/gitlab/lorenzo.cavazzi.tech/example-with-_-in-the-middle/-/jobs/119065), [example 2](https://dev.renku.ch/gitlab/lorenzo.cavazzi.tech/test-oww_.wewe1322/-/jobs/119062)).

The changes are:
* Any valid symbol is converted to `-`
* The slugify function tries to convert Unicode accented chars to ascii with the specified rules.

I'll update the preview as soon as the backend counterpart is ready.

lorenzo-cavazzi:
As discussed offline, the final solution won't include the accented letters conversion.
The function is there, I just invoke `slugFromTitle` without the flag for the conversion.

**Preview updated**: https://lorenzotest.dev.renku.ch/projects/new

P.S: I'll create a follow-up issue for making the fork modal UX similar to the new project creation. At the moment, the feedback to the user is not real-time (the error comes after trying to fork the project) and the wording for errors is slightly different.

ciyer (approved):
Nice fix. Please remember to make sure the issue and PR numbers are in the commit message.

PR #1117, opened 2020-11-09
fix: dataset form is much simpler now but it keeps all fields

### LINKED ISSUE #1116: Simplify Dataset Create / Edit
## Motivation

It is great that all the options for specifying dataset metadata are there, but it is a little overwhelming and unnecessary in many cases.

## Proposal

1. We should only show the Creator, Keywords, and Name fields if explicitly requested.

image
(The screenshot shows "Create", but "Edit" should work the same.)

2. The user should be set as the creator in a new dataset by default

3. Changing just the affiliation of a creator on the edit field does not seem to work.

vfried:
FIXES:
@ciyer 
- The user should be set as the creator in a new dataset by default (done)
- Changing just the affiliation of a creator on the edit field does not seem to work.(works now)

@rokroskar 
- what is "name" in "do you want to specify the name"? --> name is the name of the dataset, there is also the title of the dataset and they are different (we had a 1-hour meeting about this a few months ago, in the past name was short_name then we decided to have title and name)
- I would say "additional dataset creators" since the default creator is already set (there is a non-edit default creator now and you can just add creators)
- if the field for adding creators is shown, make sure it's clear that they are in addition to the existing creator ---> not sure how this can be done, do we mean that the creator is the owner???
- I'm not sure we need to remove keywords from the default display - if they come under the description it's fine but maybe the element can be slightly smaller than what is there currently --> I didn't change the size of the element but i changed the color, it is less prominent now...

RESULT:

image

NAME IS ONLY EDITABLE ON REQUEST...
image

CREATOR ONLY ADDS CREATOR ON REQUEST...
image

⚠️  I noticed here that the user I see has a different email than the one used... this results in something like this:
image
I assume i have two emails linked to the same userId --> this could be fixed if the user field had a field with extra emails but otherwise i can't do anything.

Testing: https://virginia.dev.renku.ch 

Closes #1116


rokroskar:
@vfried thanks for implementing this quickly! It looks really good now, I think the funny behavior I saw last time has all been resolved. 

One final thing from my side: I like that it's possible to change the name but I think it might be confusing to users what we mean by this field... I don't know if it's possible to have like a hover-over help element or something like that that would say "The `name` is used as a short-hand reference for the dataset for use with the renku CLI" or something along those lines. @ciyer what do you think?

vfried:
image


> @vfried thanks for implementing this quickly! It looks really good now, I think the funny behavior I saw last time has all been resolved.
> 
> One final thing from my side: I like that it's possible to change the name but I think it might be confusing to users what we mean by this field... I don't know if it's possible to have like a hover-over help element or something like that that would say "The `name` is used as a short-hand reference for the dataset for use with the renku CLI" or something along those lines. @ciyer what do you think?

image

rokroskar:
Looks good! I can't remember if there are any character limitations (apart from no spaces)?

ciyer (changes requested):
Code is very nice -- I like how it works.

One minor thing: I think the creator should not have a "." at the end of it in the listing

1. I am having some problems with the handling of creator. If I add a creator on dataset creation, it is shown as the creator in listing, overriding the user that created it, even though that user is also a creator.

image

image

image

2. It is not possible to remove creators.

vfried:
@ciyer 
- Problem 1 was fixed (i had done some quick change for the edit case and didn't realize that the case for creating was different) 
- Couldn't replicate problem 2 (i was able to delete a creator)
- the '.' at the end of the default creator was deleted

The reason for problem 1 is that you can create a dataset with someone that is not you as a creator, in case you edit the dataset i think you are added as a creator automatically. If we wanted we could not put a default creator in the CREATE form but i think that 99.9% of the times the person creating the dataset will want to be creator also so maybe we can just leave it as it is.

ciyer (approved):
Works very nicely now! I added a few commits in there, introducing some changes to the wording of messages and a refactoring. Please double check to make sure I didn't break anything. From my side, it is ready to merge.

PR #1123, opened 2020-11-13
add client side notifications

### LINKED ISSUE #1055: User notifications v1
Follow up to #919 

* Remove test notification creation buttons

1. Switch text to "No unread notifications"

image

2. Change the notification listing

From 
image

to

image

3. The `[more info]` should show the description. EDIT: note also the notification time next to the title

time

4. EDIT: When a notification is marked as read, it should go to an "Already read" section with the icon grayed out. That should serve as a user action log for the convenience of the user. Don't show the notification anymore in the dropdown menu

grayed2

lorenzo-cavazzi:
This PR addresses #1055 and adds client-side notifications. The elements introduced in the UI are:
* toast notifications appearing in the top right when a notification is added (not for `info` notifications)
* dropdown menu on the top right for the unread notifications (not for `info` notifications)
* notifications page with a full recap (all notifications, also when already marked as read)

Currently, notifications are produced only when starting interactive environments. In the preview, it's possible to generate fake notifications from the notifications menu. That will be removed before merging the PR

**Preview**: https://lorenzotest.dev.renku.ch

fix #1055


client/src/notebooks/Notebooks.container.js
```diff
@@ -335,9 +338,27 @@ class StartNotebookServer extends Component {
   internalStartServer() {
     // The core internal logic extracted here for re-use
     const { location, history } = this.props;
-    return this.coordinator.startServer().then(() => {
+    return this.coordinator.startServer().then((data) => {
       this.setState({ "starting": false });
-      if (history && location && location.state && location.state.successUrl)
+      const resources = Object.keys(data.resources).map(res => `${res}: ${data.resources[res]}`);
```
ciyer, line 343: Could this part be extracted into a function? Something like `postNotification(data)`? Having it inline here makes it hard to separate the code that does the work from the code that informs the user about what was done.

client/src/notifications/Notifications.present.js
```diff
@@ -0,0 +1,396 @@
...
+        <DropdownItem onClick={() => this.props.handlers.markAllRead()}>Mark all read</DropdownItem>
+        <DropdownItem divider />
+        <div className="notification-list-container">{renderedNotifications}</div>
+      </Fragment>) :
+      null;
+
+    return (
+      <Fragment>
+        <NotificationsMenuTesting {...this.props} /> {/* // ! TEST- Remove  */}
+        <Link to="/notifications"><DropdownItem>Notification page</DropdownItem></Link>
```
ciyer, line 206: I think this should simply be called `Notifications`.

client/src/notifications/Notifications.present.js
```diff
@@ -0,0 +1,396 @@
...
+          .sort((a, b) => b.timestamp - a.timestamp)
+          .map(notification => (
+            <Row key={notification.id} className="notification-page-item">
+              <NotificationPageItem notification={notification} read={false}
+                markRead={() => this.props.handlers.markRead(notification.id)} />
+            </Row>
+          ));
+        newSection = (
+          <Fragment>
+            <h1>New notifications</h1>
```
ciyer, line 267: I think this section should be called `Unread`.

ciyer (changes requested):
Looking forward to having this feature integrated! The notifications page looks good.

I have one request for a refactoring (extract a function) and some rewordings. I also have a couple of requests for different behavior.

1. Closing a toast should not mark the notification as read.

image

I have been closing the notifications precisely when I do not have time to read them / do not want them in my way. 

2. Could we make some changes to the way the notifications page is shown?

image

* Move the "Mark Read" button to the other left side and make it just a check mark, without a box
* Make the links buttons with text to describe what you see if you click on the link. Keep an indication if the link takes you to an external site.

client/src/notebooks/Notebooks.container.js
lorenzo-cavazzi: I think the whole logic in that function is about crafting the notification (apart from the redirect, that I was tempted to remove), while the logic to start the server is in the `startServer` function in the coordinator.
If you prefer, I can move lines `343-358` to a separate function, but that won't be reusable and `internalStartServer` would only contain `this.coordinator.startServer()`.
Would it work better if I add a couple of comments instead?

ciyer: Ok. Comments should be enough then.

lorenzo-cavazzi: Done. If you believe it's still confusing, we can try with the separate function.

lorenzo-cavazzi:
1. Behavior changed as requested. Closing the toast notifications doesn't silence it. Clicking on the notification URL (if available) still does it.
2. Visualization changed. I chose to keep the icon on the link button for both internal and external links to keep it coherent with the toast and the dropdown notifications. Let me know if this is ok.
3. In-line changes done.

The preview is up-to-date: https://lorenzotest.dev.renku.ch

### Further changes
The link button seems too prominent to me. There are a couple of changes we may want to apply:
- Invert colors (as in the second link in the screen below)
- Shrink the button (as in the first link). Not sure this is necessary if we invert the colors.

Let me know your feedback and feel free to suggest further changes/adjustments.

Screenshot_20201116_165633

client/src/notebooks/Notebooks.container.js
```diff
@@ -335,9 +338,28 @@ class StartNotebookServer extends Component {
...
       this.setState({ "starting": false });
-      if (history && location && location.state && location.state.successUrl)
+
+      // craft and add the notification
+      const projectData = Object.keys(data.annotations)
+        .filter(elem => elem.startsWith(ExpectedAnnotations.domain))
+        .map(elem => `${elem.substring(ExpectedAnnotations.domain.length + 1)}: "${data.annotations[elem]}"`);
+      const resources = Object.keys(data.resources).map(res => `${res}: "${data.resources[res]}"`);
+      const fullDescription = `Interactive environment will start soon.
+      Reference URL: "${data.url}", ${projectData.join(", ")}, ${resources.join(", ")}.`;
```
ciyer, line 350: Most of the project data isn't very useful.

image

Could we just say `An interactive environment for the project [project link] will be available soon.`

(Or, if the change below is made, then `An interactive environment for the project [project link] is available.`)

client/src/notebooks/Notebooks.container.js
```diff
@@ -335,9 +338,28 @@ class StartNotebookServer extends Component {
...
+      this.notifications.addSuccess(
```
ciyer, line 351: Could we add the notification at a later point. Instead of when the environment is starting, I would prefer to have it once the environment has started. But if this is complicated, it isn't absolutely necessary.

ciyer (changes requested):
We are in the final stretch for this PR! Looking very good, we just need to remove the items from the notification menu that are there for testing.

I have some requests for the wording and timing of the interactive env launch notifications.

client/src/notebooks/Notebooks.container.js
lorenzo-cavazzi: That would be a very good improvement. This is a bit difficult now due to how we poll the resources, but as soon as we move the notebook polling logic to the ui-server it should be much easier. I would postpone it to that point.

ciyer:
Sorry, I overlooked this part.

> ### Further changes
> The link button seems too prominent to me. There are a couple of changes we may want to apply:
> 
> * Invert colors (as in the second link in the screen below)
> * Shrink the button (as in the first link). Not sure this is necessary if we invert the colors.

I think we should leave the colors since these are the colors for primary buttons, but making the button smaller is a good idea.

client/src/notebooks/Notebooks.container.js
lorenzo-cavazzi: I've changed the text. The short message is similar to what you suggested, but the full name is used instead of the url. In the extra information, all the content is gone except the branch and the commit.

Screenshot_20201117_151719

lorenzo-cavazzi: I've changed the text. The short message is similar to what you suggested, but the full name is used instead of the url. In the extra information, all the content is gone except the branch and the commit.

Screenshot_20201117_151719

lorenzo-cavazzi:
I've addressed the comments and removed the buttons to generate fake notifications.
https://lorenzotest.dev.renku.ch

ciyer (approved):
Very cool! I love being able to request a new environment and then go do other things and get notified when it's (almost) ready.🎉

client/src/notifications/Notifications.present.js
```diff
@@ -0,0 +1,396 @@
...
+      return (
+        <div key={notification.id} className={className}>
+          <NotificationDropdownItem notification={notification} markRead={() => markRead()} />
+        </div>
+      );
+    });
+
+    const content = renderedNotifications.length ?
+      (<Fragment>
+        <DropdownItem onClick={() => this.props.handlers.markAllRead()}>Mark all read</DropdownItem>
```
ciyer, line 205: Actually, just using it again and I have one more request. Could we remove this menu item? It's easy to click by accident, and I don't think it is a big deal to go to the notifications page and mark all read there.

ciyer (changes requested):

client/src/notifications/Notifications.present.js
lorenzo-cavazzi: I think this will still be useful once more notifications are generated. I can image a page left open in the browser that possibly generates a few errors and it may be tedious for the user to remove each one.
Anyway, I can remove this from the dropdown menu and we can re-introduce it once (and if) we need it.

ciyer (approved):
🎉

PR #1126, opened 2020-11-16
Replace deprecated set-env commands in the GitHub actions

lorenzo-cavazzi:
Reference: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

~P.S. the second commit is just for testing, I'll remove it as soon as the chart has been published correctly.~
UPDATE: second commit removed, it worked :tada: 
Chart pushed here: https://swissdatasciencecenter.github.io/helm-charts/renku-ui-0.11.2-9af026e.tgz
Chart tested here: https://lorenzo.dev.renku.ch/ (not that this test was very useful since the env variables are used only to push the  chart :grin: )

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,7 @@ jobs:
   test-chart:
     needs: test
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


ciyer (approved):
Great, thanks!

Issue #1083, opened 2020-10-21
Download Files


ciyer:
## Motivation

As a user, I would like to be able to download files from the File browser (without having to go to GitLab). This should be for files in regular git or git-lfs.

## Proposal

Add a download button to the file display:

![image](https://user-images.githubusercontent.com/1196411/96723388-caa54c80-13ae-11eb-9e73-984d2346b018.png)

![image](https://user-images.githubusercontent.com/1196411/96696897-9704fb00-138b-11eb-8fb0-2e2d52610f98.png)

And, on hover in the file view, give a download icon:

![image](https://user-images.githubusercontent.com/1196411/98927001-48550780-24d8-11eb-9532-e9d9c1a1e443.png)

ciyer:
Part 1 is fixed in #1133.

PR #1133, opened 2020-11-18
feat: added download button to top bar and file size

### LINKED ISSUE #974: display file size in renkulab file browser
## Motivation

When users are starting an interactive environment they have to decide whether to include all LFS files or none using the check box. But unless they go to gitlab (or already know how much data is in the project because they recently added the data and it's their own project), they don't know how much data will be pulled into the session.

## Proposal

include the total LFS file size vs the project file size, and the file sizes on the file browser for each file.

## Context

This could also make it easier for us to point out when someone accidentally added large data to their project.

vfried:
Download button added to the nav bar of the file...

image

I also did some cosmetic changes and added the file size
This is how it looked before...
image

Closes #1133 
Closes #974 
Addresses the first part of #1083

### DIFF
--- a/client/src/file/File.container.js
+++ b/client/src/file/File.container.js
@@ -311,7 +311,17 @@ class ShowFile extends React.Component {
     let buttonJupyter = null;
     if (this.props.filePath.endsWith(".ipynb"))
       buttonJupyter = (<JupyterButton {...this.props} file={filePath} />);
+    
+    let fileSize = this.state.file ? this.state.file.size : undefined;
 
+    // If the file is LFS this means that to get the real file size we need to read
+    // the file string we get with the LFS info
+    if(this.props.hashElement && this.props.hashElement.isLfs && this.state.file){
+      const splitFile = atob(this.state.file.content).split("size ")
+      if(splitFile.length === 2)
+        fileSize = splitFile[splitFile.length-1]
+    }
+      
     return <ShowFilePresent externalUrl={this.props.externalUrl}
       filePath={filePath}
       gitLabFilePath={gitLabFilePath}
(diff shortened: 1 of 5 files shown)


ciyer (approved):
Looks great, works great, even for LFS files! The file size indication is very helpful. Glad it was so easy to do. Tested on public and private projects, works everywhere I tried.

vfried:
Thanks for reviewing, I don't absolutely love the download icon but I thought it was the best choice. There is also a file download button in font awesome but it doesn't look as good.

Currently waiting for approval of the product owner and then I will merge this.

vfried:
@ciyer thanks for fixing the linting, it should be fixed now on my vscode

rokroskar:
This is going to be loved by many :) 

One small comment: is it possible to make the file nav bar look the same for the lineage view and the file view? I don't see a reason why I shouldn't have the option to download the file (or see its size) if I'm looking at its lineage...

vfried:
> I don't see a reason why I shouldn't have the option to download the file (or see its size) if I'm looking at its lineage...
I thought maybe the users might think that they could download the lineage... the file size could be less confusisng i guess

vfried:
Regarding the file size.. for git lfs files we will have to fetch the file and read it before we can display the size, since we don't get it from gitlab... not sure if we want to do that or not.

vfried:
Regarding file size inside lineage we can:
1- display file size for files that are not lfs
2- display file sizes for all files, this imples fetching files that are lfs to read the file size
3- leave it as it is

Regarding the download button, shall i add it? ---> i need a confirmed desition

vfried:
With file size and download button for graph as well...

rokroskar (approved):
Looks great now, thanks!

ciyer (approved):
🎉

PR #1134, opened 2020-11-18
improvement: mark commits with autodaved content when starting a new environment

### LINKED ISSUE #1113: Better feedback to the users for autosaved content
## Motivation

Users should know in advance when there is autosaved content for a specific branch/commit.

## Proposal

The UI should already have all the needed information to give proper feedback to the user.
I suggest adding a `*` next to the short SHA on the commit dropdown menu item for which we find autosaved content.
Once selected, we can show a `Label` explicitly saying that there is autosaved content, possibly with a link to the documentation.

It was also suggested to show the commnds to reset the changes (`git reset --hard <target-commit>`?).
We can add it in 2 places:
* on the info popup showing before we the environment starts
* on the documentation we link.

## Context

Screenshot_20201105_134027

lorenzo-cavazzi:
Implement the solution suggested in #1113 to make it clear to the users which commits have autosaved content.

***Preview:*** https://lorenzotest.dev.renku.ch/
***How to test:*** Start an interactive environment on any project, modify a file and stop the environment from the environments page in the UI. Try to start a new envireonment and verify that the relevant commit is marked with a `*`. Select it to get a short explanation.

Screenshot_20201118_152646

Screenshot_20201118_152707

fix #1113

### DIFF
--- a/client/src/notebooks/Notebooks.present.js
+++ b/client/src/notebooks/Notebooks.present.js
@@ -1055,7 +1055,7 @@ class StartNotebookPipelinesContent extends Component {
 
 class StartNotebookCommits extends Component {
   render() {
-    const { commits, fetching } = this.props.data;
+    const { commits, fetching, autosaved } = this.props.data;
     if (fetching)
       return (<Label>Updating commits... <Loader size="14" inline="true" /></Label>);
 
@@ -1063,11 +1063,33 @@ class StartNotebookCommits extends Component {
     const filteredCommits = displayedCommits && displayedCommits > 0 ?
       commits.slice(0, displayedCommits) :
       commits;
+    const autosavedCommits = autosaved.map(autosaveObject => autosaveObject.autosave.commit);
     const commitOptions = filteredCommits.map((commit) => {
-      return <option key={commit.id} value={commit.id}>
-        {commit.short_id} - {commit.author_name} - {Time.toIsoTimezoneString(commit.committed_date)}
-      </option>;
+      const star = autosavedCommits.includes(commit.id.substr(0, 7)) ?
+        "*" :
+        "";
+      return (
+        <option key={commit.id} value={commit.id}>
+          {commit.short_id}{star} - {commit.author_name} - {Time.toIsoTimezoneString(commit.committed_date)}
+        </option>
+      );
     });
+    let commitComment = null;
+    if (this.props.filters.commit.id) {
+      const autosaveExists = autosavedCommits.includes(this.props.filters.commit.id.substr(0, 7)) ?
+        true :
+        false;
+      if (autosaveExists) {
+        const url = "https://renku.readthedocs.io/en/latest/user/interactive_stopping_and_saving.html" +
+          "#autosave-in-interactive-environments";
+        commitComment = (
+          <FormText>
+            <FontAwesomeIcon className="no-pointer" icon={faInfoCircle} /> For this commit we
+            found <ExternalLink url={url} iconSup={true} iconAfter={true} title="autosaved content" role="link" />
+          </FormText>
+        );
+      }
+    }
     return (
       <FormGroup>
         <Label>
@@ -1081,6 +1103,7 @@ class StartNotebookCommits extends Component {
           <option disabled hidden></option>
           {commitOptions}
         </Input>
+        {commitComment}
       </FormGroup>
     );
   }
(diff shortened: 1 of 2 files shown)


client/src/notebooks/Notebooks.present.js
```diff
@@ -1055,19 +1055,41 @@ class StartNotebookPipelinesContent extends Component {
...
+      const autosaveExists = autosavedCommits.includes(this.props.filters.commit.id.substr(0, 7)) ?
+        true :
+        false;
+      if (autosaveExists) {
+        const url = "https://renku.readthedocs.io/en/latest/user/interactive_stopping_and_saving.html" +
+          "#autosave-in-interactive-environments";
+        commitComment = (
+          <FormText>
+            <FontAwesomeIcon className="no-pointer" icon={faInfoCircle} /> For this commit we
+            found <ExternalLink url={url} iconSup={true} iconAfter={true} title="autosaved content" role="link" />
```
ciyer, line 1088: Suggested wording:
`We found <ExternalLink url={url} iconSup={true} iconAfter={true} title="unsaved work" role="link" /> for this commit.`

client/src/notebooks/Notebooks.present.js
```diff
@@ -1433,6 +1457,10 @@ class AutosavedDataModal extends Component {
             Renku has recovered {autosavedLink} for the <i>{this.props.filters.branch.name}</i> branch.
             We will automatically restore this content so you do not lose any work.
           </p>
+          <p>
+            If you don&apos;t need it, you can restore the using the following command:
```
ciyer, line 1461: Text: `If you do not need it, you can discard this work with the following command:`

client/src/notebooks/Notebooks.present.js
```diff
@@ -1433,6 +1457,10 @@ class AutosavedDataModal extends Component {
...
+            <br /><code>git reset --hard {this.props.filters.commit.short_id} && git clean -f -d</code>
```
ciyer, line 1462: Could we wrap the command in a `<Clipboard>[...]</Clipboard>` to make it easy to copy?

client/src/notebooks/Notebooks.present.js
```diff
@@ -1415,14 +1438,15 @@ class AutosavedDataModal extends Component {
...
-    const docsLink = (<ExternalLink
-      role="text"
-      url="https://renku.readthedocs.io/en/latest/user/autosave.html"
-      title="documentation page" />);
+    const autosavedLink = (
+      <ExternalLink role="text" iconSup={true} iconAfter={true} url={url} title="unsaved work" />
+    );
+    const docsLink = (
+      <ExternalLink
+        role="text" iconSup={true} iconAfter={true} title="documentation page"
```
ciyer, line 1446: Title can be just `documentation` (instead of "documentation page").

ciyer (changes requested):
This is nice. Even though the '*' does not have any documentation, I think users will quickly understand what it means (they will pick a commit with a '*' to see why that is there, then they will see the explanation that there is unsaved work here).

I have some suggested changes to the wording and a request to make it easier to copy the text. Could we also change the title of the pop-up to: `Unsaved work`?

image

BTW, is it possible to indicate on the environment listing that this commit was started from an autosave branch? And could we show the command there as well?

lorenzo-cavazzi:
I addressed all the comments apart from this:

> BTW, is it possible to indicate on the environment listing that this commit was started from an autosave branch? And could we show the command there as well?

As discussed offline, that's trickier because we don't get that information from renku-notebooks. We determine the availability of autosaved content based on the branch names, and we get the branches directly from GitLab. Since the branch is deleted as soon as the autosaved content is restored, we don't have that information anymore. We could try to save that in the UI but a simple UI refresh would permanently remove it.
Also, the user can get rid of the autosaved content in the JupyterLab environment and the UI would still mention it. This may be confusing.

Preview updated here: https://lorenzotest.dev.renku.ch/

Screenshot_20201120_121722

ciyer (approved):
Very nice!

rokroskar (approved):
This works as advertised! The only thing I'm wondering is whether there should be some indication that _some_ autosave branches exist at the top level of the interactive environment page - otherwise the user doesn't really know that there is some saved state until they look at the commit listing. We can discuss that separately though.

lorenzo-cavazzi:
That's a good point. On one hand, it may be useful information, but on the other hand, it may not be relevant to the user to know that there is autosaved content _at some previous commit_. If there are more recent commits, the old autosaved content is very probably outdated.

The one reason why this consideration may not apply is that we don't associate the autosaved content to the latest commit synchronized with the remote. We can expect users to keep working on environments started from outdated commits, ending up with an autosave for an old commit.

We can continue the discussion (and possibly add a mockup) here: #1143

rokroskar:
> The one reason why this consideration may not apply is that we don't associate the autosaved content to the latest commit synchronized with the remote.

yep, exactly. This is the sort of situation I got into when I was testing this PR and I wasn't sure at all how it should be resolved.

PR #1135, opened 2020-11-18
feat: improvement in file and lineage view bar

### LINKED ISSUE #864: File view improvements
## Motivation

Working with the file view has made a few improvements clearer.

### Different modes as radio buttons

The current way of presenting the file-view modes is confusing, since you do not see which mode you are viewing. 

image

This should be turned into a [radio button](https://getbootstrap.com/docs/4.4/components/buttons/#checkbox-and-radio-buttons).

The same for the menu above the file listing

image

### Collapse file sidebar

I would like some kinds of files to take up the full width available. For this, we should support collapsing the file sidebar.

image

vfried:
File/Lineage view bar looks like this now...
image

It's blocked by an issue that needs to be confirmed to be merged since some changes in that issue also affect the file view.
I might need to do final changes after we merge that issue into this one.

## extra comment below

Closes #864

### DIFF
--- a/client/src/file/File.container.js
+++ b/client/src/file/File.container.js
@@ -337,6 +337,7 @@ class ShowFile extends React.Component {
       projectPathWithNamespace={this.props.projectPathWithNamespace}
       hashElement={this.props.hashElement}
       fileSize={fileSize}
+      history={this.props.history}
     />;
   }
 }
--- a/client/src/file/File.present.js
+++ b/client/src/file/File.present.js
@@ -23,14 +23,14 @@ import { Badge, Card, CardHeader, CardBody, CustomInput } from "reactstrap";
 import { Button, ButtonGroup } from "reactstrap";
 import { ListGroup, ListGroupItem } from "reactstrap";
 import "../../node_modules/highlight.js/styles/atom-one-light.css";
-import { faProjectDiagram, faDownload } from "@fortawesome/free-solid-svg-icons";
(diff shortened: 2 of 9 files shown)


vfried:
Just as a comment, i did my best to align the elements inside the top bar but it wasn't so easy.

---> this could become icons if we want...
image

--> we could put the icons in a button group thing if we wanted...
image

--> we could put the copy button along with the other buttons if we wanted
image

After we can merge #1133 we will have a better feeling of how this looks.

Something that can be done is to make the height of this two things equal but not sure it will look great.

right now...
image

image

ciyer (commented):
🎉 I think the interaction is clearer and easier to follow now. I changed the text for the switch to be contents/lineage (clearer, I think), made the button group smaller, and switch to icons in the card header.

image


I think this is ready to go, I would like to retest after #1133 gets merged into this branch.

vfried:
With file size and download button for graph as well...

vfried:
> Found a minor rendering bug. When there are a large number of files and the file list scrolls, the contents/lineage switch gets clipped. Can you see if you can find a fix? If you can't find something in 1h, let us create a follow-up, but if you can fix the problem, that would be nice.
> 
> image

I was able to replicate the issue you point in safari, I just managed to push a fix for it.

ciyer (approved):
Great! Makes the file view easier to use and improves the look as well.

PR #1137, opened 2020-11-18
improvement: add preview for C++ and Fortran files

lorenzo-cavazzi:


ciyer (changes requested):
This isn't working for me, and I cannot understand why! Could you take a look?

image

And something very strange happens with the Fortran90 file 

https://sekhar.dev.renku.ch/projects/cramakri/file-type-test/files/blob/src/Fortran90.f90

For some reason, it is interpreted as a folder.

lorenzo-cavazzi:
Ah my bad, I added an extra `.`
It should be `"h"` and not `".h"`

lorenzo-cavazzi:
Ok it's fixed now.
@cchoirat you mentioned also the `org` extension, right? We can include that here, but I'm not familiar with that extension. Is that for code files or for something else (e.g. notebook or markdown files)?

ciyer (changes requested):
Ah, that was it! Now it works for most files, but something strange is still happening with the Fortran90 file.

It looks fine in GitLab: https://dev.renku.ch/gitlab/cramakri/file-type-test/-/blob/master/src/Fortran90.f90

But in RenkuLab, viewing the file causes a crash: [renkulab_deployment]/projects/cramakri/file-type-test/files/blob/src/Fortran90.f90

I looked into it quickly, but do not understand what is happening. Could you have a more detailed look?

lorenzo-cavazzi:
I spent a bit of time checking the reason, then I remembered we already had a similar problem with URLs ending with numbers (probably when we used numeric ids for the projects).
We even left a comment here https://github.com/SwissDataScienceCenter/renku-ui/blob/e766ba1ce3e48757e01fb46fbe4ea944fd4ae825/client/src/App.js#L75

This is `react-router-dom`'s fault.

I tried a couple of solutions on the fly to remove the problem. The easiest would be fixing the API error that triggers the UI error feedback, but that's a nasty workaround because the UI would actually reload completely.
I can try to dig deeper and look for a better solution

ciyer (approved):
👍

lorenzo-cavazzi:
I found where the error is generated. The UI still reloads when hitting one of those files ending with numbers, but at least it won't be broken.

ciyer (approved):
Nice fix!

PR #1139, opened 2020-11-19
fix notifications and change notebook notifications

lorenzo-cavazzi:
This includes 2 changes
1. fix: use the current location for notifications

2. refactor: change environments notifications
   * Use error notifications when starting an environment fails
   * Prevent successful notification from generating a toast when not needed (i.e. the user is on a environment page relevant for that project)
   * Change link on the successful notification from the JupiterHub url to the project environments page

***Preview***: https://lorenzotest2.dev.renku.ch

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -61,7 +61,8 @@ class App extends Component {
     const blockAnonymous = !user.logged && !this.props.params["ANONYMOUS_SESSIONS"];
 
     // setup notification system
-    const notifications = new NotificationsManager(this.props.model, this.props.client, this.props.location);
+    const getLocation = () => this.props.location;
+    const notifications = new NotificationsManager(this.props.model, this.props.client, getLocation);
 
     return (
       <Fragment>
--- a/client/src/model/RenkuModels.js
+++ b/client/src/model/RenkuModels.js
@@ -507,9 +507,14 @@ const statuspageSchema = new Schema({
 const notificationsSchema = new Schema({
   unread: { [Prop.INITIAL]: 0, [Prop.MANDATORY]: true },
   all: { [Prop.INITIAL]: [], [Prop.MANDATORY]: true },
+  dropdown: {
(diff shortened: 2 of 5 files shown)


lorenzo-cavazzi:
After some discussions, it seems that there is no agreement on the utility of notifications in its current state.
Hiding it from the final user seems to be a compromise between not exposing the feature yet and leaving the infrastructure necessary to gradually add notifications during the development of other features (especially feedback to the user in case of unexpected failures).

lorenzo-cavazzi:
Notifications have been hidden.
Preview updated: https://lorenzotest2.dev.renku.ch/

rokroskar:
Does this mean we have to undo the acceptance test change?

lorenzo-cavazzi:
No, it's not necessary.
The fix there improved the specificity of the button selection, and that is not affected by the removal of visible notifications.

ciyer (approved):
I tried the version where notifications were not toasted if the user was on a page where the information would be shown anyway. I think that works very well, actually. You don't get bothered if you would see the info, but if you wouldn't you see a notifications.

In any case, I think it is also fine to hid this feature and reenable it for a use case where it is more necessary.

rokroskar:
@ciyer you are technically correct - you don't get a notification if you are on the environments page. But then you also just don't get any notification at all once the environment _is_ running. I.e. the notification only happens when the environment launch is initiated, which means that in the case that I wait on the env page to see the yellow spinny wheel and then decide to go somewhere else, I'll _still_ have to go look back myself because the notification that the environment is now ready isn't there. To get the notification pop up when my environment is running (useful and new information!) we would need to either a) poll everywhere all the time or b) have a backend.

lorenzo-cavazzi:
Should we try to wrap up this PR and move further discussions about specific notifications to another issue (or maybe a meeting)?

rokroskar:
Sounds good to me.

PR #1144, opened 2020-11-24
ci: add spellcheck to tests

lorenzo-cavazzi:
Add spellcheck capabilities to the linter and enforce them during the CI test.
I tried to keep the list of non-standard words to the bare minimum and change the words we used wherever possible. 

The first 2 commits add rules for strings and text, which is the PR's main target.
The following 2 commits add rules for identifiers. That required limited additional effort, and I believe it's nice to enforce clear variable names.


ciyer (approved):
Wow, this must have been tedious to prepare, but it's great that we have it now!

If we have a word that we want to add to the dictionary, we should add it to `.eslintrc.json` in the `"skipWords"` section, right?

lorenzo-cavazzi:
> Wow, this must have been tedious to prepare, but it's great that we have it now!
> 
> If we have a word that we want to add to the dictionary, we should add it to `.eslintrc.json` in the `"skipWords"` section, right?

Yes, that's correct.
We can use `skipWordIfMatch` if we need to match with regex (e.g. URL or hex identifiers).

PR #1145, opened 2020-11-25
feat: progress bar for file upload

### LINKED ISSUE #972: data upload loading bar instead of spinning wheel
## Motivation

(1) For larger datasets, it's hard to tell in the upload whether it's stuck or just taking a long time. The download of datasets in the renku python `renku dataset add` and `renku dataset import` have a loading bar.

(2) if you click off the tab, does the data stop uploading?

## Proposal

(1) Change the spinning wheel to a loading bar for dataset upload.
(2) have a pop up about leaving the tab asking if you really want to stop uploading.

## Context

Working with a user to upload a 1GB dataset; we clicked to the "import" tab while it was still uploading and when we clicked back to "create" everything was gone. The spinning wheel also didn't give us information about how much longer we would have to wait.

vfried:
Added progress bar for file upload...

image

Closes #972

### DIFF
--- a/client/src/api-client/dataset.js
+++ b/client/src/api-client/dataset.js
@@ -11,33 +11,37 @@ export default function addDatasetMethods(client) {
     });
   };
 
-  client.uploadFile = (file, controller, unpack_archive = false) => {
+  client.uploadFile = (file, unpack_archive = false, setFileProgress, thenCallback, onErrorCallback, setController) => {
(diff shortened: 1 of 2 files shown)


ciyer (approved):
Very nice! It seems like a small change, but it greatly improves usability.

ciyer (approved):
👍

PR #1163, opened 2020-12-04
feat: sync tabs on authentication events

### LINKED ISSUE #1048: Looking logged in open tabs when a log out has been performed
## Description

The user still looks logged in when a log out has been performed in another tab. Even if it's not possible to get data for protected resources, the UX is very confusing because the user believes to still be logged in.
An especially confusing case is the interactive environment page because the anonymous login is triggered on a 404 response from `/servers`, showing the content of an anonymous user in a page for logged user.
E.G. on a public project, a user may have launched a few environments. On log out, those disappear but the anonymous login is triggered, therefore the API succed and the following responses are valid. The user thinks the environments were closed but in fact he is now interacting as an anonymous user and the previous environments are still running but non accessible until the login.
Furthermore, if the user try to start new environments, he won't be able to push to the repository since the proper autentication tokens are missing.

## Reproduce

* Open 2 tabs in https://renkulab.io/
* Browse to a own project (possibly public) and start an interactive environment
* Try to logout from the other tab, then go back to the previous one

lorenzo-cavazzi:
This implements a basic mechanism to synchronize tabs on relevant authentication events.

I prepared 2 previews, one with notifications and one without. I believe taking advantage of the notification system is very useful for this case. Otherwise, the user may be confused when he switches to a page that was refreshed without explanations.

***Preview without notifications***: https://lorenzotest2.dev.renku.ch/
***Preview with notifications***:  https://lorenzotest.dev.renku.ch/

***How to test***: Open at least 2 tabs on the same deployment. Try to login/logout in one of them, then check what happens in the other.

fix #1048

### DIFF
--- a/client/src/authentication/Authentication.container.js
+++ b/client/src/authentication/Authentication.container.js
@@ -35,6 +35,10 @@ const RenkuQueryParams = {
 
 const LOGOUT_EVENT_TIMEOUT = 5000;
 
+/**
+ * Manages communication of login/logout events between tabs. It uses localStorage to communicate
+ * the events between tabs, and uses sessionStorage to remember an event after a refresh within a tab.
+ */
 const LoginHelper = {
   /**
    * Add the renku login query parameters
--- a/client/src/authentication/Authentication.container.js
+++ b/client/src/authentication/Authentication.container.js
@@ -27,17 +27,110 @@
 import React, { Component } from "react";
 
 
+const RenkuQueryParams = {
+  login: "renku_login",
+  logout: "renku_logout",
+  loginValue: "1"
+};
+
+const LOGOUT_EVENT_TIMEOUT = 5000;
+
+const LoginHelper = {
+  /**
+   * Add the renku login query parameters
+   *
+   * @param {string} url - return url for the authentication backend
+   */
+  createLoginUrl: (url) => {
+    let redirectUrl = new URL(url);
+    if (!redirectUrl.search.includes(RenkuQueryParams.login))
+      redirectUrl.searchParams.append(RenkuQueryParams.login, RenkuQueryParams.loginValue);
+
+    return redirectUrl.toString();
+  },
+  /**
+   * Remove renku login parameters and set localStorage object
+   *
+   * @param {object} history - return url for the authentication backend
+   */
+  handleLoginParams: (history) => {
+    // check if user has just logged in
+    const queryParams = new URLSearchParams(history.location.search);
+    if (queryParams.get(RenkuQueryParams.login) === RenkuQueryParams.loginValue) {
+      // delete the login param
+      queryParams.delete(RenkuQueryParams.login);
+      history.replace({ search: queryParams.toString() });
+
+      // save the login time to localStorage to allow other tabs to handle the event
+      localStorage.setItem(RenkuQueryParams.login, Date.now());
+    }
+  },
+  /**
+   * Set up event listener fol localStorage authentication events. Invoke only once.
+   */
+  setupListener: () => {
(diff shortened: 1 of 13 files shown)


rokroskar:
This looks really nice! I do like the notifications here, I think they make sense. I think we should discuss the presentation options for the popups and the page, but that should be done elsewhere. 

Two minor comments:

1. clicking on the "X" while the notification is shown doesn't mark it as read. I have to click on the drawer (is that what it's called) and click the "X" there or go to the page and click the check mark. 
2. The wording on the two notifications is slightly different:

> The page was refreshed because you recently logged out on a different tab.

vs

> The page was refreshed because the user has recently logged in on a different tab.

I guess I prefer the "you recently..." vs "the user..." but I don't have a strong opinion there - it just seems like they should be the same.

ciyer:
> 1. clicking on the "X" while the notification is shown doesn't mark it as read. I have to click on the drawer (is that what it's called) and click the "X" there or go to the page and click the check mark.

It originally worked the other way (x marked the notification as read), but I asked for it to be changed because I found myself closing the notification precisely when I wanted out of my way without reading it, with the intention of reading it later.

lorenzo-cavazzi:
I believe clearing the notification when clicking on the `X` would be the most natural behavior.
After trying the other way around for a while, it still seems strange that closing it leaves the red dot. The content we provide is concise, acknowledging that something is finished instead of providing information to read carefully.

I have changed the wording as suggested (preview not updated).
It's probably better to also use the same type of notification. I guess a warning is the better solution since it's a consequence of actions done somewhere else (the result in the tab may be negative, like a 404 page if a private project page was open).

client/src/authentication/Authentication.container.js
```diff
@@ -27,17 +27,110 @@
...
 
+const RenkuQueryParams = {
+  login: "renku_login",
+  logout: "renku_logout",
+  loginValue: "1"
+};
+
+const LOGOUT_EVENT_TIMEOUT = 5000;
+
+const LoginHelper = {
```
ciyer, line 38: Could you add a little comment here like:
```
The LoginHelper manages communication of login/logout events between tabs. It uses localStorage to communicate the events between tabs, and uses sessionStorage to remember an event after a refresh within a tab.
```

ciyer (changes requested):
This is great. It works very well (better than I thought possible)! 

My only request is for a little comment in the LoginHelper class. The tests are helpful (and also very nice!), but it took me a minute to understand how the code worked, and a class comment just explaining how localStorage and sessionStorage are used would be helpful.

I made a proposal. If my understanding is 100% correct, you can take that, otherwise, fix or describe in your own words.

client/src/authentication/Authentication.container.js
lorenzo-cavazzi: I agree the component is convoluted and a description would help. The one you provided is on point.

lorenzo-cavazzi:
I updated the preview on https://lorenzotest.dev.renku.ch/ with the latest changes.
It includes toast notifications with the modified behavior as agreed offline.

ciyer (approved):
🎉

lorenzo-cavazzi:
> I think we should discuss the presentation options for the popups and the page, but that should be done elsewhere.

@rokroskar Should I create an issue for discussing that and keep the notifications disabled in the meanwhile?

rokroskar:
@lorenzo-cavazzi yes please make an issue but not necessary to disable IMO. Thanks!!

lorenzo-cavazzi:
Ok, then I'll merge this.
This is the issue for improving notifications #1167

PR #1164, opened 2020-12-07
feat: remove dataset from projects

### LINKED ISSUE #1151: Delete files in datasets
# Motivation

I might want to remove files from a dataset

# UI Mock-ups 

Coming...

vfried:
There is an issue with button placement since the title and the buttons are at the same height.
Issue...
image

Solution 1: putting the buttons on top
image

Solution 2: removing text from buttons (We could display the text inside the button on hover) **I like this one**
How it looks like...
image

On hover... (we could also add the normal tooltip on hover)
image

Closes #1151

### DIFF
--- a/client/src/project/datasets/remove/RemoveDataset.present.js
+++ b/client/src/project/datasets/remove/RemoveDataset.present.js
@@ -1,0 +1,101 @@
+/*!
+ * Copyright 2020 - Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ *  renku-ui
+ *
+ *  RemoveDataset.present.js
+ *  Presentational components.
+ */
+
+
+import React, { Fragment } from "react";
+import { Row, Col, Modal, ModalHeader, ModalBody, Button, Alert, FormText } from "reactstrap";
+import { Loader } from "../../../utils/UIComponents";
+
+function RemoveDatasetPresent(props) {
+
+  let modalContent = null;
+
+  if (props.serverErrors) {
+    let content;
+    let error = props.serverErrors.error;
+    if (typeof serverErrors == "string") {
+      content = <pre className="text-wrap">{error}</pre>;
+    }
+    else {
+      const errors = Object.keys(error).map(v => {
+        const text = typeof error[v] == "string" ?
+            `${v}: ${error[v]}` :
+            `Error message: ${JSON.stringify(error[v])}`;
+        return (<pre key={v} className="text-wrap">{text}</pre>);
+      });
+      if (errors.length === 1)
+        content = (errors[0]);
+      else
+        content = error[0];
+    }
+
+    modalContent = <Col>
+      <Alert color="danger">
+        <p>Errors occurred while deleting this dataset</p>
+        <p>{content}</p>
+      </Alert>
+    </Col>;
+  }
+  else {
+    modalContent = <Col>
+      <p>
+        Are you sure you want to remove dataset <strong>{props.dataset.name}</strong>?
+      </p>
+      <Fragment>
+        { props.submitLoader !== undefined && props.submitLoader.value ?
+          <FormText color="primary">
+            <Loader size="16" inline="true" margin="2" />
+            {props.submitLoader.text}
+          </FormText>
+          : null
+        }
+        <Button type="submit" onClick={props.removeDataset}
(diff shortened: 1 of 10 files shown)


ciyer:
I like Solution 2. I think it looks good and is clear in terms of functionality.

vfried:
I implemented the solution 2 with the tooltip that we have been using to be consistent. These tooltips don't work 100% well though (I've seen this not working well in other places as well). I can implement the on hover solution that we describe on top if we prefer this solution. We should also create an issue to try to solve the hover problem in general.

vfried:
I changed the order of the actions to be like...
image

In case the user is outside of a project only the add to project is displayed and it looks like...

vfried:
@ciyer latest changes for this are in virginiatest

ciyer (changes requested):
# Buttons

I like the icons, but, I think we want to give delete a less prominent place. Could we make the buttons like this:

image

(Edit on the far right, a "more options" button to the left. Clicking on "More options" should bring the two options: add to project and delete).

vfried:
I don't like this suggestion.
 
In my opinion the add to project button is the most important button in datasets (because we want to promote users to share datasets) so it doesn't make sense for me to hide this button inside a menu. 

Regarding edit and delete I think if we put them in a menu it could happen that the users don't know there is new functionality. 
In case we add more buttons I would use the menu option but not now. I can try changing the color of the buttons if this is preferred. Given the size of the buttons, I wouldn't say that any of them has a "prominent place".

vfried:
This is how it would look in reality...
image

vfried:
We can play with colors to make buttons less strong if this is what we want...
image

client/src/project/datasets/remove/RemoveDataset.present.js
```diff
@@ -0,0 +1,101 @@
...
+  let modalContent = null;
+
+  if (props.serverErrors) {
+    let content;
+    let error = props.serverErrors.error;
+    if (typeof serverErrors == "string") {
+      content = <pre className="text-wrap">{error}</pre>;
+    }
+    else {
+      const errors = Object.keys(error).map(v => {
```
ciyer, line 42: Is this a pattern that is used in multiple places by the renku-core? If so, we should extract this to a function
```
function displayErrorFromServerErrors(serverErrors) {
    let content;
    let error = props.serverErrors.error;
    if (typeof serverErrors == "string") {
      content = <pre className="text-wrap">{error}</pre>;
    }
    else {
      const errors = Object.keys(error).map(v => {const text = typeof error[v] == "string" ?
            `${v}: ${error[v]}` :
            `Error message: ${JSON.stringify(error[v])}`;
        return (<pre key={v} className="text-wrap">{text}</pre>);
      });
      if (errors.length === 1)
        content = (errors[0]);
      else
        content = error[0];
    }
    return content;
}

ciyer (changes requested):
Code looks good, and the feature works well (tested a public project with KG and a private project w/o KG).

I have one question/request about the code and a request about the button ordering/display.

client/src/project/datasets/remove/RemoveDataset.present.js
vfried: I think the function for the Form generator should be more generic and potentially could be improved since we get error messages for fields in particular and we should display this error messages in the fields where they happen.

Also, some changes where made there (in the form generator) that actually broke/break the error messages in some cases. I discovered this a few days ago and I was planing on investigating this a bit more when making the form generator available everywhere. 

I can extract this in a function now but i think we should revise the way we handle errors from the core service in general.

ciyer:
> I don't like this suggestion.
> 
> In my opinion the add to project button is the most important button in datasets (because we want to promote users to share datasets) so it doesn't make sense for me to hide this button inside a menu.
> 
> Regarding edit and delete I think if we put them in a menu it could happen that the users don't know there is new functionality.
> In case we add more buttons I would use the menu option but not now. I can try changing the color of the buttons if this is preferred. Given the size of the buttons, I wouldn't say that any of them has a "prominent place".

I think it makes sense to have the add to project button easily accessible, but I do not think delete should be so readily available. Deleting a dataset should be rare.

Can we do it this way, then

image

On the dataset page. "Delete" should be moved to an action on the edit page:

image

vfried:
I think operations should be simple and easy to reach in general. I am not sure that putting the delete inside the modify will be clear for the user. Considering that things are backed up in git and can be reverted and also that there is a confirmation window that asks a user if they are sure they want to delete the dataset I am not sure that we want a delete operation inside the modify.

vfried:
I am just trying to think about this in a different way... the reason why we want this button "hidden" is because we want to prevent the user from deleting datasets in general or because we want to prevent users from accidentally deleting a dataset?

--> If we want to prevent users from accidentally deleting a dataset we could implement something like the delete project in gitlab that asks you to write the name of the project that you are trying to delete.

--> If we want to prevent users from deleting datasets in general then i can hide the button somewhere.

vfried:
In case this two things go inside modify dataset i think we should do something like this... 
image

vfried:
I named this remove but i'm not sure if this should be remove or delete... any suggestions?

vfried:
4th proposal... if we want to group operations I would group the delete and edit inside a menu... I would leave the add to project button as it is and add a menu with delete and edit.

vfried:
As we spoke, here are the new changes... i also did a small refactor in error messages since i realized we didn't need all the code that i pushed, the error messages for delete are simpler.

Changes:
- new menu
- fix for errors
- renaming to delete

image

client/src/project/datasets/remove/RemoveDataset.present.js
vfried: I realized most of this code isn't needed so I deleted most of it, it's much simpler now.

ciyer (approved):
Very nice! I think this solution works very well.

PR #1174, opened 2020-12-08
fix: error message for form panel is now correct

### LINKED ISSUE #1172: Data set error message
## Description

When trying to editing a dataset fails, I get this error message:

image

The header is wrong: I was editing a dataset, not creating a project.

For clarity -- the issue is not that the error occurs, but that the error message is incorrect. It should say: `Could not save changed to data set.`

## Reproduce

This error is hard to reproduce, but the message should be easy to fix.

vfried:
This is actually inside the project form.

Fixes also problem in template up to date message...

image

Closes #1172


ciyer (approved):
👍

PR #1166, opened 2020-12-07
ci: deploy a new environment on every pull request

lorenzo-cavazzi:
* Deploy the environment on new pull requests.
* Update the environment on PR changes.
* Cleanup the environment when closing the PR.

re #1165

PR #1179, opened 2020-12-14
Add GitHub action to deploy an environment on every pr 

lorenzo-cavazzi:
_This is the same content of #1166 , just rebased and reposted to verify it works as expected_.

This PR adds a new GitHub action that:
* Deploys a new environment for every new pull request.
* Keeps it up-to-date when new commits are pushed.
* Deletes it when the PR is closed.

RenkuBot will post and update a message with the deployment status. Keep in mind that GitHub doesn't promptly notify about message edits, so it's necessary to manually refresh the PR page to see the new content.


.github/workflows/acceptance-tests.yml
```diff
@@ -0,0 +1,133 @@
...
+        GITLAB_TOKEN: ${{ secrets.DEV_GITLAB_TOKEN }}
+        KUBECONFIG: "${{ github.workspace }}/renkubot-kube.config"
+        RENKU_NAMESPACE: ci-ui-${{ github.event.number }}-renku
+        RENKU_TMP_NAMESPACE: ci-ui-${{ github.event.number }}-renku-tmp
+        RENKU_RELEASE: ci-ui-${{ github.event.number }}-renku
+        RENKU_VALUES_FILE: "${{ github.workspace }}/values.yaml"
+        RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES }}
+        RENKUBOT_KUBECONFIG: ${{ secrets.RENKUBOT_DEV_KUBECONFIG }}
+        RENKUBOT_RANCHER_BEARER_TOKEN: ${{ secrets.RENKUBOT_RANCHER_BEARER_TOKEN }}
+        renku: "@development"
```
rokroskar, line 71: if you want this to pass atm you need to set this to `@000-deploy-action`

lorenzo-cavazzi: Ok great, I updated that to verify if it works

lorenzo-cavazzi:
Closing because a newer version is available here #1196 .
I will miss the auto-updated comments :cry:

PR #1184, opened 2020-12-16
Fix wrong commit list when starting new environment

### LINKED ISSUE #1127: Wrong commit list when starting new environments (occasionally)
When the user tries to start a new environment and he selects a different branch **before** the commits are completely fetched (i.e. before the commit dropdown menu appears), the commit dropdown shows the wrong commits, specifically the commits for the `master` branch.

It's easy to reproduce this for projects with many commits (~1000), while it's very difficult if the commits are limited (<= 100).
It's possible that this bug has been there for a long time but we never found out because, until recently, only the last 100 commits were fetched (https://github.com/SwissDataScienceCenter/renku-ui/commit/a5ca3efa2ec88d58a30678ce10859f4c0939151c).

lorenzo-cavazzi:
In specific circumstances, it was possible to get outdated content after changing branch / commit.
In practice, that seems to happen only when changing the branch while commits fetch wasn't complete yet (it's not too tricky on projects with 1000+ commits).

I tried a few solutions, but I settled on disabling the branch / commit dropdown while fetching resources. Other solutions were potentially better but required a lot of changes in the code for minimal benefit.
The PR also implements better error handling when fetching commits fails.

Test environment: https://lorenzotest.dev.renku.ch

***How to test***: you can use a project with 1000+ commit, like `covid-19/covid-19-public-data`. Previously, changing the branch quickly after clicking on the "Environments --> New" tab resulted in wrong commits. That is not possible anymore.
https://lorenzotest.dev.renku.ch/projects/covid-19/covid-19-public-data/environments/new

fix #1127

### DIFF
--- a/client/src/notebooks/Notebooks.container.js
+++ b/client/src/notebooks/Notebooks.container.js
@@ -157,7 +157,7 @@ class StartNotebookServer extends Component {
     this.coordinator = new NotebooksCoordinator(props.client, this.model, this.userModel);
     // TODO: this should go away when moving all project content to projectCoordinator
     this.projectModel = props.model.subModel("project");
-    this.projectCoordinator = new ProjectCoordinator(props.client, props.model.subModel("project"));
+    this.projectCoordinator = new ProjectCoordinator(props.client, this.projectModel, props.notifications);
     this.notifications = props.notifications;
     // temporarily reset data since notebooks model was not designed to be static
     this.coordinator.reset();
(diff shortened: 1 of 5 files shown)


ciyer (dismissed):
This is great solution! As a user, the delay and limitation are completely unnoticeable.

lorenzo-cavazzi:
I rebased cause I prefer to keep the 2 separate commits in master.
Could you re-approve?

ciyer (approved):
👍

ciyer (changes requested):
We should merge https://github.com/SwissDataScienceCenter/renku-ui/pull/1176 first.

ciyer (approved):

Issue #1121, opened 2020-11-11
Revive JupyterLab Extension


ciyer:
We implemented a JupyterLab Extension, but that was coded against an earlier version of the API (pre 1.0). We should revive the extension. Initially it should just provide a command cheat sheet.

ciyer:
Superseded by https://github.com/swissdatasciencecenter/renku-jupyterlab-ts/issues/29.

PR #1185, opened 2020-12-17
Initial Renku jupyterlab extension

ciyer:
Does the first item under #1121: creates a jupyterlab extension that provides a cheat sheet for the Renku CLI.

Preview Image

image

Test instance coming soon.


ciyer:
Moved this code to https://github.com/SwissDataScienceCenter/renku-jupyterlab-ts/pull/30

PR #1186, opened 2020-12-23
speed up csv preview and improve lfs and big files

### LINKED ISSUE #1147: Slow CSV File Rendering
## Description

Viewing this file in RenkuLab causes the UI to grind to a halt.

  https://dev.renku.ch/projects/cramakri/my-test-project/files/blob/data/rendering-probolem/data_original_aziza_HIVIncidence_per1000_15_49.csv

We should investigate why because the file is not especially large.

## Solution

- [x] Find out why this file causes problems
- [x] Fix the problem
- [x] Also, add limits to the size of files we display in the RenkuLab UI. We do not want to show very large files in the UI.

lorenzo-cavazzi:
This PR improves the file preview in the following cases:

* csv: speed up the rendering by showing it as a text file, without trying to highlight the code
* lfs files: common message with download option
* big files: common message with download option and "preview anyway" option when the file size is reasonable (>0.5MB & <5MB)

An interesting idea for csv files would be to preview with an interactive table. I'll open a discussion on another issue since this is not in the scope ( --> #1187 )

A preview is available here: https://lorenzotest.dev.renku.ch

**How to test**
Try to browse csv files, as well as big and lfs files.

Here are a few test cases:
* csv: https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/new-project/files/blob/csv/medium.csv
* big file: https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/new-project/files/blob/csv/big.csv
* too big file: https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/new-project/files/blob/csv/huge.csv
* lfs: https://lorenzotest.dev.renku.ch/projects/covid-19/covid-19-public-data/files/blob/data/covid-19-chile/covid19_chile_2020-03-09.csv

fix #1147


client/src/file/File.container.js
```diff
@@ -18,29 +18,40 @@
...
-  "py", "js", "json", "sh", "r", "yml", "csv", "parquet", "cwl", "job", "prn", "rout",
+  "py", "js", "json", "sh", "r", "yml", "parquet", "cwl", "job", "prn", "rout",
   "dcf", "rproj", "rst", "bat", "ini", "rmd", "jl", "toml", "ts", "rs", "scala",
   "c", "cc", "cxx", "cpp", "h", "hh", "hxx", "hpp", // C++
   "f", "for", "ftn", "fpp", "f90", "f95", "f03", "f08" // Fortran
 ];
-const TEXT_EXTENSIONS = ["txt"];
+const TEXT_EXTENSIONS = ["txt", "csv"];
+
+const LIMITS = {
```
ciyer, line 39: Should we make this a configuration and put these values into the k8s chart? 

I cannot decide -- I can see cases where a deployment might want to change these (e.g., they know users will be connecting over a very slow connection), but maybe those are corner cases.

On the other hand, code is much simpler this way.

client/src/file/File.container.js
```diff
@@ -18,29 +18,40 @@
...
+  soft: 1024 * 1024 * 0.5,
+  softLabel: "0.5 MB",
```
ciyer, line 41: I think it would be cleaner to use a function to convert the value into a label. We may already have one in Utils, if not, here is a good one:

  https://stackoverflow.com/a/14919494

Maybe we can extend it to optionally take units, in which case the function doesn't try to figure out the best units, it just uses what is asked for. We can replace the line with:

```
humanFileSize(LIMITS.soft, units="MB")
```

client/src/file/File.present.js
```diff
@@ -184,12 +179,54 @@ class ShowFile extends React.Component {
...
+    );
+
+    // LFS or very big files
+    if (this.props.lfs || this.props.hardLimit) {
+      const reason = this.props.hardLimit ?
+        `the file is too big (more than ${this.props.hardLabel})` :
+        "the file is stored in Git LFS";
+      return (
+        <CardBody key="file preview" className="pb-0">
+          <p>The preview is disabled because {reason}.</p>
```
ciyer, line 204: Wording: `The preview is not available because {reason}`.

client/src/file/File.present.js
```diff
@@ -184,12 +179,54 @@ class ShowFile extends React.Component {
...
+          <p>You can still {downloadLink}</p>
+        </CardBody>
+      );
+    }
+
+    // Big preview-able files
+    if (!this.props.previewAnyway) {
+      const loadButton = (
+        <Button color="link" className="p-0" onClick={() => { this.props.loadAnyway(); }}>
+          load it anyway
```
ciyer, line 214: Wording: `preview it anyway`.

client/src/file/File.present.js
```diff
@@ -184,12 +179,54 @@ class ShowFile extends React.Component {
...
+        </Button>
+      );
+      return (
+        <CardBody key="file preview" className="pb-0">
+          <p>The preview may take long because the file is big (more than {this.props.softLabel}).</p>
```
ciyer, line 219: Wording: `The preview may be slow because...`

client/src/file/File.present.js
```diff
@@ -151,27 +151,22 @@ class ShowFile extends React.Component {
       return (
         <Card>
           <CardHeader className="align-items-baseline">&nbsp;</CardHeader>
-          <CardBody>{"Loading..."}</CardBody>
+          <CardBody>Downloading content... <Loader size="14" inline="true" /></CardBody>
```
ciyer, line 154: Wording: `Downloading...` (I.e., remove "content").

ciyer (changes requested):
Code looks good, and this feature is very useful.

I have some requests for wording changes.

I also found this alignment problem ("load it anyway" has a different baseline than the rest).

image

client/src/file/File.container.js
lorenzo-cavazzi: I will expand and re-use this function since we are already using it https://github.com/SwissDataScienceCenter/renku-ui/blob/c7bd5657dc7a4a39a4d77de8f920131ba72c5044/client/src/utils/HelperFunctions.js#L223-L233

I prefer to keep the unit dynamic to keep things simple while allowing the admin to configure the sizes.
We can change this later if any user requires deeper personalization

client/src/file/File.container.js
lorenzo-cavazzi: I moved the 2 thresholds to the values files. Maybe it's a marginal benefit, but the code complexity doesn't suffer much -- it's also much faster to implement it now rather than doing it later.

lorenzo-cavazzi:
I addressed all the points and updated the preview in https://lorenzotest.dev.renku.ch

The one thing that still bothers me is the "load it anyway" alignment. It's fixed on Firefox, but on Chrome it seems to be off 1px depending on the zoom level. I guess the workaround would be to use an `<a>` element instead of `<button type="link">`, even if I generally prefer to avoid that whenever the logical element is a button and not a real link.

client/src/file/File.container.js
```diff
@@ -73,12 +87,26 @@ class FilePreview extends React.Component {
...
+      return null;
+
+    // LFS files and big files
+    if (this.fileIsLfs() || (this.props.file.size > this.props.previewThreshold.soft && !this.state.previewAnyway)) {
+      return (
+        <FileNoPreview
+          url={this.props.downloadLink}
+          lfs={this.fileIsLfs()}
+          softLimit={this.props.previewThreshold.soft}
+          softLimitReached={this.props.file.size > this.props.previewThreshold.soft ? true : false}
```
ciyer, line 99: Why isn't this just? :
`softLimitReached ={this.props.file.size > this.props.previewThreshold.soft}`

client/src/file/File.container.js
```diff
@@ -73,12 +87,26 @@ class FilePreview extends React.Component {
...
+          hardLimit={this.props.previewThreshold.hard}
+          hardLimitReached={this.props.file.size > this.props.previewThreshold.hard ? true : false}
```
ciyer, line 101: Same question as for softLimitReached.

ciyer (commented):
This works great! I tried on Safari and Chrome on Mac, and the alignment of the "preview it anyway" button looked ok to me in both browsers at multiple zoom levels, so I think we can leave it as is.

I have one question about the code, but other than that, it is ready for merge for me.

helm-chart/renku-ui/values.yaml
```diff
@@ -65,6 +65,10 @@ templates:
       ref: 0.1.15
       name: Renku
 
+previewSizeThreshold:
```
rokroskar, line 68: Could you add a brief comment here to document the consequences of these values?

helm-chart/renku-ui/templates/deployment.yaml
```diff
@@ -72,6 +72,8 @@ spec:
               value: {{ toJson .Values.privacy.banner.layout | quote }}
             - name: TEMPLATES
               value: {{ toJson .Values.templates | quote }}
+            - name: PREVIEW_THRESHOLD
+              value: {{ toJson .Values.previewSizeThreshold | quote }}
```
rokroskar, line 76: ha! it's never occurred to me to use `toJson` with environment variables... 👍 I guess I never noticed it in the values above 🤔

rokroskar (approved):
Thanks @lorenzo-cavazzi! Looks and works as advertised - only one minor comment about adding a brief line of documentation in the values file.

client/src/file/File.container.js
lorenzo-cavazzi: you are right, the `? true : false` part is totally unnecessary

helm-chart/renku-ui/templates/deployment.yaml
lorenzo-cavazzi: Yeah I initially did it for `privacy` since it's a complex structure. It's a convenient way to import data structures without an exploding number of env variables.

lorenzo-cavazzi:
All the comments have been addressed, thanks for the reviews!

helm-chart/renku-ui/templates/deployment.yaml
rokroskar: indeed - though I guess it has to be used carefully to not obfuscate the application configuration too much

ciyer (approved):
🎉

rokroskar (approved):

PR #1191, opened 2021-01-05
ci: update helm url

lorenzo-cavazzi:
The `test-chart` action uses the helm version `2.16.9`. We need to update to version `>= 2.17.0` to prevent it from failing as in https://github.com/SwissDataScienceCenter/renku-ui/runs/1649868184

Reference: https://helm.sh/blog/new-location-stable-incubator-charts/


ciyer (approved):
👍 Thanks for looking into this and getting it resolved!

Issue #1183, opened 2020-12-15
Refactor project state management


lorenzo-cavazzi:
## Description

Moving `ProjectModel` to the global redux store and the functions to `ProjectCoordinator` requires a gradual approach.
The idea is to first migrate the code used in single tabs (Files, Collaboration ...) to either `ProjectCoordinator` or a dedicated coordinator wherever this makes sense (e.g. this is already the case for Interactive Environments and it may be the case for datasets #1182).
The safest approach may involve temporarily storing the data in 2 positions at the same time.
An example is [fetchCommits](https://github.com/SwissDataScienceCenter/renku-ui/blob/aa7b63d12f4042b92e73d5064fc8b4fdeab9766f/client/src/project/Project.state.js#L432-L455), that requires `"metadata.id"`. That is still fetched by the [fetchProject function](https://github.com/SwissDataScienceCenter/renku-ui/blob/aa7b63d12f4042b92e73d5064fc8b4fdeab9766f/client/src/project/Project.state.js#L163-L186) in `ProjectModel`, but the content is also saved to the global store in the [container fetchProject function](https://github.com/SwissDataScienceCenter/renku-ui/blob/aa7b63d12f4042b92e73d5064fc8b4fdeab9766f/client/src/project/Project.js#L212-L217) using the new projectCoordinator.
Once nothing references the `ProjectModel` content anymore, the need for duplication will go away and it's only a matter of cleaning up the unnecessary invocations.

Notice that, in this `fetchProject` example, the schema in the global store is different. The idea is to keep the `fetching` + `fetched` variables approach to know if the current information status. That is already extensively used in other parts in the global store.

To be effective, we should try to store all the content coming from an API in a separate object, and avoid nesting when unnecessary.
In the following case (taken from here)[https://github.com/SwissDataScienceCenter/renku-ui/blob/aa7b63d12f4042b92e73d5064fc8b4fdeab9766f/client/src/model/RenkuModels.js#L257-L286], `commits` and `metadata` both have `fetched` and `fetching` variables. The structure is slightly different, but the behavior is similar. We expect that checking those variables would give us a clear indication of the status of the related data.
```
projectGlobalSchema {
  commits: {
    list: object[],
    fetched: Date nullable,
    fetching: bool,
    error: object nullable
  },
  metadata: {
    fetched: Date nullable,
    fetching: bool,
    exists: bool,
    id: string,
    otherProp1: <...>,
    otherProp2: <...>,
    <...>
[...]

Issue #1114, opened 2020-11-05
Edit the project settings from the UI


lorenzo-cavazzi:
##  Motivation

We are gradually storing more project settings in the `.renku/renku.ini` file. Since some of them are relevant only for the interactive environments, it would be nice to allow the user to change them directly from the UI.

## Proposal

* Move the clone commands and repository URL to an overview tab. I suggest renaming the `Stats` tab to `Information` (or `Info`) and move the content there. The preview is only a broad reference, some elements should be adjusted better.
![Screenshot_20201105_140914](https://user-images.githubusercontent.com/43481553/98245736-a9795a00-1f71-11eb-877e-50313f7e76e4.png)

* Divide the Settings page into 2 sections ~(similar to the `Cards` components used in the preview above), one with Renkulab settings (the current project Tags and Description)~ (EDIT: use subtabs as suggested in the next post) and a new one with the actual project settings.
We should show the current settings and allow to edit or add them. Adding should suggest the available parameters (we have a limited set, custom options and custom image) but also allow the user to add new custom fields.
The best solution would be to have a backend service from renku-core to invoke the `renku settings` commands. [Preview: to be added]

# Progress

- [x] API Design (sprint-2020-12-04)
- [ ] UI implementation

ciyer:
I propose putting this on a subtab of the settings page.

<img width="963" alt="image" src="https://user-images.githubusercontent.com/1196411/99834233-8d75db00-2b63-11eb-8d65-23fd14ec62eb.png">

lorenzo-cavazzi:
I think the mockup you proposed is on point!
A couple of small variations I would do:
1. Move the "Pinned image" on top, just after the "Default for new pods" to keep it in line with the Environments page.
2. Add the possibility to set custom values for know fields (e.g. 16G of RAM, or any other random string) and new fields. When doing that, the value should be underlined in red/yellow with a note saying the value may not by valid in this RenkuLab deployment. We can get the list of available options and values from the `GET /notebooks/server_options` API.

The cleanest solution requires APIs from renku-core to expose `renku-config`. Technically, implementing a custom UI solution isn't difficult, but I would avoid doing that to prevent potential discrepancies between UI/renku-python

## API structure

Here is a proposal for API: https://app.swaggerhub.com/apis-docs/lorenzo-sdsc/renku-core/0.1.0
It's basically 3 commands to:
1. get the list of key/value pairs
2. set a new key or edit an existing one
3. delete an existing key

[...]

PR #1192, opened 2021-01-08
fix projects overview tabs

### LINKED ISSUE #1074: Project stats are not correct
## Description

Project stats (Renku project page > Overview > Stats) are not correct when a project is modified. It seems that they will be eventually correct after couple of hours.

## Reproduce

1. Go to a project with some files in LFS: https://dev.renku.ch/projects/mohammad.alisafaee/lego-analytics/overview/stats
2. Clone the project
3. Go to the cloned project > Overview > Stats
4. See that **Storage stats** are not correct (even after gitlab pipeline finishes)

lorenzo-cavazzi:
This PR's primary goal is to fix the project overview stats tab (see #1074).
It shows incorrect data when:
1. the project has been recently updated.
2. the user has no specific permission for a public or internal project.

Point 1 is fixed. It happened because we never refreshed the stats. In the past, we also observed that the GitLab API was occasionally returning outdated values for a while after pushing new commits. This [should be fixed now](https://gitlab.com/gitlab-org/gitlab-foss/-/merge_requests/26854).

Point 2 is partially fixed. Sadly, the GitLab project API doesn't return statistics for a public or internal project if the user doesn't have specific permissions for that. This is unexpected because it's undocumented -- I still hope [there is another way](https://gitlab.com/gitlab-org/gitlab/-/issues/296656).  
We don't show the missing stats anymore when unavailable, but we provide extra links to the GitLab UI where some of those can be inspected.

Mind that this PR includes a few other minor changes for the overview tabs:
* Avoid fetching readme on project overview sub-tabs
* Fetch commits when opening overview/commits
* Move overview/commits components to the new ProjectsOverview files
* Merge stats into a single table (it anticipates a tiny bit of #1114 for convenience)
* Move the statistics information to the project coordinator (better to anticipate steps from #1183 whenever possible)

**Preview**: https://lorenzotest.dev.renku.ch/

**How to test**
Keep an instance of https://dev.renku.ch/ and compare the following:
* Try to move around in the project overview sub-pages on the left navigation pills. You shouldn't see queries to the readme file at each click.
* Open the project overview commits for a project and wait until the full list is fetched. Then manually navigate to another project from the projects page. If you open the commits, you should see they are now correct while wrong on dev.
* Open one of your project overview stats. Then manually navigate to a public project where you have no permission. The stats will be wrong on dev, showing the previous values.
* 
Mind that the last 2 problems may not happen if you first navigate around the project, triggering fetches from GitLab for other reasons.

fix #1074


client/src/project/Project.present.js
```diff
@@ -1507,4 +1317,4 @@ class ProjectView extends Component {
 export default { ProjectView };
 
 // For testing
-export { filterPaths, ProjectViewCommitsBody };
+export { filterPaths };
```
ciyer, line 1320: Very minor, but this file should have a newline at the end.

client/src/project/overview/ProjectOverview.present.js
```diff
@@ -0,0 +1,248 @@
...
+    const storageSize = this.valueOrEmptyOrLoading(statistics.data.storage_size, statistics.fetching);
+    const repositorySize = this.valueOrEmptyOrLoading(statistics.data.repository_size, statistics.fetching);
+    const lfsSize = this.valueOrEmptyOrLoading(statistics.data.lfs_objects_size, statistics.fetching);
+    const commitsCount = this.valueOrEmptyOrLoading(statistics.data.commit_count, statistics.fetching, false);
+
+    let info = null;
+    if (!metadata.accessLevel) {
+      info = (
+        <p className="font-italic">
+          Since you do not have any permission for this project, some fields may be unavailable.
```
ciyer, line 65: Would it make more sense to say:
`Since you are not a member of this project, some information may not be available.`?

client/src/project/overview/ProjectOverview.present.js
```diff
@@ -0,0 +1,248 @@
...
+        <CardBody><Row><Col>
+          {info}
+          <Table className="table-responsive" key="stats-table" size="sm">
+            <tbody>
+              <tr>
+                <th className="align-middle" scope="row">Number of Branches</th>
+                <td className="px-3 px-lg-4 align-middle">{branches.length + 1}</td>
+                <td>
+                  <ExternalLink role="text" showLinkIcon={true}
+                    url={`${metadata.repositoryUrl}/branches`} title="Branches in Gitlab" />
```
ciyer, line 84: Very minor as well (not not actually part of this PR, but I'm just noticing it, so may as well fix):
`Gitlab` -> `GitLab`

client/src/project/overview/ProjectOverview.present.js
```diff
@@ -0,0 +1,248 @@
...
+                </td>
+              </tr>
+              <tr>
+                <th className="align-middle" scope="row">Number of Forks</th>
+                <td className="px-3 px-lg-4 align-middle">{metadata.forksCount}</td>
+                <td>
+                  <ExternalLink role="text" showLinkIcon={true}
+                    url={`${metadata.repositoryUrl}/forks`} title="Forks in Gitlab" />
```
ciyer, line 92: `Gitlab` -> `GitLab`

client/src/project/overview/ProjectOverview.present.js
```diff
@@ -0,0 +1,248 @@
...
+                </td>
+              </tr>
+              <tr>
+                <th className="align-middle" scope="row">Number of Stars</th>
+                <td className="px-3 px-lg-4 align-middle">{metadata.starCount}</td>
+                <td>
+                  <ExternalLink role="text" showLinkIcon={true}
+                    url={`${metadata.repositoryUrl}/-/starrers`} title="Stars in Gitlab" />
```
ciyer, line 100: `Gitlab` -> `GitLab`

client/src/project/overview/ProjectOverview.present.js
```diff
@@ -0,0 +1,248 @@
...
+                </td>
+              </tr>
+              <tr>
+                <th className="align-middle" scope="row">Number of Commits</th>
+                <td className="px-3 px-lg-4 align-middle">{commitsCount}</td>
+                <td>
+                  <ExternalLink role="text" showLinkIcon={true}
+                    url={`${metadata.repositoryUrl}/commits`} title="Commits in Gitlab" />
```
ciyer, line 108: `Gitlab` -> `GitLab`

client/src/project/overview/ProjectOverview.present.js
```diff
@@ -0,0 +1,248 @@
...
+                </td>
+              </tr>
+              <tr>
+                <th scope="row">Storage Size</th>
+                <td className="px-3 px-lg-4 align-middle">{storageSize}</td>
+                <td className="align-middle" rowSpan="3">
+                  <ExternalLink role="text" showLinkIcon={true}
+                    url={metadata.repositoryUrl} title="Sizes in Gitlab" />
```
ciyer, line 116: `Gitlab` -> `GitLab`

ciyer (changes requested):
Code looks very good! I have some very minor things (wording, spelling of GitLab, which wasn't introduced in this PR, but I just noticed it).

I noticed some strange behavior, though.

1. Stats disappear

First look at stats shows this:
image

But if I go to description and come back, I get this:
image

2. On the above project, in RenkuLab, it says there is 1 branch, but GitLab shows 3
image

client/src/project/overview/ProjectOverview.present.js
```diff
@@ -0,0 +1,248 @@
...
+              <tr>
+                <th className="align-middle" scope="row">Number of Commits</th>
+                <td className="px-3 px-lg-4 align-middle">{commitsCount}</td>
+                <td>
+                  <ExternalLink role="text" showLinkIcon={true}
+                    url={`${metadata.repositoryUrl}/commits`} title="Commits in Gitlab" />
+                </td>
+              </tr>
+              <tr>
+                <th scope="row">Storage Size</th>
```
ciyer, line 112: Is there a reason to drop the old structure where Project Stats and Storage Stats where shown separately? I prefer 

image

to

image

I think having two sections looks nicer and is easier to read (but I think the links in the new version are better than the buttons in the old one).

lorenzo-cavazzi: As discussed offline, we will keep the separate structure for now.

lorenzo-cavazzi:
I addressed the comments.

About the 2 bigger problems:

1. Fixed. I introduced that by mistake just before committing :facepalm: Thanks for spotting it!
2. Good point. I didn't change the previous implementation, but that number of branches is wrong. Would it work if I open a follow-up issue to tackle immediately? I prefer not to fix it here to keep things separate (I will move `getBranches` in the new project coordinator and that requires minor adjustments to the notebooks components as well)

Preview updated at https://lorenzotest.dev.renku.ch/

lorenzo-cavazzi:
I addressed the comments.

About the 2 bigger problems:

1. Fixed. I introduced that by mistake just before committing :facepalm: Thanks for spotting it!
2. Fixed. I didn't change the previous implementation, but that number of branches is indeed wrong. Now the branches are fetched once at project load. Mind that the number already excludes the autosaved branches.
We could further improve the logic to refresh periodically as we do with the other stats and commits, but that will be much simpler after migrating the function to the new project coordinator. I wouldn't include that in this PR.

Preview updated at https://lorenzotest.dev.renku.ch/

ciyer (approved):
Very nice fix!

PR #1204, opened 2021-01-18
ci: remove unneeded requirements on cleanup job

lorenzo-cavazzi:
This fix is required to properly trigger the `cleanup` action whenever a PR is merged or closed in any way.

In short, the `cleanup` action required another step to start, but that could never happen because that step doesn't run on `close` events.

P.S. This has been tested here #1203


ciyer (approved):
👍

PR #1208, opened 2021-01-19
feat: add Url helper class

### LINKED ISSUE #779: urls available in all pages
## Motivation

There are many places in the code where URLs are hardcoded. Not only it's difficult to change them, but for the non-trivial ones it's hard to guess the correct URL.

E.G. `/projects` is easy to guess, but `/projects/search?q=lorenzo.cavazzi.tech&searchIn=users` is not, and it's not coherent with the url one gets when clicking the search button on the same page
`/projects/search?q=lorenzo.cavazzi.tech&page=1&orderBy=last_activity_at&orderSearchAsc=false&searchIn=users`

## Proposal

We could store the URLs in the redux store in the `url` object.
* for the trivial ones, a hardcoded string should do it (E.G. `url.datasets: "/datasets"`)
* for the complex, we may prefer a function (E.G. `url.makeDatasetSearch: (datasetName, ascending=true) => "/datasets?q=" +  datasetName + "&orderSearchAsc=" + ascending`)

lorenzo-cavazzi:
Adds a `Url` helper class with static methods and properties that simplifies getting a URL form wherever in the application.

Most of the URLs are static strings, but some are more complex (e.g. project in subgroups, projects search page). Those can now be built on the fly and even validated.
The PR also includes a thorough list of tests acting as documentation, and a few first examples.

The goal is to remove all the duplicate codes to create URLs on the fly (sometimes faulty, especially for project pages). We can either:

1. Gradually change the URL code when addressing issues impacting it.
2. Try to remove all the duplicates and fix the problems as part of this PR.
3. Create a follow-up issue for `2)`

I'm undecided between `1)` and `2)`, let me know if you have any preference.

**How to test**: try to start a new interactive environment and, from the global `environments` page, click on the project link.

fix #779 
/deploy

### DIFF
--- a/client/src/utils/url/Url.js
+++ b/client/src/utils/url/Url.js
@@ -1,0 +1,264 @@
+/*!
+ * Copyright 2021 - Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ *  renku-ui
+ *
+ *  Url.js
+ *  Url helper class.
+ */
+
+/** Class to represent a set of rules to derive a specific URL */
+class UrlRule {
+  /**
+   * Create a set of metadata to construct and validate URLs, throwing early errors to help developers in
+   * preventing bugs.
+   *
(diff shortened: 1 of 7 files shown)


client/src/utils/url/Url.js
```diff
@@ -0,0 +1,337 @@
...
+        search.append(key, value);
+      return `${url}?${search.toString()}`;
+    };
+  }
+}
+
+const projectsConfig = new ProjectsUrlConfig();
+
+/** Helper class to handle URLs */
+class Url {
```
ciyer, line 183: There are two things I don't like about this structure:
1. The implementation for projects, for example, is split across `_validations`, `_outputs`, and `rules`.
2. A class with only static methods does not make much sense in JS -- I think an Object would fit better.

I added a commit that is an attempt to address 1. for projects. If you like this direction, could you apply it to project as well?

ciyer (changes requested):
I like these changes! I think the Url API is nice, but I have a suggestion for changing the internal implementation (this doesn't affect clients -- the tests pass without modification)

Re: commit messages - I would make the commit message for both commits "refactor" -- I think we should reserve "feat" for user-facing features.

lorenzo-cavazzi:
As discussed offline, we decided to simplify the architecture by:
1. Completely removing `_validations`, `_outputs`, and `rules`. Any reusable part was moved to a function.
2. Converting the `Url` class with static methods/fields to an object.

ciyer (approved):
Very nice!

PR #1209, opened 2021-01-21
ci: update renku release name

lorenzo-cavazzi:
Include the `renku` string in the release name to prevent malfunctioning components.

/deploy


ciyer (approved):
👍

PR #1211, opened 2021-01-21
ci: improve robustness to missing privacy values

### LINKED ISSUE #1045: make banner setup more robust to missing values
Currently, a missing value in `ui.privacy.banner.content` would make the whole UI unusable.
This should be more robust since once `ui.privacy.enabled` is set to `false` there shouldn't be any need for the rest of the privacy values to be set correctly.

It would be good to
* [ ] set a default values to `PRIVACY_BANNER_LAYOUT` in "docker-entrypoint.sh" so that we will never have `"PRIVACY_BANNER_LAYOUT": ,` in the "config.json" file.
* [ ] change the default values in "helm-chart/renku-ui/templates/deployment.yaml" ? It's probably not needed but it would make the code more elegant.

lorenzo-cavazzi:
This PR makes the `PRIVACY_BANNER_CONTENT` and `PRIVACY_BANNER_LAYOUT` env variables robust to missing values when the privacy `enabled` flag is `false`.

Tbh it was unlikely to mess up since default are assigned in the local values file, but it was tecnically possible. Considering how env variable values are consumed in "docker-entrypoint.sh", en empty `PRIVACY_BANNER_CONTENT` would have crashed the UI because of an invalid `config.json` file
> Uncaught (in promise) SyntaxError: JSON.parse: unexpected character at line 11 column 28 of the JSON data

/deploy
fix #1045

### DIFF
--- a/helm-chart/renku-ui/templates/deployment.yaml
+++ b/helm-chart/renku-ui/templates/deployment.yaml
@@ -66,10 +66,17 @@ spec:
               value: {{ .Values.global.anonymousSessions.enabled | default (printf "false") | quote }}
             - name: PRIVACY_ENABLED
               value: {{ .Values.privacy.enabled | quote }}
+            {{- if .Values.privacy.enabled }}
             - name: PRIVACY_BANNER_CONTENT
               value: {{ .Values.privacy.banner.content | default (printf "") | b64enc | quote }}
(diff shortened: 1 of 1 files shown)


ciyer (approved):
👍

ableuler (approved):

PR #1212, opened 2021-01-22
fix: adjust parameter value on repository tree api

lorenzo-cavazzi:
The highest `per_page` value supported by GitLab is 100, any value higher than that doesn't take effect.
Even if the current setting doesn't create relevant problems, it was confusing to other devs trying to debug unexpected behaviors.


ciyer (approved):
👍

PR #1214, opened 2021-01-25
Expand sentry configuration and prevent sending notebook errors

### LINKED ISSUE #1112: Don't log errors coming from the user's notebook preview
We shouldn't track errors coming from the user's notebook preview triggered when browsing files.

We may consider giving feedback to the user through notifications (not if that's a good idea)

Sentry Issue: [RENKU-UI-6B](https://sentry.dev.renku.ch/organizations/swiss-datascience-center/issues/736/?referrer=github_integration)

```
ReferenceError: require is not defined
  at None (/projects/tao.sun/preparation_dsip2020/files/blob/notebooks/co2_viz.ipynb%20line%202%20%3E%20injectedScript:1:17)
  at value/< (file/File.container.js:281:16)
...
(11 additional frame(s) were not displayed)
```

lorenzo-cavazzi:
* Move Sentry configuration to a separate object.
* Add more information to exceptions sent to Sentry, including the UI version.
* Add a mechanism to filter events before sending them.
  This has been used to prevent sending notebook preview related exceptions ( fix #1112 ).

**How to test**: try to preview a notebook that triggers an exception. Compare `dev` (exception logged in Sentry) to the auto deployed environment (no exceptions logged).
The following link provides a good test case; the exception is generated after clicking on "preview it anyway" https://dev.renku.ch/projects/cramakri/plotly-test/files/blob/notebooks/Plotly.ipynb

/deploy

### DIFF
--- a/helm-chart/renku-ui/templates/deployment.yaml
+++ b/helm-chart/renku-ui/templates/deployment.yaml
@@ -81,6 +81,8 @@ spec:
               value: {{ toJson .Values.templates | quote }}
             - name: PREVIEW_THRESHOLD
               value: {{ toJson .Values.previewSizeThreshold | quote }}
+            - name: UI_VERSION
+              value: {{ .Chart.Version | quote }}
           livenessProbe:
             httpGet:
               path: /
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -157,7 +157,7 @@
           "starrers",
           "statuspage",
           "stdout",
-          "textarea",
+          "telepresence",
           "textarea",
           "thead",
           "toastify",
--- a/client/docker-entrypoint.sh
+++ b/client/docker-entrypoint.sh
@@ -20,6 +20,7 @@ export NGINX_PATH=/usr/share/nginx/html
 
 echo "Config file contains the following settings:"
 echo "==================================================="
+echo " UI_VERSION=${UI_VERSION}"
 echo " GATEWAY_URL=${GATEWAY_URL:-http://gateway.renku.build}"
 echo " BASE_URL=${BASE_URL:-http://renku.build}"
 echo " SENTRY_URL=${SENTRY_URL}"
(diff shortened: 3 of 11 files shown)


ciyer (dismissed):
Code looks good. I checked tried out the test scenario in the description, and it behaves as expected. I also clicked around the UI a bit and there as well, everything worked as expected. 👍

helm-chart/renku-ui/templates/deployment.yaml
```diff
@@ -81,6 +81,8 @@ spec:
               value: {{ toJson .Values.templates | quote }}
             - name: PREVIEW_THRESHOLD
               value: {{ toJson .Values.previewSizeThreshold | quote }}
+            - name: UI_VERSION
```
rokroskar, line 84: you should consider using sentry releases maybe to automate this?

https://docs.sentry.io/product/releases/

rokroskar (approved):

helm-chart/renku-ui/templates/deployment.yaml
lorenzo-cavazzi: I thought about that, but I'm including the last commit short SHA and a custom prefix, resulting in a great variety of ui-versions (e.g. `telepresence@0.11.4-c67565e`).
On the other hand, it may be better to drop the details and log the whole version number (e.g. `0.11.1`) with a `dev` postfix if it's a later version (e.g. `0.11.1-dev` for 0.11.1-1ab2c435

rokroskar: yes - it's not obvious to me if the sentry releases work only with tagged releases or would it pick up individual commits as well.

lorenzo-cavazzi: Yeah, not sure either, but I guess it would somehow consider the commits https://github.com/marketplace/actions/sentry-release
I have changed that part of the code so that we set the current release client side. For versions with commits, I strip the last part and use `-dev` instead to avoid an unneeded level of details (and simplify filtering).
I could try to setup the GitLab integration by creating an action as described in the link above (but only on tags). I'm not about the benefits though. And I'd rather tackle that in a separate issue to move on with this one and #1215.

rokroskar: yes I wasn't suggesting you do it now :)

rokroskar (approved):

ciyer (approved):
👍

Issue #1165, opened 2020-12-07
Trigger a new deployment on PRs and run acceptance tests


lorenzo-cavazzi:
To improve our current workflow, we want to:
- [x] Create a new deployment for every new PR and keep it updated on changes.
- [x] Run the acceptance tests on that deployment.
- [ ] Allow the developer to specify custom versions for specific components.

This way, every PR will have a preview without requiring time-consuming tasks from the developer, and the acceptance tests will be run before anything is merged to master.
The UI sometimes needs specific WIP components from other repos. If we could pin them in a way easily readable from the GitHub action, it would be great.

lorenzo-cavazzi:
This was done a while ago

PR #1217, opened 2021-01-28
ci: run acceptance tests after deploy

lorenzo-cavazzi:
Run the acceptance tests when deploying a dedicated environment.

/deploy
re #1165


ciyer (approved):
Cool!

PR #1229, opened 2021-02-04
fix: refactor the project list components

### LINKED ISSUE #1181: Move ProjectListModel to projectsCoordinator
"/project/list/ProjectList.state.js" defines a schema used in a custom redux store. This should be moved to the global redux store, while the functions should go to the already existing projects coordinator  "/project/shared/Projects.state.js" (move all the pieces in a single folder).

It may be a good idea to verify if the logic in the `List` component could be simplified and even moved to a common component to be reused for similar scenarios (e.g. DatasetList).

- [ ] Move schemas.
- [ ] Merge with projects coordinator.
- [ ] Simplify List component
- [ ] _[optional] create a shared component_

### LINKED ISSUE #1189: Search is not preserved when changing tab
## Description

When the user searches on the projects page and he moves from one category to another ("Your projects", "Starred projects", "All projects"), the search is cleared. It would be more natural to preserve the search.

We could also add a few quick actions to prevent confusion (e.g. when the user switches to "Your projects" but he doesn't realize a previous search has been preserved and it's limiting the results)

## Mockup

Interactive preview (click on the highlighted backgrounds to interact)
https://app.moqups.com/tTKVjKfdQj/view/page/a93e72279

## Context

It would be better to address this issue after (or together with) #1181

lorenzo-cavazzi:
* Removes the local redux store from the projects list pages and use react functional components instead (fix #1181)
* Preserves the search string when changing tab (fix #1189) and when clicking any button on the page that triggers a new search.
* Prevent illegal actions by automatically redirecting to a valid search page. E.G. anonymous users can't search in their starred projects; Authenticated users can't search for users in their own projects.

This PR simplifies other changes as well. An interesting one would be allowing the user to select the number of projects shown in one page. A dropdown menu with values like `10, 20, 50, 100` would do, and that preference could be easily saved in the local storage.

/deploy


lorenzo-cavazzi:
Sorry for force-pushing a few times in a row, I tried different solutions to prevent `npm` tests from failing in GitHub actions when re-defining the window.location object.
It works fine now.

client/src/project/list/ProjectList.container.js
```diff
@@ -16,293 +16,371 @@
...
-        return "name";
-      case orderByValuesMap.CREATIONDATE:
-        return "creation";
-      case orderByValuesMap.UPDATEDDATE:
-        return "updated";
-      default:
-        return "";
-    }
-  }
+  // Initial setup
```
ciyer, line 169: The logic in this component is a bit hard to follow. I think writing custom hooks would help make it clearer.

client/src/project/list/ProjectList.present.js
```diff
@@ -16,426 +16,336 @@
...
-        <UncontrolledTooltip key="tooltip" placement="top" target="searchButton">
-          {tip}
-        </UncontrolledTooltip>
-      );
-    }
+function ProjectListRows(props) {
+  const { currentPage, getAvatar, perPage, projects, search, totalItems } = props;
+
+  if (!projects || !projects.length)
+    return (<p>We couldn&apos;t find any project matching the search criteria.</p>);
```
ciyer, line 83: I think 
```
<p>We could not find any matching projects.</p>
```

Would be better.

client/src/project/list/ProjectList.present.js
```diff
@@ -16,426 +16,336 @@
...
+      <Pagination currentPage={currentPage} perPage={perPage} totalItems={totalItems} onPageChange={onPageChange} />
+    </div>
+  );
+}
+
+function ProjectListSearch(props) {
+  const { orderByMap, params, search, searchInMap, sectionsMap } = props;
+
+  // input and search
+  const [userInput, setUserInput] = useState(params.query.toString());
```
ciyer, line 100: Here also, I think custom hooks would help get the state-management logic out of this function to make it easier to understand the logic here.

client/src/project/list/ProjectList.present.js
```diff
@@ -16,426 +16,336 @@
...
-          active={Number(this.props.selectedUserOrGroup) === Number(p.id)}>{p.username}
-        </Button>;
+  let content = null;
+  // don't show anything if users are updating, since the content would be outdated
+  if (!users.fetching) {
+    if (fetching) {
+      content = (<Loader />);
+    }
+    else if (!fetched) {
+      content = (<p>Please enter you filtering words in the field above and click on Search to get the result.</p>);
```
ciyer, line 248:
```
      content = (<p>Please enter text in the field above and click on "Search" to see the results.</p>);
```

But I wasn't able to get this message to appear.

ciyer (changes requested):
I think the code improvements here are very nice, and it's great to see the logic using more modern React features.

I have reviewed about 1/2 the PR (it's very big!) and wanted to get these comments up.

I found two bugs in the logic:

- It should be possible to search for all projects by leaving the input empty, but that does not work.

image

- In the same page as above, it is not possible to return to "Filter by Project" after changing to one of the other options.

client/src/project/Project.test.js
```diff
@@ -102,33 +102,6 @@ describe("rendering", () => {
   const loggedUser = generateFakeUser();
   const model = new StateModel(globalSchema);
 
-  it("renders list without crashing for anonymous user", () => {
```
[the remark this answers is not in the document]
lorenzo-cavazzi: That is now included in the tests in `ProjectList.test.js `

client/src/project/list/ProjectList.present.js
lorenzo-cavazzi: Theoretically, this text should never appear since entering the page triggers a fetch, therefore it would be `fetching = true`. When that is set to `false` again because the action finished, then `fetched != null`.
Technically, it's still possible that the condition is satisfied for a brief instant. We can either get rid of it and return `null` or leave it there just in case (and fix the text :grin: )

lorenzo-cavazzi:
>     * It should be possible to search for all projects by leaving the input empty, but that does not work.
It's possible when searching projects, but it's not when looking for users or groups. Is that wrong?

>     * In the same page as above, it is not possible to return to "Filter by Project" after changing to one of the other options.
Oh, that is definitely a bug! I'll fix that soon

ciyer:
> > ```
> > * It should be possible to search for all projects by leaving the input empty, but that does not work.
> > ```
> 
> It's possible when searching projects, but it's not when looking for users or groups. Is that wrong?

It does not work for me on projects either:

image


But on dev, I can leave the field empty and get all results also on users or groups.

lorenzo-cavazzi:
The tests with the up-to-date acceptance tests passed here: https://github.com/SwissDataScienceCenter/renku-ui/runs/1856588114
I've now removed the explicit reference to the `ui1229-fix-acceptance` branch when deploying the test environment. After merging this PR, we should tag a UI version and merge SwissDataScienceCenter/renku#1937 to correctly update the acceptance tests for the other PRs in the UI and the other renku repositories

ciyer (approved):
Looks good! 🎉

PR #1235, opened 2021-02-10
fix: prevent accessing stats for non existing projects

### LINKED ISSUE #1234: Avoid setting stats when the project doesn't exist
The project coordinator tries to access the project's statistics even when the project doesn't exist. We should fix by moving that part inside the `if(data)` to prevent triggering errors.

Sentry Issue: [RENKU-UI-AT](https://sentry.dev.renku.ch/organizations/swiss-datascience-center/issues/4977/?referrer=github_integration)

```
TypeError: e is undefined
  at setProjectData (project/Project.state.js:444:21)
  at e/</< (project/Project.js:215:31)
```

lorenzo-cavazzi:
The error was not affecting the usability of the UI.
It happened when trying to access a project that either doesn't exist or requires privileges.
Sentry caught that, it's good to fix it.

**How to test**: when the deployment is ready, compare accessing the following with the console open. The first should trigger an error, the second doesn't.
https://dev.renku.ch/projects/lorenzo.cavazzi.tech/idontexists
https://renku-ci-ui-1235.dev.renku.ch/projects/lorenzo.cavazzi.tech/idontexists

fix #1234
/deploy

### DIFF
--- a/client/src/project/Project.state.js
+++ b/client/src/project/Project.state.js
@@ -408,7 +408,7 @@ class ProjectCoordinator {
   }
 
   setProjectData(data, statistics = false) {
-    let metadata;
+    let metadata, statsObject;
 
     // set metadata
     if (!data) {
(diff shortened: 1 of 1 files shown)


ciyer (approved):
Nice!

PR #1237, opened 2021-02-11
fix: mardown breaks when it has a file inside

vfried:
When markdown has a file inside it breaks because of the new threshold variable.

I set this for markdown. Another option would be to use a default threshold variable for all file previews instead.

Example:
https://dev.renku.ch/projects/virginiafriedrich/project-11

/deploy

### DIFF
--- a/client/src/utils/Markdown.js
+++ b/client/src/utils/Markdown.js
@@ -82,7 +82,6 @@ function FileAndWrapper(props) {
         <input type="checkbox" id={togglerId} className="visually-hidden fake-toggle" />
         <div className="hide-show-me">
           <FilePreview
-            previewThreshold={{ soft: 1048576, hard: 10485760 }}
             file={props.file}
             {...props}
             springConfig={props.springConfig}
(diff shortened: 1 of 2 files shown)


client/src/utils/Markdown.js
```diff
@@ -82,6 +82,7 @@ function FileAndWrapper(props) {
         <input type="checkbox" id={togglerId} className="visually-hidden fake-toggle" />
         <div className="hide-show-me">
           <FilePreview
+            previewThreshold={{ soft: 1048576, hard: 10485760 }}
```
lorenzo-cavazzi, line 85: I would use the variable `params.PREVIEW_THRESHOLD` instead of 2 fixed values.
I'm not sure if this would solve the issue though. What happens if the inside files are over the threshold?

lorenzo-cavazzi (changes requested):

client/src/utils/Markdown.js
vfried: I just pushed a new fix for this, do you think it would be better?

lorenzo-cavazzi: Yes, that's a good solution :+1:

lorenzo-cavazzi (approved):
:+1:

ciyer (approved):
👍

PR #1238, opened 2021-02-11
fix: url for manual migration instructions changed

### LINKED ISSUE #1213: Change URL for manual migration
The URL for manual migration info is wrong, it should be this one... https://renku.readthedocs.io/en/latest/user/upgrading_renku.html#upgrading-your-image-to-use-the-latest-renku-cli-version

The content inside that URL should be updated since it's outdated.

vfried:
The URL for the migration instructions was wrong, the right URL is in place now.

Closes #1213 

/deploy


ciyer (approved):
👍 Thanks!

PR #1239, opened 2021-02-11
refactor: add an id to every projects link

lorenzo-cavazzi:
This PR adds an `id` to the links to "Projects" pages.
It's useful to make our acceptance tests more robust and it doesn't introduce any change visible to users.

P.S. feel free to suggest a different naming.


ciyer (approved):
👍 Great idea!

PR #1240, opened 2021-02-11
fix: datasets ordered by dates created

### LINKED ISSUE #1226: Order by date_published should be changed to order by date_created for datasets
## Description

We are sorting the search of datasets by date_published and we should actually order the datasets by date_created.

date_published refers to the date the dataset was published in dataverse/zenodo/etc and datasets created in renku don't have this date.

date_created is what we want to have instead.

vfried:
There was a bug in dataset search that was not ordering datasets per imported date.

I revised the way we handle dates in datasets to display it in a more accurate way.

**Datasets Search**
We get from the KG "date" we check if the dataset was published or not (we do this by checking if it has a datePublished or not) we display "Published date" or "Created date" and the date.

**Dataset Display**
For datasets inside projects:
- Datasets created in renku: Inside projects we get created_at from the core service we map this to created
- Imported datasets: we get datePublished from the KG and we use this since we get nothing from core

For datasets inside the kg (display from search)
- Datasets created in renku have created
- Imported datasets have published date

For datasets that are not in the KG(only inside projects):
We only get a date_created for datasets created in the KG so we display this (we map it to created)

Closes #1226 

/deploy

### DIFF
--- a/client/src/dataset/Dataset.present.js
+++ b/client/src/dataset/Dataset.present.js
@@ -130,12 +130,14 @@ export default function DatasetView(props) {
   const datasetPublished = dataset.published !== undefined && dataset.published.datePublished
     !== undefined && dataset.published.datePublished !== null;
 
+  let dateCreated = datasetPublished ? dataset.published.datePublished : dataset.created;
+
   return <Col>
     <Row>
       <Col md={8} sm={12}>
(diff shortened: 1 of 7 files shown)


lorenzo-cavazzi (approved):
Works as advertised :+1:

PR #1242, opened 2021-02-12
fix: test lfs status properly on the preview

lorenzo-cavazzi:
Some non-lfs files were mistakenly considered as lfs, preventing the preview.

Compare:
- https://dev.renku.ch/projects/lorenzo.cavazzi.tech/csci-e-88b-2021-problem1/files/blob/notebooks/git_lfs.ipynb
- https://renku-ci-ui-1242.dev.renku.ch/projects/lorenzo.cavazzi.tech/csci-e-88b-2021-problem1/files/blob/notebooks/git_lfs.ipynb

/deploy


vfried (approved):
This solution is much cleaner!

PR #1247, opened 2021-02-17
feat: file upload notification for datasets

### LINKED ISSUE #1150: File upload notification
## Motivation

Uploading files can take some time, and I would like to be able to do other things instead of watching a progress bar.

## Proposal

Use the notification infrastructure to post notifications when 1. if a file upload has an error 2. When *all* queued uploads have completed.

The notifications should only be shown if the user isn't on the file upload page.

image

image

vfried:
When all files finished uploading on a dataset add/edit

image

When a file failed

image

/deploy

Closes #1150


client/src/project/datasets/change/DatasetChange.container.js
```diff
@@ -69,33 +69,33 @@ function ChangeDataset(props) {
...
-    // }
+
+    if (!fileField.notifyFunction) {
+      fileField.notifyFunction = (success = true, error) => {
+        const datasetName = props.edit ? "dataset " + dataset.name : "new dataset";
+        const redirectUrl = props.edit ? `/projects/${props.projectPathWithNamespace}/datasets/${dataset.name}/modify`
+          : `/projects/${props.projectPathWithNamespace}/datasets/new`;
+        if (success) {
+          props.notifications.addSuccess(props.notifications.Topics.DATASET_FILES_UPLOADED,
+            `Files for the ${datasetName} in ${props.projectPathWithNamespace} finished uploading.`,
```
lorenzo-cavazzi, line 80: Should we make it more explicit that it's necessary to carry on with the operation?   Maybe something like `Files uploaded, you can finalize the dataset`. I would omit the project name

lorenzo-cavazzi (changes requested):
All good! Just a note on the text for the successful operation.

For the user, it's not clear he can move away from the page while uploading. I was tempted to suggest adding such a message, but probably that would just clog the UI even more. We could consider this later if we think it's useful.

ciyer:
I suggest the following:
- If the file is greater than some configurable threshold (default 50 MB), inform the user that they will be notified when the upload finishes.
image

client/src/project/datasets/change/DatasetChange.container.js
vfried: @ciyer what do you think about this. A more direct message could be ok for notifications but I think it would be out of tone with he rest of the UI messages.

Regarding the project name, I think it should be there because you could be creating two datasets for different projects at the same time.

The message so that the user knows they will receive a notification will be pushed soon.

vfried:
I just pushed new changes, if the file is bigger than 100MB they will see the message. Otherwise, they won't.

image

client/run-telepresence.sh
```diff
@@ -28,6 +28,7 @@ TEMPLATES='{"custom":true,"repositories":
 {"name":"Telepresence","ref":"0.1.11",
 "url":"https://github.com/SwissDataScienceCenter/renku-project-template"}]}'
 PREVIEW_THRESHOLD='{"soft":"1048576","hard":"10485760"}'
+UPLOAD_THRESHOLD='{"hard":"104857600"}'
```
ciyer, line 31: Good idea to reuse the existing format/structure!

ciyer (approved):
This works very well! I tested uploading files to two different datasets at the same time, and that worked, the notifications were easy to follow. 

Ready to merge.

🎉

helm-chart/renku-ui/values.yaml
```diff
@@ -71,6 +71,11 @@ previewSizeThreshold:
   soft: 1048576 # 1MB
   hard: 10485760 # 10MB
 
+# This defines the threshold for displaying a message when a user is trying to upload a large file
+# so that they know it's not necessary to wait for this. There is only hard limit for now.
+uploadSizeThreshold:
+  hard: 104857600 # 100MB
```
lorenzo-cavazzi, line 77: We could simplify this structure to:
```
uploadSizeThreshold: 104857600 # 100MB
```
So that we have a string here instead of a JSON, making it more robust in the following steps (e.g. we won't need to invoke the `toJson` function in the "deployment.yaml" file)

lorenzo-cavazzi (commented):
It works well :tada: 
I suggest a minor change to simplify the "upload threshold" structure before merging -- see the comment inline

helm-chart/renku-ui/values.yaml
lorenzo-cavazzi: As discussed offline, let's keep this structure and change `hard` to `soft` to make it more meaningful.
We will use `hard` when implementing #1251

lorenzo-cavazzi (approved):
:+1:

PR #1250, opened 2021-02-19
fix: markdown with incomplete links in comments

vfried:
If markdown contains a line like this it breaks leaving the project not usable:

`# <a name="litterature-review"></a> Litterature review`

This pr fixes that.

@erbou @pameladelgado 

/deploy


lorenzo-cavazzi (approved):
:+1:

PR #1252, opened 2021-02-22
Update fork UX to make it more similar to the project creation

### LINKED ISSUE #1118: Adapt the fork UI/UX to be similar to new project creation
With the changes introduced in #1115 , we should update the "fork project" component to provide similar feedback to the real-time validation we use on the "new project" page.
Consider that the current errors come from GitLab and something may be slightly different in RenkuLab (e.g. no emojis allowed in the title)

We should include
* [x] Same rules/errors for deriving the project slug
* [x] Check for duplicate slug

EDIT: we should also adjust the Fork components to
* [x] remove the duplicate content (there should be a lot in common with the new project components)
* [x] remove the local redux store

# Implementation

We should try to extract one or more shared components because that will make the coming UI aesthetic redesign easier to implement.

lorenzo-cavazzi:
I have reused quite a few components from "new project".
I needed to create:
* A mapper component to map the redux content (we can remove it later and use the `useSelector` hook)
* A container component containing the hooks and the `fork` function. Even if `fork` could be part of the state component, we don't really need to store much in the redux store. Including it in the container component makes the whole logic easier to understand.

I moved the redirect logic from the `client` functions to the react components and added notifications.

We can improve the error handling if necessary -- most of the errors are now prevented by the realtime validation, so I wouldn't focus much on that unless we notice specific errors happening frequently.

/deploy
fix #1118


client/src/project/new/ProjectNew.container.js
```diff
@@ -23,17 +23,255 @@
...
+    };
+  }
+
+  componentDidMount() {
+    // fetch if not yet available and refresh if older than 10 seconds
+    const tolerance = 10 * 1000;
+    const currentState = this.model.get("projects");
+    const now = new Date();
+    if (!currentState.namespaces.fetching) {
+      if (!currentState.namespaces.fetched || now - currentState.namespaces.fetched > tolerance)
```
ciyer, line 94: I think this is a nice solution, and I could see it being used in other places as well, so we can make a function out of this to reuse it.

```
function refreshIfNecessary(tolerance, property, action, now = new Date()) {
  if (property.fetching) return;
  const lastFetch = property.fetched;
  if (!lastFetch || now - lastFetched > tolerance) action()
}
```

And then use it like:
```
refreshIfNecessary(tolerance, currentState.featured, () => this.getProjects(), now);
```

client/src/project/new/ProjectNew.present.js
```diff
@@ -60,6 +61,87 @@ function makeRefreshButton(refresh, tip, disabled) {
...
+    "btn " + btnClass :
+    "btn-primary";
+  return (<button className={className} onClick={() => { toggle(); }}>{content}</button>);
+}
+
+function ForkProject(props) {
+  const { btnClass, btnContent, error, fork, forkError, forkedTitle, forking } = props;
+
+  const [modalOpen, setModalOpen] = useState(false);
+  const toggle = () => setModalOpen(!modalOpen);
```
ciyer, line 80: There seems to be a problem here. The first time I click the fork button after a page load, it flashes open, then closes. Subsequent clicks work, but the first one does not.

This does not happen in develop mode as often, but happens reliably in the production build. This may have something to do with the problems with tests as well.

client/src/project/new/ProjectNew.container.js
```diff
@@ -23,17 +23,255 @@
...
+    );
+  }
+}
+
+
+function ForkProject(props) {
+  const { btnClass, btnContent, forkedTitle, handlers, namespaces } = props;
+
+  const [title, setTitle] = useState(forkedTitle + " - copy");
+  const [namespace, setNamespace] = useState(""); // TODO - pick the default namespace if available
```
ciyer, line 147: This is a minor issue, I think we can merge even if it cannot be fixed, but I noticed that if I do the following:
- click fork
- change the title
- change the namespace
- click cancel
- click fork again

The changed title is kept, but the namespace is reset to the default namespace. Not a huge problem, but if it is easy to fix, then it would be nice.

client/src/project/new/ProjectNew.container.js
```diff
@@ -23,17 +23,255 @@
...
+
+      // add notification and redirect
+      let newProjectData = { namespace: forked.project.namespace.full_path, path: forked.project.path };
+      const newUrl = Url.get(Url.pages.project, newProjectData);
+      newProjectData.name = forked.project.name;
+      addForkNotification(notifications, newUrl, newProjectData, startingLocation, true);
+      //client.runPipeline(forked.project.id);
+
+      // automatically redirect only when the user hasn't changed location
+      if (mounted && history.location.pathname === startingLocation)
```
ciyer, line 226: This works nicely!

I found a minor issue with the UX here which is that if I close the popup, but stay on the same project page, then I still get redirected. I sort of expected the notification only if I was still looking at the pop-up, but I think the current behavior is fine.

ciyer (changes requested):
Overall works well, and code looks good. I have one issue that needs to be fixed -- the fork button doesn't work on the first invocation after a page load. The other points I raised are all optional, I would say.

client/src/project/new/ProjectNew.present.js
lorenzo-cavazzi: That's a good point and it may very well be the cause of the failing tests.
After some experiments, it looks like the `mapStateToProps` is re-triggered when we get responses from APIs with irrelevant content for the fork.
The cause is in the mapping function in one of the ancestor components, mapping the local project redux store to the global project view. This means we can't really find an elegant solution until the Project redux store is united with the global one.

A workaround could be storing the modal state in the mapper component `ForkProjectMapper`, but I fear we may need to store all the relevant data in the redux store to prevent losing the state.

client/src/project/new/ProjectNew.container.js
lorenzo-cavazzi: This makes sense.
I implemented a slightly different function `refreshIfNecessary(fetching, fetched, action, tolerance = 10)` with separate `fetching` and `fetched` since we may need to support legacy code or play with the fetching status when it requires other actions to be finished as well.

client/src/project/new/ProjectNew.container.js
lorenzo-cavazzi: Currently, the state is volatile and it should reset every time the modal is closed. Are you sure the title is preserved? That seems strange

client/src/project/new/ProjectNew.container.js
lorenzo-cavazzi: I agree the expected behavior is the one you described.
The notifications work with the route location, but we can easily check if the modal is still open and force showing the notification whenever that is not the case.

client/src/project/new/ProjectNew.container.js
ciyer: This is not happening anymore. Not sure what changed, but it looks good now.

client/src/project/new/ProjectNew.present.js
lorenzo-cavazzi: As discussed offline, the problem can't be solved in a very elegant way.
I've tried to move the main `Modal` component to the Project.present page to avoid flashing content. This is still not bulletproof (an unrelated project API may still return data when the modal is open and cause a re-draw), but it seems to work well in most cases.

I added a loader while we fetch the necessary data for the validation to prevent circumventing them (it's not to difficult when clicking fast and having a lot of projects).

Screenshot_20210224_190404

lorenzo-cavazzi:
All the points should be fixed, even if moving the Modal to an ancestor component is not a very elegant solution. This makes the Fork components harder to re-use. Adapting them to non-modals is fairly easy if we will ever need to do that.

ciyer (approved):
Changes look good; and everything works very well now! 🎉

lorenzo-cavazzi:
The acceptance tests passed in 5cadb46 . I reverted the change before merging the PR

PR #1253, opened 2021-02-22
fix: use full_path in projects list

### LINKED ISSUE #1248: Link in search for projects in subgroups is broken
## Description

If you search for this project
https://renkulab.io/projects/omnibenchmark/omni_data/cellbench

It appears on the search without the subgroup

https://renkulab.io/projects/all?query=cellbench

And when you access the project the link also fails
https://renkulab.io/projects/omni_data/cellbench

lorenzo-cavazzi:
* Use the `full_path` property for the project links to prevent failure with subgroups.
* Prevent showing the `Filter by` section to anonymous users since they can't change it.

/deploy
fix #1248

### DIFF
--- a/client/src/project/list/ProjectList.container.js
+++ b/client/src/project/list/ProjectList.container.js
@@ -397,7 +397,7 @@ function ProjectList(props) {
       fetching={projects.fetching}
       getAvatar={id => this.client.getAvatarFromNamespace(id)}
       getPreciseUrl={getPreciseUrl}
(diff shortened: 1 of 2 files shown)


ciyer (approved):
Good fix. We should probably have a test for subgroups to prevent bugs like this from being introduced.

I added a commit to change the name of the variable from `logged` to `loggedIn`.

lorenzo-cavazzi:
Do you mean an integration test? I agree that would be very useful.

PR #1254, opened 2021-02-24
feat(bs5): change to bs5 and first fixes

### LINKED ISSUE #1232: Move to Bootstrap 5
## Description

We have decided to start the design refresh with a move to Bootstrap 5

## Tasks

- [x] Upgrade to bootstrap 5 in the ui-design-2021
- [x] See what problems need to be fixed and either fix them or create issues to have them fixed in future sprints

## Background

One question to resolve is how we want to interact with Bootstrap. Bootstrap will be version 5 soon, and it's not clear if Reactstrap will support v5. React-Bootstrap now supports v4 and seems to have more momentum, so that might be an option. Or we can just use bootstrap directly / make our own wrappers.

- https://getbootstrap.com/docs/5.0/getting-started/introduction/
- https://github.com/reactstrap/reactstrap/issues/1748
- https://github.com/react-bootstrap/react-bootstrap/issues/5174

_Originally posted by @ciyer in https://github.com/SwissDataScienceCenter/renku-ui/issues/1161#issuecomment-760935820_

vfried:
Steps:

1. Clone https://github.com/reactstrap/reactstrap/commits/bootstrap5 (this branch is in progress)
2. To test locally go to /reactstrap
2.1  npm install, npm build
2.2 cd /node_modules/react npm link (do the same with react-dom)
3. go to renku-ui and do npm install and npm build
4. then do npm link react npm link react-dom (IN /CLIENT -->NOT INSIDE NODE MODULES)
4. all should be ok

The input group has changed slightly, I still need to fix the search in datasets and in the top bar so that it looks OK.
The switch has changed, I still need to make it work well

Closes #1232 

/deploy


ciyer (approved):
Good start! Great to see this moving and looking forward to the next steps with this.

PR #1256, opened 2021-03-01
refactor: remove environment notification

lorenzo-cavazzi:
The current notification for the environment creation is not very useful because it notifies the user when the creation API invocation is successful, not when the environment is available (i.e. the user still needs to wait before using the environment).
This PR removes it. We can revive that once we will get the information about the ready environment.

P.S. Notice that the notification for a failure is still shown. We don't need to wait for any further information in that case.

/deploy


ciyer (approved):
👍

rokroskar (approved):
Thanks @lorenzo-cavazzi !

PR #1258, opened 2021-03-02
fix: fail gracefully when a dataset is missing

### LINKED ISSUE #1003: Gracefully fail when trying to reach a dataset that doesn't exist
While testing a few dataset features, I noticed that trying to open a dataset in the UI with wrong id results in a blank page with no feedback 
E.G. https://dev.renku.ch/datasets/28067ed1-783a-41e6-9886-20651c0f044eERROR

It would be better to provide a "dataset not found" page, similar to what happens in other similar cases
https://dev.renku.ch/projects/fakename/fakeproject
https://dev.renku.ch/projects/fakegroup

<kbd>
  
</kbd>

lorenzo-cavazzi:
Adds a 404 page when the dataset doesn't exist or the user doesn't have access (currently, the access control is not yet enforced by the APIs).
It's similar to the 404 project page when the user is trying to access `/datasets/<id>`, while it's a simpler message when trying to access datasets in a project page.

**Hot to test**: try to access non-existing datasets as anonymous or logged user
1. PR: https://renku-ci-ui-1258.dev.renku.ch/datasets/fake-id
    DEV: https://dev.renku.ch/datasets/fake-id

2. PR: https://renku-ci-ui-1258.dev.renku.ch/projects/lorenzo.cavazzi.tech/dataset-test-project-copy/datasets/fake-id
    DEV: https://dev.renku.ch/projects/lorenzo.cavazzi.tech/dataset-test-project-copy/datasets/fake-id

Screenshot_20210302_173048

/deploy
fix #1003

### DIFF
--- a/client/src/dataset/Dataset.container.js
+++ b/client/src/dataset/Dataset.container.js
@@ -46,7 +46,7 @@ export default function ShowDataset(props) {
                 .map(file => ({ name: file.name, atLocation: file.path })));
             }
             else {
-              setDatasetFiles(response.data);
+              setDatasetFiles([]);
               if (response.data && response.data.error) {
                 if (response.data.error.code === -32100)
                   setFetchError({ code: 404, message: "dataset not found or missing permissions" });
--- a/client/src/dataset/Dataset.container.js
+++ b/client/src/dataset/Dataset.container.js
@@ -60,8 +60,7 @@ export default function ShowDataset(props) {
     return () => {
       unmounted = true;
     };
-  }, [props.insideProject, datasetFiles, props.datasetId,
-    dataset, props.httpProjectUrl, setDatasetFiles, props.client]);
+  }, [datasetFiles, props.datasetId, dataset, props.httpProjectUrl, setDatasetFiles, props.client]);
 
   useEffect(() => {
     let unmounted = false;
@@ -83,8 +82,7 @@ export default function ShowDataset(props) {
     return () => {
       unmounted = true;
     };
-  }, [props.insideProject, props.datasets_kg, props.datasetId, props.identifier,
-    props.client, datasetKg, dataset, props.graphStatus]);
+  }, [props.insideProject, props.identifier, props.client, datasetKg, dataset, props.graphStatus]);
 
   if (props.insideProject && datasetFiles === undefined)
     return <Loader />;
--- a/client/src/dataset/Dataset.container.js
+++ b/client/src/dataset/Dataset.container.js
@@ -32,49 +32,27 @@ export default function ShowDataset(props) {
       : undefined
     , datasetKg, datasetFiles)
   , [props.datasets, datasetKg, props.datasetId, datasetFiles]);
-  const [fetchError, setFetchError] = useState();
+  const [fetchError, setFetchError] = useState({});
 
   useEffect(()=>{
     let unmounted = false;
-    if (props.insideProject && datasetFiles === undefined && dataset && dataset.name) {
-      props.client.fetchDatasetFilesFromCoreService(dataset.name, props.httpProjectUrl)
-        .then(response =>{
+    if (datasetFiles === undefined && ((dataset && dataset.name) || (props.datasetId !== undefined))) {
+      const name = (dataset && dataset.name) ? dataset.name : props.datasetId;
+      props.client.fetchDatasetFilesFromCoreService(name, props.httpProjectUrl)
+        .then(response => {
           if (!unmounted && datasetFiles === undefined) {
             if (response.data.result) {
               setDatasetFiles(response.data.result.files
                 .map(file => ({ name: file.name, atLocation: file.path })));
             }
-            else { setDatasetFiles(response.data); }
-          }
-        }
-        );
-    }
-    return () => {
-      unmounted = true;
-    };
-  }, [props.insideProject, datasetFiles,
-    dataset, props.httpProjectUrl, setDatasetFiles, props.client]);
-
-  useEffect(() => {
-    let unmounted = false;
-    if (datasetKg === undefined && ((props.insideProject && dataset.identifier && props.graphStatus)
-    || (props.identifier !== undefined))) {
-      const id = props.insideProject ? dataset.identifier : props.identifier;
-      props.client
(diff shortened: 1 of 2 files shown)


vfried (commented):
Hey, things look good to me... i tested it and it looks like it works...

i just have a small question/resquest

client/src/dataset/Dataset.container.js
```diff
@@ -32,58 +32,59 @@ export default function ShowDataset(props) {
...
                 .map(file => ({ name: file.name, atLocation: file.path })));
             }
-            else { setDatasetFiles(response.data); }
+            else {
+              setDatasetFiles(response.data);
+              if (response.data && response.data.error) {
+                if (response.data.error.code === -32100)
+                  setFetchError({ code: 404, message: "dataset not found or missing permissions" });
+                else
+                  setFetchError({ code: 0, message: "error fetching dataset: " + response.data.error.reason });
```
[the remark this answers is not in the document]
lorenzo-cavazzi: Right, I guess the "dataset not found" case is covered by the `-32100` error code, so any other error should be about files. I've changed the text to `error fetching dataset files` to make it explicit. This is the new output in case the dataset exists and fetching the files triggers an error.

Screenshot_20210304_145359

client/src/dataset/Dataset.container.js
```diff
@@ -32,58 +32,59 @@ export default function ShowDataset(props) {
...
+      props.client.fetchDatasetFilesFromCoreService(name, props.httpProjectUrl)
+        .then(response => {
           if (!unmounted && datasetFiles === undefined) {
             if (response.data.result) {
               setDatasetFiles(response.data.result.files
                 .map(file => ({ name: file.name, atLocation: file.path })));
             }
-            else { setDatasetFiles(response.data); }
+            else {
+              setDatasetFiles(response.data);
```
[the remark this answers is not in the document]
lorenzo-cavazzi: I guess this is setting the content to the error we are getting in the response. I agree it doesn't make much sense, and it's more clear if we set it to an empty array, even if it has no impact. I've changed it

client/src/dataset/Dataset.present.js
```diff
@@ -107,26 +112,115 @@ function LinkToExternal(props) {
...
+
+  // inside project case
+  if (insideProject) {
+    const title = `Error ${fetchError.code ? fetchError.code : "unknown"}`;
+    let errorDetails = null;
+    if (fetchError.code === 404) {
+      errorDetails = (
+        <Fragment>
+          <p>
+            We could not find the dataset. It is possible is has been deleted by its owner or you don&apos;t have
```
ciyer, line 140: I believe this message is incorrect. If the user can access the page of the project containing the dataset (which must be the case to be here), then they must also have access to the dataset.

client/src/dataset/Dataset.present.js
```diff
@@ -107,26 +112,115 @@ function LinkToExternal(props) {
...
+            permission to access it.
+          </p>
+        </Fragment>
+      );
+    }
+    else if (fetchError.message) {
+      errorDetails = (<p>Error details: {fetchError.message}</p>);
+    }
+    const tip = logged ?
+      (<p className="mb-0">You can try to select again a dataset ferom the list in the previous page.</p>) :
```
ciyer, line 150: `You can try to select a dataset again from the list in the previous page.`

client/src/dataset/Dataset.present.js
```diff
@@ -107,26 +112,115 @@ function LinkToExternal(props) {
...
+          </ul>
+        </InfoAlert>
+      ) :
+      (<InfoAlert timeout={0}>{loginHelper}</InfoAlert>);
+    errorDetails = (
+      <div>
+        <h3>Dataset not found <FontAwesomeIcon icon={faSearch} flip="horizontal" /></h3>
+        <div>&nbsp;</div>
+        <p>
+          It is possible that the dataset has been deleted by its owner or you don&apos;t have permission
```
ciyer, line 186: `...or you do not have permission...` would be better.

ciyer (changes requested):
Looks good, but why do we have different styles of error messages in these two cases:

image

vs.

image

Could these two errors not be made to look more similar?

client/src/dataset/Dataset.container.js
```diff
@@ -32,58 +32,59 @@ export default function ShowDataset(props) {
...
           }
-        }
-        );
+        });
     }
     return () => {
       unmounted = true;
     };
-  }, [props.insideProject, datasetFiles,
+  }, [props.insideProject, datasetFiles, props.datasetId,
```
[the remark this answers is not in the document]
lorenzo-cavazzi: Good point. I was able to remove a couple of dependencies also in the other `useEffect` hook

client/src/dataset/Dataset.present.js
lorenzo-cavazzi: This makes sense since it's on the project page and, even if the user has access only to the dataset but not to the project (not sure if it's possible with the future authorization on datasets), this page won't appear.

lorenzo-cavazzi:
I agree the two pages could be more similar, but I'm not sure if showing the bigger `404` on a project's sub-page would be good. Also, the message should be simplified because the permissions don't play a role here (currently, if the user can access the project, he can also see the datasets -- no permissions would result in the project 404 page).

Let me know if you have an idea to find common ground to unify the two dataset error pages. I believe it would be easier to unify the 404 on datasets outside the project and the 404 on missing projects. Maybe that would be easier once the search will be global (we could point the user to the search page and automatically compile the search field).

I've addressed the other comments.

ciyer (approved):
Yes, I wonder if we need to show a "404" error, since that is quite technical and probably not of interest to our users, but that is a larger discussion and the changes here are consistent with our current approach to errors.

It looks good now! 👍

lorenzo-cavazzi:
> Yes, I wonder if we need to show a "404" error, since that is quite technical and probably not of interest to our users, but that is a larger discussion and the changes here are consistent with our current approach to errors.

I agree we may want to hide some technical details to make the UI more user-friendly. We could still expose a few extra bits with the `show more` collapsable elements we have already used in other elements.
Should we create an issue/epic for this?

PR #1259, opened 2021-03-03
refactor: route all the kg requests through the gateway

### LINKED ISSUE #1053: Requests to the knowledge-graph service to go through the Gateway
As a Renku user, I'd like to see metadata of my private and internal projects on the UI.

**Acceptance criteria:**
* all calls to the knowledge-graph service to go through the Gateway which would authorise them and add relevant auth headers to them;

lorenzo-cavazzi:
This routes all the knowledge-graph API through the gateway using the prefix `/api/kg`

Since it requires SwissDataScienceCenter/renku-gateway#382 , I manually created a deployment to test it
https://lorenzotest.dev.renku.ch

/deploy
fix #1053

### DIFF
--- a/client/src/api-client/project.js
+++ b/client/src/api-client/project.js
@@ -319,37 +319,16 @@ function addProjectMethods(client) {
       .then(result => yaml.load(result));
   };
 
-  client.fetchDatasetFromKG = (datasetLink) => {
+  client.fetchDatasetFromKG = async (id) => {
+    const url = `${client.baseUrl}/kg/datasets/${id}`;
     const headers = client.getBasicHeaders();
-    const datasetPromise = client.clientFetch(datasetLink, { method: "GET", headers });
-    return Promise.resolve(datasetPromise)
-      .then(dataset => dataset.data);
+    return client.clientFetch(url, { method: "GET", headers }).then(dataset => dataset.data);
   };
 
   client.getProjectDatasetsFromKG = (projectPath) => {
-    let url = `${client.baseUrl}/knowledge-graph/projects/${projectPath}/datasets`;
-    url = url.replace("/api", "");//The url should change in the backend so we don't have to do this
+    const url = `${client.baseUrl}/kg/projects/${projectPath}/datasets`;
     const headers = client.getBasicHeaders();
-    return client.clientFetch(url, { method: "GET", headers }).then((resp) => {
-      return resp.data;
-    }).then(resp => {
-      let fullDatasets = resp.map(dataset =>
-        client.fetchDatasetFromKG(dataset._links[0].href));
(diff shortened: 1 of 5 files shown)


client/src/api-client/project.js
```diff
@@ -319,37 +319,16 @@ function addProjectMethods(client) {
       .then(result => yaml.load(result));
   };
 
-  client.fetchDatasetFromKG = (datasetLink) => {
+  client.fetchDatasetFromKG = async (id) => {
+    const url = `${client.baseUrl}/kg/datasets/${id}`;
```
vfried, line 323: It would be cool to make this be something like...

const url = `${client.baseKgUrl}/datasets/${id}`;

So we can access the KG URL from anywhere we want

client/src/api-client/project.js
```diff
@@ -319,37 +319,16 @@ function addProjectMethods(client) {
...
-
-  //in the future we will get all the info we need for the dataset list from this call...
-  client.getProjectDatasetsFromKG_short = (projectPath) => {
-    let url = `${client.baseUrl}/knowledge-graph/projects/${projectPath}/datasets`;
-    url = url.replace("/api", "");//The url should change in the backend so we don't have to do this
-    const headers = client.getBasicHeaders();
-    return client.clientFetch(url, { method: "GET", headers }).then((resp) => {
-      return resp.data;
-    });
+    return client.clientFetch(url, { method: "GET", headers }).then(resp => resp.data);
```
vfried, line 331: It's cool you deleted this function.

We were using it to enrich datasets when fetching datasets from the KG but it's not needed anymore.

vfried (changes requested):

client/src/api-client/project.js
lorenzo-cavazzi: That could be useful, but I'd rather use a similar solution for all the APIs (`baseKgUrl`, `baseCoreUrl`, `baseGitlabUrl`) ok keep the `client.baseUrl` everywhere.
In the long term, the plan is to have a single point of contact for all the entities, so it should be back to a single `baseUrl`. But that's for the _long_ term, maybe it's not relevant here :grin:

vfried: It's true that it should be the same everywhere... i think it's ok to leave it as it is if we will change it again in the future

vfried:
Besides the comment I wrote on top all works on my side, shall I approve or do you want to wait for the blocking issue to be pushed?

lorenzo-cavazzi:
Approving would be great, so we know we already checked the code.
I'll re-deploy when the other PR on the gateway will be merged to be sure all works, but I think there is no further change required on the UI.

vfried (approved):

ciyer (approved):
👍

PR #1273, opened 2021-03-11
Image picker

### LINKED ISSUE #1206: Image Picker Component
## Motivation

We need a component that would let users pick or upload an image. The possible images would come from a stock set of images that the UI would maintain and images from a project. Additionally, the user would have the ability to upload a new image or provide a URL of an image. 

## Proposal

image

## Details

- The stock images should be resources in the UI
- The graphic-design trainee should create some stock images

ciyer:
An image picker component, with integration in the project settings page for avatar selection.

Fix #1206

# Testing

## Project Settings

The project settings page now includes the ability to add, change, or remove avatars. Try it out on some projects.

image

## Playground

The page https://renku-ci-ui-1273.dev.renku.ch/image demonstrates the features of the component.

image

You can select an image by file, URL, or pick from some stock images. Play around and see if it works the way you expect. This page will be removed before merging, since it is just here as a demonstration

/deploy


vfried:
Hey, it's great we have the image picker almost ready to go.

I found some things I would change/improve

In a project where a user doesn't have rights, we should show the image picker as disabled.
image

In the image on top, i would make "Project Avatar" not bold.

I would put the image name in a smaller size between the arrows and the image and without the [ ].

It would be nice if we could have a viewport size setting so the picker doesn't get resized when the images change.

I would put the gitlab message under the other message or along with it (the 200 limit) and I would make it have that font style.

I would make the file/URL buttons be a dropdown inside the image picker.  I would put both here.. 
image

ciyer:
Great suggestions!

> In a project where a user doesn't have rights, we should show the image picker as disabled.

image


> In the image on top, i would make "Project Avatar" not bold.
> 
> I would put the image name in a smaller size between the arrows and the image and without the [ ].
> 
> It would be nice if we could have a viewport size setting so the picker doesn't get resized when the images change.

Here is what the Project Settings page looks like now:

image

The purpose of the `[]` is to indicate content that is from the system. For example, the `[none]` , the stock images or the `[Current Avatar]`. 

> I would put the gitlab message under the other message or along with it (the 200 limit) and I would make it have that font style.

I wanted that message to replace the submit/cancel buttons, so I think it should stay where it is, but I made it flush right to make it a better replacement for submit.

image

> I would make the file/URL buttons be a dropdown inside the image picker. I would put both here..

Good idea! This is how it looks now.

image

vfried:
Hey, thanks for implementing the requested changes!

I found some other things i think we should change...

If the editor is disabled i think we should disable the arrows and also hide the gitlab sign...
image

I think it's confusing that the gitlab sign only shows when you set things to none... I understand the reason for this, since it's there for when you want to set the image to none but i think that it's wouldn't be wrong to just put it there constantly...
I would also align this to the left under the select image...
image

ciyer:
> If the editor is disabled i think we should disable the arrows and also hide the gitlab sign...

Good catch. This is now fixed.

> I think it's confusing that the gitlab sign only shows when you set things to none... I understand the reason for this, since it's there for when you want to set the image to none but i think that it's wouldn't be wrong to just put it there constantly...
> I would also align this to the left under the select image and make it a bit smaller...

As discussed, I think it makes sense to leave the button and placement as is, since they come from the "submit" component.

client/src/api-client/project.js
```diff
@@ -274,6 +283,28 @@ function addProjectMethods(client) {
...
+   * @param {*} fieldNameOrObject
+   * @param {*} fieldValue
+   */
+  client.putProjectFieldFormData = async (projectId, fieldNameOrObject, fieldValue) => {
+    const headers = client.getBasicHeaders();
+
+    const formData = new FormData();
+    formData.append(fieldNameOrObject, fieldValue);
+
+    // Do not do this! See Warning: https://developer.mozilla.org/en-US/docs/Web/API/FormData/Using_FormData_Objects
```
vfried, line 298: Not sure about what this means, do we want to leave it there as a reminder?

client/src/project/shared/ProjectAvatar.container.js
```diff
@@ -0,0 +1,101 @@
...
+      modes={[ImageInputMode.FILE]}
+      format="image/png,image/jpeg,image/gif,image/tiff"
+      disabled={settingsReadOnly}
+      setInputs={(e) => { setValue(e.target.value); }} />
+    <ProjectAvatarSubmit value={value}
+      onCancel={() => setValue(initial)}
+      externalUrl={externalUrl}
+      onAvatarChange={(f) => onAvatarChange(f).catch(e => { setAlert(e.errorData.message.avatar); })} />
+  </div>;
+
```
vfried, line 98: This should be replaced by the FormGenarator once it's possible.

client/src/project/Project.state.js
```diff
@@ -372,6 +372,18 @@ class ProjectModel extends StateModel {
...
+    return client.setAvatar(this.get("core.id"), avatarFile)
+      .then(() => { this.fetchProject(client, this.get("core.id")); });
+  }
+
+  toggleForkModal() {
+    const forkModalOpen = this.get("transient.forkModalOpen");
+    const forkModalFlipped = forkModalOpen === false ? true : false;
+    this.set("transient.forkModalOpen", forkModalFlipped);
+  }
+
```
vfried, line 386: I'm not sure what this change was here.... maybe this is a rendering error of github

client/src/project/shared/ProjectAvatar.container.js
```diff
@@ -0,0 +1,101 @@
...
+import { Button } from "reactstrap";
+import React, { useEffect, useState } from "react";
+
+import ImageInput, { ImageFieldPropertyName as Prop } from "../../utils/formgenerator/fields/ImageInput";
+import { ImageInputMode } from "../../utils/formgenerator/fields/ImageInput";
+import { ExternalLink } from "../../utils/UIComponents";
+
+const CURRENT_AVATAR_NAME = "[Current Avatar]";
+
+function ProjectAvatarSubmit({ value, onCancel, onAvatarChange, externalUrl }) {
```
vfried, line 35: Maybe i would rename this to ProjectAvatarSubmitButtons to make it a bit more clear this is a component.

vfried (commented):
The code looks good. I left some small comments. 

I think it would be nine to adapt the picker to be able to work with the form generator since i imagine we might want to use it in other places. --> i know this will be addressed in another PR.

I think when we integrate the picker to the form generator we might want to split the code in ImageInput.js but it's not necessary to do this now.  We might want to take some code out so that the field is a bit more generic.

While taking a look at the code i realized that when I shrink the screen things are misaligned, it's just a small detail

image

If the small issues are fixed I'm ready to approve the PR.

client/src/api-client/project.js
ciyer: For me that was the first thing I wanted to do. I left the note in there in case someone looks at the code and asks why aren't we setting the Content-Type header as we do in other places. But we can take it out if you think it is unnecessary.

client/src/project/shared/ProjectAvatar.container.js
ciyer: Do we want to use the FormGenerator even on cases like this where there is just one field?

client/src/api-client/project.js
vfried: I think it's good to leave it... i thought it was that but wanted to make sure about it.

client/src/project/Project.state.js
ciyer: I think it came when I rebased against master.

client/src/project/shared/ProjectAvatar.container.js
vfried: I would say yes because this would allow us to do things like restyling all buttons at the same time. But i guess we could discuss whether it makes sense.

vfried (dismissed):
Thanks for addressing the changes that I requested! 🚀

vfried (approved):

PR #1277, opened 2021-03-15
refactor: add dropdown indicator on autosuggest

### LINKED ISSUE #1271: Add a dropdown indicator to the NamespacesAutosuggest component
## Motivation

It is not always obvious to the user that there are other options for the namespace available when creating a new project or forking an existing project.

## Proposal

Add a triangle or some other indicator to the `NamespacesAutosuggest` component.

lorenzo-cavazzi:
~That was not planned for this sprint but~
I was checking something on bootstrap and I accidentally found the `custom-select` class to style an input element as a dropdown.
Sadly, I had to re-implement the same styling because the Autosuggest element has a minor bug preventing adding a class on the main `input` element when untouched.

/deploy
fix #1271

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -16726,24 +16726,15 @@
       }
     },
     "react-autosuggest": {
-      "version": "10.0.0",
-      "resolved": "https://registry.npmjs.org/react-autosuggest/-/react-autosuggest-10.0.0.tgz",
-      "integrity": "sha512-WFCeRuYv6YCY9Ch4BodXPASWJXDh32ehQQHA2eLNhOBhu3Xcank+5W4wcEN+oqdAX1fvKROZRQRF2fpNZaZaxA==",
+      "version": "10.1.0",
+      "resolved": "https://registry.npmjs.org/react-autosuggest/-/react-autosuggest-10.1.0.tgz",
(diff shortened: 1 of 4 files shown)


ciyer (approved):
Great fix!

It looks a little bit different than than what is expected on Safari and Chrome (didn't test FF), but I think this is fine, and I don't think it is worth putting in the effort to fix that.

ciyer (changes requested):
Actually, I looked a little longer, and I think we should make a minor change if it is possible. Could you move the dropdown indicator (the triangles) over by a few pixels to line up with the other dropdown indicator. See image

image


Chrome, Safari, and FF all have their indicator closer to the right edge of the widget.

lorenzo-cavazzi:
I suggest using the opposite approach: instead of trying to adjust the class on the input, we could actually assign the same class to the other `select` elements and have the same output.
I don't particularly like to modify the default select appearance, but it's just the icon on the right, while the dropdown interaction won't change. That seems the only solution to prevent very different layouts in some browsers (e.g. with Firefox on Linux the dropdown icon is _very_ different).
This would be the layout with the suggested change

Screenshot_20210315_154939

ciyer:
> I suggest using the opposite approach: instead of trying to adjust the class on the input, we could actually assign the same class to the other `select` elements and have the same output.

I like this solution! Do you have an opinion, @vfried?

ciyer (approved):
Great solution. I think this looks better than before.

Issue #1194, opened 2021-01-13
Allow encoding data in the URL to automatically fill in new project fields


lorenzo-cavazzi:
## Motivation
Some users asked about the possibility to include encoded information in a new project URL to be applied immediately (e.g. repository URL and reference, template-id, visibility level, variables...)

## Proposal
We could include these data in the URL using query parameters, save the data from there to the redux store, and change the URL back to the standard one.

The main pain points of this approach are:
- How to include strings with reserved chars (e.g. a template URL): we could base64 encode the data, potentially even using a single parameter containing a JSON formatted structure. That would require clear instructions on how to create a proper URL.
- How to fill in dynamic fields not immediately available: most of the fields in the "new project" page are dynamically created based on the fetched content and/or on the other selected options. This means setting the `description` field requires to 1) fetch the templates, 2) select a specific template, 3) fill in the input. All operations require the previous to be finished, so they must complete in order. Even worse, some fields may be available immediately, but the choice will be limited when changing other fields (e.g. the visibility may be limited by the selected namespace, which is fetched and not available immediately).
The user could further mess up by manually changing a field in the meantime.
A possible strategy is to show a temporary page while fetching all the data and filling in the give values. That requires having clear precedences between the different fields, deciding a strategy to follow when something fails (e.g. can't fetch a resource, or a field that should be filled in doesn't exist).

- [x] Design (sprint-2021-01-29)
- [ ] Implementation

lorenzo-cavazzi:
## Design

### Providing data
- We should allow the users to provide data to pre-compile any editable field. A reasonable would be to encode a json object and provide it through the `data` parameter (`btoa(JSON.stringify(data)`).
- We should get the `data` from the URL, store them in the redux store, and move from `/projects/new?data=(encoded)` to `/projects/new`. Otherwise, the user may believe he can modify a few fields and copy/paste the current URL to share the current data.
- The possible fields of the `data` object are: 
  ```
  data: {
    title, namespace, visibility, source, ref, template,
    variables: {
      <var_name_1>, <var_name_2>, ...
    }
  }
  ```
  Mind that 3 fields are missing
  * `identifier`: the user can't currently edit it
  * `template source`: we know it's selected when we get a `source` and a `ref`
  * `kg opt-out`: that field is going away soon
[...]

PR #1282, opened 2021-03-22
refactor: update login to preserve query params

lorenzo-cavazzi:
While working on #1194 , I had to update the login mechanism to preserve the query parameters on the return URL.
I decided to split this out to simplify testing and merging separately.

/deploy
re #1194

### DIFF
--- a/client/src/dataset/Dataset.present.js
+++ b/client/src/dataset/Dataset.present.js
@@ -31,6 +31,7 @@ import { ProjectsCoordinator } from "../project/shared";
 import AddDataset from "./addtoproject/DatasetAdd.container";
 import DeleteDataset from "../project/datasets/delete/index";
 import Time from "../utils/Time";
+import { Url } from "../utils/url";
 
 
 function DisplayFiles(props) {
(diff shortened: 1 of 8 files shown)


ciyer (approved):
👍 

Details:

I tested adding extra URL parameters in all of the pages containing the components modified here. For every page, I had runs where the parameters were preserved after login, but on several pages, I also had runs where the the url was URL-encoded and the `?` was converted to `%3F`. It looks like that behavior is coming from keycloak. For example, sometimes I had to log into `dev` as well as the `ci` deployment, and then this would happen.

I think the PR can be merged, but this is something we will need to keep an eye on.

lorenzo-cavazzi:
> +1
> 
> Details:
> 
> I tested adding extra URL parameters in all of the pages containing the components modified here. For every page, I had runs where the parameters were preserved after login, but on several pages, I also had runs where the the url was URL-encoded and the `?` was converted to `%3F`. It looks like that behavior is coming from keycloak. For example, sometimes I had to log into `dev` as well as the `ci` deployment, and then this would happen.
> 
> I think the PR can be merged, but this is something we will need to keep an eye on.

Ah interesting! I didn't notice this at first. We should keep it in mind.
If that will create problems, we will need to fix.

Since this part of the authentication will change anyway in the near future (and should go into the UI server component), I'll merge the PR anyway and I'll keep this in mind while re-implementing it

PR #1284, opened 2021-03-23
feat: show clone command first on project settings

### LINKED ISSUE #1283: proposal: change the order of commands on the project status page
## Motivation

I often find myself quickly navigating to the project settings page to get a clone URL and I always end up copying the ssh link. Then I do a regular git clone, and after running a few renku commands I realize none of the git hooks etc are installed, so I have to look up how to do that etc. This can be avoided if I would just use the `renku clone` command, but it's at the bottom of the list so I usually don't even see it when I'm in a rush. 

This is the list I mean:

image

## Proposal

Move the renku command to the top of the list.

ciyer:
Fix #1283.

## Screenshot

image

/deploy


lorenzo-cavazzi (approved):
:+1:

rokroskar:
Thanks for the quick fix! 😄

PR #1285, opened 2021-03-23
feat(design): project search results redesigned

vfried:
Redesign for project search results...

image

Pagination...

image

Design ref: https://xd.adobe.com/view/161ea402-fc08-41c1-b52e-1c423af506c9-434d/screen/75caace6-caf1-4183-9472-0224e227bd44/specs/

- we didn't had a design for the pagination so i adapted the one we have now 

/deploy


client/src/styles/icons/icons.scss
```diff
@@ -5,3 +5,22 @@
 	background-repeat: no-repeat;
 	background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='8.157' height='12.232' viewBox='0 0 8.157 12.232'%3E%3Cg id='Group_13150' data-name='Group 13150' transform='translate(-1768.922 3036.348)'%3E%3Cg id='Group_13126' data-name='Group 13126' transform='translate(577.519 -2034.265) rotate(-90)'%3E%3Cpath id='Path_285' data-name='Path 285' d='M993.93,1193.442a2.039,2.039,0,1,1-2.039-2.039,2.04,2.04,0,0,1,2.039,2.039' transform='translate(0)' fill='%23fff'/%3E%3Cpath id='Path_286' data-name='Path 286' d='M1007.46,1193.442a2.039,2.039,0,1,1-2.039-2.039,2.04,2.04,0,0,1,2.039,2.039' transform='translate(-5.376)' fill='%23fff'/%3E%3Cpath id='Path_287' data-name='Path 287' d='M1000.695,1200.21a2.039,2.039,0,1,1-2.039-2.039,2.039,2.039,0,0,1,2.039,2.039' transform='translate(-2.688 -2.689)' fill='%23fff'/%3E%3C/g%3E%3C/g%3E%3C/svg%3E%0A");
 }
+
+.circle {
```
ciyer, line 9: Could we use 2 spaces for indentation instead of tabs?

client/src/styles/components/_renku_search.scss
```diff
@@ -0,0 +1,90 @@
+.rk-search-bar{
```
ciyer, line 1: Could we use 2 spaces for indentation instead of tabs?

client/src/styles/components/_renku_search.scss
```diff
@@ -0,0 +1,90 @@
...
+	border-right: solid 1px $light;
+
+	.title{
+		font-size: 24px;
+	}
+}
+
+.rk-search-result:hover{
+	cursor: pointer;
+	border-top: solid 2px $light;
```
ciyer, line 64: The difference between the hover and non-hover rows is not easy to see.

E.g., Which row is the mouse over here?

image

If you look carefully, you will see that it is row 1, but it is very subtle. Could this be made a bit more obvious? Do not have a solution readily available, but maybe something like:

```border: 3px solid #adadad;```

Or maybe use the same color as the dot for the border?

ciyer (changes requested):
This looks great! The code is good. I think we should use spaces for indentation (see comments). Also one request for feature change: the XD file does not specify this, but I think the row that is under the mouse should be made more prominent. 

E.g., Which row is the mouse over here?

image

If you look carefully, you will see that it is row 1.

client/src/styles/components/_renku_search.scss
vfried: I hadn't read your suggestion... here is an example with the border with the dot color...

image

vfried:
I just pushed fixes and restyled all the scss code, we should make an issue to add scss to linting rules.

ciyer:
I think the solution to highlighting with the gray background on the results list is nice. Let us go with that. In testing, I found two other cases were we need to modify the hover styling.

In user and group searches, the styling of the selected option is the same as the hovered over option. Let us make these a bit more different (maybe changing the background color).

image


image

vfried:
I will apply this hover color to the hover in search results and users

ciyer (approved):
Looks good!

PR #1286, opened 2021-03-23
feat(design): redesign of footer

vfried:
Redesign of footer, this was not in the documents that luca gave us so i did something I think looks good. For now we will need to have a footer, in the future we could consider removing it if we wanted to.

image

image

/deploy

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -1,20 +1,21 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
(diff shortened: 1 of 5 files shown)


ciyer (approved):
Very nice solution! I modified code formatting in some files to better see the changes. I made one visible change: I made the Renku logo in the footer a little bigger and aligned the baseline with the other elements.

image

PR #1287, opened 2021-03-24
feat(design): dataset search redesigned

vfried:
New datasets search page...

image

/deploy

### DIFF
--- a/client/src/dataset/list/DatasetList.present.js
+++ b/client/src/dataset/list/DatasetList.present.js
@@ -16,18 +16,19 @@
  * limitations under the License.
  */
 
-import React, { Component } from "react";
+import React, { Component, Fragment } from "react";
 import { Link, Route, Switch } from "react-router-dom";
-import { Row, Col, Alert, Card, CardBody, Badge } from "reactstrap";
+import { Row, Col, Alert, Card, CardBody } from "reactstrap";
 import { Button, Form, FormText, Input, Label, InputGroup, UncontrolledCollapse } from "reactstrap";
 import { DropdownToggle, DropdownMenu, DropdownItem } from "reactstrap";
-import { Loader, Pagination, MarkdownTextExcerpt } from "../../utils/UIComponents";
-import { faCheck, faSortAmountUp, faSortAmountDown } from "@fortawesome/free-solid-svg-icons";
+import { stringScore } from "../../utils/HelperFunctions";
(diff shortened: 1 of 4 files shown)


client/src/dataset/list/DatasetList.present.js
```diff
@@ -16,142 +16,191 @@
...
-              {new Date(dataset.date).toLocaleDateString()}
-            </small>
-            : null
-        }
-      </CardBody>
-    </Card>;
+        </div>
+        <div className="creators text-truncate text-rk-text">
+          {
+            // Can't remember if we had decided to leave just the first 3 crators...
```
ciyer, line 53: On the Project/dataset page, we handle this using CSS: 
https://github.com/SwissDataScienceCenter/renku-ui/blob/master/client/src/project/Project.css#L40

We don't need to apply the same styling in this PR, but maybe the comment could reflect that.

client/src/dataset/list/DatasetList.present.js
```diff
@@ -16,142 +16,191 @@
...
-          </CardBody>
-        </Card>
-      </UncontrolledCollapse>
-    ];
+          </Col>
+        </Form>
+        <Col sm={12}>
+          <FormText key="help" color="rk-text">
+            {this.props.errorMessage} If you are not finding what you are looking
+            for, <Button className="pr-0 pl-0 pt-0 pb-0 mb-1" color="link" id="toggler">
```
ciyer, line 170: The spacing around the button is off here. 

image

I think the problem is that Bootstrap 5 changes the naming convention for the spacing classes:

https://getbootstrap.com/docs/5.0/utilities/spacing/

I think the className should be set to `px-0 py-0` if we would want to control l/r and t/b styling separately or just `p0` to set everything to 0.

ciyer (changes requested):
This is coming along! I just found one minor issue and have a suggestion about a comment.

client/src/dataset/list/DatasetList.present.js
vfried: I fixed the button padding, it's now p-0 (it changed from pr to pe as in pading end and pl to ps as in pading start).

Fixed a  small problem with the description height, now set to 100px max.

The creatos length, it's now limited to 3 person as it is now in the dataset project. This is what we had agreed with rok.

image

ciyer (approved):
Great!

PR #1288, opened 2021-03-25
Read encoded metadata from the URL when creating new project

### LINKED ISSUE #1194: Allow encoding data in the URL to automatically fill in new project fields

lorenzo-cavazzi:
Manage metadata from the query paramaters on the "create project page" to automatically fill in some inputs. Thin included the pieces to generate the URL and to read the data:

* Generate URL: click on the dropdown button "share link" to open a modal with the encoded URL. The user can decide which inputs to include in the metadata (screenshot 1).
* Read metadata: detect and read the metadata from the query parameters. The URL resets back to the basic one, then the data are gradually fetched and restored automatically (screenshot 2). Any error (e.g. unexisting template) or warning (e.g namespace unaccessible) will appear on top (screenshot 3).

For testing purposes, you can easily generate the encoded content of the `data` param by using `btoa` and `JSON.stringify` from the browser console (useful to generate errors or warnings on purpose), then add it after `/projects/new?data=`.
This is an example:
```
btoa(JSON.stringify({title:"test", namespace:"renku-qa", template:"Custom/python-minimal", url:"https://github.com/SwissDataScienceCenter/renku-project-template", ref:"0.1.16", visibility:"private", variables: { description: "description here" }}))
```

> eyJ0aXRsZSI6InRlc3QiLCJuYW1lc3BhY2UiOiJyZW5rdS1xYSIsInRlbXBsYXRlIjoiQ3VzdG9tL3B5dGhvbi1taW5pbWFsIiwidXJsIjoiaHR0cHM6Ly9naXRodWIuY29tL1N3aXNzRGF0YVNjaWVuY2VDZW50ZXIvcmVua3UtcHJvamVjdC10ZW1wbGF0ZSIsInJlZiI6IjAuMS4xNiIsInZpc2liaWxpdHkiOiJwcml2YXRlIiwidmFyaWFibGVzIjp7ImRlc2NyaXB0aW9uIjoiZGVzY3JpcHRpb24gaGVyZSJ9fQ==

The final URL is `/projects/new?data=<encoded_data_here>`

fix #1194
/deploy

**INFO**: if this PR works well, we should create a corresponding entry in the documentation for this new feature

### Screenshots

1)
Screenshot_20210325_085711

2)
Screenshot_20210325_085801

3)
Screenshot_20210325_091103

### DIFF
--- a/client/src/project/new/ProjectNew.container.js
+++ b/client/src/project/new/ProjectNew.container.js
@@ -31,7 +31,9 @@ import { NewProject as NewProjectPresent, ForkProject as ForkProjectPresent } fr
 import { NewProjectCoordinator, validateTitle, checkTitleDuplicates } from "./ProjectNew.state";
 import { ProjectsCoordinator } from "../shared";
 import { gitLabUrlFromProfileUrl, slugFromTitle, refreshIfNecessary } from "../../utils/HelperFunctions";
-import { Url } from "../../utils/url";
+import { Url, getSearchParams } from "../../utils/url";
+import { atobUTF8, btoaUTF8 } from "../../utils/Encoding";
+import { newProjectSchema } from "../../model/RenkuModels";
 
 const CUSTOM_REPO_NAME = "Custom";
 
@@ -282,27 +284,86 @@ function ForkProject(props) {
   );
 }
 
+/**
+ * Validate and decode query params.
+ *
+ * @param {object} params - query params
+ * @returns {object} parsed parameters, validated and ready to be be used to pre-fill fields.
+ */
+function getDataFromParams(params) {
+  if (params && Object.keys(params).length) {
+    if (params.data) {
+      const data = JSON.parse(atobUTF8(params.data));
+      if (data && Object.keys(data).length) {
+        // validate metadata
+        const validKeys = Object.keys(newProjectSchema.createEmpty().automated.data);
+        const keys = Object.keys(data);
+        for (let key of keys) {
+          if (!validKeys.includes(key))
+            throw new Error("unexpected project field in the encoded metadata: " + key);
+        }
+      }
+      return data;
+    }
+    // Unrecognized params: should we notify? Let's start without notifications.
+    return null;
+  }
+}
+
 class NewProject extends Component {
   constructor(props) {
     super(props);
+    // Create model and reset inputs
     this.model = props.model;
     this.coordinator = new NewProjectCoordinator(props.client, this.model.subModel("newProject"),
       this.model.subModel("projects"));
     this.coordinator.setConfig(props.templates.custom, props.templates.repositories);
     this.projectsCoordinator = new ProjectsCoordinator(props.client, this.model.subModel("projects"));
     this.coordinator.resetInput();
+    this.removeAutomated();
+    if (!props.user.logged)
+      this.coordinator.resetAutomated();
 
+    // create handlers
     this.handlers = {
-      onSubmit: this.onSubmit.bind(this),
+      createEncodedUrl: this.createEncodedUrl.bind(this),
       getNamespaces: this.getNamespaces.bind(this),
       getTemplates: this.getTemplates.bind(this),
       getUserTemplates: this.getUserTemplates.bind(this),
+      goToProject: this.goToProject.bind(this),
+      onSubmit: this.onSubmit.bind(this),
+      removeAutomated: this.removeAutomated.bind(this),
+      setNamespace: this.setNamespace.bind(this),
       setProperty: this.setProperty.bind(this),
       setTemplateProperty: this.setTemplateProperty.bind(this),
-      setNamespace: this.setNamespace.bind(this),
       setVariable: this.setVariable.bind(this),
-      goToProject: this.goToProject.bind(this)
     };
+
+    // Handle optional param used to pre-fill inputs
+    if (!props.user.logged)
+      return;
+    const params = getSearchParams();
+    try {
+      const data = getDataFromParams(params);
+      if (data)
+        this.coordinator.setAutomated(data);
+      const newUrl = Url.get(Url.pages.project.new);
+      this.props.history.push(newUrl);
+    }
+    catch (e) {
+      this.coordinator.setAutomated(null, e);
+    }
+  }
+
+  removeAutomated(manuallyReset = true) {
+    this.coordinator.resetAutomated(manuallyReset);
+  }
+
+  createEncodedUrl(data) {
+    if (!data || !Object.keys(data).length)
+      return Url.get(Url.pages.project.new, {}, true);
+    const encodedContent = btoaUTF8(JSON.stringify(data));
+    return Url.get(Url.pages.project.new, { data: encodedContent }, true);
   }
 
   async getNamespaces() {
@@ -417,3 +478,6 @@ class NewProject extends Component {
 
 
 export { NewProject, CUSTOM_REPO_NAME, ForkProjectMapper as ForkProject };
+
+// test only
+export { getDataFromParams };
--- a/client/src/project/new/ProjectNew.container.js
+++ b/client/src/project/new/ProjectNew.container.js
@@ -291,23 +291,20 @@ function ForkProject(props) {
  * @returns {object} parsed parameters, validated and ready to be be used to pre-fill fields.
  */
 function getDataFromParams(params) {
-  if (params && Object.keys(params).length) {
-    if (params.data) {
-      const data = JSON.parse(atobUTF8(params.data));
-      if (data && Object.keys(data).length) {
-        // validate metadata
-        const validKeys = Object.keys(newProjectSchema.createEmpty().automated.data);
-        const keys = Object.keys(data);
-        for (let key of keys) {
-          if (!validKeys.includes(key))
-            throw new Error("unexpected project field in the encoded metadata: " + key);
-        }
-      }
-      return data;
-    }
-    // Unrecognized params: should we notify? Let's start without notifications.
-    return null;
+  // Unrecognized params: should we notify? Let's start without notifications.
+  if (!params || !Object.keys(params).length || !params.data)
+    return;
+  const data = JSON.parse(atobUTF8(params.data));
+  if (!data || !Object.keys(data).length)
+    return;
+  // validate metadata
+  const validKeys = Object.keys(newProjectSchema.createEmpty().automated.data);
+  const keys = Object.keys(data);
+  for (let key of keys) {
+    if (!validKeys.includes(key))
+      throw new Error("unexpected project field in the encoded metadata: " + key);
   }
+  return data;
 }
 
 class NewProject extends Component {
--- a/client/src/project/new/ProjectNew.present.js
+++ b/client/src/project/new/ProjectNew.present.js
@@ -218,62 +218,60 @@ function Automated(props) {
   const [showWarnings, setShowWarnings] = useState(false);
   const toggleWarn = () => setShowWarnings(!showWarnings);
 
-  if (automated.received && automated.valid && !automated.finished) {
+  if (!automated.finished) {
     // Show a static modal while loading the data
-    return (<AutomatedModal removeAutomated={removeAutomated}></AutomatedModal>);
+    if (automated.received && automated.valid)
+      return (<AutomatedModal removeAutomated={removeAutomated}></AutomatedModal>);
+    return null;
   }
-  else if (automated.finished) {
-    // Show a feedback when the automated part has finished
-    // errors
-    if (automated.error) {
-      const error = (<pre>{automated.error}</pre>);
-      return (
-        <Alert key="alert" color="danger">
-          <p>
-            <FontAwesomeIcon icon={faExclamationTriangle} />&nbsp;
-            An error occurred while bootstrapping your pre-initialized project.
-          </p>
-          <p>
-            It is possible the metadata are not valid or outdated. You can try to contact who provided you
-            the RenkuLab link asking for a new one.
-          </p>
-
-          <Button color="link" style={{ fontSize: "smaller" }} className="font-italic" onClick={() => toggleError()}>
-            {showError ? "Hide error details" : "Show error details"}
-          </Button>
-          <Fade in={showError} tag="div">{showError ? error : null}</Fade>
-        </Alert>
-      );
-    }
-    // warnings
-    else if (automated.warnings.length) {
-      const warnings = (<pre>{automated.warnings.join("\n")}</pre>);
-      return (
-        <Alert color="warning">
-          <p>
-            <FontAwesomeIcon icon={faExclamationTriangle} />&nbsp;
-            Some pre-initializzation settings could not be applied.
-          </p>
-          <Button color="link" style={{ fontSize: "smaller" }} className="font-italic" onClick={() => toggleWarn()}>
-            {showWarnings ? "Hide warnings" : "Show warnings"}
-          </Button>
-          <Fade in={showWarnings} tag="div">{showWarnings ? warnings : null}</Fade>
-        </Alert>
-      );
-    }
-    // all good
+  // Show a feedback when the automated part has finished
+  // errors
+  if (automated.error) {
+    const error = (<pre>{automated.error}</pre>);
     return (
-      <Alert color="primary">
-        <p className="mb-0">
-          <FontAwesomeIcon icon={faInfoCircle} />&nbsp;
-          Some pre-initializzation settings were applied.
-          <br />You can still change any setting before you create the project.
+      <Alert key="alert" color="danger">
+        <p>
+          <FontAwesomeIcon icon={faExclamationTriangle} />&nbsp;
+          An error occurred while bootstrapping your pre-initialized project.
         </p>
+        <p>
+          It is possible the metadata are outdated or not valid.
+          Please contact the source of the RenkuLab link and ask for a new one.
+        </p>
+
+        <Button color="link" style={{ fontSize: "smaller" }} className="font-italic" onClick={() => toggleError()}>
+          {showError ? "Hide error details" : "Show error details"}
+        </Button>
+        <Fade in={showError} tag="div">{showError ? error : null}</Fade>
       </Alert>
     );
   }
-
-  return null;
+  // warnings
+  else if (automated.warnings.length) {
+    const warnings = (<pre>{automated.warnings.join("\n")}</pre>);
+    return (
+      <Alert color="warning">
+        <p>
+          <FontAwesomeIcon icon={faExclamationTriangle} />&nbsp;
+          Some pre-initialization settings could not be applied.
+        </p>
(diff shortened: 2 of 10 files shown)


client/src/project/new/ProjectNew.container.js
```diff
@@ -277,27 +279,86 @@ function ForkProject(props) {
...
+        for (let key of keys) {
+          if (!validKeys.includes(key))
+            throw new Error("unexpected project field in the encoded metadata: " + key);
+        }
+      }
+      return data;
+    }
+    // Unrecognized params: should we notify? Let's start without notifications.
+    return null;
+  }
```
ciyer, line 305: I suggest rewriting it with guarding clauses. I think this will be advantageous if we switch to notifications, since then we will know what the problem is.

```suggestion
    // Unrecognized params: should we notify? Let's start without notifications.
  const errorValue = null;
  if (params  == null) return errorValue;
  if (Object.keys(params).length < 1) return errorValue;
  if (params.data == null) return errorValue;
  const data = JSON.parse(atobUTF8(params.data));
  if (data == null || Object.keys(data).length < 1) return errorValue;
  // validate metadata
  const validKeys = Object.keys(newProjectSchema.createEmpty().automated.data);
  const keys = Object.keys(data);
  for (let key of keys) {
    if (!validKeys.includes(key))
      throw new Error("unexpected project field in the encoded metadata: " + key);
  }
  return data;
```

client/src/project/new/ProjectNew.present.js
```diff
@@ -208,6 +209,120 @@ class NewProject extends Component {
...
+  const toggleError = () => setShowError(!showError);
+
+  const [showWarnings, setShowWarnings] = useState(false);
+  const toggleWarn = () => setShowWarnings(!showWarnings);
+
+  if (automated.received && automated.valid && !automated.finished) {
+    // Show a static modal while loading the data
+    return (<AutomatedModal removeAutomated={removeAutomated}></AutomatedModal>);
+  }
+  else if (automated.finished) {
```
ciyer, line 225: `else` is unnecessary here.

I think 
```
if (!automated.finished) return null;
```
would be clearer.

client/src/project/new/ProjectNew.present.js
```diff
@@ -208,6 +209,120 @@ class NewProject extends Component {
...
+      const error = (<pre>{automated.error}</pre>);
+      return (
+        <Alert key="alert" color="danger">
+          <p>
+            <FontAwesomeIcon icon={faExclamationTriangle} />&nbsp;
+            An error occurred while bootstrapping your pre-initialized project.
+          </p>
+          <p>
+            It is possible the metadata are not valid or outdated. You can try to contact who provided you
+            the RenkuLab link asking for a new one.
```
ciyer, line 238:
```suggestion
It is possible the metadata are outdated or not valid. Please contact the source of the the RenkuLab link and ask for a new one.
```

client/src/project/new/ProjectNew.present.js
```diff
@@ -208,6 +209,120 @@ class NewProject extends Component {
...
+      );
+    }
+    // warnings
+    else if (automated.warnings.length) {
+      const warnings = (<pre>{automated.warnings.join("\n")}</pre>);
+      return (
+        <Alert color="warning">
+          <p>
+            <FontAwesomeIcon icon={faExclamationTriangle} />&nbsp;
+            Some pre-initializzation settings could not be applied.
```
ciyer, line 255:
```suggestion
            Some pre-initialization settings could not be applied.
```

client/src/project/new/ProjectNew.present.js
```diff
@@ -208,6 +209,120 @@ class NewProject extends Component {
...
+          <Fade in={showWarnings} tag="div">{showWarnings ? warnings : null}</Fade>
+        </Alert>
+      );
+    }
+    // all good
+    return (
+      <Alert color="primary">
+        <p className="mb-0">
+          <FontAwesomeIcon icon={faInfoCircle} />&nbsp;
+          Some pre-initializzation settings were applied.
```
ciyer, line 269: See above comment on "pre-initialization" vs. "initialization".

ciyer (changes requested):
Sorry for the delay -- I have been distracted with my computer set up. 

I have not yet had a chance to test the feature, but I was looking at the code and have done a partial review and noticed a few things. Take a look and see if you agree.

client/src/project/new/ProjectNew.present.js
lorenzo-cavazzi: Simplified in a similar way -- I've slightly changed the code preceding this part since it already included `if (!automated.finished)`.

client/src/project/new/ProjectNew.container.js
lorenzo-cavazzi: The code you suggested is definitely less convoluted.
I ended up committing a small variation of it, using `|| !` to merge the conditions in the first few lines so that it's coherent with line `295`. Let me know if that works

client/src/project/new/ProjectNew.present.js
```diff
@@ -208,6 +209,118 @@ class NewProject extends Component {
...
+    );
+  }
+  // warnings
+  else if (automated.warnings.length) {
+    const warnings = (<pre>{automated.warnings.join("\n")}</pre>);
+    return (
+      <Alert color="warning">
+        <p>
+          <FontAwesomeIcon icon={faExclamationTriangle} />&nbsp;
+          Some pre-initialization settings could not be applied.
```
ciyer, line 256:
```suggestion
          Some fields could not be pre-filled with the information provided in the RenkuLab project-creation link.
```

client/src/project/new/ProjectNew.present.js
```diff
@@ -208,6 +209,118 @@ class NewProject extends Component {
...
+      </Alert>
+    );
+  }
+  // all good
+  return (
+    <Alert color="primary">
+      <p className="mb-0">
+        <FontAwesomeIcon icon={faInfoCircle} />&nbsp;
+        Some pre-initialization settings were applied.
+        <br />You can still change any setting before you create the project.
```
ciyer, line 271:
```suggestion
        Some fields were pre-filled. 
        <br />You can still change any values before you create the project.
```

client/src/project/new/ProjectNew.present.js
```diff
@@ -208,6 +209,118 @@ class NewProject extends Component {
...
+
+
+function AutomatedModal(props) {
+  const { removeAutomated } = props;
+
+  const [showFadeIn, setShowFadeIn] = useState(false);
+
+  const toggle = () => setShowFadeIn(!showFadeIn);
+
+  const button = showFadeIn ?
```
ciyer, line 285: Is there a specific use case for including this? The pop-up is there for too short a time to make this useful. Or is this an "escape hatch" if the pre-filling doesn't finish for a long time?

- If this is a useful feature, I think it should be placed on the main "New project" page as part of the info dialog that says "Some fields were pre-filled...".

- If this is a fallback mechanism, I think the wording should be changed to "If pre-filling the new project form is taking too long, you can [use a blank form](link to normal new project form)".

client/src/project/new/ProjectNew.present.js
```diff
@@ -208,6 +209,118 @@ class NewProject extends Component {
...
+      </Link>
+    </p>
+  );
+  return (
+    <Modal isOpen={true} centered={true} keyboard={false} backdrop="static">
+      <ModalHeader>Fetching initialization data</ModalHeader>
+      <ModalBody>
+        <Row>
+          <Col>
+            <p>You entered a url containing initialization data.</p>
```
ciyer, line 308:
```suggestion
            <p>You entered a url containing information to pre-fill.</p>
```

client/src/project/new/ProjectNew.present.js
```diff
@@ -208,6 +209,118 @@ class NewProject extends Component {
...
+            <span>
+              Please wait, we are fetching the required metadata...&nbsp;
```
ciyer, line 310:
```suggestion
              Please wait while we fetch the required metadata...&nbsp;
```

client/src/project/new/ProjectNew.present.js
```diff
@@ -775,20 +910,177 @@ class Create extends Component {
...
+        id="create-new-project"
+        color="primary"
+        onClick={this.props.handlers.onSubmit}
+        disabled={disabled}
+      >
+        Create project
+      </Button>
+    );
+    const createLink = (
+      <DropdownItem onClick={this.toggle}><FontAwesomeIcon icon={faLink} /> Share link</DropdownItem>
```
ciyer, line 928: "Share link" is a bit unclear. How about: "Create link"?

client/src/project/new/ProjectNew.present.js
```diff
@@ -775,20 +910,177 @@ class Create extends Component {
...
+
+  return (
+    <Modal isOpen={props.show} toggle={props.toggle}>
+      <ModalHeader toggle={props.toggle}>Create shareable link</ModalHeader>
+      <ModalBody>
+        <Row>
+          <Col>
+            <p>
+              Here is your shareable link, containing the current settings for a new project.
+              Clicking on it will pre-fill the inputs automatically.
```
ciyer, line 1056:
```suggestion
              Here is your shareable link, containing the current values for a new project.
              Following the link will lead to a <b>New project<b> form with these values pre-filled.
```

client/src/project/new/ProjectNew.present.js
```diff
@@ -775,20 +910,177 @@ class Create extends Component {
...
+            </p>
+            <p>
+              You can review which settings to include in the link metadata.
```
ciyer, line 1059:
```suggestion
              You can control which values should be pre-filled..
```

client/src/project/new/ProjectNew.present.js
```diff
@@ -775,20 +910,177 @@ class Create extends Component {
...
+          onChange={e => handleCheckbox(item, e)}
+        /> {item === "templateRepo" ? "template source" : item}
+      </Label>
+    </FormGroup>
+  ));
+
+  const feedback = include.namespace ?
+    (
+      <FormText color="danger">
+        Including the namespace may lead to errors if the user does not have access to it.
```
ciyer, line 1043:
```suggestion
        Pre-filling the namespace may lead to errors since other users are not guaranteed to have access to it.
```

ciyer (changes requested):
Trying out the feature, and it has been working well so far. I have some suggestions for changes in wording.

client/src/project/new/ProjectNew.container.js
```diff
@@ -277,27 +279,83 @@ function ForkProject(props) {
...
+    const params = getSearchParams();
+    try {
+      const data = getDataFromParams(params);
+      if (data)
+        this.coordinator.setAutomated(data);
+      const newUrl = Url.get(Url.pages.project.new);
+      this.props.history.push(newUrl);
+    }
+    catch (e) {
+      this.coordinator.setAutomated(null, e);
```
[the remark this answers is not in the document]
lorenzo-cavazzi: This error should occur only when the user copy/pasted a wrong link. I think it's useful to notify the user about that to prevent confusion. I'm ok with removing it if you think it's a better UX, but I would rather not logging the error to the console -- not sure if that is a pattern we want to introduce for user-side debugging.

client/src/project/new/ProjectNew.present.js
lorenzo-cavazzi: It's actually the second case, and I don't expect it to happen often. An example could be a very slow response from a remote git repo when a custom template is provided.

I think the wording you suggested is great. Should I change the "show more" text to something like `taking too long?`?
This is the current output, let me know if I can improve it further

Screenshot_20210329_150208
Screenshot_20210329_150240

lorenzo-cavazzi:
Thank you @ciyer for taking the time to go through the wording, it was a bit incoherent :pray: 

I'm not sure about using the `console.log` message for logging an error. If the error is useful, I would use another way to display it -- either a notification or showing it somehow on the page. Otherwise, I would just drop it. I don't think the user/dev would ever check the console, nor we should encourage this pattern.

client/src/project/new/ProjectNew.present.js
ciyer: Looks good as you have solved it. 👍

client/src/project/new/ProjectNew.container.js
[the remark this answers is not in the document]
ciyer: Maybe I do not understand how you get into this code path. 

In general, I think the error handling is good as is. One case that could be improved, though, is if the URL is broken. For example:

https://renku-ci-ui-1288.dev.renku.ch/projects/new?data=eyJ0aXRsZSI6InRlc3QtdGl0bGUiLCJ0ZW1wbGF0ZSI6IlJlbmt1L3B5dGhvbi1taW5pbWFsIiwidmFyaWFibGVzIjp7ImRlc2NyaXB0aW9uIjoidGVzdCBkZXNjcmlwdGlvbiJ9fq%3D%3D

This URL has been corrupted. For most users, ignoring it as happens is the correct behavior, but if I am the developer of the URL, it is confusing because I just get the blank project form with no indication that the data parameter was ignored.

I thought that is what is happening in this code above, but maybe that logic is somewhere else.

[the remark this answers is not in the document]
lorenzo-cavazzi: Oh, I fear I added a check on `newProject.automated.finished` that prevents early errors from showing up to the user :facepalm:  -- the corrupted link is the most expected one.

The intended output was the following

Screenshot_20210329_161657

ciyer (dismissed):
Very nice 🎉 ! Ready to merge as far as I am concerned, but we can wait a day to see if any new feedback comes in.

ltalirz:
Thanks, guys!   ( @rokroskar pointed me here).

> if this PR works well, we should create a corresponding entry in the documentation for this new feature

Indeed it will be important to document how to generate the data payload, since these links will likely be created outside renku (at least in our case). 
It's therefore also helpful if the method for generating the payload remains stable, and if future changes to the parameters in the form (=keys in the dictionary) are backwards-compatible where possible.

rokroskar:
@ltalirz I just tried now and the deployment seems to be fine - you can find it at https://renku-ci-ui-1288.dev.renku.ch/ - note that the login is a bit convoluted because it authenticates against our dev cluster so you essentially log in twice.

lorenzo-cavazzi:
@ltalirz Thanks for the pointers! The project creation based on templates should be flexible enough. For sure, we will try to keep backward-compatibility in case of changes.
The main reason for links to become broken could be referencing templates that are not available anymore, or using a mutable reference (e.g. a branch instead of a tag/commit).

Generating the links outside renku seems more challenging to achieve. We serialize and encode an object with all the data for pre-filling fields, assigning the string to a single query parameter that we include in the URL. It's much easier to do that from the UI. Do you perhaps need to automatize the link generation?

ltalirz:
Sorry for the late follow-up

> Generating the links outside renku seems more challenging to achieve. We serialize and encode an object with all the data for pre-filling fields, assigning the string to a single query parameter that we include in the URL. It's much easier to do that from the UI. Do you perhaps need to automatize the link generation?

Yes - our use case will involve creating these links on the Materials Cloud Archive, where next to an AiiDA export file on a record like [this](https://archive.materialscloud.org/record/2020.153) there would be a "open in renkulab" icon.
That would link directly to the new project page, with metadata concerning the project template, title + some user-defined variables of the template already filled in.

ltalirz:
If the serialization uses standard libraries and is clearly documented, I don't see a big issue in doing this in the materials cloud archive application (might use python instead of JS).

ltalirz:
P.S. An alternative could be to `POST` to the URL with a specific JSON payload - that would get rid of the need for (de)serialization.

lorenzo-cavazzi:
> If the serialization uses standard libraries and is clearly documented, I don't see a big issue in doing this in the materials cloud archive application (might use python instead of JS).

Right, after all we just serialize a dictionary to a JSON string, and we base64 encode it. Once the underlying dictionary structure is known, the process is straightforward in any programming language.

I pushed a PR for updating the documentation here: SwissDataScienceCenter/renku#2060
It contains also a Python example.

In short, here is the process to compute the link:
```
# 1. Create a dictionary with all the required data

  raw_data = {
    "title":"test",
    "url":"https://github.com/SwissDataScienceCenter/renku-project-template",
    "ref":"0.1.17",
    "template":"Custom/python-minimal",
    "variables": {
      "description":"test description"
    }
  }

  # 2. Serialize to a string

  import json
  serialized_data = json.dumps(raw_data)

  # 3. Encode in base64
  import base64
  data = base64.b64encode(str.encode(serialized_data))

  # 4. Use the output to compose the URL
  print(data)

  > b'eyJ0aXRsZSI6ICJ0ZXN0IiwgInVybCI6ICJodHRwczovL2dpdGh1Yi5jb20vU3dpc3NEYX
    RhU2NpZW5jZUNlbnRlci9yZW5rdS1wcm9qZWN0LXRlbXBsYXRlIiwgInJlZiI6ICIwLjEuMT
    ciLCAidGVtcGxhdGUiOiAiQ3VzdG9tL3B5dGhvbi1taW5pbWFsIiwgInZhcmlhYmxlcyI6IH
    siZGVzY3JpcHRpb24iOiAidGVzdCBkZXNjcmlwdGlvbiJ9fQ=='

  # The link will be
  # https://<renkulab_url>/projects/new?data=eyJ0aXRs...biJ9fQ==
```

ciyer (approved):

ltalirz:
Just wanted to mention that @vgranata created an internal prototype using this and it seems to be working fine!

Just waiting for https://github.com/SwissDataScienceCenter/contributed-project-templates/pull/4 to be reviewed and merged before we can consider making it public.

ciyer:
@ltalirz Thanks for the update!

PR #1290, opened 2021-03-25
refactor: add copy to clipboard on registry popover

lorenzo-cavazzi:
Minor change: adds the "copy to clipboard" icon to the image source URI in the popover on the interactive environment row status.

Screenshot_20210325_151103

/deploy


ciyer (approved):
Nice!

PR #1291, opened 2021-03-29
feat: redesign of environments

vfried:
Redesign of interactive environments

Normal:
image

Compact:
image

Inside project: (might be adjusted/reviewed in the future)
image

/deploy

### DIFF
--- a/client/src/notebooks/Notebooks.present.js
+++ b/client/src/notebooks/Notebooks.present.js
@@ -20,7 +20,7 @@ import React, { Component, Fragment } from "react";
 import Media from "react-media";
 import { Link } from "react-router-dom";
 import {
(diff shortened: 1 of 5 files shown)


ciyer (changes requested):
I like how you have the design to this page. I have a few cosmetic suggestions.

## New Environment Page

There needs to be a gap between the "Start Environment" button and the footer.

Screen Shot 2021-03-31 at 1 54 59 PM

## Global Environments Page

I think the environments information should be shown in two columns, using **bold** for the titles (no colons). I also think the info and link to the commit should be merged into one element that is a popup. 

image

## Project Environments Page

Basically the same as the global version, just without the project name.

image

vfried:
I did some changes, as discussed before, the form was not addressed in this PR.

This is how it looks
image

image

In my opinion, this is much cleaner now, what do you think?

ciyer (approved):
I agree, it looks fantastic now!

PR #1296, opened 2021-03-30
Update project fork: default name and namespace handling

### LINKED ISSUE #1294: The fork form should leave the title unchanged
## Motivation

The fork form provides live validation. There is no need to modify the title.

image

## Proposal

Do not add ` - copy` to the project name. The live validation will tell you if you need to change the project name.

lorenzo-cavazzi:
This PR fixes the following:
- Keep the original project name as default when clicking on the `Fork` button.
- Fetch namespaces recursively: now we handle up to 1'000 namespaces -- even though the UX with >50 is not great.
- Automatically select the user namespace if available.

Pinging @rokroskar to verify the last 2 points.

/deploy
fix #1294

### DIFF
--- a/client/src/project/new/ProjectNew.container.js
+++ b/client/src/project/new/ProjectNew.container.js
@@ -143,7 +143,7 @@ class ForkProjectMapper extends Component {
 function ForkProject(props) {
   const { forkedTitle, handlers, namespaces, projects, toggleModal } = props;
 
-  const [title, setTitle] = useState(forkedTitle + " - copy");
+  const [title, setTitle] = useState(forkedTitle);
   const [namespace, setNamespace] = useState(""); // TODO - pick the default namespace if available
   const [projectsPaths, setProjectsPaths] = useState([]);
   const [error, setError] = useState(null);
--- a/client/src/project/new/ProjectNew.container.js
+++ b/client/src/project/new/ProjectNew.container.js
@@ -148,7 +148,7 @@ function ForkProject(props) {
   const { forkedTitle, handlers, namespaces, projects, toggleModal, user } = props;
 
   const [title, setTitle] = useState(forkedTitle);
-  const [namespace, setNamespace] = useState(""); // TODO - pick the default namespace if available
+  const [namespace, setNamespace] = useState("");
   const [projectsPaths, setProjectsPaths] = useState([]);
   const [error, setError] = useState(null);
 
--- a/client/src/project/new/ProjectNew.container.js
+++ b/client/src/project/new/ProjectNew.container.js
@@ -118,6 +118,10 @@ class ForkProjectMapper extends Component {
         fetched: state.projects.featured.fetched,
         fetching: state.projects.featured.fetching,
         list: state.projects.featured.member,
+      },
+      user: {
+        logged: state.user.logged,
+        username: state.user.data && state.user.data.username ? state.user.data.username : null
       }
     };
   }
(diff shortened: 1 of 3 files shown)


rokroskar:
Thanks @lorenzo-cavazzi this does indeed fix the problem I've been complaining about :) However, I wonder if the correct behavior wouldn't be to just present a single user namespace? A logged in user can be linked to exactly one user ID and therefore user account in gitlab, so we should never have a situation where multiple _user_ namespaces are available for project creation (or any other operation for that matter). What do you think?

ciyer:
> Thanks @lorenzo-cavazzi this does indeed fix the problem I've been complaining about :) However, I wonder if the correct behavior wouldn't be to just present a single user namespace? A logged in user can be linked to exactly one user ID and therefore user account in gitlab, so we should never have a situation where multiple _user_ namespaces are available for project creation (or any other operation for that matter). What do you think?

Could you present a screenshot or other image to illustrate what you mean? It is true that there is only going to be one user namespace, but there might be _group_ namespaces.

rokroskar:
yes, exactly, that's my point - multiple group namespaces are fine, but there should only be a single user namespace. On renku I have

image

and on gitlab I have

image

ciyer:
> yes, exactly, that's my point - multiple group namespaces are fine, but there should only be a single user namespace. On renku I have
> 
> image
> 
> and on gitlab I have
> 
> image

Ok, I have never seen that before. It must be because you are an admin.

rokroskar:
yes :)

ciyer:
> yes :)

Ok, then this is quite a corner case. I was not able to figure out how to give anyone permission to create project in my user namespace -- maybe it isn't even possible for non-admins then.

BTW, you may want to consider splitting out the admin rights to a separate account that you use only to do admin actions for security reasons. :)

client/src/project/new/ProjectNew.container.js
```diff
@@ -141,9 +145,9 @@ class ForkProjectMapper extends Component {
 
 
 function ForkProject(props) {
-  const { forkedTitle, handlers, namespaces, projects, toggleModal } = props;
+  const { forkedTitle, handlers, namespaces, projects, toggleModal, user } = props;
 
-  const [title, setTitle] = useState(forkedTitle + " - copy");
+  const [title, setTitle] = useState(forkedTitle);
   const [namespace, setNamespace] = useState(""); // TODO - pick the default namespace if available
```
ciyer, line 151: Is this comment still necessary? It looks like the default namespace is set correctly (to the user's namespace), it just happens in the `NamespacesAutosuggest` component below.

ciyer (changes requested):
Works great, thanks for putting together this fix so quickly! 

I found one comment that I think is outdated and should be removed.

lorenzo-cavazzi:
@rokroskar it looks like the GitLab UI shows you the namespaces your user would be able to access without the admin rights. I fear we can't replicate the very same behavior: I don't see any way to distinguish the data in the namespaces API response, nor any option to exclude them. I'm not actually sure how GitLab does that, and I don't see any query getting namespaces. Maybe they pre-render the options server-side.

Showing only the user namespace with a path matching the username is very easy to implement, but then an admin would see only his own user namespace but all the group namespaces. Isn't that even more confusing?
If you think it's still better than now, we can do it, but I would rather try to have a better UX for admins if we are really trying to support them in RenkuLab (i.e. instead of an endless dropdown, when we have 20+ namespaces we could show a selection window. There we could separate "your user" form "other users", and have a better filtering for "groups")

client/src/project/new/ProjectNew.container.js
lorenzo-cavazzi: Ah yes, that looks like an outdated comment. I'll remove it

ciyer (approved):
👍

rokroskar:
@lorenzo-cavazzi I see - fine for me to leave it as is then. Thanks!!

PR #1298, opened 2021-04-01
Apply the new design of the home page

ciyer:
This PR makes changes to the homepage shown when an anonymous user access the root URL of RenkuLab.

The default is to show a description of RenkuLab features, but it can be configured as well. 

image

The default looks like this:

image

# Preview

Preview available at https://renku-ci-ui-1298.dev.renku.ch

**TODO**
- [x] handle mobile screen sizes
- [x] connect the links to the correct locations

**Deferred**
This will be done in a follow-up PR.
- make the list of projects displayed on the main page configurable

# Testing

1. Test that the default content looks good and scales to different browser sizes.
2. Test configuration of the content. The easiest way to do this is to run in telepresence and then modify the `config.json` file and refresh the browser.

```
"contentMd": "# Welcome to Yoyodyne\nThis is the Yoyodyne Renku instance.",
```

In this case, you should get the default background image.

3.  Test configuration of the background image. This is also easiest by manually modifying the `config.json` file.

```
"backgroundImage": {
  "url": "https://eoimages.gsfc.nasa.gov/images/imagerecords/79000/79803/earth_night_rotate_lrg.jpg"
}
```

```
"backgroundImage": {
  "url": "https://dev.renku.ch/gitlab/cramakri/homepage-content/-/raw/master/img/earth_night_rotate_lrg_renkubg.jpg"
}
```

It makes sense to look at https://github.com/SwissDataScienceCenter/renku/pull/2066 together with this.

/deploy


lorenzo-cavazzi:
Wow the new landing page is gorgeous!

I have a couple of notes on the appearance (I still need to go through the code).

One is on the element above the footer: it looks nice on big and small pages, but the alignment of the locations is a bit off on medium size.

Screenshot_20210415_170815

A suggestion to mitigate this would be to let the first element (renku logo + "We have offices in...") take all the horizontal space (a `col-12` without max-width) and let the 2 locations float below.
I would also let them align on the left with the aforementioned element. Something like this

Screenshot_20210415_171510

The other thing is the menu on top right. When open, it looks misaligned on the right -- not sure if the idea is to align it on the right or not, but since it's just slightly off, the current vertical alignment looks a bit odd.

Screenshot_20210415_171639

Maybe it's just me, but it was difficult to identify the Login button on the top right, even though I knew where to look. Maybe using a background (or different colors) would make it stand out more?
I guess waiting for a second opinion on this may be better since it's probably just a personal preference

client/src/landing/AnonymousHome.js
```diff
@@ -0,0 +1,363 @@
+/*!
+ * Copyright 2018 - Swiss Data Science Center (SDSC)
```
lorenzo-cavazzi, line 2: 2021

lorenzo-cavazzi (commented):
The code looks good! Just a couple of minor comments inline.

As mentioned offline, I found a problem with setting the content through telepresence. I'll test with more combinations once that is fixed.

client/.eslintrc.json
```diff
@@ -166,8 +244,10 @@
           "toml",
           "tooltip",
           "tspan",
+          "tsstrasse",
```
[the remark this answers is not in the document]
ciyer: Good suggestion! I actually tried this at first, but I was trying to just disable the spellcheck, which I couldn't get to work. You are right: we can just disable all linting on this line.

ciyer:
Thanks for the review. I think all of your concerns have been addressed now.

client/run-telepresence.sh
```diff
@@ -44,6 +44,37 @@ fi
 
 if [ -z "$STATUSPAGE_ID" ]; then STATUSPAGE_ID="r3j2c84ftq49"; else echo "STATUSPAGE_ID is set to '$STATUSPAGE_ID'"; fi
 
+# Set HOMEPAGE_MAIN_CONTENTMD to some markdown to try out the main content:
+# E.g.,
+# set HOMEPAGE_MAIN_CONTENTMD '# Yoyodyne'\n'Welcome to the Yoyodyne Renku instance!'
```
lorenzo-cavazzi, line 49:
```suggestion
# set HOMEPAGE_MAIN_CONTENTMD '# Yoyodyne\nWelcome to the Yoyodyne Renku instance!'
```
I believe the extra `'` around the wrap word char slipped in by mistake.
Compare the output with the current solution (1) and with the suggested change (2)

Screenshot_20210416_141314

Screenshot_20210416_141616

client/src/landing/AnonymousHome.js
```diff
@@ -0,0 +1,364 @@
...
+    style={{
+      backgroundImage: `url(${backgroundUrl})`
+    }}>
+    <div>
+      <div><img src={logo} alt="Renku" height="68" className="d-block my-1" /></div>
+      <Row className="rk-pt-s" >
+        <Col xs={12} xl={4} >
+          <p>We have offices in both Lausanne on the EPFL campus and in Zürich at ETH Zürich.</p>
+        </Col>
+        <Col xs={12} lg={5} xl={4} className="rk-pl-s bg-primary">
```
lorenzo-cavazzi, line 296:
```suggestion
        <Col xs={12} lg={5} xl={4} className="rk-pt-s bg-primary">
```

The extra space on left (and no space on top) makes the output on medium/small screen very close to the above content and misaligned on the left (as in screen 1).
The suggested correction (screen 2) works, but the `xl={4}` at line 293 should also get removed. Otherwise, the content is better on smaller sizes but it's misaligned on bigger.

Screenshot_20210416_140350

Screenshot_20210416_140608

client/src/landing/AnonymousHome.js
```diff
@@ -0,0 +1,364 @@
...
+          <h4 className="text-info">Lausanne</h4>
+          <p className="rk-pt-s">
+            INN Building, Station 14, 1015 Lausanne<br />
+            Contact: Cindy Ravey, Executive Assistant<br />
+            +41 21 693 43 88
+          </p>
+        </Col>
+        <Col xs={12} lg={5} xl={4} className="rk-pl-s bg-primary">
```
lorenzo-cavazzi, line 304:
```suggestion
        <Col xs={12} lg={5} xl={4} className="rk-pt-s bg-primary">
```
Check comment at line 296

client/src/landing/AnonymousHome.js
```diff
@@ -0,0 +1,364 @@
...
+function Section6(props) {
+  const backgroundUrl = VisualFooter;
+  return <div id="rk-anon-home-section6"
+    style={{
+      backgroundImage: `url(${backgroundUrl})`
+    }}>
+    <div>
+      <div><img src={logo} alt="Renku" height="68" className="d-block my-1" /></div>
+      <Row className="rk-pt-s" >
+        <Col xs={12} xl={4} >
```
lorenzo-cavazzi, line 293:
```suggestion
        <Col xs={12}>
```
Check comment at line 296

lorenzo-cavazzi (changes requested):
Great! There is just a minor typo in the telepresence example command I noticed while testing (see inline).

The office locations still look misaligned on smaller browsers. If we are fine with that, feel free to disregard the other inline comments suggesting a slightly different solution.

lorenzo-cavazzi:
P.S: the acceptance tests may be failing due to a timeout. You could include the commit from here to prevent that #1317

Panaetius:
Looks gorgeous!

Not sure if it's supposed to be this way, but the empty white horizontal bar in the middle here seems odd:
image
And the repeat renku logo there and the background image seem oddly positioned, I guess that's a problem with using a 4k display?

client/run-telepresence.sh
ciyer: That is because I use a different shell (fish) which handles strings differently from bash/zsh. I thought that syntax would be portable, but apparently not. I'm ok with switch to the one that works on bash/zsh and adapting it for myself.

ciyer:
Thanks for the feedback!

> Not sure if it's supposed to be this way, but the empty white horizontal bar in the middle here seems odd:

That bar is where the section that highlights some interesting projects will go, but the PR has gotten very large already, so that will be implemented in a follow-up.

> And the repeat renku logo there and the background image seem oddly positioned, I guess that's a problem with using a 4k display?

That's a good catch, something that isn't easy for me to spot on my 13" display. The background shouldn't repeat -- I will fix that.

vfried (commented):
It feels really good to open renkulab and see this as the landing page!!! 🥳  

Things look really good in general but i found some things we could improve...

On my screen the background image on top is there twice (i think you fix this with background:no-repeat)
image

I also think we should do as in inside renkulab and have a max-width and fit the content to that if the screen is bigger than that.

The green text here should be on the right
image

I think we should add a bit of padding on the top here or vertically align the content so that there is not so much white space under the three sections...
image

I see under the h3 luca added some space (We could add padding here or just some <br/> so that there is space between title and paragraph). Also in the "Data drive projects...."

I would add a white border to this image...
image
Something like this could work... 
    min-width: 300px;
    border-width: 40px;
    border-style: solid;
    border-color: white;


So that it looks like this...
image

I'm not sure the collapsible menu looks good the way it is now... I think it could be better if we make it be a vertical menu either similar to this...
image

Or this...
image

Otherwise, I would just add some buttons next to learn more where we say explore projects, explore datasets, etc.

client/src/landing/AnonymousHome.js
ciyer: So the designer had this for the layout on desktop:

image

So I want to maintain that for the larger sizes, but I now have this:

image

and this

image

for smaller sizes.

ciyer:
@vfried Thanks for the review. I have made the fixes you suggest. I set a max width on the main content, but I kept the content on the left side. I could make it centered if you think that is better, though.

vfried:
Thanks for applying the changes to the homepage, I really like the way the background expands when the window expands... it's a really cool effect.

I realized there is a yellow dot in the image, I don't think that should be there... could we change this ourselves?
image

Regarding the content, I would prefer to center things (put the extra space in the right and left instead of only on the right) inside there the content can remain on the left (i think it's what Luca said we should do as well), if you prefer to leave things on the left we could also do that.

I noticed there is some blue background/rectanble opening when you open the menu, could we make this disappear?
image
I tried and you can use something like this to at least make it smaller, i think if here you set the background to transparent it should be better. (what i used is... justify-content: flex-end;display: flex;)
<div style="justify-content: flex-end;display: flex;"><div class="mt-2 collapse show" style=""><nav class="container-fluid navbar rk-anon-home rk-navbar px-0 bg-primary navbar bg-primary">

I tested this on the mobile version and it also looks great, i noticed something strange happening with the footer... we should make it blue like in the other version (or make both light gray but with darker text). I'm not sure if this is something you did or if it was a bug i introduced... we can also address footer fixes in a separate PR.
image


Everything else looks really good!!! Looking forward to having this merged :)

ciyer:
@vfried I had forgotten some of the things we discussed with the designer, but they should all be in now. I think I addressed everything you brought up, except the footer. Something happened to the footer after rebasing against the ui-design-2021 branch, but I will fix them in a follow-up pr.

vfried:
All looks good to me, I found a last small thing (I think current changes might have caused this) we could address this on a follow-up or as well...

when you access things from a phone you see this(there is a white line across things)
image

vfried (approved):
I will approve this PR, you can make changes here or make the changes in a follow-up PR.

PR #1301, opened 2021-04-06
ci: provide chart name when publishing chart

lorenzo-cavazzi:
Fixes this error https://github.com/SwissDataScienceCenter/renku-ui/runs/2276583841
It has been tested on #1300 with this successful result https://github.com/SwissDataScienceCenter/renku-ui/runs/2276677109


ciyer (approved):
👍

PR #1303, opened 2021-04-09
feat(design): project header nav and overview redesigned

vfried:
Header redesigned... this is a proposal, might make project path and description with a lighter font or in italics.

Decided to move the project visibility like a tag, but I'm not sure this is the best. I thought next to the title it didn't look so good.

This doesn't shrink well but will fix it once we like the design for it unshrinked

image

Menu...
image

Readme
image

Stats
image

Commits
image

Status
image

/deploy


ciyer (approved):
This looks very nice! Keep continuing in this direction.

PR #1304, opened 2021-04-09
ci: create ui-server pr for bumping version

lorenzo-cavazzi:
The charts and images are created together, but the `update-component-version` considers each component separately ( SwissDataScienceCenter/renku#2059 ).
This change manually opens another PR in the Renku repository for bumping the ui-server component on new tags


ciyer (approved):
👍

PR #1305, opened 2021-04-09
ci: use a smaller footprint image in docker

### LINKED ISSUE #1091: Reduce ui-server image size
Look into using `mhart/alpine-node:slim-10.22` as a base image for the ui-server.

Follow-up to #1043.

lorenzo-cavazzi:
Reduces the footprint of the ui-server chart

Screenshot_20210409_160303

**How to test**: Query the `GET /ui-server` endpoint and verify it returns a valid response

Screenshot_20210409_160423

/deploy #notest

fix #1091

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -11,7 +11,7 @@ COPY src /app/src/
 RUN npm install --silent && \
     npm run-script build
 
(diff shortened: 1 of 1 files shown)


ciyer (approved):
👍

PR #1308, opened 2021-04-12
fix: update environments section name

### LINKED ISSUE #1306: renku config options don't appear properly on the UI
## Description

In the file `./renku/renku.ini` there are options for `renku config`s. For example, setting the `default_url = "/vnc"` will direct the 'connect' button to the vnc:

image

However, other options are not picked up. For example, after running `renku config set interactive.mem_request "4G"` and `renku config set interactive.cpu_request "1"`, these defaults are not selected in the UI:

image

## Reproduce

1. Create an empty project and run  `renku config set interactive.mem_request "4G"` and `renku config set interactive.cpu_request "1"` from the terminal
2. Wait for the image to build and go to the environments page
3. See that the defaults are not shown.

lorenzo-cavazzi:
The UI now supports both the new and the previous environments options section name in the renki.ini file .
This problem should not exist anymore once #1114 will be implemented.

/deploy
fix #1306

### DIFF
--- a/client/src/notebooks/Notebooks.state.js
+++ b/client/src/notebooks/Notebooks.state.js
@@ -29,7 +29,8 @@ import { parseINIString } from "../utils/HelperFunctions";
 const POLLING_INTERVAL = 3000;
 const IMAGE_BUILD_JOB = "image_build";
 const RENKU_INI_PATH = ".renku/renku.ini";
-const RENKU_INI_SECTION = `renku "interactive"`;
+const RENKU_INI_SECTION_LEGACY = `renku "interactive"`;
+const RENKU_INI_SECTION = "interactive";
 
 const PIPELINE_TYPES = {
   anonymous: "registries",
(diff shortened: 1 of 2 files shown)


gavin-k-lee:
Thanks! I just had a look and it looks like the [interactive] section in renku.ini is working as expected, but when I try to change the config settings using `renku config set cpu_request "1"` for example, the changes go into the [renku] section. Is this intentional?

lorenzo-cavazzi:
> Thanks! I just had a look and it looks like the [interactive] section in renku.ini is working as expected, but when I try to change the config settings using `renku config set cpu_request "1"` for example, the changes go into the [renku] section. Is this intentional?

Keep in mind the settings for the environments should be set using the prefix `interactive.`, like in `interactive.<setting_name>`.
The command `renku config set cpu_request "1"` should actually be
```
renku config set interactive.cpu_request "1"
```

This is the reference in the docs for all the available values https://renku.readthedocs.io/projects/renku-python/en/latest/commands.html?available-configuration-values#available-configuration-values

gavin-k-lee (approved):
I just tested it and it looks great! Thanks

ciyer:
> Thanks! I just had a look and it looks like the [interactive] section in renku.ini is working as expected, but when I try to change the config settings using `renku config set cpu_request "1"` for example, the changes go into the [renku] section. Is this intentional?

I make this mistake all the time as well! This will be fixed once https://github.com/swissdatasciencecenter/renku-ui/issues/1114 is finished.

ciyer (approved):
Thanks for quickly putting this fix together. I tested with both old and new settings and the UI handles both correctly. 🎉

PR #1314, opened 2021-04-14
build: set autoscaling for uiserver

### LINKED ISSUE #1092: Set-up autoscaling for ui-server
See the comment https://github.com/SwissDataScienceCenter/renku-ui/pull/1043#pullrequestreview-506107116

It references a [commit in renku-notebooks](https://github.com/SwissDataScienceCenter/renku-notebooks/commit/c9b919a55fe15661a07f52d7bfc926e46ea43410) and [k8s documentation](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/#autoscaling-on-multiple-metrics-and-custom-metrics)

lorenzo-cavazzi:
This should set the autoscaling for the UI server component.
Something is probably wrong because I tried to enable the policy and to load the UI server pod CPU to >50%, but it doesn't spawn a second pod. Maybe @ableuler could help?

fix #1092

/deploy

### DIFF
--- a/helm-chart/renku-ui-server/values.yaml
+++ b/helm-chart/renku-ui-server/values.yaml
@@ -54,7 +54,7 @@ autoscaling:
   enabled: true
   minReplicas: 1
   maxReplicas: 5
-  targetCPUUtilizationPercentage: 80
+  cpuUtilization: 80
 
 resources: {}
   # We usually recommend not to specify default resources and to leave this as a conscious
--- a/helm-chart/renku-ui-server/values.yaml
+++ b/helm-chart/renku-ui-server/values.yaml
@@ -58,8 +58,8 @@ autoscaling:
 
 resources:
   limits:
-    cpu: 500m
-    memory: 2Gi
+    cpu: 100m
+    memory: 128Mi
   requests:
     cpu: 100m
     memory: 128Mi
--- a/helm-chart/renku-ui-server/values.yaml
+++ b/helm-chart/renku-ui-server/values.yaml
@@ -51,7 +51,7 @@ ingress:
 
 ## Configure autoscaling
 autoscaling:
-  enabled: false
+  enabled: true
   minReplicas: 1
   maxReplicas: 5
   targetCPUUtilizationPercentage: 50
--- a/helm-chart/renku-ui-server/values.yaml
+++ b/helm-chart/renku-ui-server/values.yaml
@@ -49,6 +49,13 @@ ingress:
   #    hosts:
   #      - chart-example.local
 
+## Configure autoscaling
+autoscaling:
+  enabled: false
+  minReplicas: 1
+  maxReplicas: 5
+  targetCPUUtilizationPercentage: 50
+
 resources: {}
   # We usually recommend not to specify default resources and to leave this as a conscious
   # choice for the user. This also increases chances charts run on environments with little
--- a/helm-chart/renku-ui-server/values.yaml
+++ b/helm-chart/renku-ui-server/values.yaml
@@ -54,7 +54,7 @@ autoscaling:
   enabled: true
   minReplicas: 1
   maxReplicas: 5
-  targetCPUUtilizationPercentage: 50
+  targetCPUUtilizationPercentage: 80
 
 resources: {}
   # We usually recommend not to specify default resources and to leave this as a conscious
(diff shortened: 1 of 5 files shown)


helm-chart/renku-ui-server/values.yaml
```diff
@@ -49,6 +49,13 @@ ingress:
   #    hosts:
   #      - chart-example.local
 
+## Configure autoscaling
+autoscaling:
+  enabled: true
+  minReplicas: 1
+  maxReplicas: 5
+  targetCPUUtilizationPercentage: 50
```
aledegano, line 57: Isn't this a bit low as default? Usually scaling out at ~80% is enough to ensure enough capacity will be created before the current running processes are swamped.

lorenzo-cavazzi: Good to know! I'm not very expert on this, I was taking inspiration from this https://github.com/SwissDataScienceCenter/renku-notebooks/blob/master/helm-chart/renku-notebooks/values.yaml#L32-L36
Perhaps that should be changed as well

ableuler:
Great that you're adding that from the start. I'd recommend taking a look at the `autoscaling/v2beta2` api. I think it's much nicer and allows you to scale based on memory too.

ableuler:
> Something is probably wrong because I tried to enable the policy and to load the UI server pod CPU to >50%, but it doesn't spawn a second pod. Maybe @ableuler could help?

It averages over one minute, see [here](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/autoscaling/horizontal-pod-autoscaler.md#autoscaling-algorithm) could that be the reason for it not scaling out?

lorenzo-cavazzi:
@ableuler I've updated to `autoscaling/v2beta2` and fixed the deployment (the ui-server component wasn't updated correctly).
Not sure if I'm testing in the correct way: I tried to put the cpu under pressure by running `yes > /dev/null &` a few times until I got to >90%, but I still get no more than 1 container in the UI server pod even if I wait a few minutes

Screenshot_20210414_184703

rokroskar:
@lorenzo-cavazzi it might help to see what the horizontal pod autoscaler is doing; you can see its status via `kubectl`, e.g. `kubectl -n renku-ci-ui-1314 get hpa -oyaml`. The utilization you see via `top` might be different than what the metrics server reports - the latter is what is used by the HPA. You can use `kubectl top pod <podname>` to get the numbers from the metrics server.

helm-chart/renku-ui-server/values.yaml
rokroskar: We set this as the default in the notebooks service for no reason other than it's what is used in the documentation about HPA. It will need to be adjusted, surely.

ableuler:
> Something is probably wrong because I tried to enable the policy and to load the UI server pod CPU to >50%, but it doesn't spawn a second pod.

For completeness here: the problem was that there were no resource limits set on the deployment spec. After adding those, the service scales as expected.

lorenzo-cavazzi:
Thank you @rokroskar , those commands are very useful!
As mentioned by Andreas, the specific problem was solved by setting the resources (see commit c92e00f132aebbc15480497b865c8eb4c9f32772 ).
Now scaling up (and down) seems to work well

Screenshot_20210415_093141

I ended up setting low resources in the `requests` section, while the `limits` are a bit higher. Let me know if we follow any other best practice.

rokroskar:
Ah great! I assumed that might be the problem but when I checked the pod it had limits/requests already :)

helm-chart/renku-ui-server/values.yaml
```diff
@@ -49,17 +49,20 @@ ingress:
...
+autoscaling:
+  enabled: true
+  minReplicas: 1
+  maxReplicas: 5
+  cpuUtilization: 80
+
+resources:
+  limits:
+    cpu: 500m
+    memory: 2Gi
```
ableuler, line 62: Is that realistic? A smaller number here would help the scheduler to place these pods correctly. If you're worried that you might run out of memory, you could enable scaling on memory too.

lorenzo-cavazzi: Like `1Gi` or something closer to the requested amount?

ableuler: Yes, preferably closer to the request. If you're using 5 times the requested amount and memory gets tight on the node, this pod will be a prime candidate for eviction.

lorenzo-cavazzi: Ok I set them down to
```
cpu: 100m
memory: 128Mi
```

ableuler (approved):

PR #1315, opened 2021-04-14
feat(design): redesign of project files tab

vfried:
This is not 100% polished, there are still changes we could do to make it look better.

I made an attempt to improve the way the button group looks (Content/lineage) but I'm not 100% sure i like it.
There is also the option to make the background light blue.

Result...
image

/deploy


ciyer (changes requested):
Both code and results look great! I have a few small suggestions for changes in the code. I noticed that the scrolling behaviors of the file list and main content do not combine very well for large files (example https://renku-ci-ui-1315.dev.renku.ch/projects/time-series/data/files/blob/data/generated/linear/0_1_norm(0,%201).csv), but I think that is not part of this PR and something we can fix elsewhere.

vfried:
@ciyer 

I did the small changes that were requested. 

After trying a bit with buttons I decided to leave the buttons as they are now, I would like to explore this a bit more but I think we can do this in a separate PR, same with the scrolling behavior, I would prefer to do this seaparatelly as well.

I think we should also consider adding some option so that the users can expand and shrink the files tab on the side since when names are long it's hard to read the files names.

ciyer (approved):
Great! Yes, please create issues for the unresolved items.

PR #1316, opened 2021-04-16
feat(design): redesign of datasets inside projects

vfried:
Redesign of datsets inside projects

image

image

image

image

/deploy

### DIFF
--- a/client/src/dataset/Dataset.present.js
+++ b/client/src/dataset/Dataset.present.js
@@ -19,14 +19,14 @@
 import React, { useState } from "react";
 import {
   Alert, Button, Card, CardBody, CardHeader, Col, DropdownItem, DropdownMenu, DropdownToggle, Row,
-  Table, UncontrolledButtonDropdown,
+  Table, UncontrolledButtonDropdown, Badge
 } from "reactstrap";
 import { Link } from "react-router-dom";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import {
-  faEllipsisV, faExternalLinkAlt, faInfoCircle, faPen, faPlus, faSearch, faTrash
+  faEllipsisV, faInfoCircle, faPen, faPlus, faSearch, faTrash
 } from "@fortawesome/free-solid-svg-icons";
-import { ErrorAlert, FileExplorer, InfoAlert, Loader, RenkuMarkdown } from "../utils/UIComponents";
+import { ErrorAlert, FileExplorer, InfoAlert, Loader, RenkuMarkdown, ExternalLink } from "../utils/UIComponents";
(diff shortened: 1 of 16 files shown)


ciyer (changes requested):
Code looks good! I have some cosmetic changes to suggest.

image

# Cosmetics

1. Show the source, author, keywords metadata as a table. Make the row header bold, no ":" necessary
2. So the keywords as badges, the way we show tags elsewhere
3. Move description above "Dataset Files" and apply a white background to that section
4. Move the "Published date" information to the same line as the title of the dataset

# Bugs

Children should be indented in the files list. Currently, it looks like all the files are in the root directory.

# Improvements?

If these changes are easy to do, then do them, otherwise we can create a new issue for this:
- Do not show "Source" and "DOI" separately. There should just be one field displayed, `Source`. **If** the there is a DOI, show it here, otherwise, **if** the dataset is imported from another project, show the source here, otherwise show nothing
- In the "Used In Projects" section, only show *other* projects that use this dataset.

vfried:
Cosmetics

DONE - Show the source, author, keywords metadata as a table. Make the row header bold, no ":" necessary
So the keywords as badges, the way we show tags elsewhere
DONE - Move description above "Dataset Files" and apply a white background to that section
DONE ** - Move the "Published date" information to the same line as the title of the dataset ( I decided to put this in the table instead)

Bugs 
Children should be indented in the files list. Currently, it looks like all the files are in the root directory. --> this is not a bug, it was implemented in the files view, once that is merged all should be OK.

Improvements?

PARTLY DONE - Do not show "Source" and "DOI" separately. There should just be one field displayed, Source. If there is a DOI, show it here, otherwise, if the dataset is imported from another project, show the source here, otherwise, show nothing --> there is no way to know if a project was imported from another project (at least at the moment), the source is only a link to the dataset "detached" from a project I'm not sure we want to hide source when we have doi but I implemented this. We could get a request to add the source even when there is a DOI.
DONE - In the "Used In Projects" section, only show other projects that use this dataset.

vfried:
image

vfried:
When the dataset is outside of the project the projects table is displayed...
(files are fixed in another PR, this is a generic component)

image

vfried:
With card on table info, i think it looks better without this... 
image

ciyer:
This looks good! I agree with you that we *should not* show the metadata table as a card. It looks better on the gray background. Could you fix revert that, and then it is ready to merge!

ciyer (approved):

vfried:
I added some small changes, if the description is empty we don't show the empty description card.

If there are no files we show the empty files card but with a message that says that the dataset doesn't have files.

image

ciyer (approved):
👍

PR #1317, opened 2021-04-16
ci: increase helm test timeout

lorenzo-cavazzi:
It seems that the CI tests are often failing due to a 40 minutes timeout in the `helm test` command.
We should increase that, as already done in the main renku repository SwissDataScienceCenter/renku#2041

/deploy


ciyer (approved):
👍

Issue #1219, opened 2021-01-29
UX (sessions) : Improved session list


ciyer:
## Description

Make the session listing more powerful. We can drop the session list from the project page -- there should just be one global session list.

## Mock-ups

![image](https://user-images.githubusercontent.com/1196411/107025617-f86da000-67a9-11eb-8363-d443eff02597.png)

## API

The UI will need to able to get from the session the following information:

- uncommitted work yes/no?
- clone ahead/behind origin?
- last push to autosave branch

And the UI will need to be able to invoke the following commands on the session:
- push to autosave branch
- renku save (with message provided)

## Progress

- [x] API design (sprint-2021-02-19)
- [ ] Status handling
- [ ] Saving

rokroskar:
Just an idea - but maybe clicking on "sessions" in the project could take the user to the global project page but restrict the search to show only the project they just came from? With an obvious way to remove the filter so they can see all the sessions.

rokroskar:
regarding "last push to autosave branch" - this implies that these pushes are periodic, but this is not the case. Or did you have something else in mind?

rokroskar:
I just tried accessing git information via curl using the session cookie and the git extension endpoint and it works  - so I don't think this issue is really blocked by https://github.com/SwissDataScienceCenter/renku-python/issues/1659

ciyer:
- That's a good idea to use the git extension endpoint. 
- The pushes to the autosave branch should be periodic.

rokroskar:
We decided to not make them periodic initially for a few reasons, but the main one is that in the case that large files need to be pushed it could have detrimental effects on the user's session. In order to do periodic pushes, you need to have a mirror of the repo somewhere because you need to change the branch, commit, and push. I don't see a good way of doing this automatically, but maybe I'm missing something obvious?

ciyer:
If we do not intend to save periodically, then it should not be called "autosave". We should change it to "recovery" or something similar.

rokroskar:
sure, we can give it a different name - the more important question I think is whether we want to enable periodic saves or not. Maybe we can actually do better by having dedicated re-usable volumes instead, for example.

ciyer:
Note for implementor: check that the API provided by the Jupyter git extension in available if the user has launched RStudio (it probably is, but we should just test).

rokroskar:
[...]

Issue #1152, opened 2020-12-03
Simplified launch of environments


ciyer:
## Description

With the ability to specify environment settings defaults, there is no need to ask the user so many questions to launch an environment. 

There should be a start/connect button that the user can just click to start an environment with the default settings (no form to fill out). If an environment is already running for the project, then the user can connect to it. There is an "advanced" start option that will let the user fill out the form.

The form should also be simpler. The advanced options can be hidden by default.

## Differences to mock-up

In the mock up, there is no environments tab, but we will make that change later in another issue. Let us start by simplifying launch first.

In the mock up there is a the expectation that start becomes connect if a session is running, but that feature is being blocked by the UI server, so let us leave the button as start always and when clicked, the UI will decide if it should start a new server or connect.

The start form should have a back to project button.

## Mock Ups

![image](https://user-images.githubusercontent.com/1196411/106929331-872fdd80-6714-11eb-89c8-844774fb48f3.png)

lorenzo-cavazzi:
Tip: remember to check for autosaved content from the new `GET /autosaves` endpoint SwissDataScienceCenter/renku-notebooks/issues/584

PR #1318, opened 2021-04-19
feat(design): small fixes for environments inside projects

vfried:
Deleted side tabs, so now when you enter environments inside projects you see this...
image

Form switch fix...
image

Shall we delete the info message?
Shall we make the width of all forms be 12 (max)?

/deploy

### DIFF
--- a/client/src/notebooks/Notebooks.present.js
+++ b/client/src/notebooks/Notebooks.present.js
@@ -81,7 +81,8 @@ class Notebooks extends Component {
       <Fragment>
         <Row className="pt-2 pb-3">
           <Col className="d-flex mb-2 justify-content-between">
-            <NotebooksTitle standalone={this.props.standalone} />
+            <NotebooksHeader standalone={this.props.standalone}
+              urlNewEnvironment={this.props.urlNewEnvironment}/>
           </Col>
         </Row>
         <NotebookServers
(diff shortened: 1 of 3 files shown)


ciyer:
I like these changes! I think getting rid of the tabs is a nice, simplifying improvement. As for your questions, I do not think we should make those changes.

1. I think the info message is helpful and better than seeing an empty sessions list.
2. Let us not worry about the layout of this version of the sessions list because it will change very soon: https://github.com/SwissDataScienceCenter/renku-ui/issues/1219

This PR looks like it branches off another PR with some unrelated changes. I will wait until those changes are merged to review and approve the code.

vfried:
@ciyer thanks for taking a look at this PR.

I just deleted the first commit from this PR, it's "clean" now so that the code can be reviewed.

Regarding the info message, I'm ok with it but then we should add it on every list that is empty to be consistent.

ciyer:
> I just deleted the first commit from this PR, it's "clean" now so that the code can be reviewed.

Code looks good! 🎉 
 
> Regarding the info message, I'm ok with it but then we should add it on every list that is empty to be consistent.

I think it makes sense everywhere where the user can create something (Project list, Dataset list, assuming user has write access to project, etc.)
 
> **I have another concern with this PR** should we make the width of the form be 12? the form inside datasets is but that is a bigger form.

I want to switch to a 2-column layout for the form, as here:

  https://github.com/swissdatasciencecenter/renku-ui/issues/1152

image

If it is an easy fix, we can go ahead and do it now, otherwise, we can merge this PR and make that improvement in another PR.

ciyer (approved):
👍

vfried:
Let's create a new PR for the two columns form, I want to go ahead and merge these PRs so we can rebase against master

PR #1319, opened 2021-04-20
feat(design): notifications redesign

vfried:
Redesign of notifications.
For preview go to the new dataset form (test notifications are triggered here)

Bold text looks blurry but this should be fixed with fonts added to a different PR.

Toasts
image

Menu
image

List
image

/deploy

### DIFF
--- a/client/src/notifications/Notifications.css
+++ b/client/src/notifications/Notifications.css
@@ -16,39 +16,25 @@
 /* Customization */
 
 .Toastify__toast {
-  opacity: 0.90;
+  width: 350px;
+  max-width: 100%;
+  font-size: .875rem;
+  pointer-events: auto;
+  background-color: rgba(255,255,255,1) !important;
+  background-clip: padding-box;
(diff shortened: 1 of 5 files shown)


ciyer (changes requested):
Code looks good, but there are a few cosmetic things:

1. The background for the dismiss button should be white like the rest of the header. It is semi-transparent. image
2. I'm not convinced that the "Notifications Menu" interaction is the way to go, but it is an interesting idea that I hadn't thought of. I ok with leaving it in for a bit to see how it feels. However, after dismissing notifications, the notifications menu item is disconnected from the menu. 
image
3. Could we remove the timer indicator? 
image Watching it makes me nervous. :) Could we just have the toast displayed for a bit and then fade out using transparency?

ciyer (commented):
The issues that were raise have been fixed, thanks! Many files have been changed, though -- do you know the reason for that?

ciyer (approved):
👍

Issue #1264, opened 2021-03-09
Use ui-server to poll notebook session state


ciyer:
# Description

Use the ui-server to poll for changes in notebook session state and pass them to the UI asynchronously. The information provided to the UI should also include the disk, CPU, and memory usage from the jupyterServer custom resource.

This information will be used in the UI during launch (since a session may not start if disk space limits are reached) and on running sessions.

lorenzo-cavazzi:
This is mostly a duplicate of #2016 . Closing this since the other contains more details

PR #1323, opened 2021-04-21
Add debugging tools to uiserver

lorenzo-cavazzi:
* Add tools for debugging when running through Telepresence
* Add libraries for supporting WebSockets and update the ones already available
* Lint code with the current rules (prevents warnings from ESlint in editors like VScode)

This is a preparation for #1264 , but it doesn't specifically address that issue.

**HOW TO TEST**: clone this branch locally, run `npm install` on the `/server` folder, and start telepresence using the `/server/run-telepresence.sh` script. It will show instructions on how to run in debug mode and how to set the configuration file in VScode to use the integrated remote bedugging for NPM.


ciyer (approved):
This is very nice! I added a commit to fix a couple of typos.

PR #1333, opened 2021-04-29
refactor: update property on datasets kg query

### LINKED ISSUE #1275: Changes in the GET /knowledge-graph/datasets/:id
## Motivation

The KG team is moving away from the `isPartOf` property returned by the `GET /knowledge-graph/datasets/:id`. Instead, there's a new property named `usedIn` which contains exactly the same data as `isPartOf`. The KG team hopes the change should suggest that projects listed under the `usedIn` field are not necessarily projects which contain the dataset with the requested `id` but somehow use a different instance of the same dataset.

## Proposal
Could UI simply start to use the `usedIn` field instead of `isPartOf`? The KG team would like to remove the `isPartOf` field in one of the next releases.

lorenzo-cavazzi:
Replace the `isPartOf` property with `usedIn` on the query `GET /knowledge-graph/datasets/:id` as request from the KG team.

/deploy
fix #1275

### DIFF
--- a/client/src/api-client/test-client.js
+++ b/client/src/api-client/test-client.js
@@ -155,7 +155,7 @@ const methods = {
     ],
     "url": "https://dev.renku.ch/datasets/79215657-4319-4fcf-82b9-58267f2a1db8",
     "sameAs": "https://dev.renku.ch/datasets/79215657-4319-4fcf-82b9-58267f2a1db8",
-    "isPartOf": []
+    "usedIn": []
   }
 };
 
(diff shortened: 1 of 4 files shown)


lorenzo-cavazzi:
**how to test**: fork a project with a dataset and be sure it gets indexed in the KG.
Open a dataset and check the `Projects using this dataset` section appears on the bottom referencing at least 1 project.

https://renku-ci-ui-1333.dev.renku.ch/projects/lorenzo.cavazzi.tech/dataset-import/datasets/missing-lfs-objects/

Screenshot_20210430_095515

ciyer (approved):
Glad to see that this turned out to be easy!

PR #1335, opened 2021-04-30
feat(design): collaboration tab redesign

vfried:
image

image

image

image

image

image

/deploy


ciyer (approved):
I think these are nice improvements! Let us merge them; if we decide to continue with our own collaboration pages, then this is a good starting point for further improvements.

PR #1343, opened 2021-05-06
Implement a style guide

### LINKED ISSUE #1322: Style Guide
# Description

Create a page in the UI that explains the style guide. This page should not have a link, but we can use it internally to aid our development.

ciyer:
Implement a style guide for the new Renku UI.

Fix #1322.

The style guide can be accessed under https://renku-ci-ui-1343.dev.renku.ch/style-guide

/deploy

### DIFF
--- a/client/src/styleguide/StyleGuide.js
+++ b/client/src/styleguide/StyleGuide.js
@@ -1,0 +1,238 @@
+/*!
+ * Copyright 2021 - Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ *  renku-ui
+ *
+ *  StyleGuide
+ *  The StyleGuide explains the components of the RenkuLab UI.
+ */
+
+import React, { Fragment } from "react";
+
+import { Col, Nav, NavItem, Row, Table } from "reactstrap";
+import { Route, Switch } from "react-router-dom";
+
+import {
+  RenkuNavLink, TimeCaption,
+} from "../utils/UIComponents";
+
+import ButtonsGuide from "./ButtonsGuide";
+import ColorsGuide from "./ColorsGuide";
+import FormsGuide from "./FormsGuide";
+import ListsGuide from "./ListsGuide";
+
+
+function Overview(props) {
+  return <Fragment>
+    <h2>Overview</h2>
+    <p>The style guide explains the different elements of the RenkuLab UI, how they should
+      look and when to use what element.
+    </p>
+  </Fragment>;
+}
+
+function FontsGuide(props) {
+  return <Fragment>
+    <h2>Fonts / Typography</h2>
+    <p><b>Font</b> <a href="https://www.ffonts.net">Calcutta</a></p>
+    <h2>Styles</h2>
+    <Table>
+      <thead>
+        <tr>
+          <th scope="col">Use</th>
+          <th scope="col">Tag</th>
+          <th scope="col">Details</th>
+          <th scope="col">Appearance</th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr>
+          <th scope="row">Page titles</th>
+          <td>h2</td>
+          <td>weight: 500; size: 32px;</td>
+          <td><h2>Lorem ipsum</h2></td>
+        </tr>
+        <tr>
+          <th scope="row">Section titles</th>
+          <td>h3</td>
+          <td>weight: 500; size: 28px;</td>
+          <td><h3>Lorem ipsum</h3></td>
+        </tr>
+        <tr>
+          <th scope="row">Normal text</th>
+          <td>p/div</td>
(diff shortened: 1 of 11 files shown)


client/src/styleguide/StyleGuide.js
```diff
@@ -0,0 +1,226 @@
...
+          <h2>
+            Style Guide
+          </h2>
+          <div className="text-rk-text">
+            An guide to the RenkuLab UI elements.
+          </div>
+        </div>
+        <div className="d-flex flex-column align-items-end justify-content-between">
+          <div>
+            Lorem ipsum dolor sit amet
```
lorenzo-cavazzi, line 164: Did this end here by mistake?

lorenzo-cavazzi:
Great to see this reference! I'm already using it for my first PR with the new design.
I noticed one minor detail (there is an unexpected misaligned "Lorem ipsum" floating on the right in the preview).

Otherwise, it looks ready for me but I'll let @vfried review it since she has already worked extensively on the new design.

client/src/styleguide/StyleGuide.js
ciyer: No, that was intentional. I just wanted a little bit of text there.

lorenzo-cavazzi: I see.
To my eyes, the content on the right looks misaligned -- there are 3 elements (the text, the top nav, and the page nav) that end at different points.

Screenshot_20210506_184914

ciyer: I see what you mean. I made the text right justified so it lines up with the "Overview" tab.

vfried (commented):
Thanks for doing this, i think this will help us be consistent across all renkulab. I really like the way this is looking, I have a few things to point out:

1. I see things from the search bar and search results missing (I would maybe add them), there are buttons with specific style and also the search result content.

2. Bootstrap has a guide and a tool to lighten and darken colors... here is the codepen for it https://codepen.io/emdeoh/pen/zYOQOPB and here is the reference https://getbootstrap.com/docs/5.0/customize/color/ I thi
nk this way of making colors lighter and darker is a bit nicer.

Here, you can see things for the renku green... i would use colors from the mix function instead of the lighten and darken.
image

Note that for the very dark blue there aren't soo many choices, but i still like the mix more than the lighten/darken functions.

Success color:
we could use something like this... 
image

I lightened a bit the color from Luca presentation here... and put it in the codepen from bootstrap, then i chose this colors
image

We could also just use the renku logo color for success and primary, i think if we add icons and things like that it should be OK.

All other colors seem good to me, we could add some more tones of grey, i think i use the rk-text color quite a lot. 

It would be nice to have also an info color but I'm not sure what i would put there. Maybe for info we should just use a sliglty gray background in white and white background in gray. Or a lighter primary (rk-blue) color.


3. Buttons: 
- I would add link style inside the buttons
- I would add outlined buttons, I've used them in lists (notifications) when i want to make things less prominent/colorful.
- The loader should be blue (this is something we might need to fix outside of this)

4. Third level nav should be included inside the second level nav (like collaboration tab).

5. I would consider lighter buttons inside forms...not for the submit buttons but for buttons inside the form (inside datasets there is an example for this) If we wanted to use this i could make a more generic button and make sure that click, hover, etc look good on them.

6. I would consider making the background for selected elements inside lists gray as in the pagination... 
image

7. Titles inside datasets and projects is h3 instead of h2.

ciyer:
Thanks for the thorough review! I made the changes to the buttons page that you suggested (outline buttons, regrouping of the rows, fix color of loader). 

Re: third-level navs

I do not want third-level navs. I think that is too deep and confusing to navigate. I didn't document them because I think we should get rid of them.

> 7. Titles inside datasets and projects is h3 instead of h2.

Can you give me an example? I see h2 when I look at the DOM.

I agree with many of your other comments, but I think those should be addressed in follow-up PRs. See detailed responses below:

> 1. I see things from the search bar and search results missing (I would maybe add them), there are buttons with specific style and also the search result content.

> 5. I would consider lighter buttons inside forms...not for the submit buttons but for buttons inside the form (inside datasets there is an example for this) If we wanted to use this i could make a more generic button and make sure that click, hover, etc look good on them.

> 6. I would consider making the background for selected elements inside lists gray as in the pagination...
>    image

I think these are good all points, but, as discussed, let us leave these sections as they are right now. We should revisit and update those in other PRs (maybe when other work is completed)

> 2. Bootstrap has a guide and a tool to lighten and darken colors... here is the codepen for it https://codepen.io/emdeoh/pen/zYOQOPB and here is the reference https://getbootstrap.com/docs/5.0/customize/color/ I think this way of making colors lighter and darker is a bit nicer.

That is a good way to lighten and darken colors generally, but the goal here is not to give unlimited flexibility, instead the point is to sepcify our colors and try to provide one variation on it designed for using a backround with dark text and one as a background for light text (this isn't not feasible for every color, though).  

But maybe it is overkill and not necessary. We can drop it if you think it is confusing or not useful.

> We could also just use the renku logo color for success and primary, i think if we add icons and things like that it should be OK.

I'm against using the renku logo color for situations where we want to assign meanings to the colors (success, failure, etc.) I think we should work to ensure the logo colors remain neutral and use other colors if we have a meaning in mind.

> All other colors seem good to me, we could add some more tones of grey, i think i use the rk-text color quite a lot.

:+1:
 
> It would be nice to have also an info color but I'm not sure what i would put there. Maybe for info we should just use a sliglty gray background in white and white background in gray. Or a lighter primary (rk-blue) color.

The color that has been assigned to info is `rk-pink`. Should we change that?

vfried:
- Third level navigation ---> since we are replacing the collab features for the iframe at the moment we don't need anything, if in the future we need them we will have to find a solution to have them or not to have them. So let's not document them.

- h2 h3 example: (project name h2 dataset name h3)
image

- Colors: as I stated before I don't really like the success color, I find it a bit too bright this of course a personal opinion. I think this is something that we should ask Luca in the future this is 100% design work in my opinion. Regarding the info color, if the logo colors are neutral we should maybe make the info some other color (i would be up for a gray or some type of light blue). I feel on the color aspect we got under-delivered, we got a nice palette color but for doing UI work it's insufficient. If we keep working with Luca we should clearly communicate that he needs to fix this or otherwise we will have to get some other designer that does this for us the way we need it.

- Info color some other color---> yes

PR #1344, opened 2021-05-07
feat: show environments in iframe

### LINKED ISSUE #1220: UX (session) : sessions in iframe
## Motivation

We want to provide controls for a session to make them easier to use. The session should be embedded in an iframe and the controls will be around.

## Mock-ups / Design

image

## Progress

- [x] API Design (sprint-2021-01-29)
- [ ] Initial UI Implementation (#1344)
- [ ] Improvements
    - show commit message
    - dropdown to switch endpoints (e.g., from JL to RStudio)
    - cheatsheet integration

lorenzo-cavazzi:
Embed running sessions in RenkuLab using an `<iframe>`, allowing use to also access logs and docs on the same page.

From the test environment, it seems that SwissDataScienceCenter/renku-notebooks#606 is not a requirement since the iframe's content is loaded correctly even without that PR.

Screenshot_20210510_162201

/deploy renku=ui-design-test-update-iframe
fix #1220

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event.action != 'closed'
     steps:
-    - uses: rokroskar/workflow-run-cleanup-action@v0.3.0
+    - uses: rokroskar/workflow-run-cleanup-action@v0.3.3
       env:
         GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
 
--- a/.github/workflows/test-and-ci.yml
+++ b/.github/workflows/test-and-ci.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     if: "!startsWith(github.ref, 'refs/tags/')"
     steps:
-    - uses: rokroskar/workflow-run-cleanup-action@v0.3.0
+    - uses: rokroskar/workflow-run-cleanup-action@v0.3.3
       env:
         GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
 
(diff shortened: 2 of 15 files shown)


lorenzo-cavazzi:
I addressed the points in the review, apart from compressing the project data above the session.

That would require a bit of work since the `<Project>` component would need to be aware of the specific page open on the Sessions tab. Since that will completely change soon (see #1219 on detaching the sessions from the project ), I'm wondering if it's worth putting in the effort, or rather postpone that to a later point.

The suggested change wouldn't expand the actual iframe area (scrolling down a bit in the current output brings to the same result). Would it make sense to think about a solution to maximize the iframe, expanding it further also on the left and right sides?

ciyer:
Thanks for addressing those requests. I made a small commit to extract a shared function for formatting the resource list.

As far as compressing the project info, yes, let us hold off on this. However, I noticed that if I switch to the "Log" tab, it causes the page to lose scroll position and return to the top. I can switch between the Jupyter and documentation tabs without this happening. Is this easy to fix?

lorenzo-cavazzi:
> I made a small commit to extract a shared function for formatting the resource list.

Great, thank you!

> Is this easy to fix?

Short answer: yes. I pushed a small change for that.

Long answer: switching to the logs refreshes the content, temporarily showing a spinning wheel that reduces the page height.
Since we force a minimum height for the `iframe`, we can easily do the same for the logs. This prevents losing the vertical position and shouldn't negatively impact since the logs usually take (much) more space once loaded.

ciyer (approved):
The changes you made make it a much smoother experience. Thanks!

lorenzo-cavazzi:
Acceptance tests need to be adapted before merging this.
We should wait for #1347 to be closed, and update the tests on top of SwissDataScienceCenter/renku#2119

PR #1351, opened 2021-05-17
feat: handle non blocking errors on fork

### LINKED ISSUE #1341: Improve error messages when fork fails
## Motivation

The fork operation is a two step process: 
1. Fork the project in GitLab
2. Run the CI pipeline to build images

Sometimes the first succeeds, but the second fails. The error message currently just says `An error occurred while forking: Renku API error`. This does not tell the full story.

See https://renku.discourse.group/t/error-during-forking-process-but-it-actually-finished-correctly/321

## Proposal

If the error occurs with the CI pipeline, we should show a different (non-error) message.

`The project was forked, but the images were not be built. You may need to build images before starting an interactive session.`

lorenzo-cavazzi:
This improves the UX when a non-fatal error occurs while forking (i.e. the pipeline or webhook creation fails)

**How to test**: it's a bit tricky. The easiest way is running telepresence on the PR environment and manually modifying the code in the `client.forkProject` function to trigger the error on the target API (try with the main fork API, then with `runPipeline` and `createGraphWebhook`)

Screenshot_20210517_144720

P.S. mind that the error message is slightly different than the suggested one, to keep it coherent with the other cases. Feel free to suggest changes if it's not clear (or change it directly in the PR).

/deploy renku=ui-design-test-fix
fix #1341

### DIFF
--- a/client/src/api-client/project.js
+++ b/client/src/api-client/project.js
@@ -230,11 +230,25 @@ function addProjectMethods(client) {
 
     // Wait 1 second before starting the pipeline to prevent errors
     await new Promise(r => setTimeout(r, 1000));
+
     // Start pipeline -- no need to wait for the outcome, the environment page handles this
-    const pipeline = await client.runPipeline(forkedProject.data.id);
+    let pipeline;
+    try {
+      pipeline = await client.runPipeline(forkedProject.data.id);
+    }
+    catch (error) {
+      pipeline = error;
+    }
 
     // Create KG webhook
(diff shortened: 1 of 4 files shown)


ciyer (approved):
I think this will solve UX issue and clarify what happened when a fork only partially succeeds. I added a commit to change the messages slightly.

lorenzo-cavazzi:
I pushed a minor change to prevent linting errors

ciyer (approved):

PR #1353, opened 2021-05-17
feat: support svg preview and fix image path

### LINKED ISSUE #1339: Fix image rendering for images with relative paths in markdown/readme files
## Description

Readmes containing images with relative paths are not being rendered in readme files, we should fix this.

## Reproduce

Go to this project and see that the images aren't being rendered
https://dev.renku.ch/projects/lorenzo.cavazzi.tech/aiida-2

We had fix this issue here https://github.com/SwissDataScienceCenter/renku-ui/pull/1008 but it looks like something isn't working.

lorenzo-cavazzi:
This PR adds the following:
* Support SVG files in both file preview and markdown preview.
* Fix markdown preview when the relative path starts with a slash.

**How to test it**: try adding images (both `svg` and other extensions) with relative paths starting either with or without `/`
Here are a few examples, compare them with the PR preview:
* https://dev.renku.ch/projects/lorenzo.cavazzi.tech/test-images/files/blob/images/RENKU_Symbole_BLACK_PROD.svg
* https://dev.renku.ch/projects/lorenzo.cavazzi.tech/test-images

/deploy
fix #1339


client/src/utils/Markdown.js
```diff
@@ -183,7 +186,10 @@ function RenkuMarkdownWithPathTranslation(props) {
     let currentBlock = filesRefs.find(block => file.src.endsWith(block.refPath));
     if (currentBlock && currentBlock.data) {
       if (currentBlock.type === REF_TYPES.IMAGE_PREV) {
-        file.src = "data:image;base64," + currentBlock.data.content;
+        const subType = currentBlock.data.file_name.endsWith(".svg") ?
+          "/svg+xml" :
+          "";
+        file.src = `data:image${subType};base64,${currentBlock.data.content}`;
```
ciyer, line 192: Would it be possible to extract this code, together with the image extensions, to share code between this file and `File.container.js`?

ciyer (changes requested):
I tested and the problem is fixed. I think it would be nice to share the image rendering code to ensure we don't extend the code in one place and forget about the other.

client/src/utils/Markdown.js
lorenzo-cavazzi: Good point! I created a shared function to handle the string required as `src` attribute

ciyer (approved):
👍

PR #1354, opened 2021-05-19
Handle markdown in GitLab description

### LINKED ISSUE #1346: GitLab project description is not formatted correctly
## Motivation

The UI doesn't allow the user to set a project description when creating one, but it still allows to change it on the settings page. When we display it in the project overview, the content is not formatted properly

**RenkuLab**
image

**GitLab**
Screenshot_20210511_075845

## Proposal
We should either:
- Format the content properly, or
- ~Remove the support entirely~ _the previous solution sounds more reasonable_
P.S. double-check the output in the project list, where keeping the plain text may be better (we don't want to render an image or anything else too disruptive for the layout)

lorenzo-cavazzi:
The GitLab description field is now translated to HTML when it contains markdown.
The first commit addresses the project overview page (it also changes a bit the header layout, see screenshots), while the second changes the preview in the project list.

**How to test**: try to search for some projects with a description and open the overview page.
E.G. https://dev.renku.ch/projects/lorenzo.cavazzi.tech/test-comment

Here are a few screenshots: on the left the updated version, on the right the previous one.

* PROJECT HEADER - large page
  Screenshot_20210519_155114

* PROJECT HEADER - small page
  Screenshot_20210519_155138

* PROJECT LIST
  Screenshot_20210519_164819

/deploy renku=ui-design-test-fix
fix #1346

### DIFF
--- a/client/src/project/Project.present.js
+++ b/client/src/project/Project.present.js
@@ -47,7 +47,9 @@ import { SpecialPropVal } from "../model/Model";
 import { ProjectAvatarEdit, ProjectTags, ProjectTagList } from "./shared";
 import { Notebooks, StartNotebookServer } from "../notebooks";
 import Issue from "../collaboration/issue/Issue";
-import { CollaborationList, collaborationListTypeMap, itemsStateMap } from "../collaboration/lists/CollaborationList.container";
+import {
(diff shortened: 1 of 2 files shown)


ciyer (approved):
This works well, and the code looks good!

PR #1359, opened 2021-05-25
fix: handle namespaces that start with numbers

### LINKED ISSUE #1257: Can't access a project when the namespace starts with a number
## Description

When trying to access a project in a namespace starting with a number, the UI hangs with a loader and the text "Identifying project number <namespace>..."

Screenshot_20210301_150410

## Reproduce

Try to open https://dev.renku.ch/projects/123numeric/numeric-group-project

## Context

Checking the Network, it seems that the query to the `/api/project` endpoint ignores everything after the `/`.
The UI is probably treating this case as a numeric id project. Furthermore, the support for that case seems to be broken.

ciyer:
Fix #1257

# Testing

Namespaces that start with numbers are now correctly handled.

Compare the following:

New: https://renku-ci-ui-1359.dev.renku.ch/projects/123testgroup/testing-project
Old: https://dev.renku.ch/projects/123testgroup/testing-project (endless spinning wheel)

New: https://renku-ci-ui-1359.dev.renku.ch/projects/123testgroup/987test-project
Old: https://dev.renku.ch/projects/123testgroup/987test-project (endless spinning wheel)

If you wish, create new project in a namespace that starts with a number and see that you get properly redirected to the project after it is created.

/deploy

### DIFF
--- a/client/src/project/Project-Route.test.js
+++ b/client/src/project/Project-Route.test.js
@@ -87,6 +87,18 @@ describe("basic route extraction", () => {
     expect(pathComponents.baseUrl).toEqual("/projects/namespace/project-name");
     expect(pathComponents.namespace).toEqual("namespace");
   });
+  it("handles project-with-numeric-namespace routes", () => {
+    const pathComponents = splitProjectSubRoute("/projects/1namespace/project-name");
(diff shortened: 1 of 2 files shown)


lorenzo-cavazzi (approved):
Nice fix!

PR #1360, opened 2021-05-25
feat(design): new grid display for project search

### LINKED ISSUE #1327: Add a new display option for projects
Users should be able to choose between seeing project on a list or grid.

The list is ready, we should do the grid as in Luca's designs.

image

I think we could make this the only option on smaller screens but we can explore this idea.

vfried:
New option for users to switch to/from grid display in the project search...

image

- At the moment we are not using infinite scrolling but we could implement this.
- We should make a component from this and replicate this in other parts of the ui.
- I had to use a component for the grid since bootstrap doesn't support this (I used masonry since it's what bootstrap recommends using)

/deploy renku=ui-design-2021

Closes #1327


client/src/project/list/ProjectList.container.js
```diff
@@ -65,7 +65,7 @@ const DEFAULT_USERS_GROUPS = { fetched: null, fetching: null, list: [] };
 const DEFAULT_PARAMS = {
   query: "",
   page: 1,
-  perPage: 10, // TODO: change to 10
+  perPage: 12, // TODO: change to 10
```
ciyer, line 68: Should this be 10 or 12?

client/src/project/list/ProjectList.container.js
```diff
@@ -269,6 +269,8 @@ function useUserProjectSearch(client, params, targetUser, setParams, setProjects
 
     // Prepare fetching user or group projects
     setProjects(p => ({ ...p, fetched: null, fetching: true }));
+
```
ciyer, line 272: A `log` call that slipped in.

client/src/styles/components/_renku_search.scss
```diff
@@ -121,3 +125,29 @@
     }
   }
 }
+
+.rk-search-result-grid {
+  display: -webkit-box; 
+  display: -ms-flexbox;
```
ciyer, line 131: Do we need this? I think flexbox is supported by all browsers we target.

client/src/styles/components/_renku_search.scss
```diff
@@ -121,3 +125,29 @@
...
+
+.rk-search-result-grid {
+  display: -webkit-box; 
+  display: -ms-flexbox; 
+  display: flex;
+  margin-left: -1em;
+  width: auto;
+
+  .rk-search-result-grid_column {
+    padding-left:1em;
```
ciyer, line 137: For spacing consistency.

```suggestion
    padding-left: 1em;
```

ciyer (changes requested):
This looks great! Especially with images. I found a few things in the code that I was unsure about, but they should all be small fixes.

client/src/project/list/ProjectList.container.js
vfried: I decided to set it to 12 because that way we get 4 columns with 3 elements inside... otherwise it looks a bit weird

client/src/styles/components/_renku_search.scss
vfried: I will remove it wasn't sure we should have it or not, I took that css from mansonry's tutorial.

client/src/project/list/ProjectList.container.js
ciyer: Makes sense -- just get rid of the comment then.

ciyer, line 68:
```suggestion
  perPage: 12,
```

client/src/styles/components/_renku_search.scss
```diff
@@ -121,3 +125,29 @@
     }
   }
 }
+
+.rk-search-result-grid {
+  display: -webkit-box; 
+  display: -ms-flexbox; 
+  display: flex;
```
ciyer, line 132:
```suggestion
  display: flex;
```

vfried:
@ciyer all was fixed, let's wait until we can deploy+test

ciyer (approved):
Looks great! I'm hoping this will motivate users to add images, tags, and description, since the display looks even better when that information is available.

PR #1361, opened 2021-05-26
Embed GitLab as an iframe in the collaboration pages

### LINKED ISSUE #1330: Embed GitLab in an iFrame for the Collaboration Tab
# Description

Instead of implementing our own collaboration pages, we will re-use the GitLab UI by embedding it in an iframe. We should do this for _Issues_, _MRs_, and _Branches_.

# Mock-Up

(ignore the "forks" and "connections" sections for now -- these will come later). 

image

ciyer:
Do not merge until acceptance tests have been updated

# Description

Instead of implementing our own collaboration pages, just embed GitLab 

image

# Testing

To test, create, view, and update issues and merge requests in a project.

Links directly to issues and MRs are handled as well, though there is no where in the UI that generates these.

* `projects/cramakri/collaboration-test/collaboration/mergerequests/1`
* `projects/cramakri/collaboration-test/collaboration/issues/1`

Will both take you to a specific issue in the iframe.

# Research

## Catch embedding failures

This is not possible from just Javascript. It is not possible to detect that embedding as an iframe is not allowed. When the UI Server is available, we can do this by making a call to the GitLab instance and checking the values of the content-security-policy header to see if they are compatible with the current domain. (There is also the x-frame-options, but this is deprecated and superseded by CSP).

## Handling navigation events within GitLab

This is only possible if GitLab and the UI have the same origin. This is the case for our deployments, but to develop, it will be a bit complicated, since we will need to run a GitLab instance on the same origin we are using for UI development

Fix #1330

/deploy renku=ui-1361-tests


ableuler:
Looks good to me! Two thoughts:
1. In the future, a RenkuLab instance will likely be less tied to a specific GitLab instance, and thus the number of cases where embedding of GitLab is not possible will likely increase. This is just something to keep in mind.
2. Whenever we show an Iframe, would it be possible to have a little link/button on top to `open frame content in separate tab`?

ciyer:
> Looks good to me! Two thoughts:
> 
> 1. In the future, a RenkuLab instance will likely be less tied to a specific GitLab instance, and thus the number of cases where embedding of GitLab is not possible will likely increase. This is just something to keep in mind.

Yes, and unfortunately, it is not possible just in the browser to detect if embedding GitLab will fail. Once we have the UI server, we can do this and be able to offer a better explanation to the user when that happens.

I think we can reconsider implementing of a RenkuLab Issue/MR UI in the future. Depending on how the roadmap develops, that may be a good use of our development resources, but for the near term, I think we will be busy with other things so, even though this is admittedly not the greatest UX, it will be better than the alternative.

> 2. Whenever we show an Iframe, would it be possible to have a little link/button on top to `open frame content in separate tab`?

I added a little icon in the menu that will open the GitLab in a separate tab (see screenshot).

ableuler:
@ciyer These explanations make sense to me. For the iframe-to-tab button I was wondering if it could a link that would open whatever the user has browsed to in the iframe in a new tab rather than the fixed static links.

ciyer:
> @ciyer These explanations make sense to me. For the iframe-to-tab button I was wondering if it could a link that would open whatever the user has browsed to in the iframe in a new tab rather than the fixed static links.

That is unfortunately only possible if the URLs for gitlab and the UI have the same origin. Otherwise, the URL displayed in the iframe is opaque to the RenkuLab UI.

ableuler:
> > @ciyer These explanations make sense to me. For the iframe-to-tab button I was wondering if it could a link that would open whatever the user has browsed to in the iframe in a new tab rather than the fixed static links.
> 
> That is unfortunately only possible if the URLs for gitlab and the UI have the same origin. Otherwise, the URL displayed in the iframe is opaque to the RenkuLab UI.

Ah, I see, wasn't aware of this, but it makes sense.

vfried (changes requested):
I really like this change! I think it will push a bit our users to interact with Gitlab and I think it will take a big burden from us since developing collaboration tools is a big task.

I have a small change to request... could we add some vertical spacing into the tabs?

image

So that they look like this...
image

I see that the iframe is inside several elements, making it have extra "white" space on the borders...

Should be like...
image

Right now it is...
 
image

I see many row/col elements there and the padding/margin on them is causing this --> i might be the person to blame for this since I did some work on this tab before
image

The buttons to go to GitLab are not 100% clear to me but (what happens when they are clicked... but I'm not sure I have a good solution for them, I think once users get used to them they will know what they are there for since we have them in several places. I would consider adding a tooltip to them.

ciyer:
@vfried Thanks for the review. All of your comments make sense, and I have addressed them all:

**Added spacing in menu**
image

**Added tooltip**
image

**Remove row whitespace**
image

vfried (approved):
Thanks for addressing the changes I requested, things are looking really good!!! 

I will approve and when tests pass you can merge 🚀

PR #1363, opened 2021-05-26
fix(project): project header improved and responsive

vfried:
The project header needed some spacing/improvements and was not very responsive. I decided to build a flex grid to place items and it's now responsive.

image

image

When no description is set...
image

We have an issue in this case when it shrinks... 
image

I thought we can add a button like this... but saying add a description if there is no description
image

I could explore adding more less pace, switching the order of items when it shrinks, changing the weight/style of the item fonts if we wanted to as well.

/deploy renku=ui-design-2021


ciyer (approved):
Great!

PR #1364, opened 2021-05-26
feat(design): improvements in datasets

vfried:
Small improvements in datasets...

image

I changed the circles to triangles, we could also do squares.

### DIFF
--- a/client/src/dataset/list/DatasetList.present.js
+++ b/client/src/dataset/list/DatasetList.present.js
@@ -43,7 +43,7 @@ class DatasetListRow extends Component {
 
     return <Link className="d-flex flex-row rk-search-result"
       to={`${datasetsUrl}/${encodeURIComponent(dataset.identifier)}`}>
-      <span className={"circle me-3 mt-2 " + color}></span>
+      <span className={"triangle me-3 mt-2 " + color}></span>
       <Col className="d-flex align-items-start flex-column col-10 overflow-hidden">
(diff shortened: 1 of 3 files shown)


client/src/dataset/list/DatasetList.present.js
```diff
@@ -43,7 +43,7 @@ class DatasetListRow extends Component {
 
     return <Link className="d-flex flex-row rk-search-result"
       to={`${datasetsUrl}/${encodeURIComponent(dataset.identifier)}`}>
-      <span className={"circle me-3 mt-2 " + color}></span>
+      <span className={"triangle me-3 mt-2 " + color}></span>
```
ciyer, line 46: I suggest we name the class for its meaning — so `rk-ds` (and have `rk-proj` for projects). This will give us freedom to change the shape that is rendered in the css without having to modify the code elsewhere.

ciyer (changes requested):
Great idea, and this looks nice. I have a naming suggestion that will give us more flexibility to change things if the designer proposes a different solution.

client/src/dataset/list/DatasetList.present.js
ciyer: After looking at it longer, I think we should avoid using triangles because we use them for the warning sign.

vfried:
We decided to let go of this idea for now, we might want to ask the designer to make proper icons for datasets/projects/etc but we don't need this for now.

PR #1366, opened 2021-05-28
fix: remove flash error when loading datasets

vfried:
@rokroskar found a bug when accessing datasets in the dataset search.

When accessing a dataset for a second you can see an "error unknown" sign, this was due to a bag if, this has been fixed.

To test you can search for datasets and open one.

Do the same in this deployment and the error shouldn't be there.

I removed an unnecessary call to the core service that was also failing since it was trying to fetch a dataset outside a project.

/deploy


ciyer (approved):
👍

PR #1372, opened 2021-06-01
feat(design): generic component to map items into a list applied to projects and datasets

vfried:
I created a List component that receives a list of items with different fields and maps them to a grid or list (depending on what the user wants).

image

It looks like the project grid, there is a benefit here... if we want to implement a generic search in the KG (for projects, datasets, etc) we could use this generic component to display a list of mixed elements.

I applied this to datasets and projects. I see the project row is used in the landing page as well, we could delete this code after we adapt the landing page to using this.

/deploy renku=ui-design-2021

### DIFF
--- a/client/src/dataset/list/DatasetList.present.js
+++ b/client/src/dataset/list/DatasetList.present.js
@@ -162,6 +162,30 @@ class DatasetsRows extends Component {
 
     const { currentPage, perPage, search, totalItems, gridDisplay } = this.props;
 
+    const datasetItems = datasets.map(dataset => {
+      const projectsCount = dataset.projectsCount > 1
+        ? `In ${dataset.projectsCount} projects`
+        : `In ${dataset.projectsCount} project`;
+
+      return {
+        id: dataset.identifier,
+        url: `${datasetsUrl}/${encodeURIComponent(dataset.identifier)}`,
+        stringScore: stringScore(dataset.identifier) % 3,
+        title: dataset.title || dataset.name,
+        description: dataset.description !== undefined && dataset.description !== null ?
+          <Fragment>
+            <MarkdownTextExcerpt markdownText={dataset.description} singleLine={gridDisplay ? false : true}
+              charsLimit={gridDisplay ? 200 : 100} />
+            <span className="ms-1">{dataset.description.includes("\n") ? " [...]" : ""}</span>
+          </Fragment>
+          : null,
+        timeCaption: new Date(dataset.date),
+        labelCaption: projectsCount + ". Created",
+        creators: dataset.published !== undefined && dataset.published.creator !== undefined ?
+          dataset.published.creator : null,
+      };
+    });
+
     return <ListDisplay
       itemsType="dataset"
       search={search}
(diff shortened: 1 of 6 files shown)


client/src/dataset/list/DatasetList.present.js
```diff
@@ -206,50 +157,82 @@ class DatasetSearchForm extends Component {
...
+    const { currentPage, perPage, search, totalItems, gridDisplay } = this.props;
+
+    return <ListDisplay
+      itemsType="dataset"
+      search={search}
+      currentPage={currentPage}
+      gridDisplay={gridDisplay}
+      totalItems={totalItems}
+      perPage={perPage}
+      items={datasets.map(dataset => {
```
ciyer, line 172: I think this could be extracted to a variable above. It's a lot of code to look at in a property.

client/src/project/list/ProjectList.present.js
```diff
@@ -134,39 +85,36 @@ function ProjectListRowBar(props) {
...
+  const { currentPage, perPage, projects, search, totalItems, gridDisplay } = props;
+
+  return <ListDisplay
+    itemsType="project"
+    search={search}
+    currentPage={currentPage}
+    gridDisplay={gridDisplay}
+    totalItems={totalItems}
+    perPage={perPage}
+    items={projects.map(project => {
```
ciyer, line 97: This is a lot of code in a property. It could be extracted to a variable.

client/src/utils/List.js
```diff
@@ -0,0 +1,142 @@
...
+    <Col className="d-flex justify-content-end align-self-center flex-shrink-0">
+      {mediaContent ?
+        <img src={mediaContent} alt=" " className="card-img-bottom"/>
+        : null
+      }
+    </Col>
+  </Link>;
+}
+
+function ListDisplay(props) {
```
ciyer, line 103: This class could benefit from some documentation about the properties. What is expected in each of the properties.

client/src/project/list/ProjectList.present.js
```diff
@@ -134,39 +85,36 @@ function ProjectListRowBar(props) {
...
+      const path = project.path;
+      const url = Url.get(Url.pages.project, { namespace, path });
+      return {
+        id: project.id,
+        url: url,
+        stringScore: stringScore(project.path_with_namespace) % 3,
+        title: project.path_with_namespace,
+        description: project.description ?
+          <Fragment>
+            <RenkuMarkdown markdownText={project.description} fixRelativePaths={false} singleLine={true} />
```
ciyer, line 108: Why is `MarkdownTextExcerpt` not used here?

ciyer (changes requested):
Structuring this as a component will make it much easier to get a uniform display of lists across pages! 

I have a couple of suggestions for places were code could be restructured a bit and where some documentation would help, but I think this is overall very good.

client/src/project/list/ProjectList.present.js
vfried: Copy-paste --> I just used what was used in the project before but I will change it to the Excerpt

vfried:
@ciyer thanks for your review,  I addressed all changes, all made sense to me.

Let me know if the documentation for the list component is clear enought.

ciyer (approved):
👍

PR #1375, opened 2021-06-02
fix: minor ui fixes to header and footer

vfried:
Fixes to header and footer

- Unselected text is now grey and selected is white (I applied a fix since the selected was only showing when the user was in "your projects" but not in starred or all projects.
- Added spaces as luca asked
- Applied style to footer

image
image

The search bar in the header has been cleaned up, unfortunately, I couldn't change the style of the blue button (I will try harder but I don't want to use too much time on that now)
image

Todo, change order when small so that projects, environments and datasets are all in the same column
image

/deploy renku=ui-design-2021


ciyer (approved):
👍

PR #1377, opened 2021-06-02
feat(design): use of colours to identify entities

### LINKED ISSUE #1376: Give meaning to the dot colors in the new design
Use colors to designate entities (projects = green, datasets = pink; *later* workflows = yellow).

vfried:
We decided that each entity has a color, this will be useful when we have different entities returned by the same search results.

Green --> projects
Pink --> datasets
Yellow --> collaboration (until the tab is replaced by the iframe)

image

image

Closes #1376

### DIFF
--- a/client/src/styles/icons/icons.scss
+++ b/client/src/styles/icons/icons.scss
@@ -27,8 +27,13 @@
   background: #009568;
 }
 
-.circle.collaboration {
+.circle.workflow {
   background: #D1BB4C;
+}
+
+// TODO - DELETE WHEN COLLABORATION IFRAME IS MERGED
+.circle.collaboration {
+  background: rgba(128, 128, 128, 0.893);
 }
 
 .circle.dataset {
--- a/client/src/styles/icons/icons.scss
+++ b/client/src/styles/icons/icons.scss
@@ -23,14 +23,14 @@
   display:inline-table;
 }
 
-.circle.green {
+.circle.project {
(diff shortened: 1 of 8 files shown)


ciyer:
I think yellow will probably be workflows, once those are also indexed into the KG.

client/src/styles/icons/icons.scss
```diff
@@ -23,14 +23,14 @@
   display:inline-table;
 }
 
-.circle.green {
+.circle.project {
   background: #009568;
 }
 
-.circle.yellow {
+.circle.collaboration {
```
ciyer, line 30: I think we can make this workflow or wait until we can actually integrate workflows.

ciyer (changes requested):
This is a nice solution! The only thing that I think should be changed is the integration of collaboration. That is about to go away, and I think the next entity that we will want to so search for are workflows.

client/src/styles/icons/icons.scss
vfried: Yes, i thought that would be the case... shall I use some gray dots for collaboration while we have them? Also... I could make things be a different shape if we wanted to (for colorblind people)

vfried:
@ciyer i addressed the changes that were requested, i made an additional change to name the entities inside the items on top of the list, this will allow us to do lists with mixed items.

ciyer (approved):
👍

PR #1378, opened 2021-06-02
refactor: fix the extra scrollbar on firefox

lorenzo-cavazzi:
On firefox, any page with empty content before the footer results in an extra scrollbar added to the right (see the screenshot below).

This change should fix it without having any negative impact on chrome.

/deploy #notest

Screenshot_20210602_165513

### DIFF
--- a/client/src/App.css
+++ b/client/src/App.css
@@ -32,9 +32,7 @@ html {
   position: relative;
   min-height: 100%;
 }
-body {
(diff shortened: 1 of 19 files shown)


lorenzo-cavazzi:
The second commit fixes missing margins added through the old bootstrap classes

Screenshot_20210602_174008

ciyer (approved):
👍

Issue #1373, opened 2021-06-01
Side-by-side deployment of Renku new and old UIs


ciyer:
# Description

For a transition period, we want allow people to try out the new UI while still leaving the existing UI as the default.

# Configuration / Deployment

There are several options to deploy the two UIs side-by-side.

1. Create a new k8s deployment (beta.renkulab.io) and deploy the new UI there

This would work similarly to how we do the personal dev deployments. beta.renkulab.io would be configured to share gitlab with renkulab.io, but it would have parallel infrastructure for everything else.

**Pros** Easy to set up, easy to run the Selenium tests against both UIs
**Cons** Different URLs for the two versions, cannot seamlessly switch (e.g., sessions started in the old UI would only be visible there)
**Assessment: This is probably the best approach if the resources we need are available.**

2. K8s canary deployment for the new UI

The new and old UIs would be available under the same URL, but a cookie would determine which UI the user gets routed to.

https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#canary

**Pros** Can seamlessly switch between the new and old UI
**Cons** Some work necessary to get it set up; Selenium tests could be complicated

**Assessment: This is a bit more complicated than 1, but would probably work.**

3. Have both UIs available in the UI pod, use nginx to route to the appropriate UI

The new and old UIs would be available under the same URL and Nginx would look at a cookie to determine which one to serve.

~~https://www.nginx.com/blog/performing-a-b-testing-nginx-plus/~~
https://stackoverflow.com/questions/27376235/routed-traffic-with-cookies-nginx

Pros: All setup is restricted to the UI component, can seamlessly switch between the new and old UIs
Cons: Some work necessary to get it set up; Selenium tests could be complicated

**Assessment: About the same as 2 in terms.**

# Old UI Banner

We need to add a banner to the old UI to let people know that they can try out the new one.

![image](https://user-images.githubusercontent.com/1196411/120338172-43309c00-c2f4-11eb-83ca-7e94fff944e3.png)

# New UI Landing

We need to add a new transitionary landing page for the Renku UI that explains the big changes

(Design coming)

ableuler:
I would vote against option 1, mostly for the KG parts. I don't think we can build the full graph for both deployments (is it @jachro), and having dataset search etc. yield different results might be confusing.

[...]

PR #1379, opened 2021-06-02
enable canary deployment

rokroskar:
To test, deploy this branch and set the cookie `other-ui` to `always` or `never`. 

cc @SwissDataScienceCenter/yat

partially addresses #1373  - does not implement the banner.


ableuler (commented):
Looks good on first sight, haven't tried it yet though. If this canary option is going to stick around (which I believe it will as it's something very useful in general), it would be nice to avoid the duplication between the regular and the canary deployment spec. Not sure what's best here, maybe a range loop?

rokroskar:
hmm I don't know - they're very similar here but they might as well be very different for a different purpose. I'm not sure there's really much of a benefit in avoiding this duplication right now.

helm-chart/renku-ui/templates/ingress.yaml
```diff
@@ -37,3 +37,29 @@ spec:
...
+  labels:
+    app: {{ template "ui.name" . }}-canary
+    chart: {{ template "ui.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+  annotations:
+    kubernetes.io/ingress.class: nginx
+    nginx.ingress.kubernetes.io/canary: "true"
+    nginx.ingress.kubernetes.io/canary-by-cookie: "other-ui"
+spec:
```
ableuler, line 56: Hmm, how does TLS termination actually work here? I've seen THAT it works, but I wonder how 😅. Is it handled according to the main ingress spec still?

helm-chart/renku-ui/templates/ingress.yaml
```diff
@@ -37,3 +37,29 @@ spec:
...
+  rules:
+    - host: {{ .Values.canary.host }}
```
ableuler, line 58: I think we could use `.Values.global.renku.domain` here (and do the same in our main renku ingress).

ableuler (changes requested):

helm-chart/renku-ui/templates/ingress.yaml
rokroskar: oh good point, I forgot about that one

helm-chart/renku-ui/templates/ingress.yaml
rokroskar: good question - I actually don't know 😆 Probably because it's set per host and channeling the request to the correct service happens afterwards?

helm-chart/renku-ui/values.yaml
```diff
@@ -125,3 +125,10 @@ homepage:
   tutorialLink: https://renku.readthedocs.io/en/latest/tutorials/01_firststeps.html
   # Projects to highlight on the homepage
   projects: []
+
+canary:
+  enabled: false
```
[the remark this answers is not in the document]
rokroskar: idk, i see it the other way (this is not required for a functioning UI) but you're right that the @SwissDataScienceCenter/renku-ui-maintainers  should decide the default for the release...

rokroskar:
> the duplicated section will get very stale

that's true, I just don't know of a clean way of doing this. I guess we could have a meta template file that gets injected and we modify the behavior via alias definitions? e.g. 

```
{{- $name := include "ui.fullname" . -}}-canary
{{ .Files.Get deployment.yaml.tpl }}
```

and `deployment.yaml.tpl` would have e.g.

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: {{ $name }}
...
```

_maybe_ something like that would work and not be too crazy to maintain without tons of templating directives all over the place?

helm-chart/renku-ui/values.yaml
```diff
@@ -125,3 +125,10 @@ homepage:
...
+  image:
+    repository: renku/renku-ui
+    tag: "0.11.11"
+  host: example.com
```
ciyer, line 134: Where does this line come from? Is `example.com` correct?

ciyer (changes requested):

helm-chart/renku-ui/values.yaml
rokroskar: I've gotten rid of this actually -- it's fetched from `global.renku.domain` (haven't pushed the changes yet)

ciyer:
I tested this, and it works well, which is very cool!

We need to clarify the mechanics, though. I'm wondering if it makes more sense to apply this change to the 0.11.x branch (which is what is currently deployed in prod) and designate the latest build from the ui-design-2021 branch as the canary. To me, this approach would have several advantages:

1. The canary is the "canary" (right now, the tried-and-true version is the "canary", which is a bit confusing)
2. The "default" UI that you would get with no cookie set is the right default (right now, we would need to do a bit of extra work to get that behavior).

The current approach also has some advantages (e.g., if we add a new value to the chart, handling that is a bit more natural), but I feel like applying this change to the 0.11.x branch is still the better way to go.

Thanks for putting this together!

rokroskar:
@ableuler I've refactored the deployment template to avoid duplication. I think this will eventually need to be duplicated anyway to allow for different kinds of deployments, but for now I guess it's ok. 

@ciyer certainly, we can do it like that, I think that makes more sense. I'll change the reference branch

ciyer:
> @ciyer certainly, we can do it like that, I think that makes more sense. I'll change the reference branch

Great. I will add the UI changes to add a banner and set the cookie to this PR after you rebase against the HEAD of 0.11.x.

rokroskar:
@ciyer this branch is now based off `0.11.x`

rokroskar:
@ciyer - @ableuler [suggested](https://github.com/SwissDataScienceCenter/renku-ui/pull/1379#discussion_r644817981) we make canary enabled by default - what do you think about this? I think it's better to leave it off, but am open to changing it. It can be switched on for a specific release if needed, but I think it should be intentional.

ciyer (approved):
This works very well, thanks for putting it together!

ableuler:
For me the canary is rather a property of a given UI-version than a property of a Renku deployment just like a given version of a microservice might serve two API versions at the same time. But if you both prefer to leave it off per default I can definitely live with that.

rokroskar:
@ableuler I agree - my point is just that the "steady-state" should be with it off and when the UI makes a release that needs it on, it can be done for that release, with the expectation that it will be toggled off again in the near future.

ciyer:
> @ableuler I agree - my point is just that the "steady-state" should be with it off and when the UI makes a release that needs it on, it can be done for that release, with the expectation that it will be toggled off again in the near future.

Yes, I think we can merge this as is now. I am getting a PR ready that includes the banner to notify users about the preview. There, I will turn on and configure the canary.

Issue #1374, opened 2021-06-02
Transitional Landing Page


ciyer:
# Description

A special landing page for the new UI Preview

# Intro

Explain what they are seeing and how to return to the old UI if they wish

![image](https://user-images.githubusercontent.com/1196411/120471232-c3aed580-c3a4-11eb-87cd-81c44acd7895.png)

# Motivation for preview

We are working on many improvements to the UI and want to start making them available. The new UI offers equivalent functionality to the old UI and has been tested. We are not yet switching everyone over because: 

1. We want to let existing users transition at their own pace
2. Some additional changes are in the pipeline, and we know that some find it annoying if the UI changes multiple times in a short period

(mockup coming)

# Changes compared to the old version

Describe how the new UI is different from the old one.

(mockup coming)

# Changes that are in the pipeline

Some further changes are coming that we think will make the experience of using the UI much nicer. Describe these changes.

(mockup coming)

PR #1383, opened 2021-06-07
Turn on the preview version banner

### LINKED ISSUE #1373: Side-by-side deployment of Renku new and old UIs

ciyer:
Fix #1373

# Testing

Connect to the deployment. 
- Click on the "Try out a preview of the new RenkuLab UI" link. You should be redirected to the new UI.
- Close your browser and come back to the URL. You should still see the new UI
- Delete the "other-ui" cookie and refresh. You should be back to the old UI.

# Appearance

image

/deploy


lorenzo-cavazzi:
Something doesn't work, if I click on the banner the UI crashes with an error on `homeCustomized`
```
TypeError: e.homeCustomized is undefined
```

Screenshot_20210608_080425

More in general, would it make sense to add the banner also on the logged user home?
I generally don't even look at the landing page and click directly on login, I'm wondering if other users do the same

lorenzo-cavazzi (changes requested):
Something doesn't work, if I click on the banner the UI crashes with an error on `homeCustomized`
```
TypeError: e.homeCustomized is undefined
```

Screenshot_20210608_080425

More in general, would it make sense to add the banner also on the logged user home?
I generally don't even look at the landing page and click directly on login, I'm wondering if other users do the same

ciyer:
Can you try it with a private browser session and tell me if you have the same crash?

ciyer:
> Can you try it with a private browser session and tell me if you have the same crash?

Don't worry about it, I have found out the source of the problem (caching). I will fix it.

ciyer:
> Something doesn't work, if I click on the banner the UI crashes with an error on `homeCustomized`

I was able to recreate and fix this problem.

> More in general, would it make sense to add the banner also on the logged user home?
> I generally don't even look at the landing page and click directly on login, I'm wondering if other users do the same

I think you are right, but I would prefer to get this released, so that users that are very eager to try out the new UI have a way to do so, and implement your suggestion in another Issue (#1384) / PR.

helm-chart/renku-ui/values.yaml
```diff
@@ -124,7 +124,7 @@ homepage:
   # Projects to highlight on the homepage
   projects: []
 canary:
-  enabled: false
+  enabled: true
   image:
     repository: renku/renku-ui
-    tag: "0.11.12"
+    tag: "0.11.11-n045.h8e3df79"
```
lorenzo-cavazzi, line 130: Is this intended?

rokroskar:
@ciyer could we have a way to get back to the old UI? right now it seems to be a one-way sort of thing?

ciyer:
> @ciyer could we have a way to get back to the old UI? right now it seems to be a one-way sort of thing?

That will come with #1374. It is not possible to address that in this PR because the changes need to go to different branches.

helm-chart/renku-ui/values.yaml
ciyer: That is the current tag of the ui-design-2021 branch. Before we tag the 0.11.13 release, we should tag the ui-design-2021 as `1.0.0-preview1`, and we can use that tag in this slot.

lorenzo-cavazzi (approved):
:+1:

PR #1385, opened 2021-06-08
feat(design): minimalist project header

### LINKED ISSUE #1371: Project Header Streamlining
# Description

For the project main (home) tab, we want more comprehensive project header

image

For the other tabs, we want a more minimal header

image

Note: Keep public/private in the minimal header.

vfried:
Non minimalist...
image

Minimalist...
image

Discussion: 
- Shall we leave the image on minimalist or we remove it?
- Is the "small" font too small?
- responsivness---> Do we want to order elements in a specific way when it shrinks?

/deploy renku=ui-design-2021

Closes #1371


ciyer:
I think what you have is an improvement to the header and makes several improvements! But, I think the difference between the OverviewHeader and the minimal one should be greater. Here are the changes that my commit makes:

Overview
image

Minimal
image

vfried:
I'm ok with the change... I'm not so sure the users will like it but I think it's good to try this.

It makes me feel a bit weird that the height of the project nav changes with this... I would like it to always stay at the same height, but this is a minor thing.

ciyer:
> I'm ok with the change... I'm not so sure the users will like it but I think it's good to try this.

I agree -- I think we need to try it out a bit to see if it is an improvement or not. 

> It makes me feel a bit weird that the height of the project nav changes with this... I would like it to always stay at the same height, but this is a minor thing.

Yes, I agree that is annoying, and it would be preferable to keep the nav at the same location. To do this, we would need to move it higher so that changes to the header content do not affect its placement.

> Something else that feels weird to me is to use the title in one place and the slug in another... we could maybe just leave this in the minimalist header (title and slug only)
> image
> 
> But besides this minor thing, it think I would like to try a more minimalist header... the other one feels like too much.

👍 

> BTW... I'm not 100% sure i like the image being there, but i understand it makes sense to have it.

I like the image!

ciyer:
A re-ordering of the elements to make the interaction of the two headers work better.

## Maximal Header

image


## Minimal Header

image

vfried:
Is this ready to be re-reviewed/merged or do we need to make extra changes here?

ciyer:
> Is this ready to be re-reviewed/merged or do we need to make extra changes here?

We wanted to try a few more changes: make the description larger, move the tags to the left side.

ciyer:
Latest version

**Normal Project**

image

**Long Description and Many Tags**

image

**No Description, User Can Modify Settings*

image


**No Description**

image

client/src/project/Project.present.js
```diff
@@ -89,11 +90,14 @@ class ProjectVisibilityLabel extends Component {
   render() {
     switch (this.props.visibilityLevel) {
       case "private":
-        return <span><Badge color="secondary"><FontAwesomeIcon icon={faLock} /> Private</Badge>&nbsp;</span>;
+        return <span className="d-inline align-text-top ms-3"><Badge color="secondary">
```
lorenzo-cavazzi, line 93: The `align-text-top` is causing a vertical misalignment. The screenshot is from Firefox but the same happens with Chrome

Screenshot_20210623_112646

Removing it seems to be enough to re-align the content

client/src/project/Project.present.js
```diff
@@ -230,6 +234,58 @@ class ForkProjectModal extends Component {
...
+      </Row>
+    </Fragment>);
+}
+
+function ProjectViewHeaderOverviewDescription({ settingsReadOnly, description, settingsUrl }) {
+  if (description)
+    return <RenkuMarkdown markdownText={description} fixRelativePaths={false} className="p-mb-0 fs-6"/>;
+  if (!settingsReadOnly) {
+    return <div className="p-mb-0 fs-6"><i>(This project has no description.
+      You can provide one on the <Link to={settingsUrl}>settings tab</Link>.)</i></div>;
```
lorenzo-cavazzi, line 284: In the preview screenshots you shared, this content is in italics, while here it's not.

Screenshot_20210623_112150

My personal preference is for italics (makes it clear it's an info) without the parentesis (we don't use them anywhere else), but I'm ok with keeping it unchanged if you think it's better

lorenzo-cavazzi (changes requested):
Nice to see this change coming along!

I noticed the visibility label is slightly misaligned.
I've also added a comment for the "no description" message. Feel free to discard it if you disagree.

client/src/project/Project.present.js
ciyer: Hadn't noticed this because it is aligned on Safari and Chrome, but I will fix it so it works on Firefox.

client/src/project/Project.present.js
ciyer: I took the screenshot from telepresence, you probably tested before the commit got deployed on the CI. I just checked the CI deployment, and it is in italics for me. 

I will try your suggestion, and try using an "info" background as well and pick the one that looks best.

client/src/project/Project.present.js
ciyer: I was not able to get it to look *exactly* the same on all three browsers, but it should now look ok everywhere.

**Old**

image

**Fixed**
image-improved

lorenzo-cavazzi (approved):
Very nice!
The acceptance tests should be adapted since we were using the previous headers (both the buttons and the description)

Issue #1329, opened 2021-04-27
Switch to a two column layout in the sessions/notebooks/environments form.


vfried:
Switch to a 2-column layout for the form, as here:

  https://github.com/swissdatasciencecenter/renku-ui/issues/1152

![image](https://user-images.githubusercontent.com/1196411/116199633-a0ef2880-a737-11eb-82c1-1dae1d41209c.png)

If it is an easy fix, we can go ahead and do it now, otherwise, we can merge this PR and make that improvement in another PR.

_Originally posted by @ciyer in https://github.com/SwissDataScienceCenter/renku-ui/issues/1318#issuecomment-827367670_

ciyer:
This was fixed in #1386

PR #1386, opened 2021-06-09
 feat: edit project sessions settings

### LINKED ISSUE #1114: Edit the project settings from the UI

lorenzo-cavazzi:
_INFO: ready to review, but the `pinned image` is still missing and will be added soon_

The project settings are now split in 2 pages, where `General` contains the previous settings and `Sessions` allows the logged users to:
* Check the project and RenkuLab options and default values
* Allor user with > developer permissions to change the options according to the values supported on RenkuLab

**Changes/discrepancies with the original issue**
- Change multiple values and confirm later: I tried using different colors to distinguish current settings vs yet-to-be-applied changes. The result was very confusing. I opted for applying any change directly on click. The UX looks much easier, even though we may get extra commits.
- Set arbitrary values: I tried to add an "unlock values" option to allow the user to set any values for all the options. That was very confusing and there is no added value since all the invalid values are later discarded when launching a new session. If we want to provide an interface for setting any custom option, we should do that in a separate tab (in `General` on a new `Advanced` tab) with a simple interface without any validation.

**How to test**: pick a project and go to "Settings --> Sessions". Try to change values and start a new session.
This sample projects also contains a few "wrong" settings
https://renku-ci-ui-1386.dev.renku.ch/projects/lorenzo.cavazzi.tech/project-config/settings/sessions

image

/deploy renku=ui-design-test
fix #1114

### DIFF
--- a/client/src/project/settings/ProjectSettings.present.js
+++ b/client/src/project/settings/ProjectSettings.present.js
@@ -27,11 +27,11 @@ import React, { Component, Fragment, useState } from "react";
 import { Link } from "react-router-dom";
 import {
   Alert, Button, Col, Collapse, Form, FormGroup,
-  FormText, Input, Label, Row, Table, Nav, NavItem, UncontrolledTooltip
+  FormText, Input, InputGroup, Label, Row, Table, Nav, NavItem, UncontrolledTooltip
 } from "reactstrap";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import {
-  faExclamationTriangle, faInfoCircle, faTimesCircle
+  faCheck, faEdit, faExclamationTriangle, faInfoCircle, faTrash, faTimesCircle
 } from "@fortawesome/free-solid-svg-icons";
 
 import { ACCESS_LEVELS } from "../../api-client";
@@ -239,14 +239,10 @@ function ProjectSettingsSessions(props) {
     );
   }
 
-  // Handle ongoing operations
+  // Handle ongoing operations and errors
   if (config.fetching || options.fetching)
     return (<SessionsDiv><Loader /></SessionsDiv>);
 
-  if (newConfig.updating)
-    return (<SessionsDiv><NewConfigStatus {...newConfig} /></SessionsDiv>);
-
-  // Handle errors
   if (config.error && config.error.code)
     return (<SessionsDiv><SessionConfigError config={config} /></SessionsDiv>);
 
@@ -263,13 +259,19 @@ function ProjectSettingsSessions(props) {
 
   const knownOptions = (
     <SessionConfigKnown availableOptions={projectData.options} defaults={projectData.defaults}
-      devAccess={devAccess} globalOptions={globalOptions} setConfig={setConfig}
+      devAccess={devAccess} globalOptions={globalOptions} setConfig={setConfig} disabled={newConfig.updating}
+    />
+  );
+
+  const advancedOptions = (
+    <SessionConfigAdvanced devAccess={devAccess} defaults={projectData.defaults.project}
+      options={projectData.options.unknown} setConfig={setConfig} disabled={newConfig.updating}
     />
   );
 
   const unknownOptions = (
     <SessionConfigUnknown devAccess={devAccess} defaults={projectData.defaults.project}
-      options={projectData.options.unknown} setConfig={setConfig}
+      options={projectData.options.unknown} setConfig={setConfig} disabled={newConfig.updating}
     />
   );
 
@@ -283,31 +285,40 @@ function ProjectSettingsSessions(props) {
       <NewConfigStatus {...newConfig} />
       {text}
       {knownOptions}
+      {advancedOptions}
       {unknownOptions}
     </SessionsDiv>
   );
 }
 
 function SessionConfigKnown(props) {
-  const { availableOptions, defaults, devAccess, globalOptions, setConfig } = props;
+  const { availableOptions, defaults, devAccess, disabled, globalOptions, setConfig } = props;
 
-  return availableOptions.known.map(option => {
+  const elementsList = availableOptions.known.map(option => {
     return (
-      <SessionsElement key={option}
-        option={option} // key cannot be accessed as a property, hence the duplication here
-        globalDefault={defaults.global[option]}
-        projectDefault={defaults.project[option]}
-        rendering={globalOptions[option]}
-        setConfig={setConfig}
-        devAccess={devAccess}
-        configPrefix={NotebooksHelper.sessionConfigPrefix}
-      />
+      <Col key={option} xs={12} lg={6}>
+        <SessionsElement
+          configPrefix={NotebooksHelper.sessionConfigPrefix}
+          devAccess={devAccess}
+          disabled={disabled}
+          globalDefault={defaults.global[option]}
+          option={option}
+          projectDefault={defaults.project[option]}
+          rendering={globalOptions[option]}
+          setConfig={setConfig}
+        />
+      </Col>
     );
   });
+  return (<Row>{elementsList}</Row>);
 }
 
 function SessionConfigUnknown(props) {
-  const { defaults, devAccess, options, setConfig } = props;
+  const { defaults, devAccess, disabled, setConfig } = props;
+  // Remove "image" from the options since that's handled separately
+  const options = props.options.length ?
+    props.options.filter(option => option !== "image") :
+    [];
 
   const [showUnknown, setShowUnknown] = useState(false);
   const toggleShowUnknown = () => setShowUnknown(!showUnknown);
@@ -326,7 +337,7 @@ function SessionConfigUnknown(props) {
     const value = defaults[option];
 
     const reset = devAccess ?
-      (<SessionsOptionReset onChange={() => resetValue(option)} option={option} />) :
+      (<SessionsOptionReset disabled={disabled} onChange={() => resetValue(option)} option={option} />) :
       null;
 
     return (
@@ -338,9 +349,9 @@ function SessionConfigUnknown(props) {
 
   // Collapse unknown values by default
   return (
-    <Fragment>
+    <div className="mb-2">
       <Collapse isOpen={showUnknown}>
-        <h4>Unrecognized settings</h4>
+        <h5>Unrecognized settings</h5>
         <p>
           The following settings are stored in the project configuration but they are not
           supported in this RenkuLab deployment.
@@ -350,7 +361,7 @@ function SessionConfigUnknown(props) {
       <Button color="link" className="font-italic btn-sm" onClick={toggleShowUnknown}>
         [{showUnknown ? "Hide " : "Show "} unrecognized settings]
       </Button>
-    </Fragment>
+    </div>
   );
 }
 
@@ -379,7 +390,11 @@ function NewConfigStatus(props) {
   const { error, keyName, updated, updating, value } = props;
 
   if (updating) {
-    return (<div><span>Updating {keyName}, please wait... <Loader size="14" inline="true" /></span></div>);
+    return (
+      <Alert color="info">
+        <span>Updating {keyName}, please wait... <Loader size="14" inline="true" /></span>
+      </Alert>
+    );
   }
   else if (error) {
     return (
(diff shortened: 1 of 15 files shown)


client/src/project/settings/ProjectSettings.present.js
```diff
@@ -0,0 +1,521 @@
...
+    return (<Alert color="info"><FontAwesomeIcon icon={faInfoCircle} /> &quot;{keyName}&quot; {text}</Alert>);
+  }
+
+  return null;
+}
+
+function SessionsDiv(props) {
+  return (
+    <div className="mt-2">
+      <h3>Sessions settings</h3>
```
ciyer, line 408:
```suggestion
      <h3>Session settings</h3>
```

client/src/project/settings/ProjectSettings.present.js
```diff
@@ -0,0 +1,521 @@
...
+    <SessionConfigUnknown devAccess={devAccess} defaults={projectData.defaults.project}
+      options={projectData.options.unknown} setConfig={setConfig}
+    />
+  );
+
+  // Information based on user's access level
+  const text = devAccess ?
+    (<p>
+      Any change will be permanently saved in the project options and used as default.
+      <br /> Mind that users can still manually modify any option before starting a session.
```
ciyer, line 280: I don't think this text is necessary. The feedback after a change makes it clear that it has been applied.

ciyer (changes requested):
This is looking very good! I like the solution you designed for communicating default values.

I have made some suggestions about wording in the code. I have a few other things I identified:

Trying to change the Amount of Memory or Number of CPUs failed for me.

image

Our standard for forms is for the label to be above the element. Could we adopt that here as well?

image

Maybe the default values could be provided in the help text

Would it be possible to make the form take up two columns on larger screens (this isn't a necessity, but nice to have)? (see #1329)

rokroskar:
This is great @lorenzo-cavazzi ! One comment - could we not call the default environment by the path but rather by the environment? I.e. 

```
Default environment: [JupyterLab (icon); RStudio (icon)]
```

I'd also like to be able to input some custom environment (e.g. streamlit/voila). 

In addition, it would be great to be able to specify _all_ the options I want the users to see (in case I have an image with multiple environments). Like I might have Jupyter and VNC but not RStudio so it makes no sense to show the RStudio option at all.

ciyer:
I think these are all good suggestions!

> This is great @lorenzo-cavazzi ! One comment - could we not call the default environment by the path but rather by the environment? I.e.
> 
> ```
> Default environment: [JupyterLab (icon); RStudio (icon)]
> ```

Since we want to let users specify arbitrary values here, we also need to support uploading icons and the icons would need to be stored somewhere. 
 
> I'd also like to be able to input some custom environment (e.g. streamlit/voila).

This is in the requirements for the issue, but it was asked to delay this feature for another PR because we need to have a better UX concept for it (see the first comment in the conversation).

But maybe we could add this possibility *just* for the "environment" field, where it is of the most use, in this PR, and implement a more general solution in another PR.

> In addition, it would be great to be able to specify _all_ the options I want the users to see (in case I have an image with multiple environments). Like I might have Jupyter and VNC but not RStudio so it makes no sense to show the RStudio option at all.

This is a good suggestion. But I think it would be better to make a separate PR for this improvement, in which case, I will make an issue, unless @lorenzo-cavazzi says it is easy to add here.

ciyer:
> I'd also like to be able to input some custom environment (e.g. streamlit/voila).

I think custom environments are a good solution for environments like VS code, but not the right way to handle dashboards.

We need another category of information in projects: Apps. No one wants to connect they the "streamlit endpoint"; they want to connect to the "Molecule Explorer app". Apps should contain a title, description, and resource requirements in addition to endpoint. With that information, users can make informed choices about what they want to access, and the UI can provide a better UX for interacting with them.

rokroskar:
Yes! totally agree. We've been brainstorming how to make "apps" available as separate from interactive jupyter sessions. 

> Since we want to let users specify arbitrary values here, we also need to support uploading icons and the icons would need to be stored somewhere.

The icons were just an idea - they are not necessary in this first implementation _or_ they could just be limited to things we already know about like Jupyter/RStudio. But it would be great if we can somehow specify the "human readable" name (e.g. JupyterLab) as well as the actual path (`/lab`) with only the former being visible to someone browsing the project.

client/src/project/settings/ProjectSettings.present.js
lorenzo-cavazzi: removed all the text for admin

lorenzo-cavazzi:
@ciyer We can switch to the traditional form layout. Things would look similar the current start session page where the buttons are wrapped under the label. The question mark popover would also disappear and the explanation would go on `Help text` as well. This looks different than the mockups around sessions and session settings. Should I ignore that and consider the mockup generic?

image

All the other points have been addressed.

lorenzo-cavazzi:
@rokroskar good point, I could adapt the interface to accept other values. I would say we should instead change what the `GET server_options` tells us about that field. Currently, it's marked as an `enum` with the listed options. Should we switch to a `string` instead? I think this should be controlled at the project level now, and be treated as a free-text by renku-notebooks -- it should probably still provide a `/lab` as default

I see how defining this enum at the project level may be useful. Maybe we can store that information in the `renku.ini` file as well?

I like also the "label" idea for some string values, but that sounds more complicated since we should find a solution for the values coming from the options API

rokroskar:
Right, what I had in mind in the renku.ini file is something like

```
sessions:
  environments:
    - name: JupyterLab 
       path: /lab
    - name: "Awesome dashboard"
       path: /streamlit
```

I fear that the current renku.ini won't support this though and we'd need to change to yaml since ini can only have a single level... if the UI relies on renku-notebooks to provide this info, then renku-notebooks needs to read and parse the configuration file as well. I suppose it could do this via the gitlab API directly.

lorenzo-cavazzi:
Switching to `yaml` and embedding structured objects would be great! We could currently workaround this by storing encoded JSON objects in the `ini` file.

We use the `config` API form renku-core to get (and set) the project config, but we use the `server_options` API from renku-notebook to understand the type of option and the defaults.
We get this
```
"defaultUrl": {
    "default": "/lab", 
    "displayName": "Default Environment", 
    "options": [
      "/lab", 
      "/rstudio"
    ], 
    "order": 1, 
    
  },
```
We should change to `"type": "string"` and remove the options, then the UI will show an input field.

ciyer:
>  This looks different than the mockups around sessions and session settings. Should I ignore that and consider the mockup generic?

Yes, you should treat the mock-up as generic. Sorry about that, the mock-up predates the new UI design and was never updated to incorporate the design.

ciyer:
> All the other points have been addressed.

Thanks! After testing some more, I have another request. When a value is being changed, you no longer see the contents of the form anymore just "Updating...". Could we show the form with all fields set to disabled while an update is being made?

lorenzo-cavazzi:
> Could we show the form with all fields set to disabled while an update is being made?

Yes, that's a better solution.
I would also add an `Alert` on the top informing that the operation is ongoing -- not sure if it is always necessary, but the update seems to occasionally take some time (I think that's the case for big projects, but I should double-check it)

lorenzo-cavazzi:
I have addressed all the points (except the `default url`) and added the initial support for the pinned images.
As discussed offline, I'll move to another issue the support for picking the image by the branch/commit -- I left some commented code in the `SessionPinnedImage`, I'll clean it up later if that's ok.

See below for some screenshots and feel free to point out any other improvements we could add. I'm always a bit verbose in the feedback, I can remove any unneeded alert/text.

For the `default url` (@ciyer @rokroskar ), we either need to fix the information we get from the renku-notebook options API (currently, it's indistinguishable from other fixed enumerated properties like "CPU" and "Memory"), or we can treat it as a special case and create a couple of dedicated UI components -- though I would still change the misleading information from renku-notebooks since the UI is the only consumer (afaik) and we won't use them.

-----

**Screenshots**

Multi-column layout with multiline form components
Screenshot_20210611_144708

Pinned docker image
Screenshot_20210611_153530

Visible (disabled) components during an update
Screenshot_20210611_160358

ciyer (changes requested):
This is looking very nice! I like how the view is responsive. A few remaining small details.

1. Show the user selection while updating

After picking a value, the form is disabled (which is good!) but it shows the previous state, not the state it will have after update. Could this be fixed to reflect the state I will have after update?

image

2. "Default RenkuLab Value"

The help text says "Default RenkuLab Value" could this be changed to `Default value:` (I could not find this string in the code...)

3. Two column layout

I like the two column layout. 🎉 Could the columns be made a little closer? They are quite far apart. 

4. Field ordering

Could the fields be reordered a little?

| Column 1           | Column 2         |
| --------------- | --------------- |
| Default Env       | (empty)             |
| Memory            | CPUs                  |
| GPUs                | LFS                     |

5. User specified default environment

I think letting the user specify their own default environment is a priority (and more important than allowing setting a number of CPUs that is not supported by this RenkuLab). Will this be done in a follow-up PR?

lorenzo-cavazzi:
I addressed points 1-3. For the column space, I made the columns closer on xl screens. We can further adjust it, but keep in mind that some sentences may wrap if we make them even closer (like the "unsupported value" string in the screen below).

As discussed offline, the order of the options comes from a renku-notebooks API, so I would rather adjust it at the source.
We can still make an exception for the "default URL" field since it needs to be fixed also from other points of view. I would include that in this PR, after we agree on how to proceed.

I personally think the `info` icons were helpful to distinguish the type of information, but I'm fine with the current solution.
Screenshot_20210614_111918

client/src/project/settings/ProjectSettings.present.js
```diff
@@ -0,0 +1,703 @@
...
+    false;
+
+  // Collapse unknown values by default when none are already assigned
+  const [showImage, setShowImage] = useState(false);
+  const toggleShowImage = () => setShowImage(!showImage);
+
+  const warningMessage = devAccess ?
+    (<Alert color="warning">
+      <FontAwesomeIcon className="cursor-default" icon={faExclamationTriangle} color="warning" /> Changing
+      the following settings may lead to corrupted sessions.
```
ciyer, line 559: I suggest a different message here:

```
Fixing an image can yield improvements, but it can also lead to sessions not working in the expected way. [Please consult the documentation](https://renku.readthedocs.io/en/latest/user/interactive_customizing.html) before changing this setting.
```

client/src/project/settings/ProjectSettings.present.js
```diff
@@ -0,0 +1,703 @@
...
+    );
+  }
+  else if (updated) {
+    const text = value == null ?
+      "unset." :
+      (<span>updated to <code>{value}</code></span>);
+    return (<Alert color="info">&quot;{keyName}&quot; {text}</Alert>);
+  }
+
+  return null;
```
ciyer, line 414: I think we should only show a message here if there was an error. For in progress and on success, I think the feedback in the form is enough.

ciyer (changes requested):
This is close to being ready! I found some small things in the new round of testing and made comments in the code locations. 

Also:

# Form should be disabled if user cannot change values

There is a message that I do not have the rights to change these settings, but I can still interact with the form, so that is confusing. All the elements should be disabled if the user cannot change them.

image

client/src/project/settings/ProjectSettings.present.js
lorenzo-cavazzi: I've changed it as suggested.
I personally preferred a clear message on top, since the spinning wheel next to the option is barely noticeable, but I'm ok with this solution if you think it's better.

lorenzo-cavazzi:
Ah yes, the non-disabled buttons were definitely a bad UX.
I initially thought about removing also the helper/muted text for non-dev users, but I think that provides useful information anyway. Let me know if you have an opinion on that.

Screenshot_20210616_090536

I've also addressed the other points, although I don't particularly like the minimalistic feedback the user gets during the update.

ciyer (approved):
👍

PR #1393, opened 2021-06-14
Transitional landing

### LINKED ISSUE #1374: Transitional Landing Page

ciyer:
Fix #1374 

# Appearance

image

image

image

# Testing

It is probably best to test on https://sekhar.dev.renku.ch/, where it is possible switch between the new and old UIs.

- Check the text and images are a good explanation and illustration of the new features
- Check that switching between the new and old UIs works


client/src/landing/AnonymousHome.js
```diff
@@ -373,13 +374,21 @@ function AnonymousHome(props) {
...
+      // TODO Reactivate after the transition
+      // (props.homeCustomized.custom.enabled) ?
+      //   CustomizedAnonymousHome(p) :
+      //   StandardHome(p)
+      (normalHome) ?
         CustomizedAnonymousHome(p) :
-        StandardHome(p)
+        (normalHome) ?
+          StandardHome(p) :
+          null
```
lorenzo-cavazzi, line 389: This is temporary, but there seems to be an error in the logic.
The second `(normalHome)` will always evaluate as `false` when tested, so that `StandardHome(p)` can't appear. If that is correct, I would use a `null` instead of
```
(normalHome) ?
  StandardHome(p) :
  null
```

lorenzo-cavazzi (changes requested):
Great addition!

I have noticed the content is misaligned on the left. I compared it with the current solution. The title `Try out the new UI` seems to be correctly aligned in the `ui-design-2021` branch (it's `Bringing everyone together!` there), while the gray-background content may be a bit off also there. I believe it's just a matter of extra `<div>` elements. Using `<Fragments>` instead may help in preserving the left/right alignment.

Screenshot_20210615_084413

client/src/landing/AnonymousHome.js
ciyer: > This is temporary, but there seems to be an error in the logic.
> The second `(normalHome)` will always evaluate as `false` when tested, so that `StandardHome(p)` can't appear. If that is correct, I would use a `null` instead of
> 
> ```
> (normalHome) ?
>   StandardHome(p) :
>   null
> ```

😃 -- the reason for this silliness is I wanted to leave all the old code around (it will get revived when the transition is finished), but if the components aren't referenced, the linter complains. If you think it is important, I can try to create a cleaner solution to this.

ciyer:
> I have noticed the content is misaligned on the left. I compared it with the current solution.

Good eye! I added some padding that was causing the problem. I have fixed it now.

image

client/src/landing/AnonymousHome.js
lorenzo-cavazzi: No don't worry. I was a bit confused because I thought the code to restore was the commented one, but I see removing it would require commenting out other parts to prevent the linter from complaining

lorenzo-cavazzi:
Very nice, it looks great! Just another small detail on the `<div>` elements.
I've noticed they have extra padding on the right (10px). Is that on purpose? On smaller screens, that makes some elements look not aligned on the right -- in the screenshot, it's already visible for the text (+10px) and even more for the images (+20px).

Screenshot_20210616_083819

P.S: The `min-width` in the grey `<div>` makes the content overflow on smaller screens. That doesn't happen on the current landing page, but I see it's useful to prevent overlapping the text and the background dots. Should we apply the min-width to the whole landing component to preserve the correct scrolling? Or maybe removing the `min-width` and apply a `min-height` on smaller devices to prevent overflowing and text overlapping at the same time?

Screenshot_20210616_084215
Screenshot_20210616_083851

ciyer:
> I've noticed they have extra padding on the right (10px). Is that on purpose? On smaller screens, that makes some elements look not aligned on the right -- in the screenshot, it's already visible for the text (+10px) and even more for the images (+20px).

Yes, the padding is necessary to keep the images from running into each other. With the designer, we decided on left-justified, ragged right, so it's ok if the right side is not aligned.

> P.S: The `min-width` in the grey `<div>` makes the content overflow on smaller screens. That doesn't happen on the current landing page, but I see it's useful to prevent overlapping the text and the background dots. Should we apply the min-width to the whole landing component to preserve the correct scrolling? Or maybe removing the `min-width` and apply a `min-height` on smaller devices to prevent overflowing and text overlapping at the same time?

Interesting, this does not happen on Safari or Chrome, but I could re-create on Firefox. I have fixed it (see image below), but I did not re-deploy.

image

lorenzo-cavazzi (approved):
Nice addition for the initial rollout of the new design :+1:

PR #1394, opened 2021-06-15
feat: add stop button and dropdown list to show session page

### LINKED ISSUE #1370: Session Controls
# Description

We recently embedded the sessions in the UI using an iframe.

# Expand button

We should add an "expand" button that extends the surface dedicated to the session (see red circle in the image below). We could extend to full-screen (or better full-browser) all the sessions area (blue rectangle). Consider also binding the resize action to pressing "esc" on the keyboard if it doesn't conflict with anything else in JupyterLab

image

# Stop button

It is necessary to go back to the session list to stop the session. This is annoy. There should be a button to stop the session in the header we show (it should confirm before stopping).

# List item menu

The menu
image

lorenzo-cavazzi:
The user can now stop the environment without going back to the environment list. That also handles errors through the notifications, although it doesn't happen very frequently.
A dropdown menu has been added for the interactions ("stop" and "open in a new tab").

Screenshot_20210615_163530

As discussed, the button to maximize the iframe won't be added immediately since other changes are on the way and they may be enough.
I would still reduce the left and right margins from 2em to 0.5em (or even 0.25em)

/deploy renku=ui-design-2021
fix #1370


ciyer (changes requested):
Code changes are good, and the feature works well. Only one small request, could you change the color of the stop button to the Renku Blue (`btn-primary`). We use green for buttons we *want* the user to press, blue for ones the can press.

image

lorenzo-cavazzi:
@ciyer fixed

ciyer (approved):
👍

PR #1395, opened 2021-06-17
build: release 1.0.0-beta1

ciyer:


lorenzo-cavazzi (approved):
Version 1 on the way :tada:

lorenzo-cavazzi (approved):

Issue #1390, opened 2021-06-11
Landing page update


ciyer:
# Description

Update the landing page to take advantage of some new UI elements.

Main changes:
* removed "starred projects"
* use the tiles for showing projects
* include links to the help page (Learn more), tutorial (do the tutorial), and new project page (create a new project)

# Mock-up

![image](https://user-images.githubusercontent.com/1196411/121707287-b2697580-cad6-11eb-96fe-720163079fe4.png)

PR #1399, opened 2021-06-21
feat(project): title and slug in search result

vfried:
Rok asked me to do this change... instead of slug we display the project title inside the search result. The slug is displayed but smaller.

image

image

Possible changes:
- Make description and slug the same size
- In case there is space in the card put the slug under the title

/deploy renku=ui-design-2021

### DIFF
--- a/client/src/project/list/ProjectList.present.js
+++ b/client/src/project/list/ProjectList.present.js
@@ -91,7 +91,8 @@ function ProjectListRows(props) {
       id: project.id,
       url: url,
       itemType: "project",
-      title: project.path_with_namespace,
+      title: project.name,
+      slug: project.path_with_namespace,
       description: project.description ?
         <Fragment>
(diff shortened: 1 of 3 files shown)


ciyer (approved):
This looks good!

I noticed that the "Your Projects" section still has the old slug display, but we can leave it, since that page will change soon.

image

lorenzo-cavazzi:
Sorry for interfering, are we sure we want to keep the current solution on the homepage?
Even with a not-too-long username, the project list is not very readable unless the screen is large.
I would use the project name also there. An alternative could be allowing the slug to take more than a single line.
Screenshot_20210623_152219

ciyer:
> Sorry for interfering, are we sure we want to keep the current solution on the homepage?

This does not look good, but #1390 is scheduled to be worked on soon, and it will be fixed there.

vfried:
I think we are not using the same component there/in the search result... we should remember to delete that component when we change the welcome page.

PR #1404, opened 2021-06-24
feat: add link to registry images

lorenzo-cavazzi:
This adds a link on the image session setting to the container registry page in GitLab.

P.S. the screenshot may look slightly different than the preview because it includes the changes to the buttons from #1403 

image

/deploy renku=ui-design-2021

### DIFF
--- a/client/src/project/settings/ProjectSettings.present.js
+++ b/client/src/project/settings/ProjectSettings.present.js
@@ -38,7 +38,7 @@ import _ from "lodash/array";
 import { ACCESS_LEVELS } from "../../api-client";
 import { ProjectAvatarEdit, ProjectTags, } from "../shared";
 import { NotebooksHelper, ServerOptionBoolean, ServerOptionEnum, ServerOptionRange } from "../../notebooks";
-import { Clipboard, Loader, RenkuNavLink } from "../../utils/UIComponents";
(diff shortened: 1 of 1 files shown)


client/src/project/settings/ProjectSettings.present.js
```diff
@@ -707,11 +708,20 @@ function SessionPinnedImage(props) {
     </Fragment>);
   }
 
+  const imagesUrl = `${repositoryUrl}/container_registry`;
+  const imagesLink = (<div>
+    <FormText>
+      You may want to check the {" "}
+      <ExternalLink role="link" title="available images for this project" url={imagesUrl} />.
```
ciyer, line 715:
```suggestion
      A URL of a RenkuLab-compatible Docker image. For an image from this project, {" "}
       consult  <ExternalLink role="link" title="the list of images for this project" url={imagesUrl} />.
```

ciyer (changes requested):
Thanks for putting this together so quickly! It makes it much easier to find an image.

While testing, I identified an an improvement that would be good to make; it can be done in this issue, if easy, otherwise, we should create a new issue:

If there are advanced settings set for this project, the advanced settings section should be open by default.

lorenzo-cavazzi:
The suggested change for the images is actually very easy to implement (I initially implemented that, then removed it later).
Both points have been addressed in the latest commit.

ciyer (approved):
👍

PR #1406, opened 2021-06-28
fix: prevent duplication of falsy options

### LINKED ISSUE #1122: Default value added to enum options
On https://limited.renku.ch/ , the default value of GPU seems to be added as an extra `0` despite that value being already available in the list of choices.
It's possible that the default value is a `"0"` instead of a `0`, or the other way around.
Double-check the `enum` options

image (2)
image (3)
image (4)

From here: https://swiss-data-science.slack.com/archives/CAJMXA6PQ/p1605257685165200

lorenzo-cavazzi:
Remove the duplication on notebooks options that could be tested to false.
That may happen when receiving the number `0` as one of the possible values.

Screenshot_20210628_152245

**How to test**: This would require a deployment with GPUs. One way to manually fake it would be to add the following code before the `return data;` at line 82 in the file "client/src/api-client/notebook-servers.js" 
```
data = {
  ...data, gpu_request: {
    "default": 0,
    "displayName": "Number of GPUs",
    "options": [
      0,
      1,
      2
    ],
    "order": 5,
    "type": "enum"
  }
};
```

/deploy renku=ui-design-2021
fix #1122

### DIFF
--- a/client/src/notebooks/Notebooks.container.js
+++ b/client/src/notebooks/Notebooks.container.js
@@ -451,7 +451,7 @@ class StartNotebookServer extends Component {
   setServerOptionFromEvent(option, event, providedValue = null) {
     const target = event.target.type.toLowerCase();
     let value = providedValue;
-    if (!providedValue && providedValue !== 0) {
+    if (!providedValue != null) {
       if (target === "button")
         value = event.target.textContent;
 
--- a/client/src/notebooks/Notebooks.container.js
+++ b/client/src/notebooks/Notebooks.container.js
@@ -451,7 +451,7 @@ class StartNotebookServer extends Component {
   setServerOptionFromEvent(option, event, providedValue = null) {
     const target = event.target.type.toLowerCase();
     let value = providedValue;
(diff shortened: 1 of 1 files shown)


client/src/notebooks/Notebooks.container.js
```diff
@@ -451,7 +451,7 @@ class StartNotebookServer extends Component {
   setServerOptionFromEvent(option, event, providedValue = null) {
     const target = event.target.type.toLowerCase();
     let value = providedValue;
-    if (!providedValue) {
+    if (!providedValue && providedValue !== 0) {
```
ciyer, line 454:
```suggestion
    if (!providedValue != null) {
```

I think the logic here is clearer if we test for `null`, especially if we consider, in addition to `providedValue === 0` the possibility of what happens when `providedValue === false`.

ciyer (changes requested):
Behavior looks good! Made a suggestion for the code.

ciyer (approved):
👍

PR #1412, opened 2021-07-05
feat: make apearance of switch state clearer

### LINKED ISSUE #1357: UI Design Update -- Fix switch state
## Description

The state of these switches is counter-intuitive.

image

"Lineage" is selected, but to my eye, it looks like "Contents" is selected.

ciyer:
Fix #1357

## Look

**Contents Selected**

image

**Hover over Lineage**

image

**Add Dataset**

image

**Style Guide**

image

**Notebook Launch**

image

/deploy

### DIFF
--- a/client/src/notebooks/Notebooks.present.js
+++ b/client/src/notebooks/Notebooks.present.js
@@ -1800,18 +1800,18 @@ class ServerOptionEnum extends Component {
       return (<Badge color="primary">{this.props.options[0]}</Badge>);
 
     return (
-      <ButtonGroup>
+      <ButtonGroup className="rk-btn-group-light" >
         {options.map((optionName, i) => {
-          let color = "outline-primary";
+          let color = "rk-white";
           if (optionName === selected) {
             color = this.props.warning != null && this.props.warning === optionName ?
               "danger" :
-              "primary";
+              "rk-white";
           }
           const size = this.props.size ? this.props.size : null;
           return (
             <Button
-              key={optionName} color={color} size={size} disabled={disabled}
+              key={optionName} color={color} size={size} disabled={disabled} active={optionName === selected}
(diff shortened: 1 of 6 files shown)


vfried (commented):
Things look good, I'm not 100% sure the colors are clear for the user since we have similar buttons but the white means "unselected"

image

When i tried to use the new switch my first intention was to click on the white button, nothing was changing and i thought the button was "broken" since the button you should click looks like a deactivated button to me.

image

We can merge this and if it's confusing we can change it again I guess.

ciyer:
I made the suggested changes and applied the styling to the new dataset form as well.

vfried:
image

Well done, I like the way this looks and it's clear as well.

I don't like these buttons here... but it's more about the combination of colors, maybe this is something that Luca can redesign for us. We could also remove the switch from that file or put it somewhere else.

image

Shall we change this here as well....
image

image

vfried:
Thanks for addressing the changes I requested. Things are looking really good. 

I just have two small thing to point out... 

1. I see there is a weird white border when an element is selected, could we get rid of it?
image

2. there is an issue with a switch in the new sessions:
- no clicks:
image

- click on: 0.5
image

- click on 0.1
image

- click on 0.5 
image

I imagine this is not expected behaviour, could we fix it?

ciyer:
> 2. there is an issue with a switch in the new sessions

Yes, I created an issue for this: #1424

I tried to make a quick fix, but I need a bit more time to understand the problem, and the code is unrelated to the switch state code, so I think it is better to handle this in a separate PR.

vfried:
All looks good buuuuut... there is one last issue i found here: when things are collapsed the bar is bigger (height) than when they are not.

vfried (approved):
Things look good now, thanks for addressing the changes!

Issue #1345, opened 2021-05-10
Adapt the UI around starting new sessions


lorenzo-cavazzi:
## Motivation

There have been some changes in the renku-notebooks APIs that require adaptation on the UI side, especially around the autosave branches.
Specifically, these are the most relevant changes:
- We have an API endpoint to get autosave information https://github.com/SwissDataScienceCenter/renku-notebooks/pull/607
-  Old autosaves are automatically deleted when starting from a more recent commit https://github.com/SwissDataScienceCenter/renku-notebooks/pull/611

## Mock-ups / Design

TO-DO: we need a more intuitive user experience now that autosave branches will be automatically deleted.
Also, we don't need to consider those branches directly since we get the same information from the API
Since we have information from many different sources on the page, consider dropping the refresh button for single entities (e.g. branches and commits) and implement a single refresh content button.

## Progress

- [ ] API Design (sprint-??)
- [ ] UI Implementation

PR #1415, opened 2021-07-06
fix: do not autostart sessions if no image is available (#1413)

### LINKED ISSUE #1413: The "play" button does not check for image readiness
When trying to launch a new session with the ⏯️  button, the status of the image build pipeline is ignored. If the image is not ready, the session is launched with the fallback image. This should not be the case and the user should be made to wait until the image build is complete.

ciyer:
# Testing

To test, run try to use the start button on projects where the image is still building or the image build failed.

These projects can help:

* https://renku-ci-ui-1415.dev.renku.ch/projects/renku-qa/slow-build takes 10 minutes to build the Dockerfile
* https://renku-ci-ui-1415.dev.renku.ch/projects/renku-qa/broken-build has a problem with building the Dockerfile

Fix #1413

/deploy renku=ui-design-2021

### DIFF
--- a/client/src/notebooks/Notebooks.container.js
+++ b/client/src/notebooks/Notebooks.container.js
@@ -438,8 +438,33 @@ class StartNotebookServer extends Component {
         const data = this.model.get();
         const fetched = data.notebooks.fetched && data.options.fetched && data.pipelines.fetched;
         if (fetched) {
-          this.setState({ autostartReady: true });
-          this.startServer();
+          const mainPipeline = data.pipelines.main;
+          if (mainPipeline && mainPipeline.status === "success") {
+            this.setState({ autostartReady: true });
+            this.startServer();
+          }
+          else if (mainPipeline && (mainPipeline.status === "running" || mainPipeline.status === "pending")) {
+            this.setState({
+              autostartReady: true,
+              autostartTried: true,
+              launchError: {
+                frontendError: true,
+                errorMessage: `The session could not start because the image is still building.
+                Please wait for the build to finish, or start the session with the base image.`
+              },
+              starting: false });
+          }
+          else {
+            this.setState({
+              autostartReady: true,
+              autostartTried: true,
+              launchError: {
+                frontendError: true,
+                errorMessage: `The session could not start because no image is available.
+                Please select a different commit or start the session with the base image.`
+              },
+              starting: false });
+          }
         }
         else {
           setTimeout(() => { this.triggerAutoStart(); }, 1000);
@@ -500,7 +525,7 @@ class StartNotebookServer extends Component {
   startServer() {
     //* Data from notebooks/servers endpoint needs some time to update properly.
     //* To avoid flickering UI, just set a temporary state and display a loading wheel.
-    this.setState({ "starting": true });
+    this.setState({ "starting": true, launchError: null });
     this.internalStartServer().catch((error) => {
       // Some failures just go away. Try again to see if it works the second time.
       setTimeout(() => {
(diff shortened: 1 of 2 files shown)


vfried:
The code looks good to me... I would make two small changes here... After that, I'm ready to approve the PR.

1) If the image is not available i would deactivate the start button (this is assuming the user shouldn't start a session here). I clicked on the button and it started some environment for me for some reason.

2) The error message should be displayed in "danger" color and also on top of the "start session" button instead of under tit

image

rokroskar:
Shouldn't we wait for the image build to finish without changing to the "advanced" options? The progress bar should just show "waiting for build to finish" or something similar and only switch to the advanced options in case the image build fails (with the appropriate message)

vfried:
@rokroskar that's a good point... I think we should do what you say instead of redirecting the user. If this is not so easy to do we could do it in a follow-up PR.

ciyer:
> The code looks good to me... I was only able to test things with the broken image, the slow build didn't work for me.

I did not explain that very well. For the slow build, you need to first do something forces the Dockerfile to rebuild without reusing cached layers. For example, change the sleep duration -- I will add a comment to the project's README to explain that.

> I would make two small changes here... After that, I'm ready to approve the PR.
> 
> 1. If the image is not available i would deactivate the start button (this is assuming the user shouldn't start a session here). I clicked on the button and it started some environment for me for some reason.
> 2. The error message should be displayed in "danger" color and also on top of the "start session" button instead of under tit

Good suggestions! For 1, I think it *should* still be possible to start a session (because you may have a mistake somewhere and do not want to wait until the image builds to fix it), but it should be clearer that you will be getting a different images. And you are right about 2, I will make that change as well.

ciyer:
> Shouldn't we wait for the image build to finish without changing to the "advanced" options? The progress bar should just show "waiting for build to finish" or something similar and only switch to the advanced options in case the image build fails (with the appropriate message)

This will lead to bad UX since we do not know how long the image build will take. If the build takes 40 minutes, and we just show "waiting for build to finish," it will look like things are broken.

I think the start button should either start the session immediately, or, if it is determined that this is not possible, give the user the option to wait until the session can be started (with the ability to get information about what is delaying the session launch), or start the session with different settings.

rokroskar:
what if we showed a small "cancel this session request" link when waiting for the build to finish? That would take you to the normal session launch page. As it is now it's super confusing because you've clicked start but then you're just dropped into some other page without explanation.

ciyer:
> what if we showed a small "cancel this session request" link when waiting for the build to finish? That would take you to the normal session launch page. As it is now it's super confusing because you've clicked start but then you're just dropped into some other page without explanation.

I agree, that interaction was confusing because the explanation for why it happened was buried at the bottom of the page. I have changed it to be this now:

image

or, for a broken image:

image

I have also changed the start buttons in this case to be:

image

rokroskar (approved):
functionality and messaging make sense 👍 thanks for the quick fix!

vfried (approved):
Things look good to me now 👍 

Some things to point out about the notebooks form in general:
- I think the start buttons in the form should be green (I thought action buttons, in general, were gonna be that color)
- The form looks messy in general, think we should remove some color from it and maybe add some spacing between fields, it has too many blue buttons. I know we have a PR but every time I interact with the form I have the same feeling 🤔

ciyer:
> Things look good to me now 👍
> 
> Some things to point out about the notebooks form in general:
> 
> * I think the start buttons in the form should be green (I thought action buttons, in general, were gonna be that color)
> * The form looks messy in general, think we should remove some color from it and maybe add some spacing between fields, it has too many blue buttons. I know we have a PR but every time I interact with the form I have the same feeling 🤔

I agree. Could you add those comments to this ongoing issue about the launch page: https://github.com/SwissDataScienceCenter/renku-ui/issues/1345

PR #1420, opened 2021-07-12
Header takes up less vertical space

### LINKED ISSUE #1389: Design: make header smaller
# Description

The current header is very large:

By switching to just the frog, we can make the header take up less space, while retaining its functionality. 

# Implementation

- We should aim to have the header size the "standard" bootstrap header size. 
- Try making the icon respond to screen width: for small, just the frog, for large, renku + frog.

ciyer:
# Changes

## Desktop

image

## Mobile

image

/deploy renku=ui-design-2021

Fix #1389

### DIFF
--- a/client/src/App.css
+++ b/client/src/App.css
@@ -9,7 +9,7 @@
 
 .App-header {
   background-color: #222;
-  height: 150px;
+  height: 70px;
   padding: 20px;
   color: white;
 }
--- a/client/src/landing/NavBar.js
+++ b/client/src/landing/NavBar.js
@@ -222,24 +222,24 @@ class LoggedInNavBar extends Component {
       <header className="navbar navbar-expand-lg navbar-dark rk-navbar p-0">
         <Navbar color="primary" className="container-fluid flex-wrap flex-lg-nowrap renku-container">
           <NavbarBrand href="/" className="navbar-brand me-2 pb-0 pt-0">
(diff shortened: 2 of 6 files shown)


vfried:
This looks really good, i'm on a smaller screen now and it can really make a difference to have a bit more space in the screen.

Two things to point out:
- padding in the search bar when things are collapse should not be 4... to achieve this you can change the "pe-4" for "pe-0 pe-lg-4" that should do it.
- It would be nice if the amount of space on top and bottom of the frog is the same and to have a bit more space under the search field, etc... we could maybe add some internal padding for this or make the padding on top smaller.

It would be good if Luca reviews this as well, to see that it looks good. But it's not 100% necessary.

ciyer:
> This looks really good, i'm on a smaller screen now and it can really make a difference to have a bit more space in the screen.
> 
> Two things to point out:

I made these changes:

image

image

vfried:
All looks good buuuuut... there is one last issue i found here: when things are collapsed the bar is bigger (height) than when they are not.

ciyer:
> All looks good buuuuut... there is one last issue i found here: when things are collapsed the bar is bigger (height) than when they are not.

Yes, I need to add a little bit of vertical space because the elements were too squashed together, and it didn't look good.

lorenzo-cavazzi (approved):
Looks nice! :+1: 

I've re-run the acceptance tests cause they were missing the reference to the renku branch `renku=ui-design-2021`

PR #1421, opened 2021-07-13
feat: extend the list of extensions for code files in lineage display

ciyer:
# Screenshot

https://dev.renku.ch/projects/lee.gavin.k/plurilingue/files/lineage/outputs/num.txt

image

### DIFF
--- a/client/src/file/Lineage.present.js
+++ b/client/src/file/Lineage.present.js
@@ -67,7 +67,14 @@ function getNodeLabel(node, NODE_COUNT, lineagesUrl) {
 function nodeToClass(node, currentNodeId, label) {
   const nodeId = node.id;
   const nodeType = node.type;
-  const FORMATS = { "py": true, "r": true, "ipynb": true };
+  // see also File.container.js > CODE_EXTENSIONS
+  const CODE_EXTENSIONS = [
+    "bat", "jl", "js", "py", "r", "rmd", "rs", "scala", "sh", "ts",
+    "c", "cc", "cxx", "cpp", "h", "hh", "hxx", "hpp", // C++
+    "f", "for", "ftn", "fpp", "f90", "f95", "f03", "f08" // Fortran
+  ];
+  const FORMATS = { "ipynb": true };
+  CODE_EXTENSIONS.forEach(e => FORMATS[e] = true);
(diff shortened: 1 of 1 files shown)


client/src/file/Lineage.present.js
```diff
@@ -67,7 +67,14 @@ function getNodeLabel(node, NODE_COUNT, lineagesUrl) {
...
   const nodeType = node.type;
-  const FORMATS = { "py": true, "r": true, "ipynb": true };
+  // see also File.container.js > CODE_EXTENSIONS
+  const CODE_EXTENSIONS = [
+    "bat", "jl", "js", "py", "r", "rmd", "rs", "scala", "sh", "ts",
+    "c", "cc", "cxx", "cpp", "h", "hh", "hxx", "hpp", // C++
+    "f", "for", "ftn", "fpp", "f90", "f95", "f03", "f08" // Fortran
+  ];
+  const FORMATS = { "ipynb": true };
+  CODE_EXTENSIONS.forEach(e => FORMATS[e] = true);
```
vfried, line 77: What would be the difference where between FORMATS and CODE_EXTENTIONS... besides in the meaning, i don't understand why there was a difference made here.

vfried (commented):
Hey, i trust there is some change made here but i can't find the difference between the graph in dev and the one I get when i use this branch, could you explain to me how to test this?

client/src/file/Lineage.present.js
ciyer: I introduced `CODE_EXTENSIONS` to make the code more similar to what is in File.container.js, which has a similar, but not exactly the same, purpose. I left `FORMATS` to allow the other logic to say the same.

ciyer:
> Hey, i trust there is some change made here but i can't find the difference between the graph in dev and the one I get when i use this branch, could you explain to me how to test this?

This is what the graph looks like in dev:

image

All the nodes are white, meaning they are just "files." In the fixed version they are pink indicating that they are code. 

To test, you should just check that the file extensions for R (both lower and uppercase), python, and sh are all identified as code when the lineage is shown.

vfried (approved):
Thanks for your clarification, code looks good and things are working 👍

PR #1418, opened 2021-07-09
Fix start button behavior for projects with pinned images

ciyer:
# Test

Four cases to test:

1. Pinned image: https://renku-ci-ui-1418.dev.renku.ch/projects/cramakri/pinned-image-test (should start smoothly)
2. Slow build: https://renku-ci-ui-1418.dev.renku.ch/projects/renku-qa/slow-build (should start smoothly if image has built)
3. Slow build: https://renku-ci-ui-1418.dev.renku.ch/projects/renku-qa/slow-build (should show warning if image is building)
4. Broken build: https://renku-ci-ui-1418.dev.renku.ch/projects/renku-qa/broken-build (should show warning)

/deploy renku=ui-design-2021

PR #1428, opened 2021-07-16
build: release 0.11.14

ciyer:
/deploy


rokroskar:
can we get `beta4` with #1418 and then merge this?

PR #1441, opened 2021-08-02
feat(files): adjustable width in files display inside projects

### LINKED ISSUE #1423: Adjustable lateral window on files section
Hey Renku team,

The new UI looks amazing, I truly love it! I have been playing with it, and I have a small request to make. It would be cool to make the following lateral window, where all files are shown, adjustable, mainly to allow expanding it, as it is currently difficult to read the file names, even on a large screen. Here you have an image showing what I am referring to:

image

Thank you so much, keep on making Renku cooler every day!

vfried:
I added a pure css solution to making the files display list have variable widths:

image

/deploy renku=ui-design-2021

Closes #1423


lorenzo-cavazzi (changes requested):
This is very nice!
There are a couple of things to adjust before merging
- On smaller screens (`xs`), we probably want the keep the older behavior with a width 100%
- Until the user resizes the container, the size is a percentage of the available width. This doesn't work well when the screen isn't large enough cause the content is partially cut (see screen). Should we add a fixed `min-width`? I would pick a size that keeps the 2 buttons on the top (Connenct / Lineage) visible (180px should be the bare minimum, I would use 200px)
  Screenshot_20210804_084625

- On firefox, the resize indicator is not very visible (see screen) and the user can interact with the size only on the bottom right corner. We may try to make the container boundary more visible. Maybe a different background color on the files container could work? Other ideas are welcome
  Screenshot_20210804_084756

vfried:
Thanks for your review, I decided to set the initial width and the min-width as 200px, I think that's a good size for it.

Regarding the image in Firefox, unfortunately, the CSS property for changing the background image is not implemented by the browser. It's the only browser that doesn't support it, I would say it's a fair tradeoff.

image

lorenzo-cavazzi:
Ok I see. I pushed a change to improve the visibility of the file tree boundaries and simplify seeing the borders, let me know if you think it's better.
That also includes the full width on small windows -- I found a solution to keep a similar mechanism when users make the file tree container big. That makes the content container wrap so that its content won't be too small and unreadable.
Feel free to change anything if you think it's not good or you find a better solution.

- **Small screen**
  Screenshot_20210810_110013

- **Medium screen**
  Screenshot_20210810_105934

- **Medium screen with large file tree container**
  Screenshot_20210810_105958

lorenzo-cavazzi (dismissed):

ciyer:
Very nice code to solve this, and it works well on Chrome and Firefox... but it doesn't work on Safari, and I do not understand why.

ciyer (changes requested):
Very nice code to solve this, and it works well on Chrome and Firefox... but it doesn't work on Safari, and I do not understand why. I think we should make an attempt to get it working on Safari, but if, after 2h, we do not have a solution, let us go ahead and merge.

ciyer (approved):
Works very well now! I think the resize icon is a little too thick, I think, but we can go with this for now and ask the designer to provide us a better icon as a replacement.

Please squash to merge.

Issue #1380, opened 2021-06-03
Add Renku cheatsheet to sessions page


ciyer:
# Description

The sessions page should display the cheatsheet as was implemented in the JupyterLab plugin: https://github.com/SwissDataScienceCenter/renku-jupyterlab-ts/pull/30

<img width="1144" alt="image" src="https://user-images.githubusercontent.com/1196411/102483423-01bd7480-4065-11eb-8634-151eef0bd7de.png">

PR #1466, opened 2021-08-25
feat(sessions): added cheat sheet to session page (#1380)

ciyer:
Brings the cheat sheet from the JupyterLab plug-in into the sessions page.

image

## Testing

To test, start a session and click on the book icon to see the cheat sheet.


client/src/notebooks/Notebooks.present.js
```diff
@@ -293,6 +295,23 @@ function SessionDocs(props) {
...
+  const { tab } = props;
+
+  if (tab !== SESSION_TABS.commands)
+    return null;
+
+  // ? Having a minHeight prevent losing the vertical scroll position.
+  // TODO: Revisit after #1219
+  return (
+    <Fragment>
+      <div className="p-2 p-lg-3 border-top" style={{ minHeight: 800 }}>
```
lorenzo-cavazzi, line 308: `border-top` seems to show a double border in Firefox since a border is already added by the parent `<div>`

lorenzo-cavazzi (changes requested):
This looks good!

Just a couple of notes:
- There is an extra top border (see inline comment)
- When the window width is small, some of the columns take very little space, making the content hard to read. Wrapping them could be better a better solution.
  Screenshot_20210825_171756

Using the `Card` element from React may have simplified handling the wrapping logic, although this is a port from JupyterLab UI components, so rewriting all from scratch would have taken more time -- it probably doesn't worth the effort.

PR #1473, opened 2021-08-30
fix: use link on the logo to prevent reload

lorenzo-cavazzi:
Clicking on the logo used to reload the application because a standard anchor element was used instead of a react `<Link>`.
This PR fixes it.

**How to test**: browse to any page (e.g. `/projects`) and try to go back to the home by clicking on the logo. Compare the behavior with that on https://dev.renku.ch

/deploy renku=ui-design-2021


ciyer (approved):
Nice fix!

PR #1482, opened 2021-09-02
fix(datasets): creators payload sent properly

### LINKED ISSUE #1470: Creators data lost when editing/creating a dataset
## Description

While creating/editing a dataset, the "Creators" data are not correctly sent to the API. The Name appears to be truncated, while the Email and Affiliation both go lost.
Checking the Network tab, it seems that the payload sent to the backend is wrong.

## Reproduce

Try to add a Creator in a dataset.

In the example, the data for a new creator (Creator2, email2@email.com, SDSC) is correctly entered, but the payload from the Network tab on the right shows that only the truncated name is sent.

Screenshot_20210827_161754

vfried:
There was an error in a useEffect function, it has been fixed now.

/deploy renku=ui-design-2021

Closes #1470


ciyer (approved):
Such a simple fix -- it all works well now!

PR #1483, opened 2021-09-02
Allow popups on session iframes

### LINKED ISSUE #1465: RStudio popup windows blocked when displayed in iframe
## Description

With the new UI, the session is displayed in an iframe by default. This design does not allow popup windows. RStudio relies heavily on popup windows (e.g., git operations, pdf report viewer, switching to another project inside RStudio, ...) All these windows are currently blocked and RStudio seems frozen or the corresponding feature is not accessible. When it is switched into full view with the "Open in new tab" button, everything works as expected.

## Reproduce

1. Create a Renku project with the Renku/renkulab-r base template
2. Make sure to switch to the new UI
3. In the project, click "Start"
4. When RStudio appears in the iframe, edit somehow a file, say README.md and save changes
5. Go to the "Git" tab at right, stage the README.md file and click the "Commit" button
6. The Popup blocked dialog box appear. Clicking on "Try again" does not solve the problem

Thus, it is impossible to commit with the RStudio UI in iframe mode. Switching to full frame, the feature is working as expected. It is the same for several other RStudio functions that rely on popup windows for its UI.

## Solution

Either allow popup windows from sessions in iframe (hard to do?) or go back to full frame session, at least by default, if not completely. The iframe session is **not** a good idea for different reasons, imo.

ciyer:
# Testing

Follow the instructions in #1465 check that this works now.

Fix #1465

/deploy renku=ui-design-2021

### DIFF
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -213,6 +213,7 @@
           "papermill",
           "pathname",
           "poller",
+          "popups",
           "prepend",
           "presentational",
           "Pupikofer",
--- a/client/src/notebooks/Notebooks.present.js
+++ b/client/src/notebooks/Notebooks.present.js
@@ -328,7 +328,8 @@ function SessionJupyter(props) {
         "";
       content = (
         <iframe id="session-iframe" title="session iframe" src={notebook.data.url} className={localClass}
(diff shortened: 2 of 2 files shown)


lorenzo-cavazzi (approved):
That works well! :tada: 
Although I would have something to say on the R studio UX with those window-popups :roll_eyes:

PR #1488, opened 2021-09-03
fix(datasets): proper error handling when dataset add fails

### LINKED ISSUE #1077: If dataset create succeeds but dataset add fails on dataset creation the error is not handled well
When this happens, the user sees a message like the entire dataset creation failed when only the add files operation failed.

We should show a message that allows the user to try to add the files again without having to upload the files again.

It would be better to have the create operation work in an atomic way in this case: first upload the files and on dataset creation send the file ids but it's not the way things are working at the moment.

vfried:
When a dataset is created or edited it could happen that this succeeds but the file add function fails.

When this happens the error is not managed properly, a message is displayed making the user think that the entire operation failed.

This fix provides the user with a better picture of what happens.

1. If the dataset is created but the file add fails,  we redirect the user to the dataset and show an error message.
image

2. If the dataset is edited but the file add fails,  we show the user a message allowing them to retry on spot (no redirection necessary).
image

### Testing
1-To make things fail: api-client/dataset.js --> line 227 make "name" become "namee" (example)
2-Add a dataset (you can see the error message), then edit a dataset (to see the error message on edit)

Closes #1077

### DIFF
--- a/client/src/dataset/Dataset.present.js
+++ b/client/src/dataset/Dataset.present.js
@@ -298,7 +298,7 @@ function ErrorAfterCreation(props) {
   return props.location.state && props.location.state.errorOnCreation ?
     <Alert color="danger">
       <strong>Error on creation</strong><br/>
-      The dataset was crated but there was an error adding files to it.<br/>
+      The dataset was created, but there was an error adding files to it.<br/>
       Please {editButton} the dataset to add the missing files.
     </Alert>
     : null;
--- a/client/src/dataset/Dataset.present.js
+++ b/client/src/dataset/Dataset.present.js
@@ -287,6 +287,23 @@ function DatasetError(props) {
   );
 }
 
+function ErrorAfterCreation(props) {
+  const editButton = <Link className="float-right me-1 mb-1" id="editDatasetTooltip"
+    to={{ pathname: "modify", state: { dataset: props.dataset } }} >
+    <Button size="sm" color="danger" >
+      <FontAwesomeIcon icon={faPen} color="dark" /> Edit
+    </Button>
+  </Link>;
+
+  return props.location.state && props.location.state.errorOnCreation ?
+    <Alert color="danger">
+      <strong>Error on creation</strong><br/>
+      The dataset was crated but there was an error adding files to it.<br/>
+      Please {editButton} the dataset to add the missing files.
+    </Alert>
+    : null;
+}
+
 export default function DatasetView(props) {
 
   const [addDatasetModalOpen, setAddDatasetModalOpen] = useState(false);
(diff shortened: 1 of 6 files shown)


client/src/dataset/Dataset.present.js
```diff
@@ -287,6 +287,23 @@ function DatasetError(props) {
...
+    to={{ pathname: "modify", state: { dataset: props.dataset } }} >
+    <Button size="sm" color="danger" >
+      <FontAwesomeIcon icon={faPen} color="dark" /> Edit
+    </Button>
+  </Link>;
+
+  return props.location.state && props.location.state.errorOnCreation ?
+    <Alert color="danger">
+      <strong>Error on creation</strong><br/>
+      The dataset was crated but there was an error adding files to it.<br/>
```
ciyer, line 301: Spelling + comma

```suggestion
      The dataset was created, but there was an error adding files to it.<br/>
```

client/src/project/datasets/change/DatasetChange.container.js
```diff
@@ -65,7 +65,7 @@ function ChangeDataset(props) {
       titleField.help = datasetFormSchema.title.help;
       titleField.parseFun = undefined;
       image.value = {
-        options: [{ [Prop.URL]: props.dataset.mediaContent }],
+        options: [{ [Prop.URL]: props.dataset ? props.dataset.mediaContent : undefined }],
```
ciyer, line 68: `undefined` should only be used for variables that have not been assigned.

```suggestion
        options: [{ [Prop.URL]: props.dataset ? props.dataset.mediaContent : null }],
```

client/src/project/datasets/change/DatasetChange.container.js
```diff
@@ -220,7 +230,17 @@ function ChangeDataset(props) {
       .then(response => {
         if (response.data.error !== undefined) {
           handlers.setSubmitLoader({ value: false, text: "" });
-          handlers.setServerErrors(response.data.error.reason);
+          if (response.data.error.errorOnFileAdd) {
+            props.edit ?
+              handlers.setServerErrors(
+                <div>
+                  The dataset was edited but there was an error adding uploaded files to it, please try again.
```
ciyer, line 237: Is this an accurate explanation of the error situation?

```suggestion
                  Any metadata changes were successfully applied, but some uploaded files could not be added to the dataset. Please try adding files again.
```

ciyer (changes requested):
The interaction works well and provides good feedback about the error situation. I have a few small suggestions for improvements to the code.

ciyer (approved):
👍

PR #1489, opened 2021-09-03
feat: show error page when gitlab is not available

### LINKED ISSUE #805: Gracefully fail on /user API not available
## Motivation
The GitLab update on `dev` made all the dev environments temporarily unavailable. Now that #782 is in place, it should be easy to display an error message when the `/user` API fails.

## Proposal
Set a variable in the `user` object in the redux store when the `/user` API fails (e.g. `user.error` with the content of the error message, to be reset every time the API is invoked again).
Display a nice error page in these cases.

## Context
Currently, the user gets stuck in the "Checking user data" temporary page.
Screenshot_20200304_114510

lorenzo-cavazzi:
This adds a very basic warning page when GitLab is not available.
We could try to fetch the data from StatusPage and verify if a maintenance is ongoing, and provide a better feedback to the user.

**Hot to test**: the only way is to trigger an error in the `GET /user` API.
Running telepresence to modify the `client.getUser` function in "client/src/api-client/user.js" is the easiest option.

/deploy renku=ui-design-2021
fix #805

Screenshot_20210903_174229

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -42,6 +42,7 @@ import { Cookie, Privacy } from "./privacy";
 import { NotificationsManager, NotificationsPage } from "./notifications";
 import { StyleGuide } from "./styleguide";
 import { Url } from "./utils/url";
+import { Unavailable } from "./Maintenance";
 
 import "./App.css";
 import "react-toastify/dist/ReactToastify.css";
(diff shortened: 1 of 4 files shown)


ciyer (changes requested):
One minor comment to the code. I also have an issue with testing, but I will check with you on that.

lorenzo-cavazzi:
Oh I messed up a line of code before pushing.
I fixed that, and made the code a bit more clear

ciyer (approved):
👍

PR #1491, opened 2021-09-07
feat: open in new tab in sessions is now more visible

### LINKED ISSUE #1400: Session "Open in new Tab" should be more visible
## Motivation

The "Open in New Tab" action is somewhat hidden. It should be more visible.

image

## Proposal

Swap the locations of the "Open in New Tab" and the "Stop" buttons. That is, make "Open in New Tab" the visible button and "Stop" a menu item.

vfried:
Things look like this now
image

I feel like there is a lot of information there and a lot has bold/underline information, i could try and clean up a bit this bar to make the actions stand out a bit more if desired.

I don't really like the icon we are using on links and i also feel it should go at the end of the text instead of at the beginning.

Closes #1400

### DIFF
--- a/client/src/notebooks/Notebooks.present.js
+++ b/client/src/notebooks/Notebooks.present.js
@@ -145,31 +145,41 @@ function SessionInformation(props) {
   };
   const resourceList = formattedResourceList(resources);
 
+  // Create dropdown menu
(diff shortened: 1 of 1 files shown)


ciyer (approved):
I like the change! I prefer having all the the actions as a button -- I think it makes the interaction clearer -- so I made that modification.

image

PR #1492, opened 2021-09-08
fix(sessions): improvements to messagging when no image is available

### LINKED ISSUE #1448: No image available
## Description

The warning about no image being available seems to persist even when the image is available. That should be refreshed.

image (1)

Users also pointed out that hiding the docker image availability info makes it difficult to understand what the user should do when the image is not available: it's not clear it's being build and waiting would be enough.

vfried:
When an image is not available but then becomes available the "no image available" sign fades.
image

I also added a message (because of the comment linked to the issue) to tell users the image is not available.
The message has a link to expand info on top (When it's not expanded).
image

image

Closes #1448

/deploy renku=ui-design-2021

### DIFF
--- a/client/src/notebooks/Notebooks.present.js
+++ b/client/src/notebooks/Notebooks.present.js
@@ -145,19 +145,6 @@ function SessionInformation(props) {
   };
   const resourceList = formattedResourceList(resources);
 
-  // Create dropdown menu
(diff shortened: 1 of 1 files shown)


ciyer (approved):
I added a commit with some minor wording changes. The solution is clear and works well! 🎉

PR #1494, opened 2021-09-09
fix: fork rights for private projects reviewed

### LINKED ISSUE #1431: Guest members of a project should not be allowed to fork
The fork button/option should be either grayed out or it should show a proper error message when a user that has `Guest` or lower permissions to a private project tries to fork.

Currently the fork button works even if your permissions are so low (i.e. Guest) that you cannot access any of the code or stuff in the container registry from the original project. The fork does not succeed and a message that says `An error occurred while forking: Renku API error` appears when you try to fork.

I think it is better to just fully not allow the user to get into this situation at all or test for this case and provide a more meaningful message. The lowest permission one needs to fork and use a private repo in renku is `Reporter`.

So to replicate this issue, to do the following:
- make a private project with `Account A` called `account-a-private-project`
- while logged in with `Account A` invite `Account B` into `account-a-private-project` with permissions `Guest`
- log in with `Account B`, try to fork the `account-a-private-project` project you were just added to

vfried:
When the user doesn't have enough rights to fork a project the button is disabled

image

/deploy renku=ui-design-2021

Closes #1431


ciyer (approved):
This works for me! Maybe later we can consider adding an explanation as to why the fork button is disabled, but this is better than the status quo, where the user gets a useless error message.

PR #1495, opened 2021-09-09
Changes to support allow releasing 1.0.0 as default UI

### LINKED ISSUE #1469: Release UI 1.0.0
# Tasks

- [x] Banner for returning to old UI (see below)
- [x] UI changes page (see below)
- [ ] Make UI 0.11.x the canary and 1.0.0 the main UI

# Banner

Add a banner notifying the user that the UI has changed.

image

# Changes Page

Add a page, taken from the landing, explaining what has changed. Next to the button to return to the old UI, there should be a link to discourse so the user can provide feedback if they prefer the old UI.

image

ciyer:
# Screenshots

## Welcome Banner

image

## What's New

image

image

Fix #1469

/deploy renku=ui-1_0_0-release


client/src/help/Help.present.js
```diff
@@ -207,6 +213,52 @@ class HelpFeatures extends Component {
...
+function HelpChanges() {
+
+  function SwitchToOldVersion() {
+    const ninetyDays = 60 * 60 * 24 * 90;
+    document.cookie = `ui-0-11-x=never; max-age=${ninetyDays}; path=/`;
+  }
+
+  const oldUiUrl = `/?v=${new Date().getTime()}`;
+
+  return <Fragment>
```
vfried, line 225: I would put this text below also in WhatsNew1_0_0 because in my opinion, it's specific to that, maybe it can be split inside there in two parts, and keep this just as a wrapper to put what's new in versions, this could also be used in next releases if we have something important to publish.

vfried: Maybe we can just take out from here the ongoing improvements and put them in OngoingImprovements1_0_0 so we have a structure for future changes.

client/src/help/Help.present.js
```diff
@@ -207,6 +213,52 @@ class HelpFeatures extends Component {
...
+        <p>
+          We think you will prefer the new UI with the changes we have made. But, if there is an interaction that
+          you preferred in the previous version, you can still {" "}
+          <a href={oldUiUrl} className="text-rk-pink" onClick={SwitchToOldVersion}>
+            return to the 0.11 version
+          </a> of the UI for the time being.
+          If you do prefer the previous UI, though, {" "}
+          <ExternalLink url={discourseUrl} role="text" title="please let us know why" />,
+          so we can make the necessary improvements.
+        </p>
```
vfried, line 256: I would add here that the old UI will be available for a period of time since we don't want people to just not use the new UI and then just shut it down. 

And I will put here first to let us know and then that they can change to the old UI, since in my opinion is more relevant that they give us feedback rhather than that they know they can change to the old UI.

vfried (commented):
The code in general looks really good, I found two things to point out but you can also merge as it is now since they aren't so relevant.

pameladelgado:
It looks great! 🎉 

I have a couple of questions:
1. Is it normal that the welcome banner disappears after some seconds?
2. When switching to the old UI I get the same banner text as what we currently have, should the text maybe be modified too? 'This old UI will soon be completely replaced by the new UI (put link). If you still prefer the old UI please give us feedback.', or something along those lines
image
3. Should we put the 'Give us feedback' link in the welcome banner? would it be too difficult to draft a new post on discourse with a pre-loaded topic 'Re: new UI' and some commented guidelines on the description? (similar to our GH issue templates).

ciyer:
Thanks for the review!

> 1. Is it normal that the welcome banner disappears after some seconds?

That's how we set up the alert, but we can make the time the banner is visible longer, or leave it up, if you think that would be a nicer interaction.

> 2. When switching to the old UI I get the same banner text as what we currently have, should the text maybe be modified too? 'This old UI will soon be completely replaced by the new UI (put link). If you still prefer the old UI please give us feedback.', or something along those lines
>    image

Good point. I can make that change to the 0.11.x version.

> 3. Should we put the 'Give us feedback' link in the welcome banner? would it be too difficult to draft a new post on discourse with a pre-loaded topic 'Re: new UI' and some commented guidelines on the description? (similar to our GH issue templates).

This is also a good suggestion. I will look into doing it.

pameladelgado:
>That's how we set up the alert, but we can make the time the banner is visible longer, or leave it up, if you think that would be a nicer interaction.

I see. I would leave it up for the time being, or make it configurable so that it stays for a while (some days/weeks) just in case people want to give feedback or read about the new UI. But if it stays this way it's OK for me too,  your call.

ciyer:
I am changing to the old UI to display the following:

## Old UI

**Home Page**

image

**Logged-in Landing Page**
image

## New UI

**Help**

image

I made the banner stay for 10s longer on the logged-in landing page, as well.

rokroskar:
Maybe "switching" instead of "upgrading"?

vfried (approved):
👍  This looks really good and the code is also good!

.github/workflows/test-and-ci.yml
```diff
@@ -48,7 +48,7 @@ jobs:
           python-version: 3.7
       - name: Test helm chart
         env:
-          HELM_URL: https://storage.googleapis.com/kubernetes-helm
+          HELM_URL: https://get.helm.sh
```
rokroskar, line 51: ok - I made https://github.com/SwissDataScienceCenter/renku/issues/2271 because Helm v2 is used in other places as well and should just be dealt with at the same time everywhere

helm-chart/renku-ui/templates/deployment.yaml
```diff
@@ -1,121 +1,10 @@
...
-      tolerations:
-{{ toYaml . | indent 8 }}
-    {{- end }}
-      {{- if .Values.image.pullSecrets }}
-      imagePullSecrets:
-      {{- range .Values.image.pullSecrets }}
-        - name: {{ . }}
-      {{- end}}
-      {{- end }}
+{{- $fullName := include "ui.fullname" . -}}
```
rokroskar, line 1: I guess ideally we wouldn't need to have a separate file for the deployment template, since this is the default for the deployment and canary would just override the settings. But maybe this is simpler?

rokroskar (approved):

PR #1501, opened 2021-09-22
fix: restore columns auto width in markdown tables

### LINKED ISSUE #1496: Incorrect markdown rendering on renku
Markdown files (*.md) do not seem to be rendered properly when viewed from renkulab.io. You can see it for yourself here: https://renkulab.io/projects/artur.safin/DatalakesHydrodynamics/files/blob/example_run.md

Specifically 2 issues that I found:
1) For whatever reason, one of the columns in the table seems to have nearly zero width, forcing vertical character stacking, and distorting the table quite a bit as a result. 
2) Syntax highlighting in code snippets is missing.

This is what I see when I'm logged in:

image

For reference, the same file is rendered properly on gitlab: https://renkulab.io/gitlab/artur.safin/DatalakesHydrodynamics/blob/master/example_run.md

This is not a huge issue for me, given that I can always point to the gitlab version, but thought you might want to be aware.

Best,
Artur

lorenzo-cavazzi:
The addition of the `text-break` class on the `<RenkuMarkdown>` component ([Link](https://github.com/SwissDataScienceCenter/renku-ui/blob/713770f670468ff1675fb165c028a00050b1c4f0/client/src/utils/UIComponents.js#L476)) accidentally broke the way column width is computed in tables.
This fix restores the normal word break behavior on tables.

**How to test**
You can use the example in the reference issues. A better test would be creating on the fly a project where the README has a markdown table with 3-4 columns, where all of them contain just a (long-ish) word expect one that contains a long text.
Without the fix (i.e. on dev) the column with a single word should be very narrow, while this PR should display something more reasonable.
Here is an example from the reference issue

Screenshot_20210922_170612

Screenshot_20210922_170526

/deploy
fix #1496

### DIFF
--- a/client/src/App.css
+++ b/client/src/App.css
@@ -233,3 +233,7 @@ iframe.invisible {
   height: 0;
   width: 0;
 }
+
+.word-break-normal {
+  word-break: normal !important;
+}
--- a/client/src/utils/HelperFunctions.js
+++ b/client/src/utils/HelperFunctions.js
@@ -157,7 +157,7 @@ function sanitizedHTMLFromMarkdown(markdown, singleLine = false) {
     emoji: true
   };
   const showdownClasses = {
-    table: "table"
+    table: "table word-break-normal"
   };
   // Reference: https://github.com/showdownjs/showdown/wiki/Add-default-classes-for-each-HTML-element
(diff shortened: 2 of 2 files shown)


ciyer (approved):
Nice fix! CSS FTW 🎉

PR #1502, opened 2021-09-23
feat: message when user creates or edits dataset in protected branch

### LINKED ISSUE #1013: Handle migration on protected branches
When a dataset is created on a project with a protected branch, a new merge request is created in the backend.

We should display a message that describes this situation to the user and put a button to redirect the user to this merge request.

vfried:
When a user tries to add or edit a dataset in a project with a protected branch the following message is displayed:

image

Backend bugs found in the process:
@SwissDataScienceCenter/renku-python-maintainers 
- When calling cache.projects_list I don't get ref on the response, this makes me have to clone a project every time the same URL is in the list twice because I don't know which is the project_id that i need to use 
- I didn't get the branch back when importing a dataset ("dataset.add") --> I should probably get this in the job!

Can't close #1013 because if the second bug

### DIFF
--- a/client/src/project/datasets/change/DatasetChange.container.js
+++ b/client/src/project/datasets/change/DatasetChange.container.js
@@ -32,6 +32,7 @@ import { ImageFieldPropertyName as Prop } from "../../../utils/formgenerator/fie
 import FormGenerator from "../../../utils/formgenerator/";
 import { mapDataset } from "../../../dataset/index";
 import _ from "lodash";
+import { Button } from "reactstrap/lib";
 
 let dsFormSchema = _.cloneDeep(datasetFormSchema);
 
@@ -107,6 +108,11 @@ function ChangeDataset(props) {
   const onCancel = (e, handlers) => {
     handlers.removeDraft();
     props.history.push({ pathname: `/projects/${props.projectPathWithNamespace}/datasets` });
+  };
+
+  const goToCollaboration = (handlers) => {
+    handlers.removeDraft();
+    props.history.push({ pathname: `/projects/${props.projectPathWithNamespace}/collaboration/mergerequests` });
   };
 
   function setNewJobStatus(localJob, remoteJobsList) {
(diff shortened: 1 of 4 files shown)


client/src/project/datasets/change/DatasetChange.container.js
```diff
@@ -243,6 +249,19 @@ function ChangeDataset(props) {
           }
           else { handlers.setServerErrors(response.data.error.reason); }
         }
+        if (response.data.result.remote_branch) {
```
ciyer, line 252: This line needs to check that the remote_branch !== the default branch. In that case, no merge request needs to be created.

client/src/project/datasets/change/DatasetChange.container.js
```diff
@@ -243,6 +249,19 @@ function ChangeDataset(props) {
...
+          handlers.setServerWarnings(
+            <div>
+              <strong>The branch you want to merge this changes to is protected.</strong>
+              <br/><br/>
+              Review and merge the following branch: <strong>{response.data.result.remote_branch}</strong>
+              {" "}to see the changes in your project.
+              <br/><br/>
+              You can do this in the {" "}
+              <Button color="warning" size="sm" onClick={()=>goToCollaboration(handlers)}>collaboration tab</Button>
```
ciyer, line 261:
```suggestion
              <strong>This project requires use of merge requests to make changes.</strong>
              <br/><br/>
              Create a merge request to bring the changes from <strong>{response.data.result.remote_branch}</strong> {" "}
              into <strong>{deaultBranch}</strong> to see the dataset {" "}
              in your project.
              <br/><br/>
              You can do this in the {" "}
              <Button color="warning" size="sm" onClick={()=>goToCollaboration(handlers)}>collaboration / MR tab</Button>
```

ciyer (changes requested):
This will be nice to have, as it will make it easier to support git-flow-like workflows in Renku.

I have some questions about the behavior. 

1. I created a project. By default, master is protected, but as the owner, I can push to this branch. In this case, there is no need for a merge request to be created. Still, I get this warning:

image

2. On the above, the "Creating dataset, please wait" never stops spinning (but the dataset does get created)

client/src/project/datasets/change/DatasetChange.container.js
vfried: You are right, i thought i only got the remote_branch in the case the core service created a new branch, i found a place where the same thing happens and fixed this as well.

vfried:
All changes where addressed, should be good to go :)

PR #1511, opened 2021-10-04
charts: update ingress apiVersion

pameladelgado:
/deploy

### DIFF
--- a/helm-chart/renku-ui/templates/ingress.yaml
+++ b/helm-chart/renku-ui/templates/ingress.yaml
@@ -33,8 +33,10 @@ spec:
         paths:
           - path: {{ $ingressPath }}
             backend:
-              serviceName: {{ $fullName }}
-              servicePort: http
+              service:
+                name: {{ $fullName }}
+                port:
+                  number: 80
   {{- end }}
 {{- end }}
 ---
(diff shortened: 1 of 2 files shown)


lorenzo-cavazzi (approved):

helm-chart/renku-ui/templates/ingress.yaml
```diff
@@ -32,15 +32,18 @@ spec:
...
         paths:
           - path: {{ $ingressPath }}
+            pathType: Prefix
             backend:
-              serviceName: {{ $fullName }}
-              servicePort: http
+              service:
+                name: {{ $fullName }}
+                port:
+                  number: 80
```
aledegano, line 40: I think this line should be:
`name: http`
to preserve the same behavior than before (e.g. `servicePort: http`)

pameladelgado: the end behavior, if we specify the port number instead of the name, is the same and imo if we use port number we have consistency across our charts, which makes them easier to maintain.

ciyer (approved):
Thanks @pameladelgado!

lorenzo-cavazzi (approved):

PR #1518, opened 2021-10-12
Support Cloud Storage 

### LINKED ISSUE #1499: UI for mounting S3 buckets
## Motivation

Support https://sdsc.atlassian.net/wiki/spaces/RENKU/pages/1536163850/2021-04-01+S3+Buckets+in+Renku+Design+Proposal

## Mock-ups / Design

image

image
[](url)
## Implementation

The S3 configuration will probably be re-used, so we need to make it quick and convenient to re-enter data. There are a few options for this, ordered here from easiest to most work to implement:
1. Use autofill properties to ensure that the browser can store and autofill information
2. ~~Store S3 configurations in the browser using local storage. We also need a solution for picking a stored configuration then.~~
3. ~~Store S3 configuration on the server using per-user storage.~~
(2, 3 to be handled later)

We will do 1, and the input for the fields will be in a model dialog. The "Save" button on the modal will test and not return to the main UI unless testing was successful. On a failure, the error message is shown and the user can fix or cancel.

### Not displayed details
- A non-tested vs. tested S3 mount should be indicated (maybe green background for tested?)
- We should indicate the S3 mounts in the session list

## Questions

- If an S3 bucket has been configured, should we block starting until it has been tested?
- What happens if the S3 bucket fails to mount? Does the session start anyway? Can we show helpful feedback/error messages in this case (e.g., notification that S3 bucket could not mount, an indication in the session list that an S3 bucket was requested but not mounted)?

## Progress

- [ ] API Design (sprint-??)
- [ ] UI Implementation

ciyer:
Add support to the UI for mounting cloud storage buckets.

The interaction with the backend still needs to be sorted out. Error handling/information is non-existent at the moment. See https://github.com/SwissDataScienceCenter/renku-notebooks/pull/729.

# TODO

- [x] use server_options to decide if the cloud storage is available or not

# Screenshots

image

image

image

image

image

# Testing

Things that should be tested
- Enter just an endpoint or just a bucket name, that should not be allowed
- Trying to use the same bucket name is used multiple times should be prevented

# Follow-up

* Validate cloud storage credentials before start: Currently, if the credentials are wrong, the session just does not come up and no explanation is given why. We should validate the credentials in the ui-server before start, so we can prevent this situation from arising.

Fix #1499 

/deploy renku-notebooks=cloud-storage-naming-tweaks-for-testing #persist


olevski:
@ciyer the only change in the API that affects you is the following:

The schema for the request to launch a notebooks went from:

```
{
  "branch": "master",
  "commit_sha": "string",
  "image": null,
  "namespace": "string",
  "notebook": null,
  "project": "string",
  "datasets": [],
  "serverOptions": {
    "cpu_request": 0.5,
    "defaultUrl": "/lab",
    "disk_request": "1G",
    "gpu_request": 0,
    "lfs_auto_fetch": false,
    "mem_request": "1G"
  }
}
```

to 

```
{
  "branch": "master",
  "commit_sha": "string",
  "image": null,
  "namespace": "string",
  "notebook": null,
  "project": "string",
  "s3mounts": [],  <<<<< only change
  "serverOptions": {
    "cpu_request": 0.5,
    "defaultUrl": "/lab",
    "disk_request": "1G",
    "gpu_request": 0,
    "lfs_auto_fetch": false,
    "mem_request": "1G"
  }
}
```

And correspondingly the response that lists active sessions went from:

```
{
  "servers": {
    "additionalProp1": {
      "annotations": {
        "jupyter.org/servername": "string",
        ...
      },
      "image": "string",
      "name": "string",
      "resources": {
        "cpu": "string",
       ...
      },
      "datasets": [], 
      "started": "2022-01-21T13:16:29.404Z",
      "state": {},
      "status": {},
      "url": "string"
    },
}
```

to 

```
{
  "servers": {
    "additionalProp1": {
      "annotations": {
        "jupyter.org/servername": "string",
        ...
      },
      "image": "string",
      "name": "string",
      "resources": {
        "cpu": "string",
       ...
      },
      "s3mounts": [],    <<<<< only change
      "started": "2022-01-21T13:16:29.404Z",
      "state": {},
      "status": {},
      "url": "string"
    },
}
```

olevski:
In summary the only thing that changed is that `datasets` was renamed to `s3mounts`. And I am pretty sure you are not even using the `datasets` field in the response that lists active servers.

client/src/notebooks/Notebooks.state.js
```diff
@@ -873,7 +878,8 @@ class NotebooksCoordinator {
   // * Change notebook status * //
   startServer() {
     const options = {
-      serverOptions: this.model.get("filters.options")
+      serverOptions: this.model.get("filters.options"),
+      datasets: this.model.get("filters.objectStoresConfiguration")
```
olevski, line 882:
```suggestion
      s3mounts: this.model.get("filters.objectStoresConfiguration")
```

olevski: By my accounting this is the only thing that needs to change to work with the changes in the notebooks api.

olevski:
I am sorry for spamming. The other thing that changed is that now the s3 mounting feature can be turned on or off fully. And when the options is turned off buckets cannot be mounted at all. So the UI should not show the form for adding s3 buckets if the feature is off in the notebook service.

To find out if this feature is on/off you can hit the `server_options` endpoint. You will now get a new part in it like this:
```
{
  "cpu_request": {
    "default": 0.5,
    "displayName": "Number of CPUs",
    "options": [
      0.5,
      1,
      2
    ],
    "order": 2,
    "type": "enum"
  },
  ...
  },
  "s3Mounts": {
    "enabled": false
  }
}
```

olevski (dismissed):
I tested out a few cases on the CI deployment, everything was fine. So as far as I am concerned this is good to go.

olevski:
Btw after talking to Andrea it just occured to me that we should mention somewhere in the UI that:
1. The buckets that you specify will be mounted under `/cloudstorage/<bucket_name>`
2. If you mount multiple buckets their names have to be unique - among all you buckets your specify (regardless of provider) the bucket names have to be unique. This is simply because you cannot have two mountpoint/folders with the same name under `/cloudstorage` where all the buckets are mounted

And one more thing that Andrea rightfully brought up. The error messages from the notebook service do not trickle down to the UI. This is not a new thing, or something that has been introduced with this PR. I think the best way forward here is to merge this PR - the S3 feature is off by default so we can decide to keep it off until we roll out proper error messaging.

Here is the PR that will standardize error messages in the notebook service: https://github.com/SwissDataScienceCenter/renku-notebooks/pull/849. I just need to do a few changes/cleanup on it and we can merge this on the notebook side. But the API that the ui will use will not change at all with these edits that I mention. So if you want you can start to adapt the ui to this new format of error messages from the notebook service.

ciyer:
> Btw after talking to Andrea it just occured to me that we should mention somewhere in the UI that:
> 
> 1. The buckets that you specify will be mounted under `/cloudstorage/<bucket_name>`
> 2. If you mount multiple buckets their names have to be unique - among all buckets you specify (regardless of provider) the bucket names have to be unique. This is simply because you cannot have two mountpoint/folders with the same name under `/cloudstorage` where all the buckets are mounted

Those are good suggestions, I will add them to the UI.

> And one more thing that Andrea rightfully brought up. The error messages from the notebook service do not trickle down to the UI. This is not a new thing, or something that has been introduced with this PR. I think the best way forward here is to merge this PR - the S3 feature is off by default so we can decide to keep it off until we roll out proper error messaging.

That makes sense. The other part of the solution is mentioned in the main text of the description above: we want to validate the credentials in the ui-server when the user clicks "Save." Then we can immediately provide feedback that the user can fix. But this PR is already big enough, and that feature can be added in later.

olevski (approved):
@ciyer I only tested out launching a session and looked at how my previous two comments were addressed. So all of that is fine.

I did not review the code at all. Since I am not familiar with the code that all I will leave that to someone else from the UI team.

andre-code (approved):
Great feature 🎉👏
My only minor suggestion would be to rename the session test file to newSession.spec.ts since there is another PR that is adding a session file to display the session list.

🚀

ciyer:
> Great feature 🎉👏 My only minor suggestion would be to rename the session test file to newSession.spec.ts since there is another PR that is adding a session file to display the session list.
> 
> 🚀

Sure, good idea -- maybe that rename can be in the other PR though. :)

PR #1533, opened 2021-10-26
refactor: derive name on dataset files when missing

### LINKED ISSUE #1497: GET /knowledge-graph/datasets/:id won't provide name for the hasPart items
## Motivation

The new Dataset metadata does not contain information about the part `name`. Thus, the `GET /knowledge-graph/datasets/:id` won't be able to provide it anymore.

## Proposal

The UI may use part's `atLocation` as a name if needed.

lorenzo-cavazzi:
This PR is an attempt to fix the broken links we would have once we won't get the `name` property for the files from the KG datasets API. We can merge it now since it still uses the `name` property when available, and it derives it from `atLocation` otherwise.

In the end I followed the suggestion in #1497 , although a cleaner solution may be desirable.
We should consider refactoring the `FileExplorer` component that has grown way too much, handling generic file trees, lineage, dataset files [...] despite the different object properties. An abstraction layer would greatly simplify handling minor changes like the current one.

/deploy
fix #1497

### DIFF
--- a/client/src/dataset/Dataset.present.js
+++ b/client/src/dataset/Dataset.present.js
@@ -50,6 +50,25 @@ function DisplayFiles(props) {
     ( props.files[0].atLocation.startsWith("data/") ? 2 : 1 )
     : 0;
 
+  // ? This re-adds the name property on the datasets.
+  // TODO: consider refactoring FileExplorer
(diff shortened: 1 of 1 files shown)


ciyer (approved):
👍

PR #1537, opened 2021-10-28
feat: modify ui-server for custom CA certificates

olevski:
Allows the ui server to properly access another service (i.e. namely gitlab and/or keycloak) if that service is using self-signed certificates.

I followed the Gitlab cloud native helm chart and how they tackle this issue when addressing this. See for examples:

https://gitlab.com/gitlab-org/charts/gitlab/-/blob/master/templates/_certificates.tpl
https://gitlab.com/gitlab-org/charts/gitlab/-/blob/master/charts/gitlab/charts/webservice/templates/deployment.yaml#L102

Fixes SwissDataScienceCenter/renku#2328

This is only required for the ui server. It makes sure that when the server is fully up and running it will not have any issues accessing other services (over https) if those services are using self-signed certificates... and if those certificates are injected through this code.

/deploy


olevski:
The dependent PRs that were blocking this PR have been merged. This is ready for review and if approved for merging.

ciyer (approved):
👍 Great to have this support in the UI.

PR #1541, opened 2021-11-01
feat: filter out the internal Loud component from the list

ciyer:
# Screenshots

**The message appears as expected**

image

**The component listing does not show "Loud"**

image

/deploy


lorenzo-cavazzi (approved):
This works well :+1: 

Screenshot_20211101_142818

PR #1544, opened 2021-11-02
Simplify refreshing the UI after publishing new RenkuLab versions

lorenzo-cavazzi:
This PR solves the various problems around users accidentally working on outdated UI versions after a new release.
Please check the following ~annoying~ verbose explanations for the details.

**UI client build**
We use `react-scripts` to automate a few operations around starting the development server and compiling/bundling the code.
One limitation is the opinionated use of webpack, that makes adding customization difficult.
I used [this library named craco](https://github.com/gsoft-inc/craco) as a drop-in replacement that still invokes `react-script` (it's a drop-in replacement) but also allows to easily changes webpack config files.
The production-ready bundled JS files now include in the filename the last commit short sha, invalidating any cached file on new releases.

**UI versions**
The UI client gets the last short-sha from the `params` at load time.
It invokes an API on the UI server every 5 minutes checking the current version deployed in the backend (I slightly cheated using the UI server sha. It's always aligned with the UI client).
When a change is detected, we show a banner suggesting users to reload (see screens at the end).

**Extra**
This PR includes a small bugfix on the home page dropdown menu: on large screens, expanding it temporarily moved the RenkuLab main title to the bottom, suddenly bouncing back to the original position once the menu was completely open.

**How to test**
It's tricky since faking backend component changes is not easy.
A complete test requires deploying the UI images made from this PR to a dev namespace, loading the UI on any page, then pushing a new image made from a different commit (e.g. execute `git commit --amend` to modify the commit text and get a different sha). The UI should show the `new version` alert.

**EXAMPLE**
1. UI checking the version -- no changes
    i01

2. Change detected (see the `200` response on the Network tab on the right) -- showing the banner
    i02

3. After clicking on the `reload` bottom, the UI gets refreshed. Notice the bundled files on the network tab (elements 2 to 5) are fetched again and not cached since the name changed (code `200` instead of `304`)
    i03

/deploy
#fix 819

### DIFF
--- a/client/Dockerfile
+++ b/client/Dockerfile
@@ -8,8 +8,11 @@ WORKDIR /app
 COPY package.json package-lock.json /app/
 COPY public /app/public
 COPY src /app/src/
+COPY craco.config.js /app/craco.config.js
 
 ENV NODE_OPTIONS="--max-old-space-size=4096"
+ARG SHORT_SHA
+ENV RENKU_UI_SHORT_SHA=$SHORT_SHA
 
 # Delete this after reactstrap is released
 RUN apk update && apk upgrade && \
@@ -36,6 +39,9 @@ USER nginx
 
 
 HEALTHCHECK --interval=20s --timeout=10s --retries=5 CMD test -e /var/run/nginx.pid
+
+ARG SHORT_SHA
+ENV RENKU_UI_SHORT_SHA=$SHORT_SHA
 
 ENTRYPOINT ["/bin/sh", "/app/docker-entrypoint.sh"]
 CMD ["nginx", "-g", "daemon off;"]
--- a/client/src/api-client/index.js
+++ b/client/src/api-client/index.js
@@ -16,27 +16,27 @@
  * limitations under the License.
  */
 
-import { APIError, alertAPIErrors, API_ERRORS } from "./errors";
-import { renkuFetch, RETURN_TYPES } from "./utils";
-import processPaginationHeaders from "./pagination";
-import ApolloClient from "apollo-boost";
-// ? Consider a minimal graphql lib: https://github.com/yoshuawuyts/nanographql
+import ApolloClient from "apollo-boost"; // ? Consider a minimal graphql lib: https://github.com/yoshuawuyts/nanographql
 
+import addDatasetMethods from "./dataset";
+import addEnvironmentMethods from "./environment";
+import addGraphMethods from "./graph";
+import addInstanceMethods from "./instance";
+import addIssueMethods from "./issue";
+import addJobMethods from "./job";
+import addMergeRequestMethods from "./merge-request";
+import addMigrationMethods from "./migration";
+import addNotebookServersMethods from "./notebook-servers";
+import processPaginationHeaders from "./pagination";
+import addPipelineMethods from "./pipeline";
 import addProjectMethods from "./project";
 import addRepositoryMethods from "./repository";
-import addUserMethods from "./user";
-import addIssueMethods from "./issue";
-import addInstanceMethods from "./instance";
-import addNotebookServersMethods from "./notebook-servers";
-import addGraphMethods from "./graph";
-import addPipelineMethods from "./pipeline";
-import addDatasetMethods from "./dataset";
-import addMergeRequestMethods from "./merge-request";
 import addTemplatesMethods from "./templates";
-import addJobMethods from "./job";
-import addMigrationMethods from "./migration";
-
 import testClient from "./test-client";
+
+import addUserMethods from "./user";
+import { APIError, alertAPIErrors, API_ERRORS } from "./errors";
+import { renkuFetch, RETURN_TYPES } from "./utils";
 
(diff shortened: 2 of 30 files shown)


client/src/api-client/index.js
```diff
@@ -74,19 +69,20 @@ class APIClient {
       headers: { "X-Requested-With": "XMLHttpRequest" }
     });
 
-    addProjectMethods(this);
-    addRepositoryMethods(this);
-    addUserMethods(this);
-    addIssueMethods(this);
+    addDatasetMethods(this);
```
ciyer, line 72: Good idea to make this alphabetical!

client/src/model/GlobalSchema.js
```diff
@@ -25,19 +25,20 @@
 
 import { Schema, PropertyName as Prop } from "./index";
 import {
-  notebooksSchema, userSchema, projectsSchema, projectGlobalSchema, newProjectSchema,
-  statuspageSchema, notificationsSchema, formGeneratorSchema
+  environmentSchema, formGeneratorSchema, newProjectSchema, notebooksSchema, notificationsSchema,
```
ciyer, line 28: Again, good idea to make these things alphabetical!

server/src/routes/index.ts
```diff
@@ -30,7 +31,7 @@ function register(app: express.Application, prefix: string, authenticator: Authe
     res.send("Hello ingress!");
   });
 
-  registerApiRoutes(app, prefix, authenticator);
+  registerApiRoutes(app, prefix + config.api.prefix, authenticator);
```
ciyer, line 34: I think it makes sense to add `api` to the prefix before calling `registerApiRoutes`, but is there an advantage to making the path for the API routes a configuration? Will be also make this configurable in the UI client?

ciyer (commented):
The code looks really good. I like the use of CRACO to get more control over the toolchain. I have one question behind making the path to the API routes configurable, but otherwise everything looks good.

I am waiting on the environment to allow testing the feature still.

server/src/routes/index.ts
lorenzo-cavazzi: I thought someone may prefer a different URL structure, e.g. using something different than `/ui-server`. With the `config.api.prefix` config, it's even possible to route all the APIs to something like `api.customrenku.io` for the backend instead of `customrenku.io/<something>`. Not sure if it's very useful.

I kept the distinction between the server prefix and the API prefix because I initially thought the internal APIs may go under a different prefix (e.g. the /version may go under /ui-server/version instead of /ui-server/api/version). On a second thought, the client should not know about this implementation detail.

ciyer: I think allowing the API to be deployed under a different domain could be a possibility, though, I'm not sure how nice it would be in practice, since it would require setting up a bunch of CORS stuff. 

But I don't think the paths to the endpoints should be configurable: the should be fixed and documented, not subject to change between deployments.

So what do you think about making this line and removing the API prefix configuration?

```suggestion
  registerApiRoutes(app, prefix + "/api", authenticator);
```

ciyer (commented):
I tested this on Safari, Chrome, and FF, and it works *very* well! 🎉! I think we just need to resolve the question of the routes that "api" endpoints and then this can be merged.

server/src/routes/index.ts
lorenzo-cavazzi: You are right, configurable paths don't make much sense in this case.
As briefly discussed offline, it may still be useful to keep paths in the config file for a few reasons (reusability, filtering in middlewares, ...).
I reorganized them in a coherent way -- all the routing prefixes are now together (we have only 2, api and auth, but we will have more)

ciyer (approved):
This looks great. It's a really nice improvement! 🎉

olevski (approved):
In terms of the changes to the helm chart components (which are almost non-existent), I do not see any issues. I did not review any of the javascript code though.

client/Dockerfile
```diff
@@ -37,5 +40,8 @@ USER nginx
 
 HEALTHCHECK --interval=20s --timeout=10s --retries=5 CMD test -e /var/run/nginx.pid
 
+ARG SHORT_SHA
+ENV RENKU_UI_SHORT_SHA=$SHORT_SHA
```
olevski, line 44: @lorenzo-cavazzi why are these two lines repeated twice in the dockerfile?

olevski: I dont think this is a problem. But I think you can remove one set and I will re-approve.

lorenzo-cavazzi: Without repeating these 2 lines, it failed. Apparently, `ARG` is scoped to per stage (it resets after every `FROM` command).
https://stackoverflow.com/questions/53681522/share-variable-in-multi-stage-dockerfile-arg-before-from-not-substituted

rokroskar: but you're not actually using those at the top....?

lorenzo-cavazzi: Yes, `npm install` is using the commit sha to create the name to bundle the files.

Issue #1529, opened 2021-10-21
Improve performance of landing page


ciyer:
# Description

The following changes should be made to the data retrieved for the landing page:

- Just get 5 projects for the landing page
- Just get project name/namespace for the quick-nav use (see just-names query below)

Follow-up to #1505.

# Details

**minimal query**
```
{
  projects(membership: true, first: 100) {
    pageInfo {
      startCursor
      endCursor
    }
    nodes {
      avatarUrl
      description
      name
      namespace {
        fullPath
      }
      path
      lastActivityAt
      tagList
    }
  }
}
```

**just names query**
```
{
  projects(membership: true, first: 100) {
    pageInfo {
      startCursor
      endCursor
    }
    nodes {
      name
      namespace {
        fullPath
      }
      path
    }
  }
}
```

PR #1546, opened 2021-11-04
feat: improve performance of landing page (#1529)

andre-code:
Speed up loading of landing page by fetching less information.

### DIFF
--- a/client/src/project/shared/Projects.state.js
+++ b/client/src/project/shared/Projects.state.js
@@ -82,6 +82,29 @@ class ProjectsCoordinator {
     });
   }
 
+  async getLanding() {
+    if (this.model.get("featured.fetchingLanding"))
+      return;
+    // set status to fetchingLanding, get the projects for the landing page
+    this.model.set("featured.fetchingLanding", true);
+    const params = { order_by: "last_activity_at", per_page: 5 };
+    const promiseMember = this.client.getProjects({ ...params, membership: true })
+      .then(resp => resp.data.map((project) => this._starredProjectMetadata(project)))
+      .catch(error => []);
+
+    // set `featured` content and return only landing projects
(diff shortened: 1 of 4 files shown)


client/src/project/shared/Projects.state.js
```diff
@@ -82,6 +82,29 @@ class ProjectsCoordinator {
     });
   }
 
+  async getLanding() {
+    if (this.model.get("featured.fetchingLanding"))
+      return;
+    // set status to fetchingLanding, get the projects for the landing page
+    this.model.set("featured.fetchingLanding", true);
+    const params = { order_by: "last_activity_at", per_page: 5 };
+    const promiseMember = this.client.getProjects({ ...params, membership: true })
```
lorenzo-cavazzi, line 91: Should we perhaps move `membership: true` to the above line, so we define all the params together?

client/src/project/shared/Projects.state.js
```diff
@@ -82,6 +82,29 @@ class ProjectsCoordinator {
...
+      .then(resp => resp.data.map((project) => this._starredProjectMetadata(project)))
+      .catch(error => []);
+
+    // set `featured` content and return only landing projects
+    return promiseMember.then(values => {
```
lorenzo-cavazzi, line 96: For readability, I would either move this `.then` to the above chain (having something like `getProjects().then().catch().then()`) or use the `await` pattern.

lorenzo-cavazzi (changes requested):
The code looks great!

I added a couple of tips for improving readability and keeping the changes to `RenkuModels` coherent with the other Schemas. Otherwise, all works well :+1:

lorenzo-cavazzi (approved):
The code looks great! Let's merge it as soon as the acceptance tests pass :heavy_check_mark:

PR #1548, opened 2021-11-05
refactor: Rearrange session menu options (#1545)

### LINKED ISSUE #1545: Reduce click confusion between session buttons
# Description

image

The buttons for "Open in Tab" and "Stop" are next to each other, and it is easy to accidentally click "Stop" instead of "Open in Tab"

# Solution

Move "Stop" to the last position; add a separator in the menu to put move distance between stop and the other buttons.

andre-code:
- Move the stop option to the end of the list to avoid click problems.

- Refactor the NotebookServerRowAction component to be a functional component and use memo to avoid rendering that component if the props have not changed and it is already rendered, the component was rendered for each session state request.

Fix #1545


ciyer (approved):
Nice fix! 🎉  I added a commit to group together the logic that adds the separater together with the stop button, which I think makes it easier to follow. You can squash these commits together when you merge.

Issue #1352, opened 2021-05-17
Prevent getting outdated result in dataset search


lorenzo-cavazzi:
When the user searches for a dataset, it may occasionally take a bit longer (few seconds) to get the results, and it's possible to change the search entry and click again on the search button before the result is shown.
The latest query result will be ignored, showing the outdated result.

We can either:
- Prevent a new search while loading the content (i.e. disable the button and the search input when `fetching`)
- Discard the old result and correctly trigger and accept the latest query
- _From @erbou suggestion_: prevent new search while loading but allow to click on a `cancel` button while that is ongoing

I would go for the first solution since it's much quicker to implement and it prevents spamming searches (this probably isn't a problem for users, but it may cause loops in the acceptance tests when the query is slow to respond).
EDIT: the 3rd solution seems a very good alternative, though it would be one of a kind in the dataset search instead of a pattern applied to all the searches.

# Design

Disable entering new search terms while a search is in progress, but add a stop button for the search.

![image](https://user-images.githubusercontent.com/1196411/140494506-92958370-bfd7-4429-8ccb-bbf9ddf806e6.png)

erbou:
Other options:

* The first option with a `cancel search` button: prevent new search until content content is returned or until search is cancelled.
* A combination of first option and second option: prevent the search until a timeout, after timeout allow a new query that discards old results and correctly triggers and accepts the latest query.

My preference as a user tends toward the `cancel search` option.

ciyer:
Implement solution outlined in #1505.

PR #1549, opened 2021-11-08
feat: Add cancel datasets search (#1352)

andre-code:
Added cancellation of dataset search when a search is already in progress, which prevents the input from being used for new searches when a new one is still running. The user can cancel the current one.
Included searchId to show only the results of the current search.

### DIFF
--- a/client/src/dataset/list/DatasetList.state.js
+++ b/client/src/dataset/list/DatasetList.state.js
@@ -147,7 +147,7 @@ class DatasetListModel extends StateModel {
       page: this.get("currentPage"),
     };
     return this.client.searchDatasets(searchParams)
-      .then( async response => {
+      .then( response => {
         const currentSearchId = this.get("searchId");
         if (currentSearchId !== searchId) return;
         this.manageResponse(response);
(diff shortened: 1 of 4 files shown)


client/src/dataset/list/DatasetList.state.js
```diff
@@ -142,7 +147,11 @@ class DatasetListModel extends StateModel {
       page: this.get("currentPage"),
     };
     return this.client.searchDatasets(searchParams)
-      .then(response => this.manageResponse(response))
+      .then( async response => {
```
ciyer, line 150: I think the `async` is unnecessary here.

ciyer (changes requested):
Looks great! 

I added a commit to move the grid/list display switch to the left hand side (it looked weird if there were not enough results, the button was just floating on the right-hand end of the viewport). 

image

Code looks good; I spotted an unnecessary `async` that can be removed, but otherwise it is great.

ciyer (approved):
Looks good! Once the tests finish, you can squash everything into one commit.

PR #1553, opened 2021-11-10
feat: Get minimal project data for featured projects

andre-code:
/deploy


client/src/model/RenkuModels.js
```diff
@@ -38,7 +38,7 @@ const projectsSchema = new Schema({
   featured: {
     [Prop.SCHEMA]: new Schema({
       fetched: { [Prop.INITIAL]: null, [Prop.MANDATORY]: true },
-      landing: { [Prop.INITIAL]: [], [Prop.MANDATORY]: true },
+      fetching: { [Prop.INITIAL]: false, [Prop.MANDATORY]: true },
```
ciyer, line 41: Good catch. That must have slipped in an earlier PR!

client/src/project/shared/Projects.state.js
```diff
@@ -59,14 +59,18 @@ class ProjectsCoordinator {
       return;
     // set status to fetching, get all the projects and filter and invoke both APIs
     this.model.set("featured.fetching", true);
-    const params = { order_by: "last_activity_at", per_page: 100 };
+    let params = { query: "last_activity_at", per_page: 100 };
```
ciyer, line 62: Is there a reason to switch from `const` to `let` for this variable?

ciyer (changes requested):
With the first change and now this, the UI starts much quicker! 🎉 This works well for me, and I tested both the quick nav and forking, and everything seems to work as before, even though we are retrieving less data.

I just have one question about a code change, but otherwise, once the acceptance tests pass, I think this is ready to merge.

client/src/project/shared/Projects.state.js
andre-code: thanks for noticing, it was a mistake, it is not necessary to change the type of variable.

ciyer (changes requested):
Code looks good!

But, when testing I came across a problem. If I use the quick nav to search for a project:

image

And then try to navigate to the project, I get this.

image

ciyer (approved):
That fixed it!

ciyer (approved):
🎉

PR #1566, opened 2021-11-15
feat: Add PDF file viewer (#998)

### LINKED ISSUE #998: PDFs in repo not displaying on file view
## Description

renkulab doesn't render image: https://renkulab.io/projects/e.jablonski/workflow-example/files/blob/output/EWR.pdf
but gitlab does: https://renkulab.io/gitlab/e.jablonski/workflow-example/blob/master/output/EWR.pdf

I'm not sure if this is similar to the path issue for trying to display images in, for instance, the README.

andre-code:
Support viewing pdf files using the react-pdf library

See example case:
- https://renku-ci-ui-1566.dev.renku.ch/projects/dandrea.cordoba/test/files/blob/sample.49a87e34.pdf
- https://renku-ci-ui-1566.dev.renku.ch/projects/dandrea.cordoba/test/files/blob/output_EWR.pdf

Reference library used: https://github.com/wojtekmaj/react-pdf

Closes #998
/deploy


lorenzo-cavazzi (dismissed):
Great feature! :tada: 
I added some changes to import the library without adding the minified file manually.

One thing I don't like is the rendering scaling.
It seems that PDFs are rendered at a pre-defined size, making the content either too small for the page (that's acceptable) or too big (that's worse, there is some overflow, see the right-side of the attached image).
GitLab somehow manages to re-render the PDF on resize, keeping the width aligned with the available width. That would be ideal.

Anyway, this PR is a step forward and there is no need to spend much more time on it. Let's merge it! :heavy_check_mark: 

Screenshot_20211117_155002

ciyer (approved):
I tested this, and it works very well! We have a limit (default 1MB) on files we preview, this will prevent very large PDFs from being automatically previewed, so I do not think we need to worry about pagination, etc. Renku isn't going to be used as a PDF reader -- just to view small PDFs or preview larger ones, and for that purpose, this works very well!

🎉

PR #1568, opened 2021-11-16
Hoist project state to global model

### LINKED ISSUE #1183: Refactor project state management

ciyer:
wip

Fix #1183

/deploy renku=448c938d4d4a4b263857fba8ff2d402f72ad9fba

### DIFF
--- a/client/src/notebooks/Notebooks.container.js
+++ b/client/src/notebooks/Notebooks.container.js
@@ -279,7 +279,12 @@ class StartNotebookServer extends Component {
       false;
     this.state = {
       autostartReady: false,
-      autostartTried: false
+      autostartTried: false,
+      first: true,
+      ignorePipeline: null,
+      launchError: null,
+      showAdvanced: false,
+      starting: false
     };
 
     this.handlers = {
@@ -288,16 +293,12 @@ class StartNotebookServer extends Component {
       reTriggerPipeline: this.reTriggerPipeline.bind(this),
       setBranch: this.selectBranch.bind(this),
       setCommit: this.selectCommit.bind(this),
-      toggleMergedBranches: this.toggleMergedBranches.bind(this),
+      setIgnorePipeline: this.setIgnorePipeline.bind(this),
       setDisplayedCommits: this.setDisplayedCommits.bind(this),
       setServerOption: this.setServerOptionFromEvent.bind(this),
-      startServer: this.startServer.bind(this)
-    };
-
-    this.state = {
-      first: true,
-      starting: false,
-      launchError: null
+      startServer: this.startServer.bind(this),
+      toggleMergedBranches: this.toggleMergedBranches.bind(this),
+      toggleShowAdvanced: this.toggleShowAdvanced.bind(this)
     };
   }
 
@@ -329,14 +330,21 @@ class StartNotebookServer extends Component {
     }
   }
 
+  toggleShowAdvanced() {
+    this.setState({ showAdvanced: !this.state.showAdvanced });
+  }
+
+  setIgnorePipeline(value) {
+    this.setState({ ignorePipeline: value });
+  }
+
   async refreshBranches() {
     if (this._isMounted) {
-      if (StatusHelper.isUpdating(this.props.branches))
+      if (StatusHelper.isUpdating(this.props.fetchingBranches))
         return;
       await this.props.refreshBranches();
       if (this._isMounted && this.state.first)
         this.selectBranch();
-
     }
   }
 
@@ -460,6 +468,7 @@ class StartNotebookServer extends Component {
                 errorMessage: `The session could not start because the image is still building.
                 Please wait for the build to finish, or start the session with the base image.`
               },
+              showAdvanced: true,
               starting: false
             });
           }
@@ -476,6 +485,7 @@ class StartNotebookServer extends Component {
                 errorMessage: `The session could not start because no image is available.
                 Please select a different commit or start the session with the base image.`
               },
+              showAdvanced: true,
               starting: false
             });
           }
@@ -614,6 +624,8 @@ class StartNotebookServer extends Component {
       justStarted={this.state.starting}
       autoStarting={this.autostart && !this.state.autostartTried}
       launchError={this.state.launchError}
+      ignorePipeline={this.state.ignorePipeline}
+      showAdvanced={this.state.showAdvanced}
     />;
   }
 }
--- a/client/src/notebooks/Notebooks.container.js
+++ b/client/src/notebooks/Notebooks.container.js
@@ -600,9 +600,9 @@ class StartNotebookServer extends Component {
     const augmentedState = {
       ...this.props.notebooks,
       data: {
-        fetched: this.props.project.commits.fetched,
-        fetching: this.props.project.commits.fetching,
-        commits: this.props.project.commits.list,
+        fetched: this.props.commits.fetched,
+        fetching: this.props.commits.fetching,
+        commits: this.props.commits.list,
         branches: this.props.branches,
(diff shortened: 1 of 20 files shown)


client/src/project/Project.state.js
```diff
@@ -95,126 +95,144 @@ class ProjectModel extends StateModel {
...
-  fetchGraphWebhook(client, user) {
-    if (!user) {
-      this.set("webhook.possible", false);
-      return;
-    }
-    const userIsOwner = this.get("core.owner.id") === user.data.id;
-    this.set("webhook.possible", userIsOwner);
-    if (userIsOwner)
-      this.fetchGraphWebhookStatus(client, this.get("core.id"));
+  setProjectData(data, statistics = false) {
```
lorenzo-cavazzi, line 98: this shouldn't be necessary anymore since all the relevant data are now in the main store on the `project.medatata` object

client/src/project/Project.state.js
```diff
@@ -95,126 +95,144 @@ class ProjectModel extends StateModel {
...
-        let progress;
-        if (resp.progress == null)
-          progress = GraphIndexingStatus.NO_PROGRESS;
-
-        if (resp.progress === 0 || resp.progress)
-          progress = resp.progress;
-
-        this.set("webhook.progress", progress);
-        return progress;
+  fetchProjectDatasetsFromKg(client) { //from KG
```
lorenzo-cavazzi, line 111: this seems to be a leftover since the same function has been moved to the newer coordinator

client/src/project/Project.state.js
```diff
@@ -95,126 +95,144 @@ class ProjectModel extends StateModel {
...
   }
 
-  fetchGraphWebhookStatus(client) {
-    this.set("webhook.created", false);
-    this.setUpdating({ webhook: { status: true } });
-    return client.checkGraphWebhook(this.get("core.id"))
-      .then((resp) => {
-        this.set("webhook.status", resp);
+
+  fetchProjectDatasets(client, forceReFetch) {
```
lorenzo-cavazzi, line 128: leftover, same as above (already moved to the newer coordinator)

client/src/project/Project.present.js
```diff
@@ -1108,15 +1101,20 @@ class ProjectStartNotebookServer extends Component {
...
+        fetchingBranches={branches.fetching}
         refreshBranches={fetchBranches}
         externalUrl={externalUrl}
         successUrl={notebookServersUrl}
         blockAnonymous={blockAnonymous}
+        notebooks={projectCoordinator.model.baseModel.get("notebooks")}
         notifications={notifications}
-        scope={{ namespace: this.props.core.namespace_path, project: this.props.core.project_path,
-          defaultBranch: this.props.core.default_branch }}
+        project={projectCoordinator.get()}
```
lorenzo-cavazzi, line 1113: This seems a bit excessive since the Notebook component is using only the commits data. I would change this to `commits={projectCoordinator.get("commits")}` and adjust the `StartNotebookServer` component on `propsToChildProps` on `this.props.project`.

P.S. I didn't check if the commits are already coming as a property like the branches. In that case it doesn't require the `projectCoordinator.get`

lorenzo-cavazzi (changes requested):
This PR is brilliant! It's taking me some time to review, I can't image the effort required to make all the changes :grin: 

I identified a couple of leftovers, i.e. properties still read from the old store (apart from the migrations data that will be moved in another PR). I left some comments inline.

client/src/project/Project.state.js
```diff
@@ -480,77 +519,71 @@ class ProjectCoordinator {
       };
     }
 
-    this.model.setObject({ metadata: metadata, statistics: statsObject, filters: filtersObject });
+    this.model.setObject({ metadata,
+      filters: filtersObject,
+      forkedFromProject,
+      statistics: statsObject,
+      _transition: data
```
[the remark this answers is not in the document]
ciyer: Yes, I added it as part of the transition process, but then forgot to remove it. Good catch!

client/src/notebooks/Notebooks.container.js
```diff
@@ -600,9 +600,9 @@ class StartNotebookServer extends Component {
     const augmentedState = {
       ...this.props.notebooks,
       data: {
-        fetched: this.props.project.commits.fetched,
-        fetching: this.props.project.commits.fetching,
-        commits: this.props.project.commits.list,
+        fetched: this.props.commits.fetched,
```
lorenzo-cavazzi, line 603: Great! Would you mind adding the `commits` to the JSDocs as a `@param`?
I think it's the only new parameter in this component

lorenzo-cavazzi (changes requested):
I have just a tiny request before merging. Otherwise, it looks ready to me!

lorenzo-cavazzi (approved):
:tada: :rocket:

PR #1578, opened 2021-11-18
feat: Add project overview suggestions (#1512)

### LINKED ISSUE #1512: Help users add datasets to new projects
## Motivation

It would be good to assist users in getting their projects started. One thing we can do is to make suggestions for new projects, such as recommend that the user may want to add a dataset or edit the README.

## Mock-ups / Design

When a project is loaded, the UI should check how many commits there are. If there are fewer than four commits, the UI should show an extra _Starting your new project_ section on the `Overview > General` tab.

image

Design notes:
- If there is more than one commit, consider checking if any of them touch the README. In that case, we do not need to show the "Readme" section of the "Starting your new project"

## Progress

- [ ] API Design (sprint-??)
- [ ] UI Implementation

andre-code:
Show suggestions in the project overview
* Show readme suggestion if README.md file has no more than 1 commit.
* Show suggestion to create a dataset if the project has no datasets.

Closes #1512 
/deploy


client/src/project/Project.present.js
```diff
@@ -263,6 +263,58 @@ function ProjectViewHeaderOverviewDescription({ settingsReadOnly, description, s
...
+        Zenodo</a> or
+      <a href="https://dataverse.harvard.edu/" className="mx-1" target="_blank" rel="noreferrer noopener">
+        Dataverse</a>.</p></li>;
+
+  return (
+    <InfoAlert timeout={0}>
+      <div className="mb-0" style={{ textAlign: "justify" }}>
+        <span><FontAwesomeIcon icon={faInfoCircle} /> </span>
+        <strong>Welcome</strong> to your new Renku project! There are no commits thus far in this project,
+        so here are some suggestions for getting started. <br/>
```
ciyer, line 308: The wording here is exactly as in the design, but after actually using the feature, I want to make a change:

```suggestion
        <strong>Welcome</strong> to your new Renku project! It looks like this project is just getting started, so here are some suggestions to help you. <br/>
```

client/src/project/Project.present.js
```diff
@@ -263,6 +263,58 @@ function ProjectViewHeaderOverviewDescription({ settingsReadOnly, description, s
...
+  if (cantCommitsReadme > 1 && hasDatasets) return "";
+
+  const gitlabIDEUrl = props.externalUrl !== "" && props.externalUrl.includes("/gitlab/") ?
+    props.externalUrl.replace("/gitlab/", "/gitlab/-/ide/project/") : null;
+  const addReadmeUrl = `${gitlabIDEUrl}/edit/${props.core.default_branch}/-/README.md`;
+
+  const suggestionReadme = cantCommitsReadme > 1 ? null
+    : <li><p style={{ fontSize: "14px" }}>
+      <a className="mx-1" href={addReadmeUrl} target="_blank" rel="noopener noreferrer">
+        <strong>Add README.md</strong>
```
ciyer, line 284: The wording here is exactly as in the design, but after actually using the feature, I want to make a change:

```suggestion
        <strong>Edit README.md</strong>
```

ciyer (changes requested):
This looks good, and I tried it out on a couple of projects.

I have one name change in the code to suggest, and, after actually using the feature, I want to make some changes to the wording of the user notifications.

client/src/project/Project.present.js
```diff
@@ -263,6 +263,67 @@ function ProjectViewHeaderOverviewDescription({ settingsReadOnly, description, s
...
+    : <li><p style={{ fontSize: "14px" }}>
+      <a className="mx-1" href={addReadmeUrl} target="_blank" rel="noopener noreferrer">
+        <strong className="suggestionTitle">Edit README.md</strong>
+      </a>
+      Use the README to explain your project to others, letting them understand what you want
+      to do and what you have already accomplished.
+    </p></li>;
+
+  const suggestionDataset = hasDatasets ? null
+    : <li><p style={{ fontSize: "14px" }}>
```
ciyer, line 300: See comment above about font size.

client/src/project/Project.present.js
```diff
@@ -263,6 +263,67 @@ function ProjectViewHeaderOverviewDescription({ settingsReadOnly, description, s
...
+      <Link className="mx-1" to={props.newDatasetUrl}>
+        <strong className="suggestionTitle">Add some datasets</strong>
+      </Link>
+      Datasets let you work with data in a structured way, facilitating easier sharing.
+      You can create a new dataset with data you already have, or importing one from another
+      Renku project or from a public data repository such as
+      <a href="https://zenodo.org/" className="mx-1" target="_blank" rel="noreferrer noopener">
```
ciyer, line 307: This is nice! It would be good to use the `<ExternalLink role="text">...` component for this.

client/src/project/Project.present.js
```diff
@@ -263,6 +263,67 @@ function ProjectViewHeaderOverviewDescription({ settingsReadOnly, description, s
...
+        Zenodo</a> or
+      <a href="https://dataverse.harvard.edu/" className="mx-1" target="_blank" rel="noreferrer noopener">
```
ciyer, line 309: Comment on using `ExternalLink` applies here as well.

ciyer (changes requested):
This is looking very good, and having tests is great! 🙌 

I have a few suggestions to help modularity and reuse existing components.

client/src/project/Project.present.js
```diff
@@ -263,6 +263,65 @@ function ProjectViewHeaderOverviewDescription({ settingsReadOnly, description, s
...
+  if (!commits.fetching)
+    props.projectCoordinator.fetchCommits();
+
+  const loadingCommits = !commits.fetched;
+  const loadingCommitsReadme = isRequestPending(props, "commitsReadme");
+  const loadingDatasets = typeof (props.core?.datasets) === "string";
+
+  if (loadingCommits || loadingCommitsReadme || loadingDatasets) return "";
+
+  const cantTotalCommits = props.model?.get("project.commits.list")?.length ?? 0;
```
ciyer, line 280:
```suggestion
  const countTotalCommits = props.model?.get("project.commits.list")?.length ?? 0;
```

client/src/project/Project.present.js
```diff
@@ -263,6 +263,65 @@ function ProjectViewHeaderOverviewDescription({ settingsReadOnly, description, s
...
+  const hasDatasets = props.core?.datasets?.length > 0;
+  const countCommitsReadme = props.projectCoordinator.get("data.readme.countCommits");
+
+  if (cantTotalCommits > 4 || (countCommitsReadme > 1 && hasDatasets)) return "";
```
ciyer, line 284:
```suggestion
  if (countTotalCommits > 4 || (countCommitsReadme > 1 && hasDatasets)) return "";
```

client/src/project/Project.present.js
```diff
@@ -263,6 +263,65 @@ function ProjectViewHeaderOverviewDescription({ settingsReadOnly, description, s
...
+
+  const gitlabIDEUrl = props.externalUrl !== "" && props.externalUrl.includes("/gitlab/") ?
+    props.externalUrl.replace("/gitlab/", "/gitlab/-/ide/project/") : null;
+  const addReadmeUrl = `${gitlabIDEUrl}/edit/${props.core.default_branch}/-/README.md`;
+
+  const suggestionReadme = countCommitsReadme > 1 ? null
```
ciyer, line 290: This should be > 0, I think.

```suggestion
  const suggestionReadme = countCommitsReadme > 0 ? null
```

ciyer (changes requested):
This is now very close to being ready!

While testing, I found two cases with problems with the suggestions box.

1. README has been modified, but I still get the suggestion to modify the README

image

(I think I found the code where this bug is caused and suggested a fix)

2. User does not have write access to project, but suggestions still appear

image

We should only show the suggestions if the user is the maintainer of the project. There are examples of testing this in `Project.js` if you look for `accessLevel >= ACCESS_LEVELS.MAINTAINER`

client/src/project/Project.present.js
andre-code: Thanks for the revision, it should be 1 because the automatic initial commit should not be included.
The reason the suggestion component was not updating after a new commit is because it was being calculated when the project was loaded, I have modified it to calculate the number of readme commits once the total amount of commits in the project changes.

ciyer: Yes, you are correct!

client/src/project/Project.present.js
```diff
@@ -263,6 +263,70 @@ function ProjectViewHeaderOverviewDescription({ settingsReadOnly, description, s
...
+  useEffect(() => {
+    async function getReadmeCommits () {
+      const request = await props.projectCoordinator.fetchReadmeCommits();
+      setCountCommitsReadme(request);
+    }
+    if (props.projectCoordinator?.get("metadata.id") !== null)
+      getReadmeCommits();
+  }, [countTotalCommits]); // eslint-disable-line
+
+  if (isLoadingData || countTotalCommits > 4 || (countCommitsReadme > 1 && hasDatasets)) return "";
```
ciyer, line 289: I think it would be better to return null here. There is no need to add an empty string to the DOM.

```suggestion
  if (isLoadingData || countTotalCommits > 4 || (countCommitsReadme > 1 && hasDatasets)) return null;
```

client/src/project/Project.present.js
```diff
@@ -506,11 +570,14 @@ function ProjectViewGeneral(props) {
       {forkedFrom.metadata.core.path_with_namespace || "no title"}
     </Link>;
   }
+  const projectSuggestActions = props.visibility.accessLevel < ACCESS_LEVELS.MAINTAINER ? ""
```
ciyer, line 573: Nice, this does fix the problem, and I do not see the suggestions when I am not maintainer. But, I think code-wise, it would be better to 
1. Make this check in the ProjectSuggestActions component
2. Return `null` if there is nothing to show (rather than an empty string).

ciyer (changes requested):
This is coming together!

The behavior on projects where I am not a maintainer is now correct: I do not see suggestions. 

But the behavior on projects where I have updated the readme still isn't correct. I think the logic to detect if the readme has been modified needs to be looked at again.

See https://renku-ci-ui-1578.dev.renku.ch/projects/cramakri/mathjax

I can help put together a test case if you are having difficulties recreating the issue.

client/src/project/Project.present.js
ksanao: We have been reviewing nudging users towards dataset creation and it turns out that this does NOT fit all Renku use cases, for example:

- Omero users cannot create Renku datasets (we do not support this yet for streaming data)
- There is no dataset in projects that generate data outputs without using input files (simulation etc.) - we do not yet support workflow output asa dataset

ciyer: Thanks for that feedback; what you are saying makes sense. 

I think that is support to our approach of _suggesting_, but not requiring, that users make datasets. Once it is clear that the project is underway (has more than 4 commits), we do not show the messages anymore, since we assume that users have made an informed decision about whether or not they want to do the things we suggest.

client/src/project/Project.present.js
andre-code: sure, updated to return null

client/src/project/Project.present.js
andre-code: agree, I moved the validation to the ProjectSuggestActions component

andre-code:
Added validation to show readme suggestion if readme commit is the initial commit when there is only one readme commit

andre-code:
✨  Included new styles: 


Screenshot 2021-12-02 at 15 39 14

Screenshot 2021-12-02 at 15 39 32


Screenshot 2021-12-02 at 15 39 58

ciyer (approved):
Feature works great! And tests pass! Let's merge it! 🎉

I know it took a bit of time to get here, but the feature is now robust and handles many situations well, so I think it was worth it.

PR #1579, opened 2021-11-19
Support template variables metadata

### LINKED ISSUE #1320: Support object variables from the template manifest
On the project creation page, we show the corresponding variables when the user picks a template. Currently, we accept any input. Once SwissDataScienceCenter/renku-python#2034 is addressed, we can support `<select>` and typed inputs. We should consider the default values too, when available.

We will need to update the code to the new response from the `GET /templates.read_manifest` endpoint -- basically, the `variables` field will be an array of objects. Check the renku-python issue for the details

lorenzo-cavazzi:
This PR implements the UI side to support template variables metadata, keeping backward compatibility with old template repositories.
The input elements bound to variables are now styled based on the underlying type, and enumerations are presented as dropdowns. Mind that there are differences between different browsers --  we can implement customization if needed.
The default value is pre-filled, unless a different value is otherwise defined (e.g. when sharing project creation links containing pre-filled data).

Screenshot_20211119_145555

**How to test**
This is a sample template repository where variables have metadata: https://dev.renku.ch/gitlab/lorenzo.cavazzi.tech/core-it-template-variable-test-project
If the reviewer wants to try some variations of that templates repo, the available metadata are: `description`, `default`, `type` and `enum` (only when "type" is also "enum").

**Further ideas**
* Should we adjust the components' sizes to occupy the whole width?
* Now that the page easily gets filled, it's easier to miss the pre-validation error we show close to the problematic field. Should we perhaps add a short error close to the "Create project" button as well? I myself missed the "Missing title" error a few times.

/deploy renku=1.0-next renku-graph=cli-v1 renku-core=2034-extended-template-variables
fix #1320

### DIFF
--- a/client/src/project/new/ProjectNew.present.js
+++ b/client/src/project/new/ProjectNew.present.js
@@ -33,7 +33,7 @@ import {
   UncontrolledPopover, PopoverHeader, PopoverBody, Row, Table, UncontrolledTooltip
 } from "reactstrap";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import { faInfoCircle, faLink, faQuestionCircle, faSyncAlt } from "@fortawesome/free-solid-svg-icons";
+import { faInfoCircle, faLink, faPaste, faQuestionCircle, faSyncAlt } from "@fortawesome/free-solid-svg-icons";
 import {
   ButtonWithMenu, Clipboard, ErrorAlert, ExternalLink, FieldGroup, Loader, WarnAlert
 } from "../../utils/UIComponents";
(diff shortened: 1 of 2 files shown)


ciyer (changes requested):
This is nice and works very well! I agree that there should be some text next to the submit button that explains validation errors. 

I think we should use the `fa-undo` icon to for the "set value to default" button.

image

lorenzo-cavazzi:
I added the global error message and switched to the undo icon as suggested in the review.

Here are a couple of examples:

Screenshot_20211123_170301

Screenshot_20211123_170336

Issue #1521, opened 2021-10-14
Route all the APIs through the ui-server


lorenzo-cavazzi:
Now that the ui-server handles authentication, we should route all the `/api` requests through `/ui-server /api`.
An example of routing with tokens is available in `server/src/routes/apis.ts`, using "http-proxy-middleware". Consider also alternatives like "node-http-proxy".

- [ ] route all the API through the ui-server
- [ ] handle token expirations in the ui-client (i.e. when the response contains a header `ui-server-auth:expired`, it should redirect to the login endpoint. Check also that `ui-server-auth` is returned properly).

PR #1583, opened 2021-11-23
feat(server): improve k8s probes to restart pod on long failures

lorenzo-cavazzi:
While working on #1521 ,  I made a couple of changes to improve the resiliency of the UI server. Specifically, temporary interruptions in the connectivity to the storage service (Redis) will cause the Kubernetes liveness probe to fail -- new UI server instances will be created when necessary in the attempt to re-establish the connection from scratch.

I made a separate PR because it would be nice to have this merged to a new release before the routing change.

re #1521 
/deploy renku=448c938d4d4a4b263857fba8ff2d402f72ad9fba


ciyer (approved):
This works nicely! I after deleting the Redis pod, the ui-server realized that it was unavailable, and after restarting the Redis pod, the ui-server came back up! 🎉

Issue #1572, opened 2021-11-17
Detect when gitlab is non-embeddable


ciyer:
## Motivation

If GitLab is non-embeddable, the isn't very helpful

![image](https://user-images.githubusercontent.com/1196411/142145418-20755316-48b6-4d59-9f5a-b34612e7cdec.png)

## Proposal

The UI client should ask the UI server if the gitlab for a project is embeddable. The UI server should make a request to the gitlab instance and inspect `Content Security Policy` header of the response, which will indicate if embedding will work or not.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-src

If not, the UI should show a message.

![image](https://user-images.githubusercontent.com/1196411/142146085-85404b54-bc3c-452e-9236-c2e67eaa4407.png)

PR #1584, opened 2021-11-24
feat: Validate URL before using it  in iframe (1572)

andre-code:
Added endpoint in serverui to validate if URL can be used in an iframe.
This validation is used in the Collaboration component (Issues and Merge requests).

/deploy


server/src/routes/apis.ts
```diff
@@ -49,6 +50,27 @@ function reigsterApiRoutes(app: express.Application, prefix: string, authenticat
     res.json(data);
   });
 
+  app.get(prefix + "/check-url/:url", async (req, res) => {
+    const data: CheckURLResponse = {
+      isIframeValid: false,
+      url: req?.params?.url,
+    };
```
lorenzo-cavazzi, line 57: I think returning just `true/false` would be enough. But you added just a little extra information, nothing against keeping it as you already did :slightly_smiling_face:

server/src/routes/apis.ts
```diff
@@ -49,6 +50,27 @@ function reigsterApiRoutes(app: express.Application, prefix: string, authenticat
...
+    try {
+      const externalUrl = new URL(req?.params?.url);
+      const requestExternalURL = await fetch(externalUrl.toString());
+      if (requestExternalURL.status >= 400)
+        data.error = "Bad response from server";
+      else if (requestExternalURL?.headers.has("X-Frame-Options"))
+        data.error = `X-Frame-Options: ${requestExternalURL?.headers.get("X-Frame-Options")}`;
```
lorenzo-cavazzi, line 64: I fear this is not the right way to test if the page is embeddable in an iframe. You should check the `content-security-policy` header as suggested in the reference issue #1572

lorenzo-cavazzi (changes requested):
This will be a very useful feature!

Unfortunately, I fear using the `X-Frame-Options` header won't work in our case. The reference issue #1572 suggests using `content-security-policy` instead, and that seems to be a more reliable way.
Giving a look at the various options, it may be a bit more complicated https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy

As an example, in the response from the API invoked in the PR deployment, I got the following data:

```
x-frame-options:'DENY'
content-security-policy:"frame-ancestors 'self' https://dev.renku.ch https://renku-ci-ui-1584.dev.renku.ch"
```

This means it's embeddable, even if the API response says otherwise.
You can verify it by running `PR=1579 ./run-telepresence.sh` from the master branch of renku-ui and accessing a collaboration page.

server/src/routes/apis.ts
```diff
@@ -49,6 +50,27 @@ function reigsterApiRoutes(app: express.Application, prefix: string, authenticat
     res.json(data);
   });
 
+  app.get(prefix + "/check-url/:url", async (req, res) => {
```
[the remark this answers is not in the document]
ciyer: My vote is for `allows-iframe`

server/src/routes/apis.ts
ciyer: I also think json response here is nice -- it will allow us to extend the information if necessary (not sure what, but maybe we want to pass the content of the CSP to the browser).

server/src/routes/apis.ts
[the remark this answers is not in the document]
andre-code: Great, it was updated to `allows-iframe`

server/src/routes/apis.ts
andre-code: sure, the logic has been updated to use the content-security-policy header to check if the URL can be embedded in an iframe

ciyer (changes requested):
This is starting to look very good! I have a couple of minor whitespace issues to keep git happy.

I am not sure I understand the logic of some of the tests. I would have assumed that validateCSP return false in a few example, but they are true. Maybe you could explain the reason why embedding should be allowed in those cases?

server/tests/utils/url.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    // invalid by protocol
+    expect(validateCSP(url, "frame-ancestors http://*").isIframeValid).toBe(false);
+
+    // invalid None rule
+    expect(validateCSP(url, "frame-ancestors 'none'").isIframeValid).toBe(false);
+
+    // valid full match url
+    const originUrl = new URL(config.server.url);
+    expect(validateCSP(url, `frame-ancestors ${originUrl.origin}`).isIframeValid).toBe(true);
```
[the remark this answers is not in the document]
andre-code: yes, because the URL is the URL (https://www.example.org) to which we are making the request of the headers, and their headers indicate in the iframe-ancestor from which URL allows embedding in iframes(https://dev.renku.ch), is these test we add these URL and compare them with the URL of our server(config.server.url) that is from where the request is finally going to be made.

server/src/routes/apis.ts
```diff
@@ -49,6 +51,35 @@ function reigsterApiRoutes(app: express.Application, prefix: string, authenticat
...
+      if (requestExternalURL.status >= 400) {
+        validationResponse.error = "Bad response from server";
+      }
+      else if (!requestExternalURL?.headers.has("content-security-policy")) {
+        validationResponse.isIframeValid = true;
+        validationResponse.detail = "Header does not contain Content-Security-Policy (CSP)";
+      }
+      else {
+        // check content-security-policy
+        const validation = validateCSP(req?.params?.url, requestExternalURL?.headers.get("content-security-policy"));
```
lorenzo-cavazzi, line 71:
```suggestion
        const validation = validateCSP(req.params.url, requestExternalURL.headers.get("content-security-policy"));
```

server/src/routes/apis.ts
```diff
@@ -38,7 +40,7 @@ const tmpProxMiddleware = createProxyMiddleware({
 });
 
 
-function reigsterApiRoutes(app: express.Application, prefix: string, authenticator: Authenticator): void {
+function registerApiRoutes(app: express.Application, prefix: string, authenticator: Authenticator): void {
```
lorenzo-cavazzi, line 43: Ah, thanks for spotting the typo!

server/src/utils/url.ts
```diff
@@ -0,0 +1,104 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { CheckURLResponse } from "../routes/apis.interfaces";
+import config from "../config";
+import { parse } from "tldts";
```
lorenzo-cavazzi, line 21: This is really just cosmetics: I tried to keep the import from libraries separated by the import from local files, so we can immediately identify what we defined locally. Let's just add a blank line:

```suggestion
import { parse } from "tldts";

import config from "../config";
import { CheckURLResponse } from "../routes/apis.interfaces";
```

server/src/routes/apis.ts
```diff
@@ -22,7 +22,9 @@ import { createProxyMiddleware } from "http-proxy-middleware";
 import config from "../config";
 import { Authenticator } from "../authentication";
 import { renkuAuth } from "../authentication/middleware";
-
+import fetch from "cross-fetch";
```
lorenzo-cavazzi, line 25: Cosmetics only: let's move the import from an external library together with the others (after line 20 `import { createProxyMiddleware } from "http-proxy-middleware";`)

server/src/utils/url.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export function validateCSP(url: string, csp: string): CheckURLResponse {
+  const response: CheckURLResponse = {
+    isIframeValid: false,
+    url: url,
+  };
+
+  try {
+    const externalUrl = new URL(url);
+    const originUrl = new URL(config.server.url);
```
lorenzo-cavazzi, line 31: Technically, you need the UI client url instead of the UI server URL. This may be a problem in cases where the UI server is reachable from `https://myrenkulab.io` and the UI client from `https://ui.myrenkulab.io`.
I think it's fine if we ignore that case since it's not really a thing, but please add a note to clarify it, maybe something like:
```
// We should use the UI client URL since it will be embedded there and the URL may be different
```

server/tests/utils/url.test.ts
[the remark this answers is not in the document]
lorenzo-cavazzi: It's a bit confusing, but that's meant to be the external gitlab URL.
I think we can use a more realistic URL instead of example.com to make it more explicit.
I suggest using:
```
https://gitlab.com/myNamespace/myProject/-/issues
```
(that should be set at line 25))

server/tests/utils/url.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    // valid subdomain match url
+    expect(validateCSP(url, `frame-ancestors *.renku.ch`).isIframeValid).toBe(true);
+
+    // valid subdomain-composed match
+    config.server.url = "https://test.dev.renku.ch";
```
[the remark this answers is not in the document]
lorenzo-cavazzi: The `validateCSP` function uses it. Technically it's not 100% correct since we should use the UI client URL instead of the UI server -- it may be a problem if someone has different origins for client and server, like:
UI server: https://myrenkulab.io
UI client from https://ui.myrenkulab.io

I don't think this is expected. We may still make it more robust by either:
- Including the UI client URL in the config (through the values), or
- Adding the current UI URL as a second parameter to the function, and use that instead of config.server.url

Between the 2, I would go for the second

lorenzo-cavazzi (changes requested):
Great PR! It seems to work as expected, and the multiple tests are great for understanding how it works (altough I'm still a bit confused by the part where you replace `www`... but it works)

I added some comments, but almost all of them are just tiny changes on cosmetics rather than real problems in the code. So it's almost ready to ship :smile:

server/tests/utils/url.test.ts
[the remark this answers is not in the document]
andre-code: Thx, great suggestion to make more clean the example, I updated the test

server/tests/utils/url.test.ts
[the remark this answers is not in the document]
andre-code: We decided to leave this server url, we add the note to update it when the client url is different from the server url

server/tests/utils/url.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    // invalid None rule
+    expect(validateCSP(url, "frame-ancestors 'none'").isIframeValid).toBe(false);
+
+    // valid full match url
+    const originUrl = new URL(config.server.url);
+    expect(validateCSP(url, `frame-ancestors ${originUrl.origin}`).isIframeValid).toBe(true);
+
+    // valid subdomain match url
+    expect(validateCSP(url, `frame-ancestors *.renku.ch`).isIframeValid).toBe(true);
```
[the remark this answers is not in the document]
andre-code: I updated the test to make it more clear with a  more realistic example

server/src/utils/url.ts
andre-code: A comment has been added to remember to update the url once it is different from the server url.

client/src/collaboration/lists/CollaborationList.container.js
```diff
@@ -31,6 +33,42 @@ const collaborationListTypeMap = {
...
+
+  useEffect( () => {
+    async function validateUrl() {
+      const isValid = await isValidUrlForIframe(props.iframeUrl, props.serverUrl);
+      setIsUrlValid(isValid);
+      if (!isValid)
+        props.onIFrameLoad();
+    }
+    validateUrl();
+  }, []); // eslint-disable-line
```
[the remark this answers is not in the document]
andre-code: if we remove the dependencies or add props as dependency or onIFrameLoad function it will be executed n times causing repeated calls to the API even if the iframe URL has not changed, so we can add the dependency to the iframe URL and the comment to disable the warning

As I understand, the onIFrameLoad function triggers the hook because this function is from hook useState used in the CollaborationList component which is rendered every time the props change.

ciyer:
I tested this, and it works very well! 🤩 

image

client/src/api-client/test-client.js
```diff
@@ -173,5 +173,7 @@ for (let key in methods) {
   };
 }
 client.baseUrl = "some-url";
+client.uiserverUrl = "https://dev.renku.ch/ui-server";
+client.isValidUrlForIframe = () => true;
```
lorenzo-cavazzi, line 177: Could you move this above, together with the other methods with a mocked response, like `getTemplatesManifest` and the others?.
Lines 168-174 already take care of making a valid function response out of it (response also comes as a Promise)

server/tests/utils/url.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    expect(validateCSP(url, `frame-ancestors *.dev.renku.ch`).isIframeValid).toBe(true);
+
+    // valid subdomain-composed match
+    config.server.url = "https://dev.renku.ch";
+    expect(validateCSP(url, `frame-ancestors *.dev.renku.ch`).isIframeValid).toBe(false);
+
+    // validate subdomain-composed match and protocol
+    config.server.url = "https://test.dev.renku.ch";
+    expect(validateCSP(url, `frame-ancestors http://*.dev.renku.ch`).isIframeValid).toBe(false);
+    config.server.url = "https://test.dev.renku.ch";
```
lorenzo-cavazzi, line 64: This is the same as the previous one. Is this intended or is it a typo?

lorenzo-cavazzi: Ok I see that one test is on `http` and the other `https`.
Should we remove this line since the value for `config.server.url` stays the same?

lorenzo-cavazzi (changes requested):
Great! I have a final _tiny_ change request on the `test-client.js` file. You added the response for `isValidUrlForIframe` at the end, but it would be nice to keep it coherent with the other mocked responses for the other client functions (see inline).

Otherwise, great PR! Iframes and security policies are a beast :grin:  
Can't wait to see this PR merged! :rocket:

lorenzo-cavazzi (approved):
:rocket:

lorenzo-cavazzi (approved):

ciyer (approved):

PR #1587, opened 2021-11-24
Route all requests to APIs through UI server

lorenzo-cavazzi:
This PR routes all the UI client API through the UI server.
It handles also the credentials expirations (e.g. tokens not refreshable anymore) by adding a specific header that the client reads and uses to redirect to the login page. This should happen very rarely and it's similar to the current behavior.
Should we want a different UX, we could redirect to a new UI page where we explain what happened and suggest to re-login.

P.S. Mind that the expired credentials also cause the invoked API to return a `500` message. This prevents very weird behaviors, happening when GitLab APIs respond with a `30x` code. The browser (at least Firefox) doesn't re-read the headers, causing an infinite login loop.

**How to test**
* Keep the network tab open and verify the APIs go through `<domain>/ui-server/api` instead of `<domain>/api`
* In the CI deployment, I lowered the token validity to maximum 10 minutes. Try to login, do some action, then leave the page open for >10 minutes and click on anything later. As soon as an API is invoked, the user should be redirected to the login

/deploy renku=ui1521-tests-unavailable-iframe renku-gateway=ui1521-anonid-nopath
#1521

### DIFF
--- a/client/src/api-client/errors.js
+++ b/client/src/api-client/errors.js
@@ -27,8 +27,16 @@ const API_ERRORS = {
   forbiddenError: "FORBIDDEN",
   notFoundError: "NOT_FOUND",
   internalServerError: "SERVER_ERROR",
-  networkError: "NETWORK_PROBLEM"
+  networkError: "NETWORK_PROBLEM",
+  authExpired: "AUTH_EXPIRED"
 };
+
+function throwAuthError(response) {
+  let error = new APIError();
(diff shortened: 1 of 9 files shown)


ciyer (dismissed):
Very nice; code is clear throughout and the changes make sense!

ciyer (approved):
One comment about maybe using a library to handle cookies, but I'm ok with merging this as is. Making this switch is certainly a milestone!

PR #1605, opened 2021-12-13
feat: Allow to stop the session even if it is not fully started

### LINKED ISSUE #787: [Environments] Stop failed/pending environments
## Motivation

Users often have problems when starting an environment due to various reasons:
* error in the entrypoint of the dockerfile
* unavailable resources
* registry unavailable for a newly ported project
* etc

as a result the environment is stuck in Failed or Pending state without the user being able to do something about it.

## Proposal

Add the option to stop the failed/pending environment.

## Screenshot

unschedulable

andre-code:
- Added status stopping to control when displaying the option to stop the session.
- Added display message when in the process of stopping a session.

e2

Closes #787 

/deploy

### DIFF
--- a/client/src/notebooks/Notebooks.container.js
+++ b/client/src/notebooks/Notebooks.container.js
@@ -461,8 +461,7 @@ class StartNotebookServer extends Component {
 
           // give extra context to logged users for building images
           const loggedPipelines = data.pipelines.type === NotebooksHelper.pipelineTypes.logged ? true : false;
-          const pendingPipeline = mainPipeline &&
-            (mainPipeline.status === "running" || mainPipeline.status === "pending");
+          const pendingPipeline = mainPipeline && ["running", "pending", "stopping"].includes(mainPipeline.status);
 
           if (loggedPipelines && pendingPipeline) {
             this.setState({
--- a/client/src/notebooks/Notebooks.present.js
+++ b/client/src/notebooks/Notebooks.present.js
@@ -337,7 +337,7 @@ function SessionJupyter(props) {
     else if (invisible) {
       return null;
     }
-    else if (status === "pending") {
+    else if (status === "pending" || status === "stopping") {
       content = (<Loader />);
     }
   }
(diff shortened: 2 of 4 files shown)


ciyer (dismissed):
This works very well! Just one small wording change request, then we can merge!

lorenzo-cavazzi (changes requested):
The stopping feature works well :+1: 

For the `"stopping"` status, I like the idea of saving that information so that users can see it's stopping -- even if it may sometimes go lost, most of the time stopping a session is quick and the users see better feedback (until SwissDataScienceCenter/renku-notebooks#809 is closed)

Manually testing worked in all the situations :tada: 
For completeness, I'm wondering if it makes sense to add the case for `status === "stopping"` in all the places in the Notebook.present.js where you have also a `status === "pending"` (e.g. at line 340, 1331, 1418, 1666). I think adding a `|| status === "stopping"` is enough.

andre-code:
> The stopping feature works well 👍
> 
> For the `"stopping"` status, I like the idea of saving that information so that users can see it's stopping -- even if it may sometimes go lost, most of the time stopping a session is quick and the users see better feedback (until [SwissDataScienceCenter/renku-notebooks#809](https://github.com/SwissDataScienceCenter/renku-notebooks/issues/809) is closed)
> 
> Manually testing worked in all the situations 🎉 For completeness, I'm wondering if it makes sense to add the case for `status === "stopping"` in all the places in the Notebook.present.js where you have also a `status === "pending"` (e.g. at line 340, 1331, 1418, 1666). I think adding a `|| status === "stopping"` is enough.

Thank you very much for the suggestion, it makes a lot of sense.
 I have added the "stopping" status in all those cases.

lorenzo-cavazzi (approved):
This works very well! :tada:

jonas-a-haller:
I have this running on https://renkulab.io/ but there is no such option, as to stop the "pending" sessions. What can I do to fix this?

andre-code:
> I have this running on https://renkulab.io/ but there is no such option, as to stop the "pending" sessions. What can I do to fix this?

Hi @jonas-a-haller  this feature will be available in the next release on renkulab.io.

Also, unfortunately there were transient problems on renkulab.io this morning and pending sessions should have been stopped by the admins, but they have been fixed.

PR #1608, opened 2021-12-14
feat: Add session share link with more options

### LINKED ISSUE #1596: feat: allow more options for autostart
The session "play" button has been a really nice addition to the UI, and it is now even possible to provide users with a URL that they can share with others to automatically "land" in a running session. Currently, this behavior only allow starting from the latest commit on the main branch. It would be nice if, for the purposes of automatically generating links to be shared outside renku, we could add additional query parameters to the `<renku>/<namespace>/<project>/sessions/new?autostart=1` URL that would allow one to pre-select a branch and a commit for something like

```
<renku>/<namespace>/<project>/sessions/new?autostart=1?branch=feature?commit=12345abcde
```

In addition, one could imagine a drop down menu that would allow a user to copy the autostart link for easier sharing, like is currently done for pre-filling the new project form.

# Mockup

image

andre-code:
- Added button to create a share link for sessions.
- Added handler when the branch name or commit key is not valid or incomplete

### Successful case

ok-session-share-link

### When the commit or brach name is invalid or incomplete

Screenshot 2021-12-15 at 16 01 25

Closes #1596

/deploy renku=ui1608-tests-sessions

### DIFF
--- a/client/src/notebooks/Notebooks.container.js
+++ b/client/src/notebooks/Notebooks.container.js
@@ -28,6 +28,7 @@ import {
 import { StatusHelper } from "../model/Model";
 import { ProjectCoordinator } from "../project";
 import { Url } from "../utils/url";
+import { Loader } from "../utils/UIComponents";
 
 
 /**
@@ -123,6 +124,9 @@ class ShowSession extends Component {
   render() {
     if (this.props.blockAnonymous)
       return <NotebooksDisabled location={this.props.location} />;
+
+    if (!this.model.get("notebooks.fetched"))
+      return <Loader />;
 
     return (
       <ShowSessionMapped
@@ -360,6 +364,7 @@ class StartNotebookServer extends Component {
         errorMessage: `The session could not be started because the commit or the branch name was not found.`
       },
     });
+    this.coordinator.startNotebookPolling();
   }
 
   async refreshBranches() {
--- a/client/src/notebooks/Notebooks.container.js
+++ b/client/src/notebooks/Notebooks.container.js
@@ -316,7 +316,8 @@ class StartNotebookServer extends Component {
   componentDidMount() {
     this._isMounted = true;
     if (!this.props.blockAnonymous) {
-      this.coordinator.startNotebookPolling();
+      if (!this.autostart)
+        this.coordinator.startNotebookPolling();
       this.refreshBranches();
     }
   }
@@ -403,7 +404,8 @@ class StartNotebookServer extends Component {
       }
 
       // set branch
-      this.coordinator.setBranch(branchToSet ?? {});
+      // do not retrieve the books yet, wait for the commit to be set
+      this.coordinator.setBranch(branchToSet ?? {}, !this.autostart);
       if (!branchToSet) {
         this.coordinator.setCommit({});
         return;
--- a/client/src/notebooks/Notebooks.container.js
+++ b/client/src/notebooks/Notebooks.container.js
@@ -408,14 +408,14 @@ class StartNotebookServer extends Component {
         this.coordinator.setCommit({});
         return;
       }
-      // force get commit of branch selected when autostart session
-      this.refreshCommits(this.autostart);
+
+      this.refreshCommits();
     }
   }
 
-  async refreshCommits(force = false) {
+  async refreshCommits() {
     if (this._isMounted) {
-      if (!this.projectModel.get("commits.fetching") || force) {
+      if (!this.projectModel.get("commits.fetching")) {
         const branch = this.model.get("filters.branch");
         await this.projectCoordinator.fetchCommits({ branch: branch.name });
       }
--- a/client/src/notebooks/Notebooks.container.js
+++ b/client/src/notebooks/Notebooks.container.js
@@ -349,6 +349,18 @@ class StartNotebookServer extends Component {
     this.setState({ ignorePipeline: value });
   }
 
+  setErrorInAutostart() {
+    this.autostart = false;
+    this.setState({
+      "starting": false,
+      launchError: {
+        frontendError: true,
+        pipelineError: false,
+        errorMessage: `The session could not be started because the commit or the branch name was not found.`
+      },
+    });
+  }
+
   async refreshBranches() {
     if (this._isMounted) {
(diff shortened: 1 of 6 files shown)


lorenzo-cavazzi (changes requested):
Very nice feature!
I found a problem when playing around. If the url contains a wrong commit or branch (e.g. deleted branch/commit on the project, or just an error on copy/paste by the user), the UI seems to be stuck on the "Starting session" page.
It would be nice giving feedback to the user.

Screenshot_20211214_140508

andre-code:
> Very nice feature! I found a problem when playing around. If the url contains a wrong commit or branch (e.g. deleted branch/commit on the project, or just an error on copy/paste by the user), the UI seems to be stuck on the "Starting session" page. It would be nice giving feedback to the user.
> 

Thanks for the suggestion, I added the feedback when the commit key or branch is not valid 😉

lorenzo-cavazzi (changes requested):
I fear this solution may have a few issues on projects with many branches/commits (check the inline content).
Instead of using the `force` parameter on `refreshCommits`, the fetch should either include the target branch the first time (not sure if possible, maybe it's triggered somewhere else) or wait until it's finished and then fetch again if a different target branch was defined in the autostart.

I think it may be necessary (or at least much easier) doing that check in the `triggerAutoStart` function instead of modifying the fetch/selectBranch and fetch/selectCommit functions .
The `triggerAutoStart` function already contains a 1000ms timeout to check if every requirement for the autostart is ready. You could use that to check if
* the branches/commits are not being fetched (otherwise wait for the next loop)
* check if an autostart branch has been defined and the currently selected branch is that one. Otherwise set it using the `selectBranch` function, or create the error if not available (mind the previous condition -- all the other fetches should not be ongoing)
* same check for the commit, using `selectCommit` -- after the previous `selectBranch` it would require a few loops since the commits fetching would be ongoing again

client/src/notebooks/Notebooks.container.js
```diff
@@ -367,29 +385,37 @@ class StartNotebookServer extends Component {
...
+      // force get commit of branch selected when autostart session
+      this.refreshCommits(this.autostart);
     }
   }
 
-  async refreshCommits() {
+  async refreshCommits(force = false) {
     if (this._isMounted) {
-      if (!this.projectModel.get("commits.fetching")) {
+      if (!this.projectModel.get("commits.fetching") || force) {
```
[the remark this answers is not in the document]
andre-code: Good point 🙌, I find it effective to avoid fetching commits from the default branch when there is an autostart session so I added that validation in the View component.

In case there is no autostart session, the commits are retrieved from the default branch (current behavior).
In case of an invalid branch in the autostart session, it displays an error, stops the autostart, and retrieves the commits from the default branch so that all the features that depend on those commits continue working.

Also to improve the behavior of the starting session progress bar, I added the validation to don't fetch the notebooks before setting the branch and commit when is an autostart session.

client/src/notebooks/Notebooks.present.js
```diff
@@ -2059,6 +2074,74 @@ class CheckNotebookIcon extends Component {
...
+  const [includeBranch, setIncludeBranch] = useState(false);
+  const [includeCommit, setIncludeCommit] = useState(false);
+  const [url, setUrl] = useState("");
+  const data = { namespace: props.filters?.namespace, path: props.filters?.project };
+
+  useEffect(() => {
+    if (!data.namespace || !data.path)
+      return;
+    let urlSession = Url.get(Url.pages.project.session.autostart, data, true);
+    const parameters = JSON.parse(props.notebooks.lastParameters);
```
lorenzo-cavazzi, line 2087: No need to convert back that stringified object from `lastParameters` to get branch and commit. That prop is meant to be used only by the `NotebooksCoordinator` to discard outdated data.

You can used `filters` here: `props.filters.branch.name` and `props.filters.commit.id`.

lorenzo-cavazzi (changes requested):
Great, your solution is better than the suggested one!

There are a couple of minor details to iron out before merging:
* When the autostart fails, the page keeps showing `Verifying available session` without retrieving them. [EXAMPLE](https://renku-ci-ui-1608.dev.renku.ch/projects/lorenzo.cavazzi.tech/project-renku-0-16-2/sessions/new?autostart=1&branch=abc)
Screenshot_20211217_121841
* When an autostart successfully completes, the page shows a message about "session not available" for a few seconds.
[EXAMPLE](https://renku-ci-ui-1608.dev.renku.ch/projects/lorenzo.cavazzi.tech/project-renku-0-16-2/sessions/new?autostart=1)
Screenshot_20211217_111757
It would be better to show the `<Loader>` -- I think that was the case before [EXAMPLE](https://dev.renku.ch/projects/lorenzo.cavazzi.tech/project-renku-0-16-2/sessions/new?autostart=1
)
Screenshot_20211217_112935

client/src/notebooks/Notebooks.present.js
andre-code: sure, updated

andre-code:
@lorenzo-cavazzi  Updated PR to start polling session if autostart fails (first case reported) and display the loader while getting session status (second case reported)

client/src/notebooks/Notebooks.present.js
```diff
@@ -2056,6 +2071,78 @@ class CheckNotebookIcon extends Component {
...
+  };
+
+  useEffect(() => {
+    if (!data.namespace || !data.path)
+      return;
+    let urlSession = Url.get(Url.pages.project.session.autostart, data, true);
+    urlSession = includeCommit ? `${urlSession}&commit=${data.commit}` : urlSession;
+    urlSession = includeBranch ? `${urlSession}&branch=${data.branch}` : urlSession;
+    setUrl(urlSession);
+  }, [ includeCommit, includeBranch, props.notebooks.lastParameters, data ]);
```
lorenzo-cavazzi, line 2092:
```suggestion
  }, [ includeCommit, includeBranch, data ]);
```

lorenzo-cavazzi (approved):
It looks good to me! :tada: 
Thanks for polishing all the details.

P.S. I'm re-running the acceptance tests, the previous errors were confusing since I could work with sessions and manually replicate the steps from the tests :shrug:

PR #1610, opened 2021-12-16
fix(client): show the correct session status on notebooks preview

### LINKED ISSUE #1606: Notebook status does not show correctly
## Description

In the file viewer, when looking at a notebook, there is an icon that lets the user start a session or open the notebook in a session. 

This icon is not currently working (probably due to the project state refactoring)

image

image

lorenzo-cavazzi:
There were a few issues with the `CheckNotebooksStatus` component, leading to different results based on a few factors (status of the sessions, user default branch already loaded when browsing the file, ...)

This should restore the previous behavior, and correct the wrong links.
With #1345, the feature will become more useful since it will be possible to identify the "correct" commit for the session instead of checking the latest

/deploy
fix #1606


ciyer (approved):
Code looks good, the button is much better aligned, and everything works as expected!

I think it would make sense for the button to take the user to the iframe-based session instead of opening a separate window, but I will make a new issue for that.

PR #1612, opened 2021-12-17
fix: Order maintenance by schedule date asc

### LINKED ISSUE #1609: statuspage maintenance in popup is ordered incorrectly
The banned announcing upcoming maintenance windows shows the wrong maintenance first - see the current state at renkulab.io:

image

It should show the gitlab maintenance pt1 first. In addition, there should be an indication that _multiple_ maintenances are upcoming so the user is aware that this is not the only one to consider.

andre-code:
Added sorting by schedule date in the maintenance summary.

Tested in Chrome, Safari and Firefox browsers.

Closes #1609 

/deploy ui.statuspage=lhwy5ln8yrnn


ciyer (approved):
Ah! I added a commit to make the sorting function implementation more explicit.

PR #1617, opened 2021-12-21
feat: add visibility when fork a project (#1272)

### LINKED ISSUE #1272: Add a visibility option to the project fork dialog
## Motivation

A common use case for a fork is that the fork should have less visibility than the parent. Making this change is possible, but tedious, since it requires going to the GitLab UI.

## Proposal

Add a `Visibility` option to the modal dialog. The initial value should be the same as the project being forked, and only as visible or less visible options should be available.

* For a public project: public, internal, private
* For an internal project: internal, private
* For a private project: private

andre-code:
Added a Visibility option to the modal dialog. The initial value should be the same as the project being forked, and only as visible or less visible options should be available.

For a public project: public, internal, private
For an internal project: internal, private
For a private project: private

Display an informative text in the fork modal dialog when the project was forked but the custom visibility can't be added.

### Successful fork

fork-project

### Error setting visibility

error-fork-project

### Error when setting visibility when closing the modal
error-fork-project-notification

closes #1272 
/deploy

### DIFF
--- a/client/src/project/new/ProjectNew.present.js
+++ b/client/src/project/new/ProjectNew.present.js
@@ -32,7 +32,7 @@ import {
   Modal, ModalBody, ModalFooter, ModalHeader, Row, Table, UncontrolledTooltip
 } from "reactstrap";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import { faInfoCircle, faLink, faSyncAlt } from "@fortawesome/free-solid-svg-icons";
+import { faExclamationTriangle, faInfoCircle, faLink, faSyncAlt } from "@fortawesome/free-solid-svg-icons";
 
 import {
   ButtonWithMenu,
@@ -171,9 +171,9 @@ function ForkProjectStatus(props) {
   else if (props.forkVisibilityError) {
     return (
       <p>
-        <FontAwesomeIcon icon={faInfoCircle} />
+        <FontAwesomeIcon icon={faExclamationTriangle} />
         {" "} The project has been forked but an error occurred when setting the visibility
-        {" "}{props.forkVisibilityError}
+        {props.forkVisibilityError}
       </p>
     );
   }
--- a/client/src/project/new/ProjectNew.present.js
+++ b/client/src/project/new/ProjectNew.present.js
@@ -634,7 +634,7 @@ class Visibility extends Component {
       main = (
         <Fragment>
           <br />
-          <Label className="font-italic">Verifying... <Loader inline={true} size={16} /></Label>
+          <Label className="font-italic">Determining options... <Loader inline={true} size={16} /></Label>
         </Fragment>
       );
     }
--- a/client/src/project/new/ProjectNew.present.js
+++ b/client/src/project/new/ProjectNew.present.js
@@ -99,7 +99,7 @@ function ForkProjectHeader(props) {
 }
 
 function ForkProjectBody(props) {
-  const { fetching, forkError, forking } = props;
+  const { fetching, forkError, forkVisibilityError, forking } = props;
   if (fetching.namespaces || fetching.projects) {
     const text = fetching.namespaces ?
       "namespaces" :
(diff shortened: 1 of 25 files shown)


client/src/project/shared/Projects.state.js
```diff
@@ -158,6 +158,41 @@ class ProjectsCoordinator {
...
+
+    if (namespace?.kind === "user") {
+      return computeVisibilities(["public", projectVisibility]);
+    }
+    else if (namespace?.kind === "group") {
+      // get group visibility
+      const group = await this.client.getGroupByPath(namespace.full_path).then(r => r.data);
+      return computeVisibilities([group.visibility, projectVisibility]);
+    }
+    return availableVisibilities;
```
lorenzo-cavazzi, line 194: This is a very nice function!
We have something on the same line in `ProjectNew.state.js` at line 466 (`async getVisibilities(namespace)`). I'm wondering if it would make sense to merge the 2, or extrapolate a common function. Maybe it's not that straightforward tho...

client/src/project/new/ProjectNew.present.js
```diff
@@ -602,7 +630,7 @@ class Visibility extends Component {
         </Fragment>
       );
     }
-    else if (meta.namespace.fetching) {
+    else if (meta.namespace.fetching || !meta.namespace.visibilities || !input.visibility) {
```
lorenzo-cavazzi, line 633: Are you sure about `!input.visibility` here? Since `input.visibility` is set when the user selects the visibility, if that case ever happens the user can't click it anyway.
I would remove it from here.

client/src/project/shared/Projects.state.js
lorenzo-cavazzi: P.S. this is just a "nice to have". If you think it's feasible (and reasonably easy) let's include it in the PR. Otherwise, feel free to leave this out.

lorenzo-cavazzi (changes requested):
It works well in all cases I tried! :tada: 

Just a tiny change on using an `input` field a a condition to show the element to set that input.

I also added a note on a function you created. Feel free to ignore it if it's not easy to derive a common function

client/src/project/shared/Projects.state.js
ciyer: At the very least, I think it would be nice to merge the `computeVisibilities` functions. It may be necessary to introduce a `maxVisibility(projectVisibility, groupVisibility)` function.

ciyer:
This is great and works well! I tried several cases, and all were handled properly. On _minor_ request. Not critical, but nice to have if easy. Could the message shown when updating the visibility be changed to `Determining options` instead of `Verifying`? 

image

client/src/project/shared/Projects.state.js
andre-code: Thanks for the suggestion, I refactor the function to be used also by the NewProjectCoordinator class.

client/src/project/new/ProjectNew.present.js
andre-code: It is necessary to avoid choosing invalid values (null) in the visibilities select, for example when recalculating the visibilities list.
This validation does not block the visibility selection because after the list calculation a valid default value is set.

lorenzo-cavazzi: I think that value is never set to `null`. It's initialized as `""` in the RebkuModel, and it should always be default to a valid value when re-computed.
We can leave it here, it's not a big deal, I just thought it's a bit confusing for someone who doesn't know the magic happening in the state file pre-setting the `input.visibility`

lorenzo-cavazzi: Ah nevermind, I just noticed `resetVisibility` is now a thing and it temporarily sets the visibility to `null`. Ignore this comment :smile:

lorenzo-cavazzi (approved):
Very nice addition to the fork feature! :rocket:

ciyer (approved):
Very nice!

PR #1620, opened 2021-12-23
feat(client): support multiple renku-core instances

### LINKED ISSUE #1507: Support multiple versions of the core-service
## Motivation

Motivated by the transition to the core-service 1.0.0, which introduces new features and incompatibilities, we need to able to support multiple versions of the core-service in the UI.

Certain operations (e.g., new project) should always go to the latest available version of the core-service. Within a project, operations should go to a compatible version of the core-service, though users should be informed if there is a newer version and be given the option to upgrade.

## Design

We need to identify and implement a strategy for:
- discovering which service a call needs to be routed to
- remembering that information
- routing calls to the right service

[coming soon]

## Progress

- [x] API Design (sprint-??)
- [ ] UI Implementation

lorenzo-cavazzi:
This PR introduces support for simultaneous renku-core instances.
When browsing a project, some operations are blocked until the project version is assessed, and the core APIs invocation is delayed and properly routed. For each new project version, the backend availability is checked and cached for further re-use when browsing other projects on the same version.

Mind that this PR contains a few tangential changes -- minor fixes/improvements made on the fly when adapting the code.

/deploy renku=1.0-next renku-graph=cli-v1 renku-core=2507-lock-status-endpoint
fix #1507

### DIFF
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -219,6 +219,7 @@
           "noreferrer",
           "nowrap",
           "nullable",
+          "onloadend",
           "papermill",
           "pathname",
           "pdfjs",
--- a/client/src/api-client/dataset.js
+++ b/client/src/api-client/dataset.js
@@ -40,21 +40,32 @@ export default function addDatasetMethods(client) {
     "Accept": "application/json"
   };
 
-  client.uploadFile = (file, unpack_archive = false, setFileProgress, thenCallback, onErrorCallback,
-    setController, onFileUploadEnd) => {
+  client.uploadFile = (
+    versionUrl = null,
+    file,
+    unpack_archive = false,
+    setFileProgress,
+    thenCallback,
+    onErrorCallback,
+    setController,
+    onFileUploadEnd,
+  ) => {
     const data = createFileUploadFormData(file);
     data.append("processData", false);
 
     let currentPercentCompleted = -1;
     let httpRequest = new XMLHttpRequest();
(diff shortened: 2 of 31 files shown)


ciyer (changes requested):
Lots of big improvements in here! Both, better functionality and code improvements.

Everything works very well. I tested very old projects -- there I correctly get informed that I cannot modify this project.

On old, but compatible projects, I could add and modify datasets and edit the session settings.

I have a few suggestions about changes to messages that I made in the code.

1. The message shown for outdated, but compatible projects should be changed

image

I think we can remove the line that says "A new version of Renku is available". I think the warning triangle next to the project name is enough of a notification. 

2. The text of the status messages needs to be modified

image

I think the messages should be:
- [RenkuLab UI Compatibility] This project is compatible with the RenkuLab UI, but upgrading to the latest <b>renku version</b> is highly recommended.
- [Renku Version] A new version of renku is available. The update should take about ... seconds

3. Problems with datasets from old projects

I have a metadata 7 / renku 0.11.3 project with a dataset. The dataset does not appear in the project dataset view, but it is in the KG. However, if I try to view the dataset details from dataset search, the UI crashes.

Overall, I think this is very close, just a few minor issues to fix! 🎉

lorenzo-cavazzi:
Thanks for the thorough review! I missed quite a few things :sweat_smile: 

I added all the suggested changes.
Here is an example of the status page for an outdated but usable project.

Screenshot_20220107_092148

ciyer (approved):
😍 . Ready to merge!

PR #1621, opened 2021-12-23
Present estimate of migration duration

### LINKED ISSUE #1602: Show expected migration time in UI
## Motivation

The migration of a project can time some time. We should let the user know how long we expect the migration to take so they understand how long they may have to wait.

## Mock-ups / Design

*If* the project is being upgraded from a version < 1.0.0 to a version >= 1.0.0, show the expected migration time information. This can be retrieved from the KG.

image

ciyer:
# Images

image

image

# Testing

This PR shows an estimate of the migration time **if** the user is a member of the project and the project is being updated to metadata version 9. The first restriction is because the computation requires knowing the number of commits and that information is only available if the user is a member of the project. The second restriction is because the method of estimating only works for projects migrating to metadata version 9.

Here are some projects that can be used to test:
- https://renku-ci-ui-1621.dev.renku.ch/projects/cr-migration-test/project-with-core-1-0-1-group-v2/ (no estimate shown, since project is already on metadata v. 9)
- https://renku-ci-ui-1621.dev.renku.ch/projects/cr-migration-test/test-8 (estimate shown, since project can be updated to v9 automatically)
- https://renku-ci-ui-1621.dev.renku.ch/projects/cr-migration-test/test-7 (estimate shown, since project can be updated to v9, but must be done manually)

Fix #1602

/deploy renku=1.0-next renku-graph=cli-v1 renku-core=v1.0.1

### DIFF
--- a/client/src/project/status/MigrationUtils.js
+++ b/client/src/project/status/MigrationUtils.js
@@ -34,7 +34,7 @@ function GeneralErrorMessage({ error_while, error_what, error_reason }) {
       or create an issue in {" "}
       <ExternalIconLink url="https://github.com/SwissDataScienceCenter/renku/issues" icon={faGithub} title="GitHub" />.
     </p>
-    <div><strong>Error Message</strong><pre>{error_reason}</pre></div>
+    <div><strong>Error Message</strong><pre style={{ whiteSpace: "pre-wrap" }}>{error_reason}</pre></div>
   </ErrorAlert>;
 }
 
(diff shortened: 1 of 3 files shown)


lorenzo-cavazzi (approved):
This works well :tada: 
I have tested it with a few projects. I couldn't find many big projects to use as examples (and migrating them from renkulab is time-consuming), but the estimate looked reasonable in all the cases.

P.S. This probably requires a rebase on the reference branch `1.0-next` that I've recently updated since it was missing a few changes from `master`.

PR #1622, opened 2021-12-23
feat: Inline math formulas (#1600)

### LINKED ISSUE #1600: Inline math formulas
## Motivation

We recently introduced support for math formulas in the markdown files preview. All works well except for some inline formulas.
It would be nice to explore if adding that support is easy or not.

## Proposal

Check te documentations and catch the inline formulas if that's an easy task.

## EXAMPLES
Mind the final line

**Renkulab**
Screenshot_20211207_114645

**GitLab**
image

**RAW**
## The backbone: Gaussian manipulations

This library primarily provides the functionality to quickly manipulate, integrate and sample from Gaussians. If 

```math
\phi(\mathbf{x}) = N(\mu, \Sigma),
```

is a Gaussian density, this library allows to quickly compute the resulting functional form

```math
u(\mathbf{x}) = \beta\exp\left(-\frac{1}{2}\mathbf{x}^\top\Lambda \mathbf{x} + \nu^\top \mathbf{x}\right)\phi(\mathbf{x}).
```

For the resulting measure certain integrals can then be computed quickly

```math
\int f(\mathbf{x}) {\rm d}u(\mathbf{x}),
```

where $`f`$ is can be up to fourth order of $`\mathbf{x}`$. Furthermore, some functionality for mixture measures and density is provided.

andre-code:
- support inline math code between ``$` and `$``

Project example:
https://renku-ci-ui-1622.dev.renku.ch/projects/lorenzo.cavazzi.tech/math-example

/deploy

Closes #1600


lorenzo-cavazzi (commented):
The inline looks good, but there are still a few problems with the formulas.
In the example you linked, the second formula is the following
```math
u(\mathbf{x}) = \beta\exp\left(-\frac{1}{2}\mathbf{x}^\top\Lambda \mathbf{x} + \nu^\top \mathbf{x}\right)\phi(\mathbf{x}).
```

It's rendered correctly on GitLab 
Screenshot_20220103_104229
but it's not in our UI (look at the extra `<=ft` and `right`)
Screenshot_20220103_104300

P.S: I'm not sure if it's easy to fix this using `replace` or we should rather give a look at other alternative libraries for rendering the formulas.

andre-code:
> The inline looks good, but there are still a few problems with the formulas. In the example you linked, the second formula is the following
> 
> ```
> u(\mathbf{x}) = \beta\exp\left(-\frac{1}{2}\mathbf{x}^\top\Lambda \mathbf{x} + \nu^\top \mathbf{x}\right)\phi(\mathbf{x}).
> ```
> 
> It's rendered correctly on GitLab Screenshot_20220103_104229 but it's not in our UI (look at the extra `<=ft` and `right`) Screenshot_20220103_104300
> 
> P.S: I'm not sure if it's easy to fix this using `replace` or we should rather give a look at other alternative libraries for rendering the formulas.

Many thanks for the review @lorenzo-cavazzi 
According to this information https://docs.gitlab.com/ee/user/markdown.html#math the `$ and ```math delimiters are Gitlab's own that use Latex so the error was in using ascii math instead of the Latex typesetting, with the correct configuration the formulas are rendered as expected.

lorenzo-cavazzi (approved):
Switching to Latex for the `math` block codes has worked!
All the previous errors (left) are gone and the formulas are rendered correctly (right) :tada: 

Let's merge it as soon as the acceptance tests pass :tada: 

Screenshot_20220112_171446

Issue #1623, opened 2022-01-03
Support renku-core API versioning


lorenzo-cavazzi:
## Motivation

The renku-core service now supports API versioning. We should use it in the UI now that it has been backported to v8.
Each `/version` endpoint will show information about the API version supported.

```
"result":{
    "supported_project_version":9.0,
    "minimum_api_version":"0.9",
    "maximum_api_version":"1.1",
    "latest_version":"1.0.6"
}
```

The main difficulty is that older implementations won't be updated to the latest API version. This may become difficult to handle since we will need to keep backward compatibility.

## Implementation

The core API URLs should include also the API version.
Currently, it includes only the project version (apart from `GET /cache.migrations_check`) in this format `<clientUrl>/renku/<versionUrl>/apiName` (e.g. `<clientUrl>/renku/9/version`).
The new format will be `<clientUrl>/renku/<versionUrl>/<apiVersionUrl>/apiName` (e.g. `<clientUrl>/renku/9/1.1/version`).

We should set an API version for each project version, something like the following:

```
const versionMatrix = {
  "8": "0.9",
  "9": "1.1"
}
```

and handle the cases where there are significant changes directly in the `client.<apiName>` logic (if possible).

In the future, we should consider using the `GET /versions` API to discover all the available versions in advance and also verify if the compatibility matrix is valid in the current deployment.

lorenzo-cavazzi:
There has been a few changes lately and this issue is now outdated.
We will support versioned API on renku-core as defined here #2718

PR #1624, opened 2022-01-03
TEMP: support renku-core API versioning

lorenzo-cavazzi:
This is only a stub.
re #1623


lorenzo-cavazzi:
The reference issue #1623 was modified in a way that this approach is not valid anymore.

PR #1629, opened 2022-01-06
Sanitize Jupyter notebooks to remove problematic HTML

ciyer:
# Testing


lorenzo-cavazzi (dismissed):
Nice solution! It works in all cases I tried!

lorenzo-cavazzi (approved):

PR #1635, opened 2022-01-12
Cypress tests for UI

ciyer:
Initial scaffolding for Cypress-based UI tests (using mocks).

# Testing

Run `npm install .` in the `e2e/` folder to get the latest dependencies. Then run `npm run e2e:local` to see the cypress tests and try them out. In this mode, only the tests in the `local` folder will work. Those in the `live` folder need to be run against one of the other scripts (for example, `npm run e2e`).


lorenzo-cavazzi (commented):
This looks great!
And the code is easy to follow :+1: 

Would it make sense to add a few words in the README file to explain how to start this and differentiate local vs live e2e?

e2e/package.json
```diff
@@ -4,10 +4,13 @@
...
   "main": "index.js",
   "scripts": {
+    "cypress:local": "cypress open --config-file cypress.local.json",
     "e2e": "cypress open",
+    "e2e:local": "start-server-and-test start:local 3000 cypress:local",
     "e2e:run": "cypress run",
     "e2e:headless": "cypress run --headless",
-    "lint": "eslint . --ext .ts"
+    "lint": "eslint . --ext .ts",
+    "start:local": "npm --prefix ../client run start"
```
andre-code, line 13: great fix 🚀

andre-code (commented):
great job 🌟  

Just a tiny suggestion, I see that the Fixtures class could easily grow so it would be great to split it into classes by data domains or something similar ... like UserFixture, ProjectFixture. Feel free to make that suggestion or we can include it in future additions to this class 😉

ciyer:
Thanks for the feedback everyone. I have implemented both the requests:
* updated README to explain how to use the local tests
* refactored Fixtures into multiple files so no individual file grows without bounds

lorenzo-cavazzi (approved):
:rocket:

andre-code (approved):
👏

Issue #1634, opened 2022-01-12
Activate Sentry performance 


ciyer:
# Description

We should activate the performance tracking in Sentry.

  https://docs.sentry.io/platforms/javascript/guides/react/performance/

PR #1648, opened 2022-01-18
feat(client): add sentry performance (#1634)

andre-code:
This PR update sentry configuration to support performance monitoring using sentry.

To visualize some tracked transaction 👉  [sentry dashboard](https://sentry.dev.renku.ch/organizations/swiss-datascience-center/performance/?project=5) there are some examples in the andrea namespace

### Note

By default the monitoring is disabled `SENTRY_SAMPLE_RATE=0`, when runs telepresence and sentry is enabled the default value is 1.0, which means to track all transactions.

In production is recommended to set `sampleRate=0.2` in `helm-chart/renku-ui/values.yaml`. This would track 20% of all requests in Sentry performance monitoring. It is 0 by default

/deploy renku=1.0-next renku-graph=cli-v1 renku-core=v1.0.3 extra-values=ui.sentry.enabled=true,ui.sentry.namespace=renku-ci-ui-1648,ui.sentry.url=https://182290b8e1524dd3b7eb5dd051852f9f@sentry.dev.renku.ch/5,ui.sentry.sampleRate=1.0


client/src/utils/sentry/Sentry.js
```diff
@@ -92,6 +96,7 @@ function sentryInit(url, namespace = null, userPromise = null, version = null, t
   // Save data
   sentryUrl = url;
   sentryNamespace = namespace;
+  sentrySampleRate = parseFloat(sampleRate) || 0;
```
ciyer, line 99: I think it would be good to add a clamp to set values < 0 to 0 and values > 1 to 1.

ciyer (changes requested):
Code looks good! Just one minor suggestion for catching out-of-bounds values.

client/src/App.js
```diff
@@ -209,4 +210,4 @@ class App extends Component {
   }
 }
 
-export default App;
```
lorenzo-cavazzi, line 212: Since Sentry is initialized in the index file, would it make sense to keep the whole Sentry logic there? That would move `Sentry.withProfiler(App)` to somewhere else, ideally made in a function inside `utils/helpers/sentry` and invoked in the index under the same condition used to configure Sentry  (`if (params.SENTRY_URL) {`)
In that way, we won't even import the Sentry library when it's not needed

client/src/App.js
```diff
@@ -209,4 +210,4 @@ class App extends Component {
...
+export default Sentry.withProfiler(App);
```
lorenzo-cavazzi, line 213: Since Sentry is initialized in the index file, would it make sense to keep the whole Sentry logic there? That would move `Sentry.withProfiler(App)` to somewhere else, ideally made in a function inside `utils/helpers/sentry` and invoked in the index under the same condition used to configure Sentry  (`if (params.SENTRY_URL) {`)
In that way, we won't even import the Sentry library when it's not needed

helm-chart/renku-ui/templates/deployment-template.yaml
```diff
@@ -55,16 +55,18 @@ spec:
...
-              {{- if .Values.sentry.enabled }}
+            {{- if .Values.sentry.enabled }}
             - name: SENTRY_URL
               value: {{ .Values.sentry.url | quote }}
             - name: SENTRY_NAMESPACE
               value: {{ .Values.sentry.namespace | default (printf "%s" .Release.Namespace) | quote }}
-              {{- end }}
-              {{- if .Values.maintenance }}
+            - name: SENTRY_SAMPLE_RATE
+              value: { { .Values.sentry.sampleRate | quote } }
```
lorenzo-cavazzi, line 64:
```suggestion
              value: {{ .Values.sentry.sampleRate | quote }}
```
I think this is throwing the error in the deployment.

lorenzo-cavazzi (changes requested):

andre-code:
Thank @ciyer  and @lorenzo-cavazzi  for the review, I implemented all the suggestions.

client/src/utils/helpers/sentry/Sentry.js
```diff
@@ -92,6 +108,7 @@ function sentryInit(url, namespace = null, userPromise = null, version = null, t
   // Save data
   sentryUrl = url;
   sentryNamespace = namespace;
+  sentrySampleRate = _.clamp(parseFloat(sampleRate) || 0, 0, 1);
```
ciyer, line 111: Nice solution!

ciyer (commented):
Code looks great!

client/src/utils/helpers/sentry/Sentry.js
```diff
@@ -100,11 +117,15 @@ function sentryInit(url, namespace = null, userPromise = null, version = null, t
     environment: sentryNamespace,
     release: getRelease(uiVersion),
     beforeSend: (event) => hookBeforeSend(event),
-    denyUrls: sentryDenyUrls
+    denyUrls: sentryDenyUrls,
+    integrations: [new TracingIntegrations.BrowserTracing(
+      { tracingOrigins: ["localhost", /^\//, /ui-server\/api/] },
```
Panaetius, line 122: Do we need to add `renkulab.io` (i.e. `.Values.global.renku.domain` from the chart), maybe instead of `localhost`? What scenarios are there were localhost would be user?

Panaetius (approved):
Should we set https://docs.sentry.io/platforms/javascript/performance/instrumentation/automatic-instrumentation/#maxtransactionduration ? The default of 600s seems rather large, maybe 30s is enough.

I've approved for now since I think everything would work as is, just the two comments/suggestions to consider.

client/src/utils/helpers/sentry/Sentry.js
andre-code: thx @Panaetius  for the review 🙌. So far there is no particular usage for localhost so I updated it to send only the URL of the UI server and the value of maxTransactionDuration

ciyer (dismissed):
🎉

lorenzo-cavazzi (approved):

PR #1656, opened 2022-01-21
fix(client): restore projects data used by datasets

lorenzo-cavazzi:
This fixes a bug spotted by @gavin-k-lee that prevents datasets to be added to projects.

**How to test**: look for a dataset using the dataset search, open any and click on the "Add to project" button on the top right.
The user should be able to look for a project again and import the dataset (if the target project is up-to-date).
Trying the some on dev won't work -- the user won't even be able to select a target project.

P.S. We should probably check also the source project version. That may not be straightforward in the current implementation.
I would suggest re-working a bit the "add to project" logic, removing the overcomplicated FormGenerator Autosuggest (it's half-bugged, not checking for the input data to be available and not filtering some unusable target projects), adding the required logic for checking migrations on both sides.

/deploy renku=1.0-next renku-graph=development renku-core=v1.0.2

### DIFF
--- a/client/src/api-client/project.js
+++ b/client/src/api-client/project.js
@@ -153,7 +153,13 @@ function addProjectMethods(client) {
             namespace {
               fullPath
             }
-            path
+            path,
+            httpUrlToRepo,
+            userPermissions {
+              adminProject,
+              pushCode,
+              removeProject
+            }
           }
         }
       }`;
--- a/client/src/dataset/Dataset.container.js
+++ b/client/src/dataset/Dataset.container.js
@@ -27,6 +27,8 @@ export default function ShowDataset(props) {
 
   const [datasetKg, setDatasetKg] = useState();
   const [datasetFiles, setDatasetFiles] = useState();
+  const [fetchedKg, setFetchedKg] = useState(false);
+
   const dataset = useMemo(() =>
     mapDataset(props.datasets ?
       props.datasets.find(dataset => dataset.name === props.datasetId)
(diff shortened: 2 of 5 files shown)


ciyer (changes requested):
Code looks good. Thanks for putting this fix together.

When testing the UI, if I actually attempt to add a dataset to a project, I get this error:

image


Is this just a misconfiguration of the deployment?

lorenzo-cavazzi:
That's an error I've never seen.
I think the dataset is this one https://renku-ci-ui-1656.dev.renku.ch/datasets/62b66820055e4a40bdb0d219d73030e6 and it's easy to reproduce by importing it in any other proejct.
I cannot find the reference project `test-2022-01-21-14-11-17` though. That has probably been removed. Maybe that's the reason? 

I guess in that case the dataset should not appear anymore in the search? @SwissDataScienceCenter/renku-graph-maintainers 
Screenshot_20220124_151639

jachro:
> That's an error I've never seen. I think the dataset is this one https://renku-ci-ui-1656.dev.renku.ch/datasets/62b66820055e4a40bdb0d219d73030e6 and it's easy to reproduce by importing it in any other proejct. I cannot find the reference project `test-2022-01-21-14-11-17` though. That has probably been removed. Maybe that's the reason?
> 
> I guess in that case the dataset should not appear anymore in the search? @SwissDataScienceCenter/renku-graph-maintainers Screenshot_20220124_151639

Yes, that's probably what has happened, the project got removed. Generally, acceptance tests removes project at the end of each test and the name of the project suggests that it's a test created project.

ciyer:
I see. So the problem is unrelated to this PR, and should hopefully get fixed when the KG handles deletions. I will approve then!

ciyer (approved):
👍

PR #1659, opened 2022-01-24
Migration simplification

ciyer:
# Description

Simplify the way migration suggestions are presented, showing just one section that explains if the project is compatible with RenkuLab and if the project needs to be migrated.

# Screenshots

## Old, unsupported project

image

https://renku-ci-ui-1659.dev.renku.ch/projects/cr-migration-test/test-7/overview/status

## Old, supported project

image

https://renku-ci-ui-1659.dev.renku.ch/projects/cr-migration-test/test-8/overview/status

## Minor update available, supported project

image

https://renku-ci-ui-1659.dev.renku.ch/projects/cr-migration-test/test-project-internala/overview/status

## Notice for user without sufficient permissions

image

/deploy renku=1.0-next renku-graph=development renku-core=v1.0.2 extra-values=core.sentry.dsn=https://9ba8a200c9374766b1f8e229ab1cf857@sentry.dev.renku.ch/2,core.sentry.environment=enku-ci-ui-1567,ui.sentry.enabled=true,ui.sentry.namespace=renku-ci-ui-1567,ui.sentry.url=https://182290b8e1524dd3b7eb5dd051852f9f@sentry.dev.renku.ch/5


lorenzo-cavazzi (dismissed):
This simplification will be appreciated for sure by the users!
The code is clean and a big thumb up for including the e2e tests :heavy_check_mark: 

P.S. With some projects, I got somewhat unexpected results. This one has project renku version 0.16.2, but metadata v9.
https://renku-ci-ui-1659.dev.renku.ch/projects/lorenzo.cavazzi.tech/1-0-2-non-indexed-kg-with-ds/overview/status
Anyway, clicking on update fixed everything, and for the standard user this probably doesn't look strange at all
Screenshot_20220125_165547

lorenzo-cavazzi:
P.S. I merged `1.0-next` because it contains a fix that prevents failures with one of the Dataset acceptance tests.

ciyer:
Thanks for the review. Yes, there are some projects that are difficult to handle. Most projects have metadata and dockerfile versions in sync, but it is possible to get the metadata and dockerfile versions out of sync, and it isn't clear what we should so the users in these cases. I chose to present it as seen, but I am open alternatives if you have a better idea.

lorenzo-cavazzi:
I tried to play a bit with the data, but I think your solution is the best.
Showing more details would be confusing for the users, and it's probably unnecessary since the `update` button solved all the issues in the cases I tried.

rokroskar:
Thanks for the update to this @ciyer!

* I have some questions about the wording - on the first one (unsupported project) we say

> This project is not compatible with the RenkuLab UI.
You can work with datasets in interactive sessions, but to create or modify datasets, you will need to update the renku version.
The update should take about 30 seconds.

I think there are a few issues here: 

1. it focuses on datasets but other features won't work, e.g. modifying settings. What about "You can still launch interactive sessions, but other functionality will be severely limited."
2. Why say "update should take about 30 seconds" when we can't update the project? 
3. the instructions link is not obviously a link - imho the "these instructions" should be the link and underlined. Instead of having this revealing element, why not just "We cannot update this project automatically, but you can `launch a session` and follow `these instructions` to do so manually."

* Why do we need to differentiate between the second two cases? From the user's point of view aren't they basically the same? "A new version of Renku is available - you should update now to take advantage of new features." 

* On the last one, the "Do you prefer manual instructions" is underlined, but it isn't underlined above. 
* The "ask a maintainer" link is broken
* in the last one, "renku" is bold, but it isn't in the others

rokroskar:
One more nit - why show both "you don't have permission to update this project" _and_ "if you prefer manual instructions' --> if I can't update it I presumably don't care about the instructions?

ciyer:
Thanks for the feedback. You point out several good improvements.

1. The message has been changed to mention general limitations, not just issues with datasets.
2. If manual update is necessary, we show the instructions immediately (no need to expand anything)
3. If auto-update is available, the manual instructions are rendered underlined
4. The "ask a maintainer link" has been fixed, and update instructions are not shown for non-maintainers
5. For a minor update, we show "A new version of renku in bold"

See the updated screenshots. 

> 2. Why say "update should take about 30 seconds" when we can't update the project?

I think it is still interesting to know, even if you cannot carry out the update. That information can be passed on when requesting that the project be updated ("Please update the project, it will only take 30s...")

> * Why do we need to differentiate between the second two cases? From the user's point of view aren't they basically the same? "A new version of Renku is available - you should update now to take advantage of new features."

The two situations are not the same. In the "old, supported project" case, the project is not at the latest major version. It is supported at the moment, but will not be forever, and users should update those projects as soon as is convenient. In the "minor update available" case, the project is reasonably current, and there are some improvements available, but it is not critical that the user update.

Updating a project is potentially disruptive, and I think we should provide information to help users make informed decisions about when they should update. If we are always shouting at users to immediately update after every minor version bump, they will start ignoring us. I think it is better to make clear when an update in pressing, and when it is nice-to-have.

rokroskar:
This looks great. One final thing I noticed - you mentioned you don't want to bombard users with requests to update, but maybe in that case we shouldn't show the yellow warning triangle by the project name in the case of a minor version? I wonder if something more subtle would be possible (like a dot next to the status?) - but we can do that in a later follow-up.

rokroskar (approved):

lorenzo-cavazzi (approved):
Let's merge this :rocket: 

In general, I feel having a short message also for anonymous users wouldn't be bad -- imho the "RenkuLab compatible" message was better than nothing in a case like [the following](https://renku-ci-ui-1659.dev.renku.ch/projects/lorenzo.cavazzi.tech/1-0-2-non-indexed-kg-with-ds/overview/status) (left for logged user, right for anonymous).
But we can still modify it later if it turns out to be confusing for the user :slightly_smiling_face: 

Screenshot_20220131_085601

PR #1663, opened 2022-01-26
Add version to the job APIs URL

lorenzo-cavazzi:
The URL for jobs API doesn't include the version. This causes problems when adding/editing datasets on projects with an older metadata version (specifically, the answer from the job APIs is `undefined` cause it's querying the wrong service).

**How to test**: Use an older project with metadata V8, and try to add a datasets (E.G. clone `ralf.grubenmann/test-8`)

/deploy renku=1.0-next renku-graph=development renku-core=release/v1.0.3
re SwissDataScienceCenter/renku-python#2604


ciyer (approved):
This works, thanks!

PR #1671, opened 2022-01-31
feat(client): handle missing pipelines

### LINKED ISSUE #1667: The check on missing docker image fails for authenticated users
## Description

When a docker image is missing, the user should get a warning when trying to start a session -- either beforehand (if the user is starting it from the session tab) or as a warning popping out later (when using the quick-start button).

This seems to work for anonymous users (the docker registry image is checked), but it's failing for logged users.
This is the case when the response from `GET /projects/<id>/pipelines?sha=<commit-sha>` is an empty array.

It's probably a regression since this was fixed earlier.

lorenzo-cavazzi:
This PR handles missing pipelines by:
* Displaying the correct warning (missing docker image) when trying to start a new session
* Adding a "Build branch image" button.
The latter tries to build an image when there is no pipeline. Mind that it works only per-branch, not per commit https://docs.gitlab.com/ee/api/pipelines.html#create-a-new-pipeline

It also harmonizes the messages and buttons showed on non-available images.

Screenshot_20220131_183027

/deploy renku=1.0-next
fix #1667

### DIFF
--- a/client/src/notebooks/Notebooks.container.js
+++ b/client/src/notebooks/Notebooks.container.js
@@ -26,8 +26,9 @@ import {
   CheckNotebookIcon, ShowSession as ShowSessionPresent
 } from "./Notebooks.present";
 import { StatusHelper } from "../model/Model";
-import { Url } from "../utils/helpers/url";
 import { Loader } from "../utils/components/Loader";
+import { Url } from "../utils/helpers/url";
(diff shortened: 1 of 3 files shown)


ciyer (approved):
This is a good solution to a problem that has been an annoyance for a while now!

I added a commit that changes the wording of the messages a little and swaps the position of the "start with base image" and "start session" buttons.

PR #1673, opened 2022-02-01
Fix problems with relative links

### LINKED ISSUE #1655: relative links (e.g. images and notebooks) not rendering on renkulab READMEs
## Description

When embedding content (e.g. images, notebooks) in READMEs using relative paths, they are not rendered on renkulab despite being displayed normally on gitlab. Images are rendered normally when using absolute paths (URLs).

This happens both on the normal and dev instances.

## Reproduce

1. Create project
2. Add readme
3. Embed  image with relative path: `image`, or embed notebook with relative path: `[notebook](./notebooks/test.ipynb)

## Examples

* [Fails to render on renkulab.io](https://renkulab.io/projects/renku-stories/digirhythm)
  +  Other examples: [MATLAB](https://renkulab.io/projects/learn-renku/working-with-matlab), [cellpose-napari](https://renkulab.io/projects/cyril.matthey-doret/cellpose-napari), [Desktop](https://renkulab.io/gitlab/learn-renku/renku-desktop/-/blob/master/README.md) where `[Part 3.](./notebooks/desktop-icons.ipynb)` redirects to `Cannot GET /ui-server/projects/learn-renku/renku-desktop/files/blob/notebooks/desktop-icons.ipynb`
* [Also fails  on dev instance](https://renku-ci-ui-1567.dev.renku.ch/projects/cyril.mattheydoret/digirhythm-test)
* [Absolute image paths are rendered normally](https://renkulab.io/projects/learn-renku/teaching-on-renku/autograde)

ciyer:
# Testing

Look at this project 

  https://renku-ci-ui-1673.dev.renku.ch/projects/renku-qa/relative-links

Check that the images and links that should work do, and those that should not do not.

Do the same when viewing the README in the file view:

  https://renku-ci-ui-1673.dev.renku.ch/projects/renku-qa/relative-links/files/blob/README.md

And here is an example in a subfolder.

  https://renku-ci-ui-1673.dev.renku.ch/projects/renku-qa/relative-links/files/blob/subfolder/README.md

Fix #1655 

/deploy renku=1.0-next

### DIFF
--- a/client/src/utils/components/markdown/RenkuMarkdownWithPathTranslation.js
+++ b/client/src/utils/components/markdown/RenkuMarkdownWithPathTranslation.js
@@ -6,6 +6,7 @@ import { faFile } from "@fortawesome/free-solid-svg-icons";
 import "katex/dist/katex.min.css";
 import { FilePreview } from "../../../file";
 import { sanitizedHTMLFromMarkdown } from "../../helpers/HelperFunctions";
+import { Url } from "../../helpers/url";
 
 const patterns = {
   fileRefFull: /!\[(.*?)\]\((.*?)\)/g, //with !
@@ -223,7 +224,7 @@ function RenkuMarkdownWithPathTranslation(props) {
   }
 
   const previewLinks = divWithMarkdown.getElementsByTagName("a");
-  const fullBaseUrl = props.client.baseUrl.replace("/api", "/projects")
+  const fullBaseUrl = Url.get(Url.pages.projects)
     + "/" + props.projectPathWithNamespace + "/files/blob/";
 
   for (let link of previewLinks) {
--- a/client/src/utils/components/markdown/RenkuMarkdownWithPathTranslation.js
+++ b/client/src/utils/components/markdown/RenkuMarkdownWithPathTranslation.js
@@ -238,5 +238,14 @@ function RenkuMarkdownWithPathTranslation(props) {
   </div>;
 
 }
-export default RenkuMarkdownWithPathTranslation;
+
+function GuardedRenkuMarkdownWithPathTranslation(props) {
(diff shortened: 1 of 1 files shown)


client/src/utils/components/markdown/RenkuMarkdownWithPathTranslation.js
```diff
@@ -223,7 +223,7 @@ function RenkuMarkdownWithPathTranslation(props) {
   }
 
   const previewLinks = divWithMarkdown.getElementsByTagName("a");
-  const fullBaseUrl = props.client.baseUrl.replace("/api", "/projects")
+  const fullBaseUrl = props.client.baseUrl.replace("/ui-server/api", "/projects")
```
lorenzo-cavazzi, line 226:
```suggestion
  const fullBaseUrl = props.client.uiserverUrl.replace("/ui-server", "/projects")
```
Not a big change tbf since "baseUrl" is manually defined as "uiserverUrl" + "/api" :grin: 

Actually, the best solution would be using the `params["BASE_URL"]` variable that contains the pure UI URL. That would be robust to deploying UI and ui-server to different URLs. It sounds confusing, but `client.baseUrl` uses `params["UISERVER_URL"]`

lorenzo-cavazzi (approved):
Nice fix!

We could also switch to using `params["BASE_URL"]` if we don't want to use `.replace()` anymore. But we can also merge this right away.

client/src/utils/components/markdown/RenkuMarkdownWithPathTranslation.js
ciyer: Yes, I didn't like that bit of code much either. I switched to using `Url.get(Url.pages.projects)`, which is, I think, an even better solution.

lorenzo-cavazzi (approved):

Issue #1665, opened 2022-01-27
Show only a warning when forking fails non-critically


lorenzo-cavazzi:
## Description

Forking actually invokes a few extra APIs (KG indexing, running pipelines, permissions...), and some of them may fail for different reasons.
This should not block the forking process, but just show a warning to the user (and a redirect link). The user is therefore aware something didn't work but he can still visit the forked project in the UI.

This should already be the case, but apparently it doesn't always work.
The relevant code to fix is https://github.com/SwissDataScienceCenter/renku-ui/blob/master/client/src/project/new/ProjectNew.container.js#L257-L280

## Reproduce

Simulate a pipeline error when invoking `client.runPipeline`.
The response should have a `400` error status and the following JSON body
```
{"message":{"base":["Commit not found"]}}
```

Issue #1585, opened 2021-11-24
Project components cleanup


lorenzo-cavazzi:
# Update (March 2022)

We changed the plan for how to approach the component cleanup. Instead of doing small improvements, we want to focus our energy on achieving larger gains. Our plan instead is to:

1. Start using https://redux-toolkit.js.org/ and RTK Query which will standardize fetching data from a server and storing it in Redux; this will handle fetching/loading and error states in a consistent way
2. Use hooks for accessing redux state so we do not need to pass properties through many layers of components
3. Move to TypeScript for components which will, among other things, give us a better handle on the properties used by the components

The first step is to upgrade to React 17, which is done in #1763.

# Original Text

As a follow-up to #1568 , we should harmonize the Project components. Specifically:
- [ ] Remove `transient` from the project state, use `fetching` instead
- [ ] Use `fetching` wherever needed and eliminate the stringy "updating" value (tip: look for `setUpdating` and `SpecialPropVal.UPDATING`)
- [ ] remove the `subComponents` function from the `View` component in the Project.js file; add the single components wherever needed instead of inheriting them as properties.

PR #1674, opened 2022-02-01
fix(client): restore error handling on forks

lorenzo-cavazzi:
It turns out that adding notifications from the Fork modal triggers a re-rendering of the `ProjectView` component. The reason is _very_ tricky: adding a notification changes the project sub-components (like "mrView", "issueView", ...) that are added as properties (:scream: ) to the `View` component.
That seems to affect the (few) components that strongly rely on local data like `ForkProject` -- i.e. the local state gets erased when re-rendering an ancestor component.

I'm not 100% sure when this was introduced, but it will be solved by #1585
For now, disabling the notification on fork is the quickest fix.
This restores the previous warning and "Go to forked project" button when a non-critical error happens during a fork.
The screenshot shows an example as mentioned in #1665 where the pipeline creation fails.
Also, now it's much easier to recover from that state because #1671

image

**How to test**: this requires manually triggering an error. The easiest way would be running telepresence on this PR deployment and messing up the `client.runPipeline` function (e.g. set `defaultBranch` to a random string)

/deploy


ciyer (approved):
I think this is a good temporary fix! We should reactivate the notifications, but that will all be reworked in the future anyway, so let us leave them turned off for the moment. 

I added a commit that made some small changes to the wording of the messages.

Issue #1604, opened 2021-12-09
Create a new project based on a dataset


ciyer:
## Motivation

It is possible to _add_ a dataset to a project, it would be nice to add an option to **create** a new project and add the dataset.

## Proposal

The add button, which currently opens a popup, should change. It should open a **new page** that lets the user either pick an existing project or create a new project, using the new project form.

![image](https://user-images.githubusercontent.com/1196411/148747250-481e66b1-7a20-4682-84df-8c87c0e30bb1.png)

If a new project is created, that should be the first step, the second step is to add the dataset to the project.

- We may want to show some more information about the dataset than just the title in this new form.

lorenzo-cavazzi:
The `AddDataset` button currently uses the FormGenerator for the Autosuggest element.
The logic is a bit overcomplicated since we are not taking advantage of the draft (it's a single value for selecting the target project).
It would make sense to remove the FormGenerator login and to add the Autosuggest element directly there.

We should refine the whole logic by adding the following steps:
- [ ] Verify that the projects list is available before allowing the users to interact with the Autosuggest (use the fetched/fetching variables for member projects).
- [ ] Verify the status of the project containing the dataset:
  - It should exist in GitLab (it may not be the case! datasets are still there even when the underlying project is gone)
  - It should have a supported project metadata version
- [ ] Verify the status of the target project
  - It should have the very same metadata version of the origin project
- [ ] _OPTIONAL_: create a separate GraphQL query to get the member projects with relative permissions, and remove the permissions from the current one. Invoke the new one (with permissions) only when loading the AddDataset component. This would improve the performance quite a bit, since getting the permissions is time-consuming and not necessary anywhere else.

PR #1682, opened 2022-02-04
refactor(client): do not use formGenerator to import the dataset into an existing project (#1604)

andre-code:
This PR refactors the AddDataset component to not use FormGenerator.
This PR also includes:
+ Adding loading animation if the project list is not ready
+ Validation when the dataset project does not exist
+ Validation if the version of the dataset project is not supported
+ Validation if the version of the dataset project and the selected project are not the same

addDatasetToProject

# Cases to test:
- The dataset project does not exist
- The dataset project version is not supported
- The project selected needs migration
- The project selected version is not supported 
-  The project selected version and the dataset project version has a different version
- The dataset is not activated in kg

# Screenshots

**v8 dataset, v9 project**
image

**v8 dataset, v8 project**
image

**v4 dataset, v4 project**

image

/deploy #persist

### DIFF
--- a/client/src/dataset/addtoproject/DatasetAdd.container.js
+++ b/client/src/dataset/addtoproject/DatasetAdd.container.js
@@ -144,9 +144,8 @@ function AddDataset(props) {
       return false;
     }
     // check if the dataset project is supported
-    const projectVersion = checkTarget.result.core_compatibility_status.project_metadata_version;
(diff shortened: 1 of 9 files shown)


ciyer (approved):
Great refactoring and improvements! This makes the import work better, and better positions us to do further improvements.

PR #1684, opened 2022-02-08
Use the new usedIn field to get dataset projects

ciyer:
# Testing

Look at the dataset view on https://dev.renku.ch and see that projects are not displayed there. Compare with the PR deployment were projects are shown.

# Screenshots

image

/deploy #persist


lorenzo-cavazzi (approved):
Nice fix!

PR #1686, opened 2022-02-10
refactor(client): Add dataset data in redux (#1604)

andre-code:
This PR is 2/3 for ticket #1604 

Implemented:

- Refactoring to get the dataset using a DatasetCoordinator class and store the dataset data in a schema (redux).
- A separate page has been added to add a dataset to the project, including a button to select between an existing project and a new project. The new project form is not covered in this PR.

Pending:
- Cypress tests

Cases to test
- Load dataset data from the project
- Load dataset data from general dataset list
- Load dataset files in the dataset view
- Show warning if dataset is not in KG
- Display warning if the dataset does not exist
- Display error when dataset creation fails
- Display dataset data after creating the dataset
- Display dataset information and dataset project validation when importing dataset to an existing project

datasetcoordinator-refactor

/deploy #persist


client/src/dataset/Dataset.container.js
```diff
@@ -16,102 +16,97 @@
...
+        fetchFiles(dataset?.name, props.httpProjectUrl, versionUrl);
+      }
+    }
+  },
+  [
+    dataset, props.httpProjectUrl, props.insideProject, migration.core.backendAvailable,
+    migration.core.fetched, migration.core.versionUrl, props.datasetCoordinator, datasetFiles]
+  );
+
+  const loadingDatasets = dataset === null || dataset?.fetching || !dataset?.fetched;
```
ciyer, line 93: Before this was: ```props.insideProject && datasetFiles === undefined```

Is the new way of calculating loadingDatasets the same, or should it be different?

client/src/dataset/addtoproject/DatasetAdd.present.js
```diff
@@ -26,16 +26,40 @@
...
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faCheck, faExclamationTriangle } from "@fortawesome/free-solid-svg-icons";
 
 import { Loader } from "../../utils/components/Loader";
 import SelectAutosuggestInput from "../../utils/components/SelectAutosuggestInput";
+import { ButtonGroup, Table } from "reactstrap/lib";
+import { getDatasetAuthors } from "../DatasetFunctions";
 
-function ImportDatasetStatus(status, text, existingProject, history) {
+function HeaderAddDataset(dataset) {
```
ciyer, line 39: Why not make this a component? 
```suggestion
function HeaderAddDataset({dataset}) {
```

client/src/dataset/addtoproject/DatasetAdd.present.js
```diff
@@ -118,26 +153,30 @@ function DatasetAdd(props) {
...
-      </ModalBody>
-      <ModalFooter>
-        {closeButton}
-        {addDatasetButton}
-      </ModalFooter>
-    </Modal>
+    <>
+      <Row className="mb-3">
+        <Col sm={10} md={9} lg={8} xl={7}>
+          { header }
```
ciyer, line 174: If HeaderAddDataset is a component, as suggested above, you can do:
```suggestion
          <HeaderAddDataset dataset={props.dataset} />
```

client/src/model/RenkuModels.js
```diff
@@ -403,6 +403,38 @@ const notebooksSchema = new Schema({
...
+      created: { [Prop.INITIAL]: "", [Prop.MANDATORY]: true }, // dataset_core.created_at,
+      published: { [Prop.INITIAL]: null, [Prop.MANDATORY]: true }, //,
+      identifier: { [Prop.INITIAL]: null, [Prop.MANDATORY]: true }, // dataset_core.identifier,
+      keywords: { [Prop.INITIAL]: "", [Prop.MANDATORY]: true }, // dataset_core.keywords,
+      hasPart: { [Prop.INITIAL]: null, [Prop.MANDATORY]: true },
+      mediaContent: { [Prop.INITIAL]: null, [Prop.MANDATORY]: true }, // dataset_core.mediaContent
+      url: { [Prop.INITIAL]: null }, // dataset_kg.url;
+      sameAs: { [Prop.INITIAL]: null }, // dataset_kg.sameAs;
+      usedIn: { [Prop.INITIAL]: null }, // dataset_kg.usedIn;
+      insideKg: { [Prop.INITIAL]: null }, //  true;
```
ciyer, line 421:
```suggestion
      insideKg: { [Prop.INITIAL]: null }, //  true if dataset is in the KG;
```

ciyer (changes requested):
Wow, this is a really substantial refactoring that brings many needed improvements!


In testing, I noticed something.

If you go to this dataset, which is in a project, but not in the KG:

https://renku-ci-ui-1686.dev.renku.ch/projects/cramakri/dataset-test-project/datasets/more-test-datasets/

And click on the "Add to project" button, you get this view:

image

if you refresh the page, then there is a crash and white screen.

The problem is that the "Add dataset to project" only works for datasets that are in the knowledge graph.

This is what you see for the same dataset on dev, and I think this should be maintained:

image

The "Add to project" button is disabled, and there is a warning below that the dataset is not in the KG.

client/src/dataset/addtoproject/DatasetAdd.present.js
andre-code: no reason to don't be a component. I updated in the last commit

client/src/dataset/addtoproject/DatasetAdd.present.js
andre-code: sure. updated to be component :)

andre-code:
Thanks @ciyer  for taking a look at those cases and reporting them.

For case #1, the validation was missing to not have the button enabled if KG was not active so I updated it in the last commit and also added a Not Found warning when the dataset is not active since that is the error received from KG.
For case #2, I could not find the dataset in KG, I got this error on dev.renku.ch
https://dev.renku.ch/ui-server/api/kg/datasets/a78c4509-012e-49e1-a076-8d84695b387c

You can also find in the last commit the refactor for DatasetAdd component

client/src/dataset/Dataset.container.js
andre-code: yes, this  loadingDataset was used for displaying file purposes but in the new refactor I decide to complement it with  I  `props.files?.fetching || props.loadingDatasets` to be more accurate if is loading datasets or files

ciyer (dismissed):
This is great! The problem with the datasets in older projects that are in the KG is due to a change in the format of dataset identifiers, and we will fix that in #1694. 

I added a small commit to fix button sizes on the dataset page:

image

(was like this:
image
)

e2e/cypress/fixtures/datasets.json
```diff
@@ -0,0 +1,182 @@
+[
```
ciyer, line 1: I think we should use responses created on `dev.renku.ch` and with the e2e test user. Then it is easier for any team member to make changes to the fixtures.

e2e/cypress/fixtures/datasets/dataset_3117816964734179b3f3ccef5dcd43d2.json
```diff
@@ -0,0 +1,76 @@
+{
```
ciyer, line 1: Same comment applies here: I think it would be better to use `dev.renku.ch` and the e2e test user for generating the fixtures.

e2e/cypress/fixtures/datasets/dataset_4577b68957b7478bba1f07d6513b43d2.json
```diff
@@ -0,0 +1,76 @@
+{
```
ciyer, line 1: Same comment applies here: I think it would be better to use `dev.renku.ch` and the e2e test user for generating the fixtures.

e2e/cypress/fixtures/datasets/dataset_897ee9effb9b42efab454011a8fba5e4.json
```diff
@@ -0,0 +1,66 @@
+{
```
ciyer, line 1: Same comment applies here: I think it would be better to use `dev.renku.ch` and the e2e test user for generating the fixtures.

e2e/cypress/fixtures/datasets/dataset_a20838d8cd514eaab3efbd54a8104732.json
```diff
@@ -0,0 +1,80 @@
+{
```
ciyer, line 1: Same comment applies here: I think it would be better to use `dev.renku.ch` and the e2e test user for generating the fixtures.

e2e/cypress/fixtures/datasets/dataset_c23f114791984453afdffc79ea5b4033.json
```diff
@@ -0,0 +1,95 @@
+{
+  "_links":[
```
ciyer, line 2: Same comment applies here: I think it would be better to use `dev.renku.ch` and the e2e test user for generating the fixtures.

e2e/cypress/fixtures/datasets/datasets_search_3117816964734179b3f3ccef5dcd43d2.json
```diff
@@ -0,0 +1,33 @@
+[
```
ciyer, line 1: Same comment applies here: I think it would be better to use `dev.renku.ch` and the e2e test user for generating the fixtures.

e2e/cypress/support/renkulab-fixtures/projects.ts
```diff
@@ -41,44 +41,55 @@ function Projects<T extends FixturesConstructor>(Parent: T) {
...
+    project(useFixture = true, path = "", name = "getProject", result = "projects/project.json") {
+      const fixture = useFixture ? { fixture: result } : undefined;
+      cy.intercept(
+        `/ui-server/api/projects/${encodeURIComponent(path)}?statistics=true`,
+        fixture
+      ).as(name);
+      return this;
+    }
+
+    interceptMigrationCheck(name, fixture, params) {
```
ciyer, line 53: I suggest changing the signature of interceptMigrationCheck to 
```suggestion
    interceptMigrationCheck(name, fixture, params=null) {
```
Then the params could be defaulted if not provided. Otherwise it has to be passed in through a long call-chain.

e2e/cypress/support/renkulab-fixtures/projects.ts
```diff
@@ -41,44 +41,55 @@ function Projects<T extends FixturesConstructor>(Parent: T) {
...
-      const params =
-        "git_url=https%3A%2F%2Fdev.renku.ch%2Fgitlab%2Fe2e%2Flocal-test-project.git&branch=master";
       cy.intercept(`${coreUrl}?${params}`, {
         fixture: fixture
       }).as(name);
       return this;
     }
 
-    projectMigrationUpToDate(name = "getMigration") {
+    projectMigrationUpToDate(params, name = "getMigration") {
```
ciyer, line 61: If we do this a little differently, it would be easier to override the function parameters.

```suggestion
    projectMigrationUpToDate(params={repoUrl: null, name: "getMigration"}) {
      const {
        repoUrl,
        name
      } = params;
```

e2e/cypress/support/renkulab-fixtures/projects.ts
```diff
@@ -41,44 +41,55 @@ function Projects<T extends FixturesConstructor>(Parent: T) {
...
       this.interceptMigrationCheck(
         name,
-        "test-project_migration_up-to-date.json"
+        "test-project_migration_up-to-date.json",
+        params
```
ciyer, line 65: Then, continuing, we have
```suggestion
        "test-project_migration_up-to-date.json",
        repoUrl
```

e2e/cypress/integration/local/project.spec.ts
```diff
@@ -133,14 +136,15 @@ describe("display migration information for anon user", () => {
 
 describe("display migration information for observer user", () => {
   const fixtures = new Fixtures(cy);
+  const params = "git_url=https%3A%2F%2Fdev.renku.ch%2Fgitlab%2Fe2e%2Flocal-test-project.git&branch=master";
```
ciyer, line 139: I think making this a overridable is a good idea and gives us greater flexibility, but I think it would be nicer to have it assigned to a default value if none is specified. Instead of putting this here, I think it should be defaulted in the fixture (see the suggestions for `fixtures/projects.ts`)

e2e/cypress/integration/local/project.spec.ts
```diff
@@ -133,14 +136,15 @@ describe("display migration information for anon user", () => {
...
   beforeEach(() => {
     fixtures.config().versions().userTest();
     fixtures.projects().landingUserProjects().projectTestObserver();
     cy.visit("/projects/e2e/local-test-project");
   });
 
   it("displays up-to-date migration", () => {
-    fixtures.projectMigrationUpToDate();
+    fixtures.projectMigrationUpToDate(params);
```
ciyer, line 147: With the changes suggested to `fixtures/projects.ts`, this can be changed back to:
```suggestion
    fixtures.projectMigrationUpToDate();
```

e2e/cypress/integration/local/projectDatasets.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+  const projectPath = "e2e/testing-datasets";
+
+  beforeEach(() => {
+    fixtures.config().versions().userTest();
+    fixtures.projects().landingUserProjects();
+    fixtures.project(useMockedData, projectPath);
+    fixtures.projectKGDatasetList(useMockedData, projectPath);
+    fixtures.projectDatasetList(useMockedData);
+    fixtures.projectTestContents(undefined, 9);
+    fixtures.projectMigrationUpToDate("*");
```
ciyer, line 36: With the suggestions to how to change the fixtures, this would become:
```suggestion
    fixtures.projectMigrationUpToDate({repoUrl: "*"});
```

e2e/cypress/integration/local/datasets.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+import Fixtures from "../../support/renkulab-fixtures";
+import "../../support/datasets/gui_commands";
+
+describe("display a dataset", () => {
+  const fixtures = new Fixtures(cy);
+  const useMockedData = Cypress.env("USE_FIXTURES") === true;
+  beforeEach(() => {
+    fixtures.config().versions().userTest();
+    fixtures.projects().landingUserProjects();
+    fixtures.datasets(useMockedData);
```
ciyer, line 30: Would it make sense to make `useMockedData` a property of the Fixtures object? 

Making it a parameter to the fixture is better if we want to mix and have some mocked and and some non-mocked fixtures within a test. But if it is an all/nothing decision, then I think it would be better to make it a field on Fixtures.

ciyer (changes requested):
This is becoming a bit of a monster, but I think it is a great PR that bring many improvements to the code!

There is one incorrect line in handling project versions (which we talked about). That should be fixed. 

I like the cypress tests a lot. I have a few suggestions for changes there that I think will improve the code a bit, but if I misunderstood the motivations, then those might not be so valid. Take a look and let me know if you disagree with any of them.

e2e/cypress/integration/local/datasets.spec.ts
andre-code: Good point. The initial idea is to have an all/nothing decision, but also keeping it flexible is a good idea for future scenarios.  Including that property in the Fixture class makes it clean for all fixtures and easy to handle for current and new cases. thx for the suggestion

andre-code:
Thanks @ciyer for reviewing my little big monster 😅  🙌.
I liked your suggestions, those clean up the code more, here is a recount of what I have done in the last commit:
- Updated the conditional version of the metadata to cast it to numbers.
- Updated the fixture files to use only data coming from dev.renku.ch
- Include useMockedata in Fixture class, all fixtures can use it without including it as a parameter 
- Add default value for query params for migration checking and create an interface in case we want to modify that default value.

It is ready for u review ✌️

ciyer (approved):
Looks great! Let us merge this 🎉

PR #1700, opened 2022-02-17
fix: update readthedocs links to work again

noeleont:
I tried to fix all the wrong links. Please check that I referenced the correct topics. 

/deploy


noeleont:
Thanks for your feedback! As an improvement we could also define a dictionary under `src/utils/constants` for all these links instead of hard coding them into the js files. What do you think?

ciyer:
> Sure! I’ll open another PR with the changes, do you have a branch naming convention?

Sorry, just realizing that you were waiting for an answer from me: feel free to name the branch however you wish.

PR #1708, opened 2022-02-23
feat(client): show recent projects (#1628)

### LINKED ISSUE #1628: Show recently accessed projects
# Description

Display recently accessed projects. See  #1626.

- User has no recent project, then show the same as the "your projects"
- The user has no project at all, then show what we currently show for that situation

## Implementation

- Implement cypress tests for this

andre-code:
### Description
This PR include visualization of the last visited projects on the landing page (only valid projects), in case there is no last projects it displays the user projects (previous behavior)

Screenshot 2022-02-23 at 11 49 24

### Testing
Visit different projects owned or not and then go to the landing page, those projects should appear under the title **Recent Projects** in order of last visited first.

Fix #1628 


andre-code:
PR ready for review, it includes already e2e tests ✌️

client/src/project/shared/Projects.state.js
```diff
@@ -103,24 +103,48 @@ class ProjectsCoordinator {
...
+          for (const result of results) {
+            if (result?.status === "fulfilled" && result?.value?.data?.all)
+              projectList.push(this._starredProjectMetadata(result?.value?.data.all));
+          }
+          this._setLandingProjects(projectList, true);
+        }).catch( () => {
+          this.model.set("landingProjects.fetching", false);
+        });
+      }
+      else {
```
lorenzo-cavazzi, line 141:
```suggestion
      if (!lastProjectsVisited?.length || !projectList.length) {
```

It's possible the projects previously visited have been deleted or otherwise not available (permissions, ...).
It's a rare case, but the message would be confusing for someone with some projects (see screenshot).
I wouldn't put too much effort into solving this case in a very elegant way.
Checking if at least one valid project was fetched from the list should be enough (the "recent projects" still flashes for a second or two, but it's a rare case, I think it's fine)

Screenshot_20220225_101857

lorenzo-cavazzi (changes requested):
Great PR! And thumb up for the e2e tests!
I identified a corner case we could handle. It's probably rare, but we could introduce a 1-line fix to prevent it.
Otherwise, this looks good to me :rocket:

lorenzo-cavazzi (approved):

ciyer (dismissed):
Very nice! This makes the landing page much easier to use and much more useful!

ciyer (approved):

ciyer (approved):

PR #1711, opened 2022-02-25
feat(client): add project visibility in card list (#1637)

### LINKED ISSUE #1637: Display project visibility in project list
projects

### sugestion

Display current visibility labels in the project list page

Screenshot 2022-01-13 at 11 10 36
Screenshot 2022-01-13 at 11 10 27
Screenshot 2022-01-13 at 11 10 18

## Implementation

Use the symbols we have (lock, globe, etc) and show them in the dot that is already there. Add a tooltip to explain that the project is private, internal, etc.

andre-code:
### Description
This PR adds the visibility icon for projects in the card list.

projectVisibility

Fix #1637 


client/src/utils/components/List.js
```diff
@@ -20,10 +20,22 @@ import React, { Fragment } from "react";
...
 import { ProjectTagList } from "../../project/shared";
 
+const VisibilityIcon = ({ visibility }) => {
+  const icon = {
+    public: <FontAwesomeIcon icon={faGlobe} />,
+    private: <FontAwesomeIcon icon={faLock} />,
+    internal: <FontAwesomeIcon icon={faUserFriends} />
+  };
+  return <span className="fs-small text-rk-text" title={visibility}>{ icon[visibility] || "" }</span>;
+};
```
lorenzo-cavazzi, line 37: The `fs-small` class seems to make the icons *very* small. I'm not sure if it doesn't consider the relative size of the text, or maybe it's because the text is later placed in a `<div>` instead of a text element like `<h4>` or `<p>`.

Anyway, we could allow personalizing the size using the standard sizing utilities instead of the custom class `fs-small`

```suggestion
const VisibilityIcon = ({ visibility, size = "xs" }) => {
  const icon = {
    public: <FontAwesomeIcon icon={faGlobe} size={size} />,
    private: <FontAwesomeIcon icon={faLock} size={size} />,
    internal: <FontAwesomeIcon icon={faUserFriends} size={size} />
  };
  return <span className="text-rk-text" title={visibility}>{ icon[visibility] || "" }</span>;
};
```

lorenzo-cavazzi: P.S. give the code a try first, I'm not 100% it works correctly :grin:

lorenzo-cavazzi (dismissed):
Nice addition!
I would use the standard `size` properties instead of the custom `fs-small` class on the icons, since that seems to not consider the relative sizes of the `<div>`.

ciyer (approved):
I think this is nicely done! I added a small commit that makes some minor visual tweaks.

1. Make the icons a little larger and display them next to the dots
image

2. Show the visibility information underneath the dot in the list view

image

I think this is a UX improvement since the visibility is now always in a fixed location within the card/row, not in a position dependent on the width of the label, and this should address Lorenzo's concern about size as well.

PR #1713, opened 2022-02-25
feat(client): show session logs for sub containers

### LINKED ISSUE #1647: Show session logs for sub-containers
So far, we get all the session logs together. They are not exhaustive tho, since the sub-containers are missing, and usually they provide the most important logs for debugging when something goes wrong during the startup.

With SwissDataScienceCenter/renku-notebooks#869 , the API response for `GET /notebooks/logs` will change to include all the relevant logs:
```
{
  "container1": [
      "line 1 of log",
      "line 2 of log",
      "..."
   ],
   "container2": [
      "line 1 of log",
      "line 2 of log",
      "..."
   ],
}
```

We should verify if it's better to keep the logs separate, or to show a single log.
In the first case, it could be something like this (use a better visualization than the standard list-group), where the most important container is automatically selected. We expect them to generally have the same names, so we could re-write them in a more user-friendly name (e.g. `git-clone` --> `Git clone`, or even `Setup container`)

Screenshot_20220118_114227

andre-code:
### Description
This PR is to add the display of all session logs and allows to download it in a zip file.

**Note** Appearance changed slightly

renku-session-log

Closes #1647 

/deploy renku=000-acceptance-tests renku-notebooks=get-all-session-logs  #persist


olevski:
I did a quick test of the feature. Looks good to me @andre-code

olevski (approved):

olevski:
Also I think the best way to tackle this is to merge this and the dependent PR on the notebooks at the same time, make releases for notebooks and the UI at the same time and then bump the requirements.yaml in the renku repo in one single PR for both notebooks and UI.

ciyer (dismissed):
This is a very useful feature and nicely presented. I added a commit that introduces some minor cosmetic changes to use the standard background and tab styles in the popup.

client/src/notebooks/Notebooks.css
```diff
@@ -14,6 +14,10 @@
   font-style: italic;
 }
 
+li.nav-item {
+  cursor: pointer;
+}
```
lorenzo-cavazzi, line 19: Mind that this will impact **all** the elements application-wide, not only in the notebooks components -- although I believe the default for the `.nav-item` is already a pointer.

If you only need to change the behaviour for the logs tab, you could create a new class.

lorenzo-cavazzi (changes requested):
Nice solution! The zipped file for the download is a great addition!

I added a comment inline, not sure if it's relevant (feel free to ignore it if not :slightly_smiling_face:  ).

Mind that we show the logs also in a tab on the session page. They seem to be gone from there. In the screenshot, on the left it's the current PR (no logs), while on the right it's the old behavior.

Screenshot_20220309_134905

andre-code:
🙌 thx @lorenzo-cavazzi , I am happy for the zip file addition 💙 😅

And also thanks for noticing that the logs were also shown in the session view, I included them in the last commit. ✌️

lorenzo-cavazzi:
Thanks for fixing it!
I noticed the view is a bit off on that page. On both firefox and chrome , I get scroll bars halfway, as if the height was fixed (see screenshot). Could you check that?
P.S. if it's no low effort, it would be nice to add the download logs button also there, next to the "Refresh logs" button on the top.

Screenshot_20220310_092419

PR #1716, opened 2022-02-28
feat: strip '-' from dataset identifiers

### LINKED ISSUE #1694: Remove "-" from dataset identifiers in responses from api version < 9
# Description

Before v9 metadata, dataset identifiers contained "-" characters. Because the KG is always using v9 metadata, older datasets have incompatible identifiers.

# Fix

In `api-client/dataset.js > listProjectDatasetsFromCoreService` we should remove "-" from the identifiers of the results.

ciyer:
# Testing

This PR fixes the dataset identifiers returned from the core service to match the identifiers in the KG.

To test, look at a dataset in an older project. Before this fix, the dataset is not recognized as being in the KG, since the identifiers do not match, so the dataset cannot be added to projects and there is a warning. With the fix, the dataset is found in the KG.

**Without fix**
https://dev.renku.ch/projects/cramakri/test-8-cloned/datasets/v8-dataset/

image

**With fix**
https://renku-ci-ui-1716.dev.renku.ch/projects/cramakri/test-8-cloned/datasets/v8-dataset/

image

Fix #1694

### DIFF
--- a/client/src/api-client/dataset.js
+++ b/client/src/api-client/dataset.js
@@ -14,6 +14,16 @@ function addMarqueeImageToDataset(gitUrl, dataset) {
   return dataset;
 }
 
+/**
+ * Remove dashes from the dataset identifier
+ * @param {object} dataset
+ */
+function cleanDatasetId(dataset) {
+  if (dataset.identifier)
+    dataset.identifier = dataset.identifier.replace(/-/g, "");
+  return dataset;
+}
+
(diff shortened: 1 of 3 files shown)


lorenzo-cavazzi (approved):
Nice fix! :rocket:

andre-code (approved):
🚀

lorenzo-cavazzi (approved):

PR #1718, opened 2022-03-02
feat(client): use simplified session state (#1646)

### LINKED ISSUE #1646: Use the simplified state object for existing sessions
The status information for the sessions we get from the `GET /servers` API is going to be simplified, making it more meaningful and clean.
We should adapt the UI once SwissDataScienceCenter/renku-notebooks#809 is done

In short, the new status property will look like this:
```
"status": {
	"state": "single string one of running, starting, stopping, failed",
	"message": "single string"
}
```

For `state`, we may be able to change just `NotebooksHelper.getStatus` -- consider removing the function if not useful anymore.
The `message` is probably already correctly used, but in combination with other properties that are going to be removed (like `step` and `reason`. e.g. look for `details.reason`).

## INFO

Use this PR from renku notebooks SwissDataScienceCenter/renku-notebooks#898

andre-code:
This PR is for refactoring the session status display with the changes in renku-notebooks [here](https://github.com/SwissDataScienceCenter/renku-notebooks/pull/898)

The session status changed from:
````
status: {
    message: "containers with unready status: [jupyter-server oauth2-proxy git-proxy]" // string
    phase: "Pending", // Pending, Running, Error
    ready: false
    reason: "ContainersNotReady"
    step: "Ready"
}
```` 
To:
````
status: {
   message: "Containers with non-ready statuses: git-proxy, jupyter-server, oauth2-proxy, init-certificates, git-clone." // string or undefined
   state: "starting" // starting, running, stopping, failed
}
````
status session

### Testing
Error case: https://renku-ci-ui-1718.dev.renku.ch/projects/tasko.olevski/test-failing-project-1/sessions/new?autostart=1
error session status

Closes #1646 

/deploy renku-notebooks=simplify-session-status renku=000-acceptance-tests #persist

### DIFF
--- a/client/src/notebooks/Notebooks.state.js
+++ b/client/src/notebooks/Notebooks.state.js
@@ -443,11 +443,12 @@ class NotebooksCoordinator {
     return this.client.getNotebookServers(
       filters.namespace, filters.project, filters.branch, null, anonymous)
       .then(resp => {
-        let updatedNotebooks = { fetching: false, fetched: new Date() };
+        let updatedNotebooks = { fetching: false };
         // check if result is still valid
         if (!this.model.get("filters.discard")) {
           const filters = this.getQueryFilters();
           if (this.model.get("notebooks.lastParameters") === JSON.stringify(filters)) {
+            updatedNotebooks.fetched = new Date();
             const currentServers = this.model.get("notebooks.all");
             if (!(_.isEqual(resp.data, currentServers)))
               updatedNotebooks.all = { $set: resp.data };
--- a/client/src/notebooks/Notebooks.state.js
+++ b/client/src/notebooks/Notebooks.state.js
@@ -25,7 +25,7 @@
 
 import { API_ERRORS } from "../api-client/errors";
 import { parseINIString } from "../utils/helpers/HelperFunctions";
-
+import _ from "lodash";
 const POLLING_INTERVAL = 3000;
 const IMAGE_BUILD_JOB = "image_build";
 const RENKU_INI_PATH = ".renku/renku.ini";
@@ -63,26 +63,6 @@ const ExpectedAnnotations = {
 };
 
 const NotebooksHelper = {
-  /**
-   * Compute the status of a notebook
-   *
-   * @param {Object} data - either the notebook or the notebook.status object as returned
-   *   by the GET /servers API
-   */
-  getStatus: (data) => {
-    let status = data;
-    if (data.status)
-      status = data.status;
-
-    if (status.ready)
-      return "running";
-    if (status.stopping)
-      return "stopping";
-    if (status.step === "Unschedulable")
-      return "error";
-    return "pending";
-  },
-
   /**
    * Add missing annotations from the notebook servers
    *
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Review by ciyer — DISMISSED** [review]
This works very well! I think this is an improvement to the interaction and way of presenting the information.

**Review by olevski — DISMISSED** [review]
This is great @andre-code!

Also the error shown when a session fails will improve further when we merge https://github.com/SwissDataScienceCenter/renku-notebooks/pull/923

Basically after this PR is merged I am pretty sure that both https://github.com/SwissDataScienceCenter/renku-notebooks/pull/923 and https://github.com/SwissDataScienceCenter/renku-notebooks/pull/898 can be merged. https://github.com/SwissDataScienceCenter/renku-notebooks/pull/898 is more directly tied to this PR and that is what you tested against.

**Comment by olevski** [comment]
p.s. I am ready to merge the corresponding stuff on the notebook side and tag a release. But I would like to do it at the same time as when you release this in the UI. When do you plan on doing this?

**Comment by andre-code** [comment]
> p.s. I am ready to merge the corresponding stuff on the notebook side and tag a release. But I would like to do it at the same time as when you make a new release for the UI. When do you plan on doing this?

Hi @olevski thanks for your review.
Hopefully the day is today 🎉
Once passed the acceptance test I will merge this Pr and #1713 and tag a release.

File: client/src/notebooks/Notebooks.state.js [hunk 1]
```diff
@@ -463,26 +443,31 @@ class NotebooksCoordinator {
...
         }
-        this.model.setObject({ notebooks: updatedNotebooks });
+        if (updateFullObject) {
+          this.model.setObject({ notebooks: updatedNotebooks });
+        }
+        else {
+          this.model.set("notebooks.fetching", false);
+          if (this.model.get("notebooks.fetched") === null)
+            this.model.set("notebooks.fetched", new Date());
+        }
```
lorenzo-cavazzi, line 470: It's better to avoid using multiple `this.model.set` and use a single `this.model.setObject` to prevent information being updated in different transactions.

Here, I would always set both `fetching` and `fetched` cause this ends the fetching phase and the data are actually up-to-date (even if no notebook was actually changed in the store)

lorenzo-cavazzi (changes requested):
I've found a small detail to fix when saving the notebook state. In some cases, the `fetching` variable isn't falsified after finishing the fetch operation (see inline).

Should we make the "Error" message and icon red in this case?
Screenshot_20220309_114616

More in general, I've noticed that, after starting a new session, when the status is initially marked as "running" the notebook is still not reachable for a few seconds. Opening the iframe results in a 503 error -- it happens to me very often when using the autostart button.
After a few seconds, the session is actually available, but the user needs to manually refresh the page.
@olevski is it possible the notebook is marked as ready when it's not yet ready?

olevski:
> @olevski is it possible the notebook is marked as ready when it's not yet ready?
I guess it is. I have not noticed this. We have a readiness probe that will essentially disable ingress into the sessions before passes the probe - this results in a 503. I think though that the status of the session should not conflict with this. But let me check.

olevski:
I see what you mean with the 503 error. It seems that k8s happily sets all container statuses to running even though the readiness probes are not fully cleared. There is an easy way to figure this out - just by sending a request to where you are supposed to access the session. But ideally I can get this info from the manifest itself and not send a full http request on every GET request I get from the UI. I am looking into this rn.

olevski:
So with the edits on the notebook service the 503 error is gone.

But one think I notice is that with this version of the UI I see sessions from other projects when I switch projects. Or at least they kind of flash in and out.

For example:
- launch a session in project 1
- navigate to project 2 (where there are no sessions)
- go to see the sessions for project 2
- here you will see the session from project 1 for a few seconds before it disappears

I am pretty certain that this is not because of the notebooks. When I look at the network requests that are sent and then received the notebook service sends the right responses.

Could it be that the state in the UI is getting out of sync or smth?

andre-code:
> So with the edits on the notebook service the 503 error is gone.
> 
> But one think I notice is that with this version of the UI I see sessions from other projects when I switch projects. Or at least they kind of flash in and out.
> 
> For example:
> 
> * launch a session in project 1
> * navigate to project 2 (where there are no sessions)
> * go to see the sessions for project 2
> * here you will see the session from project 1 for a few seconds before it disappears
> 
> I am pretty certain that this is not because of the notebooks. When I look at the network requests that are sent and then received the notebook service sends the right responses.
> 
> Could it be that the state in the UI is getting out of sync or smth?

Thanks @olevski  for noticing, it was indeed a change introduced in the UI, I have already corrected it in the last commit.
And thanks for fixing the 503 error in record time ⚡️ 🎉.

lorenzo-cavazzi (approved):
Lgtm! :rocket:

PR #1722, opened 2022-03-08
fix(client): adjust project name

### LINKED ISSUE #1685: Adjust the project name after project creation
## Description

The project name is not set properly when creating a new project.
This happens because the renku-core API creates the project through a `git push` to GitLab, and some properties cannot be set through that. As a result, the GitLab project name is the project path.

## How to fix

After creating the project, we should immediately update the project name (if different from the project path) with the one specified by the user.
We already do something similar for the project visibility.

Info: we can easily verify if desired name and the **effective** path by checking the project creation API response. The path is returned as `slug`.

```
{
  "result": {
    "name": "test with spaces 4",
    "url": "https://dev.renku.ch/gitlab/lorenzo.cavazzi.tech/test-with-spaces-4",
    "project_id": "0436a56f63f245fe971e6484d6e1a474",
    "namespace": "lorenzo.cavazzi.tech",
    "slug": "test-with-spaces-4"
  }
}
```

## Reproduce

Try to create a project with spaces. The final name in the UI will contain only `-`.
Doing that directly on GitLab preserves the names.

andre-code:
This PR adjust the project name if it contains special characters after the project is created

Fix #1685


ciyer (commented):
This looks good! There seems to be a problem with the tests, once that is fixed, I it is ready to approve/merge.

ciyer (approved):

PR #1723, opened 2022-03-08
fix: do not show project star status until it is available

### LINKED ISSUE #1657: Project star button has a few glitches
## Description

The project start doesn't seem to show a filled star when the project is starred
image

Caveat: it does fill when manually starring the project, but it doesn't stay after a refresh
image

Also, unstarring doesn't work as expected if it's done as soon as the UI loads. It's necessary to wait 10+ seconds. Otherwise, it tries to star again.
image

These are probably caused by using the data coming from "projects" instead of "project", and it doesn't consider the value of fetcheing/fetched. 

## Reproduce

1. Try to star a project, then reload the page
2. Refresh the UI on a starred project, then try to unstar immediately

ciyer:
# Description

The project star status is not provided by the project endpoint, this means there is a delay between when the most of the data to display a project is available and when the project star status is available.

Solution: show a loader and block interaction with the star button until the star status has been retrieved.

# Test

1. Go to any project and try to star it before the star status is available. 
2. Star a project, do a browser reload on the page and try to unstar the project before the star status has been retrieved.

Fix #1657 


lorenzo-cavazzi (approved):
Great solution!

PR #1725, opened 2022-03-09
feat(client): redirect Notebook connect button to session (#1613)

### LINKED ISSUE #1613: Notebook connect button improvements
# Description

The button to connect to a notebook takes the user to a session in a separate window.

image

Instead, it should take the user to the iframe-based session.

andre-code:
This PR redirect Notebook connects button to renku session link in the same window.

notebook connect

### Testing
1. Launch a session with Notebooks
e.g. https://renku-ci-ui-1725.dev.renku.ch/projects/dandrea.cordoba/essm-egu-2022/sessions/new?autostart=1
2. Go to Notebook file
e.g. https://renku-ci-ui-1725.dev.renku.ch/projects/dandrea.cordoba/essm-egu-2022/files/blob/notebooks/Priestley-Taylor_equation.ipynb
3. Click in connect button, it should display the session Iframe and the link to go back to the file: 

connect button

Closes #1613 

/deploy renku=000-acceptance-tests renku-notebooks=get-all-session-logs #persist

### DIFF
--- a/client/src/notebooks/NotebookStart.present.js
+++ b/client/src/notebooks/NotebookStart.present.js
@@ -1182,10 +1182,15 @@ class CheckNotebookIcon extends Component {
     if (notebook) {
       const status = notebook.status?.state;
       if (status === SessionStatus.running) {
+        const annotations = NotebooksHelper.cleanAnnotations(notebook.annotations, "renku.io");
+        const sessionUrl = Url.get(Url.pages.project.session.show, {
+          namespace: annotations["namespace"],
+          path: annotations["projectName"],
+          server: notebook.name,
+        });
         tooltip = "Connect to JupyterLab";
         icon = (<JupyterIcon svgClass="svg-inline--fa fa-w-16 icon-link" />);
-        const url = `${notebook.url}/lab/tree/${this.props.filePath}`;
-        link = (<a href={url} role="button" target="_blank" rel="noreferrer noopener">{icon}</a>);
+        link = <Link to={{ pathname: sessionUrl, state: { from: location.pathname } }} >{icon}</Link>;
       }
       else if (status === SessionStatus.starting || status === SessionStatus.stopping) {
         tooltip = status === SessionStatus.stopping ?
--- a/client/src/notebooks/NotebookStart.present.js
+++ b/client/src/notebooks/NotebookStart.present.js
@@ -1173,7 +1173,7 @@ class AutosavedDataModal extends Component {
 // * CheckNotebookIcon code * //
 class CheckNotebookIcon extends Component {
   render() {
-    const { fetched, notebook, location } = this.props;
+    const { fetched, notebook, location, filePath } = this.props;
     const loader = (<span className="ms-2 pb-1"><Loader size="19" inline="true" /></span>);
     if (!fetched)
       return loader;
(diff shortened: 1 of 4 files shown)


client/src/project/Project.present.js
```diff
@@ -1052,34 +1052,37 @@ class ProjectViewFiles extends Component {
   }
 }
 
+const ProjectSessions = (props) => {
+  const history = useHistory();
```
lorenzo-cavazzi, line 1056: This is good in general, but you already have `props.history` here

client/src/project/Project.present.js
```diff
@@ -1052,34 +1052,37 @@ class ProjectViewFiles extends Component {
...
+  const locationFrom = history?.location?.state?.from;
+  const backButtonLabel = locationFrom === "notebookFile" ? "Back to notebook file" : "Back to sessions list";
+  // use goBack if the previous url is from Project File
+  const backButton = (<GoBackButton
+    label={backButtonLabel} url={props.notebookServersUrl} useGoBack={!!locationFrom} />);
```
lorenzo-cavazzi, line 1061: I like the approach here!
What about simplifying by adding the return link directly into `from`, and using it as `url` instead of changing the logic of `GoBackButton` to fit in `history`?

client/src/notebooks/NotebookStart.present.js
```diff
@@ -1166,10 +1166,15 @@ class CheckNotebookIcon extends Component {
...
+        const sessionUrl = Url.get(Url.pages.project.session.show, {
+          namespace: annotations["namespace"],
+          path: annotations["projectName"],
+          server: notebook.name,
+        });
         tooltip = "Connect to JupyterLab";
         icon = (<JupyterIcon svgClass="svg-inline--fa fa-w-16 icon-link" />);
-        const url = `${notebook.url}/lab/tree/${this.props.filePath}`;
-        link = (<a href={url} role="button" target="_blank" rel="noreferrer noopener">{icon}</a>);
+        link = <Link to={{ pathname: sessionUrl, state: { from: "notebookFile" } }} >{icon}</Link>;
```
lorenzo-cavazzi, line 1177: This is the right approach, but it may require including extra information in the URL (or even better saving it somewhere else, maybe adding a `filePath` state) to preserve the file path.
The notebook URL used for the iframe should then include a prefix `/tree/${this.props.filePath}` to open directly the notebook in the running session

lorenzo-cavazzi (changes requested):
Nice work! I think it requires a couple of changes before merging:
1. The file information should be preserved and used to open the target notebook in the relevant session.
2. The modifications in `GoBackButton` may not be necessary. Saving the original URL in the state instead of a fixed string may be easier to handle and maintain.
More details inline

andre-code:
> Nice work! I think it requires a couple of changes before merging:
> 
> 1. The file information should be preserved and used to open the target notebook in the relevant session.
> 2. The modifications in `GoBackButton` may not be necessary. Saving the original URL in the state instead of a fixed string may be easier to handle and maintain.
>    More details inline

Hey, thx for your suggestions. I updated the useHistory part. ✌️

lorenzo-cavazzi (approved):
Lgtm! :+1:

lorenzo-cavazzi (approved):

PR #1730, opened 2022-03-14
feat(client): improve starting new sessions

### LINKED ISSUE #1408: Improve checking for image availability
## Problem

Currently, we check for images availability in 2 ways when starting a new session
- If the user is not logged in, we check the registry image to be available.
- If the user is logged, we check the pipelines.

To identify the proper pipeline, we check the jobs inside each pipeline and we save the reference of the relevant pipeline. Then we keep polling it until the status is either ready or failed.
It may happen that a pipeline fails because it contains more than 1 job, but the `image_build` job succeeded. In this case, the UI shows an error, and ignoring it ends up in using a default image instead of the correct one.

## Solution

We should instead identify the job and keep polling for that. That would add jobs to the resources to keep polling (we still need to keep checking the pipelines since we may not find the one containing the target job immediately).

Once the job succeed, we should also check that the image is actually available in the registry. Occasionally, the image is not there anymore.

### LINKED ISSUE #1679: Automatically expand session options when base image is not available
Trying to start a session from the quick-start button on the top right of a project page fails for projects with a faulty image (E.G. https://dev.renku.ch/projects/lorenzo.cavazzi.tech/1-0-4-project-without-jobs ).
The feedback, in that case, is clear, with a warning message and the advanced options automatically expanded.

Screenshot_20220203_163837

However, clicking on the "New session" button from the "Sessions" tab shows a 'mild' feedback (as in the next screen), pushing the user to click on "Start with base image" instead of trying to (re)run the job or build the pipeline.

Screenshot_20220203_163534

We should expand the advanced options automatically in that case.
No extra banner is needed since the user still hasn't started any session and nothing failed yet.

lorenzo-cavazzi:
This PR fixes a number of problems with starting sessions, and it overhauls the logic to verify docker image availability.

**CI:**
Now we always check for the final image to be available in every case.
If the user is logged, we first verify that pipelines and jobs have run as expected.
If the user has permissions, we show button to restart jobs and try to fix the most common project on the fly.

**FIXES:**
Visual glitches and inconsistent results should be fixed. Some of the most visible are:
- the fields "Branch" and "Commit" are never shown empty.
- changing branch wait for all the data to be available before flashing a commit and changing it
- autostarting a notebook has a consistent behavior, and show the error when one happens

**How to test**
I created a number of projects (name starting with "1408-") to test the different cases, but covering all the cases may be tricky.
Try to mess up projects or removing previously created images is an effective way of testing.
Also, verifying if a case is missing from the tests would be good.

Here are 2 fairly common cases that used to create problem -- i.e. images marked as available but actually not available:
- Job finished, but image not available (logged user): https://dev.renku.ch/projects/lorenzo.cavazzi.tech/1408-missing-image/sessions/new
- Pinned image not available (all uswers): https://dev.renku.ch/projects/lorenzo.cavazzi.tech/custom-image/sessions/new

fix #1408
fix #1679


lorenzo-cavazzi:
I added the tests too. The PR is ready for a review

ciyer (approved):
Wow, this is quite the monster PR, but the changes are very good! I think the notebook logic is now clearer and the feedback is reasonable for many cases.

In addition to the test cases you have prepared, I tried a few of my own "difficult" projects, and everything worked well.

I think we should aim to break the `Notebooks.state.js` fie into smaller files that handle certain logic, but this PR is already so large, I think that can be done in another, refactoring PR, later.

Issue #594, opened 2019-09-04
Persistent KG popup for a private project


cchoirat:
I want to permanently deactivate the KG integration, but this keeps showing up:
![image](https://user-images.githubusercontent.com/4173925/64277863-a8066900-cf4b-11e9-954f-090f4d970ad7.png)

ciyer:
Yes, I agree, this is annoying. The solution needs two parts:
1. Make it possible on the settings page to activate the KG. The user may change their mind in the future
2. Provide the options: "Not now", and "No" (and do not ask again) to the dialog. For the "No" option, we need to store somewhere that the user has explicitly opted out.

lorenzo-cavazzi:
I have checked possible alternatives to save project-related metadata in GitLab and, after a quick discussion with @jachro, it seems that the KG may be the best candidate.
This requires an API to store/retrieve the information from the UI (graphQL endpoint). It makes sense to briefly discuss the solution before going in that direction.

rokroskar:
> it seems that the KG may be the best candidate

so you will store data in the KG for a project that doesn't want its data in the KG? :)

lorenzo-cavazzi:
Ahah yes. The project won't be indexed, but a few metadata about the project (user config data) may still be saved there. My understanding is that in the future also those projects will be (at least partially) indexed but the data won't be exposed.

The alternative is to store these metadata in the existing postgres DB (we should create an API to interact with it) or to create a new DB.

lorenzo-cavazzi:
Once SwissDataScienceCenter/renku-gateway#183 will be in place, we can use those API to store this information in the gateway Redis instance

ciyer:
Closing since we will switch to having all projects in the KG always.

PR #1732, opened 2022-03-15
Adapt to new name of core-service version endpoint

### LINKED ISSUE #1714: Use renku-core updated version/apiversion endpoint
## Motivation

The `GET /version` endpoint for renku-core will change soon. We should update the UI as well.

## Proposal

The `GET /version` endpoint will change to `GET /apiversion`.
Currently, a `GET /versions` endpoint exists as well (Mind that it will be renamed as well!), and it returns the list of all available project-version specific endpoints.
We should probably use that to discover the supported versions beforehand instead of trying to reach an endpoint and assess whether we get an error.

REF: SwissDataScienceCenter/renku-python#2593

ciyer:
# Description

Adapt to changes in https://github.com/SwissDataScienceCenter/renku-python/pull/2737.

# Testing

There are two changes in this PR.

1. The name of the endpoint that returns core-service version information has been changed to `apiversion`
2. The way we check if a project is supported by the core-service has changed. We used to use the `apiversion` endpoint to see if the project version is supported, but we now use the `api/renku/version` endpoint, which returns information about all the versions that are supported. That information is retrieved just once, and now we just check if the project's version is included in the list of supported versions.

To test, you should look at projects of several versions and check that the UI correctly reports whether the project is supported or now.

E.g.,

**Renku version 9** (supported by backend)

image

**Renku version 8** (supported by backend)

image

**Renku version 4** (not supported by backend)

image

Fix #1714

BREAKING CHANGE: Requires renku-core version > 1.1.1

/deploy renku-core=2593-version-endpoint #594 #persist


lorenzo-cavazzi (approved):
Code looks great!

I had some problems testing as anonymous users with outdated projects since I occasionally get a 502 error.
I'm not sure if it's just something in the CI deployment since it's not a tagged version from renku-core. Otherwise, all looks good.

Screenshot_20220321_153347

PR #1745, opened 2022-03-18
fix: inherit function to update project

lorenzo-cavazzi:
Fix a bug with the update project button

/deploy #persist


ciyer (approved):
Tested, and it does fix the problem. Thanks for the quick reaction!

Issue #1744, opened 2022-03-18
Cloud storage breaks settings


ciyer:
## Description

When cloud storage is turned on, the session settings page is broken

## Reproduce

Look at the settings > sessions page of any project here https://tasko.dev.renku.ch/

PR #1746, opened 2022-03-21
add tests for the project settings page with and without cloud storage

ciyer:
# Description

The bug that was reported was due to the UI not being correctly updated, but this PR adds a test for the project settings page.

Fix: #1744


lorenzo-cavazzi (approved):
Good to have a test to cover this case :+1:

Issue #1736, opened 2022-03-17
UI/UX Project forms 


andre-code:
## Motivation

Apply new form design to new project, fork project, add dataset to new project forms.

### Initial status
![New-project](https://user-images.githubusercontent.com/43388408/158759793-80c59ca5-d2af-4675-8982-34ee8327a382.png)

### Validating form
![new-project--sending form](https://user-images.githubusercontent.com/43388408/158759843-bb4bffab-6ec8-4517-aa77-318d41a4002e.png)

### Error form
![new-project-errors](https://user-images.githubusercontent.com/43388408/158759868-b561092a-3174-41c6-a4da-0de473aac890.png)

### After submit valid form
![Sending-project](https://user-images.githubusercontent.com/43388408/158762382-7bb5e7a1-814b-4bf6-8813-eeedd2218313.png)

### Fork
![fork-project](https://user-images.githubusercontent.com/43388408/158759898-a74dfa6b-f86c-4d7e-ab1c-8dfd3c65becd.png)
![fork-project-sending](https://user-images.githubusercontent.com/43388408/158762581-245c035d-d898-4d8c-b1fa-0120d4b0c760.png)

NOTE:
Use:
Template components [#1734](https://github.com/SwissDataScienceCenter/renku-ui/issues/1734)
Visibility components [#1733](https://github.com/SwissDataScienceCenter/renku-ui/issues/1733)
Request progress component [#1735](https://github.com/SwissDataScienceCenter/renku-ui/issues/1735)

Issue #1733, opened 2022-03-17
UI/UX Project Visibility common component


andre-code:
## Mock-ups / Design
![visbility2](https://user-images.githubusercontent.com/43388408/158758924-fd6c5a1c-fa32-477f-9f7f-4b84bb961448.png)

PR #1747, opened 2022-03-21
feat(client): Add project visibility component (#1733)

andre-code:
This PR adds the component for **project visibility**. It also introduces the use of typescript for new components in the client project and adds the storybook tool as a component library.

## Storybook
Storybook is a development environment tool that is used as a playground for UI components. 
It is introduced to create and test components in isolation.
### How to use it
At the moment there is not public domain to access but can be tested locally.
 - fetch and move to the branch `1733-project-visibility-component`
 - update the dependencies `npm i` 
 - run `npm run notebook`in the client folder, it should run at  http://localhost:6006

You should be able to see the new Visibility component in the Storybook and interact with it using the available controls
storybook-visibility

In the New project form:
Screenshot 2022-03-23 at 08 16 39

NOTE: the component will be included in all form in the issue https://github.com/SwissDataScienceCenter/renku-ui/issues/1736 


client/src/utils/components/visibility/Visibility.tsx
```diff
@@ -0,0 +1,140 @@
...
+/**
+ *  renku-ui
+ *
+ *  Visibility.js
+ *  Visibility input
+ */
+
+/**
+ * Project Visibility
+ * @param {Object} props - Visibility options
```
ciyer, line 52: I think it would make more sense to put this documentation on the fields of the `VisibilityInputProps` type.

That has the advantage that there no need to duplicate the type information (which could then get out of sync) or specify the options of an enum, since that would be obvious.

ciyer (changes requested):
🤩 Very nice improvement! I tested it out, and it works very well -- much clearer than before.

I added a commit that makes an unrelated cypress test more robust. The only suggestion I have in the code is that the JSDoc should be together with the type definition, rather than with the component.

ciyer (approved):
Fantastic!

PR #1766, opened 2022-04-01
build(client): update reactstrap

andre-code:
### Description
Switch to use latest Reactstrap version ^9.0.1

/deploy renku=reactstrap-test-fixes #persist


ciyer (changes requested):
Good to get back to the standard project for this dependency!

The code changes look good, but I see that tests (both local and the acceptance tests) are failing. It looks like some additional changes are necessary to fix the tests.

ciyer (approved):
Everything is working now!

PR #1768, opened 2022-04-01
Search UI preparation: introduce redux-toolkit

ciyer:
The first step of this work is to use redux toolkit.

/deploy renku-core=develop renku=tests-rp-version #persist

### DIFF
--- a/client/src/utils/helpers/EnhancedState.js
+++ b/client/src/utils/helpers/EnhancedState.js
@@ -24,24 +24,21 @@
  *  Utility UI state management.
  */
 
-import { createStore as reduxCreateStore, applyMiddleware, compose } from "redux";
-import thunk from "redux-thunk";
+import { configureStore } from "@reduxjs/toolkit";
 
 function createStore(reducer, name = "renku") {
-  const composeEnhancers =
-  typeof window === "object" &&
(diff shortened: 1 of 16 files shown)


client/src/utils/helpers/EnhancedState.js
```diff
@@ -24,24 +24,21 @@
...
-  const enhancer = composeEnhancers(
-    applyMiddleware(thunk),
-    // other store enhancers if any
-  );
-  return reduxCreateStore(reducer, enhancer);
+import { configureStore } from "@reduxjs/toolkit";
+
+function createStore(renkuStateModelReducer, name = "renku") {
+
+  // For the moment, disable the custom middleware, since it causes
```
andre-code, line 31: according to the [documentation of redux-thunk](https://github.com/reduxjs/redux-thunk )  if we use the Redux Toolkit package the thunk middleware is already included so I think it is not needed here and we have to remove redux-thunk from the package.json

andre-code (commented):
Great addition to the project 🚀
I only found a minor thing to considered now that we move to use redux-toolkit.

client/src/utils/helpers/EnhancedState.js
ciyer: Good point!

Issue #1742, opened 2022-03-17
Prioritize checking Docker image availability when starting new sessions.


lorenzo-cavazzi:
## Motivation

#1730 fixed most pending issues with starting new sessions when part of the CI fails.
Still, we could simplify it by checking first the image availability and defer further checks only when images are not available and the user has enough permissions.

## Proposal

Check the GitLab registry first in _any_ case (i.e. logged, owner, pinned, anonymous) and stop there if the image is available [EDIT: for pinned images, we can't use the local registry, there is a new renku-notebooks API. The same one cannot be used easily for locally build images, which should still rely on GitLab registry APIs]. Inspect the pipelines and jobs only when something went wrong and the user is logged or owner.
P.S: Verify the logic is in line with what renku-notebooks does.

PR #1769, opened 2022-04-02
feat(client): use new api for pinned images

### LINKED ISSUE #1767: Use the new check_docker_image endpoint to verify registry images availability
## Description

We recently changed the start session page to verify image availability for pinned images.
EDIT: the check is wrong, and it uses the same logic we use to check images in the local repository, which doesn't make sense.

We agreed on using a new `GET /notebooks/check_docker_image` endpoint from the renku-notebooks repo.
An alternative approach would be re-implementing the logic locally, maybe using a library like https://www.npmjs.com/package/docker-registry-client 

## Reproduce
Any pinned image referring to an image built within a different repo fails.

lorenzo-cavazzi:
/deploy renku-core=develop renku=tests-rp-version renku-notebooks=chore-prepare-1.6.0-release #persist

This PR changes the faulty check for the docker image availability when a project specifies a pinned image.
Instead of using the GitLab API and accessing the project registry, it uses a new renku-notebooks API `GET /notebooks/images`

Screenshot_20220402_192201

fix #1767

### DIFF
--- a/client/src/api-client/index.js
+++ b/client/src/api-client/index.js
@@ -206,9 +206,12 @@ class APIClient {
   // REF: https://stackoverflow.com/questions/4500741/suppress-chrome-failed-to-load-resource-messages-in-console
   simpleFetch(
     url,
-    method = "GET"
+    method = "GET",
+    queryParams = null
   ) {
     const urlObject = new URL(url);
+    if (queryParams)
+      urlObject.search = new URLSearchParams(queryParams).toString();
     let headers = new Headers({
       "credentials": "same-origin",
       "X-Requested-With": "XMLHttpRequest"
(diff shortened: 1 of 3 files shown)


rokroskar:
In the case of a non-pinned image, we make 4 calls to accomplish the same thing, and they all together take somewhere between 1.2-1.5 seconds. Should we not just replace those with this API call, which takes 300-600ms? And only do the full pipeline + jobs checks if the image check fails?

ciyer (dismissed):
The code looks good, and, I tested the changes against a few projects that have pinned images, and received good feedback in those cases (both anonymous and logged in). I also tested a few normal projects, and everything seems to work the same there.

ciyer:
> In the case of a non-pinned image, we make 4 calls to accomplish the same thing, and they all together take somewhere between 1.2-1.5 seconds. Should we not just replace those with this API call, which takes 300-600ms? And only do the full pipeline + jobs checks if the image check fails?

Yes, that is this issue https://github.com/SwissDataScienceCenter/renku-ui/issues/1742

ciyer (approved):
🎉

PR #1773, opened 2022-04-05
feat(client): use the progress indicator component when forking a project (#1736)

andre-code:
### Description
This PR is to use the progress indicator component when forking a project

for projeck progress indicator


andre-code:
Duplicated #1775

PR #1776, opened 2022-04-05
chore: fix spelling in pinned image error text

rokroskar:
Fixes a spelling error in pinned image error text.


ciyer (approved):
Thanks for that fix. I do not understand how the misspelling got past our linter, though...

PR #1778, opened 2022-04-07
feat(client): improve error handling on missing pipelines

lorenzo-cavazzi:
This should improve the UX when a user forks a project and something goes wrong with the pipelines.

In short:
- The pipeline API is invoked again after a short delay when it fails during a fork.
- In case of missing pipelines, the UI correctly shows the "build the branch image" button when trying to start a new session.

Screenshot_20220407_093945

Here is an example of recovering from pipeline failures.

Peek 2022-04-07 09-59

/deploy renku-core=develop renku=auto-update/renku-core-1.2.0 #persist

### DIFF
--- a/client/src/api-client/project.js
+++ b/client/src/api-client/project.js
@@ -19,6 +19,8 @@
 import { fetchJson } from "./utils";
 import yaml from "yaml-js";
 
+import { sleep } from "../utils/helpers/HelperFunctions";
+
 function getApiUrlFromRepoUrl(url) {
   const urlArray = url.split("/");
   urlArray.splice(urlArray.length - 2, 0, "repos");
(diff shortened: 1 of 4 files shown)


lorenzo-cavazzi:
**How to test:** the best way would be to run telepresence to hijack `client.runPipeline` and make it fail when forking a project. Then trying to start a new session should show the correct buttons to run the pipeline again.

I couldn't find a reliable way to reproduce the failure on the first invocation after forking, so I don't really know how we could test the efficacy of the second one introduced in this PR. It _may_ help, but that's mainly speculation based on evidence (re-running it a few seconds after the fork usually works)

ciyer (approved):
Thanks for putting this fix together; let us hope it helps!

PR #1780, opened 2022-04-07
Use react-toolkit query for search

ciyer:
Set up initial RTK-Query-based hook for getting data for the entity search.

## Testing

You can see how the query in action on https://renku-ci-ui-1780.dev.renku.ch/style-guide/search

### DIFF
--- a/client/src/features/kg-search/kg-search-api.ts
+++ b/client/src/features/kg-search/kg-search-api.ts
@@ -1,0 +1,61 @@
+import { FetchBaseQueryMeta } from "@reduxjs/toolkit/dist/query/fetchBaseQuery";
+import { createApi, fetchBaseQuery } from "@reduxjs/toolkit/query/react";
+import type { KgSearchResult, ListResponse } from "./kg-search";
+
+type SearchEntitiesQueryParams = {
+  phrase: string;
+  sort: string;
+  page: number;
+  perPage: number;
+};
+
+function getHeaderFieldNumeric(headers: Headers, field: string): number {
+  return +(headers.get(field) ?? 0);
+}
+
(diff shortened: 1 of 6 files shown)


client/src/features/kg-search/kg-search-api.ts
```diff
@@ -0,0 +1,34 @@
...
+import type { KgSearchResult } from "./kg-search.types";
+
+type SearchEntitiesQueryParams = [string, string, number, number];
+
+// Define a service using a base URL and expected endpoints
+export const kgSearchApi = createApi({
+  reducerPath: "kgSearchApi",
+  baseQuery: fetchBaseQuery({ baseUrl: "/api/kg/" }),
+  endpoints: (builder) => ({
+    searchEntitiesByPhrase: builder.query<
```
andre-code, line 12: I am not sure if an endpoint searchEntitiesByPhrase is the appropriate name, since byPhrase will also contain type, date, visibility, etc. So in the end the search is generic and not just by phrase.

andre-code (commented):
Happy to see the redux-toolkit createApi feature being used and not worrying about handling the query state 🎉🙌.
There are some interesting features pending that I would like to see... for example how we are going to handle pagination and sorting as well as other comments I left to make the api ready to use.

client/src/features/kg-search/kg-search-api.ts
ciyer: Thanks for the review -- I agree with all your comments, and I have updated the PR accordingly. I also updated the example to show how to integrate it with Pagination.

andre-code (approved):
🚀

PR #1782, opened 2022-04-08
Update toolchain (introduce TS 4.5, react 17, node 16)

ciyer:
Changes to toolchain to allow us to use TypeScript in the client.

Some things to fix now:
- [x] the notebook renderer is not compatible and needs to be fixed

Some things to fix in a follow up:
- Upgrade CKEditor

/deploy renku=tests-new-project-form-changes #persist


lorenzo-cavazzi (dismissed):
Nice update!
As discussed offline, the one problem left is with stylesheets. Otherwise, all good.

andre-code (approved):
🎉🚀

lorenzo-cavazzi (approved):

PR #1784, opened 2022-04-08
test(e2e): add tests for form to create a new project

andre-code:
### Description

This PR contains the e2e tests for the create a project form.
 


lorenzo-cavazzi (approved):
Looks good!
As discussed offline, there is something off with some tests in Firefox, but I'll check that separately

PR #1788, opened 2022-04-12
fix: do not show renkulab down when the users auth token has expired

### LINKED ISSUE #1759: Renku down page when user is not authenticated
## Description

The interface is displaying a Renku down page when the user is not authenticated, it redirects to login automatically but this page is visible to the user for a few seconds.

## Reproduce

1. Go to renkulab.io or dev.renku.ch when you are not yet authenticated
2. See error 'Renku down'
renkulab down when user is not authenticated

ciyer:
# Description

The fact that the server response was caused by an expired token was getting lost, so it was not possible to handle the error correctly at higher levels. This has been fixed.

Testing: Log into the Keycloak admin console and set the *SSO Session Idle* and *Client Session Idle* values to something very short (like 1 min). This will provoke more frequent auth token expirations, and you will be redirected to login, but you should not see the "RenklabDown" warning.

image

Fix #1759

/deploy renku=auto-update/renku-core-1.2.1 #persist

### DIFF
--- a/client/src/api-client/index.js
+++ b/client/src/api-client/index.js
@@ -111,8 +111,10 @@ class APIClient {
   ) {
     return renkuFetch(url, options)
       .catch((error) => {
-        if (error.case === API_ERRORS.authExpired)
-          return this.doLogin();
+        if (error.case === API_ERRORS.authExpired) {
+          this.doLogin();
+          return Promise.reject(error);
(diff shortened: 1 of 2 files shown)


lorenzo-cavazzi (approved):
The fix works well! :+1:

PR #1794, opened 2022-04-13
build(server): run ui-server as non-root user

lorenzo-cavazzi:
Run UI server as non-root.

To keep it coherent with the other repos, we should add also `runAsGroup: 1000`, but that creates problems with telepresence v1. I can give a quick try to telepresence v2. If that works, we could upgrade

/deploy #persist 

fix SwissDataScienceCenter/renku-notebooks#1001


ciyer (approved):
This makes sense.

PR #1797, opened 2022-04-19
fix(client): adjust session page

### LINKED ISSUE #1796: session is said to be "Running" but "Open" button leads to 503 error
## Description

Happens for every session. 
I see what the interface is trying to convey to me, but as we are about to invite some more users, I am concerned they will be confused.

Session is said to be "Running" but "Open" button leads to 503 error

## Reproduce

image

Clicking "Open" leads to

image

* https://renkulab.io
* firefox

lorenzo-cavazzi:
This fixes 2 details on the session page:
- do not show the open button when the session is not ready (fix #1796). The "Running since" label is also different until the server is actually usable -- see screenshot below.
- do not crash when the session is not available. This happened when the user tried to open a session page in the UI when the session was gone, typically a tab left open for a while.

image

/deploy #persist

### DIFF
--- a/client/src/notebooks/Notebooks.present.js
+++ b/client/src/notebooks/Notebooks.present.js
@@ -103,7 +103,7 @@ function ShowSession(props) {
   });
 
   // redirect immediately if the session fail
-  if (props.history && notebook.data.status.state === SessionStatus.failed)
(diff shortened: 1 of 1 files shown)


ciyer (approved):
This works well, thanks for getting the quick improvement together!

PR #1799, opened 2022-04-19
build(server): add runAsGroup and setup telepresence 2

lorenzo-cavazzi:
As a follow-up to #1794 , this PR adds the missing `runAsGroup` pod security context.
That requires moving to telepresence 2, which turned out to be easy for the server, but not straightforward for the client. There is likely something to adjust on the ports. I'll look into this later since it has already taken quite a bit of time.


lorenzo-cavazzi:
Superseded by #1987

PR #1824, opened 2022-05-06
feat(client): strip dash from dataset id URLs

### LINKED ISSUE #1819: Handle URLs pointing to datasets with the old dashed format
## Description

The UI handles datasets with dashes (old format) by stripping out all the unneeded `-`.
This was implemented in #1694 , but we considered only the API responses, while users may have saved links like https://renkulab.io/datasets/2b4f3e51-5fb7-4cc0-9992-ad1d0de3e68c 

A very quick and easy solution would be to redirect to the hash-stripped URL

## Reproduce

Try the following:
- working: https://renkulab.io/datasets/2b4f3e515fb74cc09992ad1d0de3e68c
- broken: https://renkulab.io/datasets/2b4f3e51-5fb7-4cc0-9992-ad1d0de3e68c

lorenzo-cavazzi:
The solution is simpler than the one suggested in the issue. We actually don't need to redirect the user since the new id is just the older one with the `-` chars stripped out.

/deploy renku=tests-new-project-form-changes #persist
fix #1819

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -130,7 +130,7 @@ function CentralContentContainer(props) {
           p => <AddDataset
             key="addDatasetNew"
             insideProject={false}
-            identifier={`${p.match.params.identifier}`}
+            identifier={p.match.params?.identifier?.replaceAll("-", "")}
             datasets={p.datasets}
             model={props.model}
           />}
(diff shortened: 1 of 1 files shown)


ciyer (approved):
👍

PR #1839, opened 2022-05-12
feat(client): speed up checking docker images availability

### LINKED ISSUE #1742: Prioritize checking Docker image availability when starting new sessions.

lorenzo-cavazzi:
This PR speeds up how we check for Docker image availability when starting a new session.
It prioritizes testing the Docker image, deferring checking pipelines and jobs -- doing it only when the image is missing.
This makes the logic in the UI similar to the one already used in renku-notebooks.

If no image is available, we either:
- [for anonymous users] notify the image is not available and keep checking for it
- [for logged users] check pipelines and jobs and provide adequate feedback -- showing quickfix buttons for users with permissions on the project

For pinned images, no big changes: we can't really check the pipelines in that case, so we just verify the image availability (without looping).

**How to test**
Ideally, the user shouldn't notice many differences. Trying to start a session and keeping the network tab open would simplify checking that we reach only the necessary APIs

fix #1742

### DIFF
--- a/client/src/model/RenkuModels.js
+++ b/client/src/model/RenkuModels.js
@@ -404,6 +404,15 @@ const notebooksSchema = new Schema({
   },
   ci: {
     [Prop.SCHEMA]: new Schema({
+      image: {
+        [Prop.SCHEMA]: new Schema({
+          available: { [Prop.INITIAL]: null },
+          fetched: { [Prop.INITIAL]: null },
+          fetching: { [Prop.INITIAL]: false },
+          registryId: { [Prop.INITIAL]: null },
+          error: { [Prop.INITIAL]: null },
+        })
+      },
       pipelines: {
         [Prop.SCHEMA]: new Schema({
           list: { [Prop.INITIAL]: [] },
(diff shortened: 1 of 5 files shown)


ciyer (changes requested):
Nice changes that does bring users to the start button faster!

I have a small request for a change in the code, and the PR needs to bring in the latest version on master. I can re-approve after that is done.

ciyer (approved):
👍

PR #1843, opened 2022-05-13
fix: missing project display

ciyer:
The dataset page should show the projects a dataset is used in, even if there is just one.

image

But the project dataset page should not (if there is just one).

image


lorenzo-cavazzi (approved):
Nice fix.
One question: is it guaranteed that the dataset will always be at least in a project?

PR #1854, opened 2022-05-17
feat: update ux add dataset form (#1740)

### LINKED ISSUE #1740: UI/UX New Dataset form
## Mock-ups / Design
New-dataset
dataset-form-after
new-dataset--sending-load

NOTE. use components
Request progress component #1735 
Author component #1738 
File Upload component #1739

andre-code:
### Description.
PR to update the UX of the Add Dataset form to be uniform like the New Project form.

add dataset form ux

#### Tests
Go to add or modify a dataset in a project.

Closes #1740 
 

client/src/project/datasets/change/DatasetChange.present.js
```diff
@@ -103,7 +103,27 @@ function DatasetChange(props) {
     submitCallback={props.submitCallback}
     title={edit ? "Modify Dataset" : undefined}
     versionUrl={props.versionUrl}
+    toggleNewDataset={props.toggleNewDataset}
+    showAddDatasetOptions={!edit}
+    addDatasetOptionSelected={"addDataset"}
```
lorenzo-cavazzi, line 108:
```suggestion
    addDatasetOptionSelected="addDataset"
```
Not really something to fix, but I noticed you sometimes specify text in properties as `prop={"text"}` and sometimes `prop="text"`. It would make sense to keep it coherent when it's a fixed string

client/src/project/datasets/import/DatasetImport.present.js
```diff
@@ -44,15 +45,36 @@ function DatasetImport(props) {
...
+      btnName="Import Dataset"
+      submitCallback={props.submitCallback}
+      model={props.datasetImportFormSchema}
+      onCancel={props.onCancel}
+      formLocation={props.formLocation}
+      notifications={props.notifications}
+      modelTop={props.model}
+      toggleNewDataset={props.toggleNewDataset}
+      showAddDatasetOptions={true}
+      addDatasetOptionSelected={"importDataset"}
```
lorenzo-cavazzi, line 59:
```suggestion
      addDatasetOptionSelected="importDataset"
```

client/src/utils/components/formgenerator/FormGenerator.present.js
```diff
@@ -139,35 +138,57 @@ function FormPanel({
...
-          }
-        </div>
-      </Form>
+      {submitLoader ? (
+        <ProgressIndicator
+          type={ProgressType.Indeterminate}
+          style={ProgressStyle.Dark}
+          title={titleProgress}
+          description="We've received your dataset information. This may take a while."
+          currentStatus={""}
```
lorenzo-cavazzi, line 168:
```suggestion
          currentStatus=""
```

client/src/utils/components/formgenerator/fields/FormLabel.js
```diff
@@ -25,12 +25,14 @@
...
 
-const FormLabel = ({ name, label, required }) => {
-  return <Label htmlFor={name} required={required}>
-    {label}
-    {required ? <span style={{ "color": "#dc3545" }}>*</span> : null}
-  </Label>;
+const FormLabel = ({ name, label, required, optional }) => {
+  return (
+    <Label htmlFor={name} required={required}>
+      <InputLabel text={label} isRequired={required} isOptional={optional} />
```
lorenzo-cavazzi, line 33: Having both `isRequired` and `isOptional` seems redundant. Would it make sense to keep only one? Although I see it's already used like this somewhere else in the code

client/src/model/Model.js
```diff
@@ -64,7 +64,9 @@ const PropertyName = {
   VALUE: "value",
   MODES: "modes",
   MAXSIZE: "maxSize",
-  FORMAT: "format"
+  FORMAT: "format",
+  OPTIONAL: "optional",
```
lorenzo-cavazzi, line 68: Isn't `OPTIONAL` just the opposite of `REQUIRED`?

client/src/model/Model.js
```diff
@@ -64,7 +64,9 @@ const PropertyName = {
...
+  IMAGECONTROLSDISABLED: "imageControlsDisabled",
```
lorenzo-cavazzi, line 69: I don't particularly like adding more properties to the model. I feel we already have a bunch that should really not be here.
I guess there is another property to pilot when it's editable (or being edited). Do we need this one too?

lorenzo-cavazzi (changes requested):
The look is much better now! Very nice changes!

I added a few comments to keep some code coherent -- mostly cosmetics, and I know the code is already messy somewhere else.

The only thing I would change is in the Model. Is it really necessary to add extra properties to the fields? (see inline comment).

client/src/utils/components/formgenerator/fields/FormLabel.js
andre-code: It was done that way to keep the case where no action is required, for example when it has a default value or the field is disabled, for example for this form the Creators case, and for example in the new project form for Template source or the identifier.

ciyer (changes requested):
Wow, the entire dataset experience is much better with these changes! 🎉 I agree with Lorenzo's suggestions to change the code, but the code is overall very good.

I have a few things regarding the appearance and experience that I wanted to mention:

1. Dataset Description

image

I think this field should be a little larger vertically -- it seems a little squashed. I think the _minimum_ size should allow display of two lines of text, but the expanding behavior of the component is otherwise nice.

Also, the help text is incorrect in this case: in this mode, you cannot use Markdown, but you can style the text using the tools on the tool bar.

2. Action display

image


When an add or modify action is running, we still show the left-hand action text (modify dataset in this case). I think we should not show this text and the progress indicator should take up the whole width.

andre-code:
> The look is much better now! Very nice changes!
> 
> I added a few comments to keep some code coherent -- mostly cosmetics, and I know the code is already messy somewhere else.
> 
> The only thing I would change is in the Model. Is it really necessary to add extra properties to the fields? (see inline comment).

Thank you very much for your review. I agree with all your suggestions, so you can find them implemented in the last 2 commits. 🙌

andre-code:
> Wow, the entire dataset experience is much better with these changes! 🎉 I agree with Lorenzo's suggestions to change the code, but the code is overall very good.
> 
> I have a few things regarding the appearance and experience that I wanted to mention:
> 
> 1. Dataset Description
> 
> image
> 
> I think this field should be a little larger vertically -- it seems a little squashed. I think the _minimum_ size should allow display of two lines of text, but the expanding behavior of the component is otherwise nice.
> 
> Also, the help text is incorrect in this case: in this mode, you cannot use Markdown, but you can style the text using the tools on the tool bar.
> 
> 2. Action display
> 
> image
> 
> When an add or modify action is running, we still show the left-hand action text (modify dataset in this case). I think we should not show this text and the progress indicator should take up the whole width.

Thank you very much for the review.
It was a bit confusing to set the CKEditor height, so the library will probably need to be updated. The change to not display the form header when the progress indicator component is visible was implemented for the data set form and the project form.

lorenzo-cavazzi (dismissed):
:rocket:

ciyer (dismissed):
This looks very nice!

lorenzo-cavazzi (dismissed):

lorenzo-cavazzi (approved):

Issue #1853, opened 2022-05-16
Create a "Developed at" footer on the Landing page


lokijuhy:
Add a "Developed at SDSC" panel above the page footer. Each logo (SDSC, ETH, and EPFL) should be clickable links to each organization's homage.

SDSC: https://datascience.ch/
ETH: https://ethz.ch/en.html
EPFL: https://www.epfl.ch/en/

[Figma](https://www.figma.com/file/AoqDhVEei3xApz8dMYYA9g/Landing-page-alternatives?node-id=0%3A1)

![Image](https://user-images.githubusercontent.com/7387902/168605752-fd325f63-af26-42d6-8e04-dd20d14fdc87.png)

Issue #1851, opened 2022-05-16
Create "Use cases" section of the landing page


lokijuhy:
![Image](https://user-images.githubusercontent.com/7387902/168600770-22794f58-fbdd-44bb-81dd-b7534bfc12c3.png)

[Figma](https://www.figma.com/file/AoqDhVEei3xApz8dMYYA9g/Landing-page-alternatives?node-id=0%3A1)

Issue #1850, opened 2022-05-16
Create "Renku Features" section of landing page


lokijuhy:
Create the "Renku Features" section of the website (not including the "Try it now" and "Sign Up" buttons)

[Figma here](https://www.figma.com/file/AoqDhVEei3xApz8dMYYA9g/Landing-page-alternatives?node-id=0%3A1)

![Image](https://user-images.githubusercontent.com/7387902/168599960-d218e1de-417c-42e2-8065-109720f5e958.png)

lokijuhy:
**Decision**: Move ahead with the homepage without any third party logos, until https://github.com/SwissDataScienceCenter/ORDES/issues/506 is resolved. If 506 takes longer than this one, a separate issue will be created to add the logos.

PR #1859, opened 2022-05-20
Improvements to home page

### LINKED ISSUE #1849: Update website header
Change title and sub-title from
> Connecting people, data, and insights

> Renku bridges the gaps...

to
> An open-source knowledge infrastructure for collaborative and reproducible data science

> Connecting people, data, and insights

[Figma](https://www.figma.com/file/AoqDhVEei3xApz8dMYYA9g/Landing-page-alternatives?node-id=0%3A1)
image

ciyer:
# Description

Implement the changes requested in:
- [x] #1849
- [x] #1853
- [x] #1850
- [x] #1851 

# Screenshots

image

image

image

image

Fix #1849 #1853 #1850 #1851


lokijuhy (changes requested):
Thanks @ciyer! I have a few points of feedback:

Features and use cases sections:
* I find the mix of left-adjusted and center-adjusted text visually strange. Could we put "Renku features: Empowering all stages of your work" all together? (and likewise for the use cases)
* I personally find the mix of green, pink, and yellow boxes a little much. I prefer the look in the Figma where all the features are green and all the use cases pink. 
* The boxes looks a bit cramped. Could we please add some more padding and spacing?

I mocked up some changes of the Features panel together with Lorenzo, and we came up with the following:
- card padding: `p-4`
- image margin: `m-4`
- bigger space between columns: remove `g-2` and leave the default
- adopt the same spacing between the rows
- make all the box colors green, and lighter than the main renku green (something like #edfff3 ? [this is not what is in the screenshot- we kept playing with it to make it less minty fresh])

Screenshot_20220523_133630

The footer:
* The footer renders spatially weird for me. Again, I think there is a mix of center- and left-adjusted spacing. Could we put all three logos on one line, or make both the SDSC and the EPFL & ETH rows left-adjusted?

ciyer:
> Thanks @ciyer! I have a few points of feedback:

Thanks for the feedback, @lokijuhy. I think I have responded to all your requests (see the updated screenshots).

andre-code (changes requested):
@ciyer I like very much how it looks now.
There are a couple of pending changes to improve how it looks on mobile devices.
1. The "data science theory/practice" image has a min-width that exceeds the width on iphone XR creating a side space, so I suggest reducing it to 300px.
`<Col md={4} lg={6} className="p-s-4 rk-bg-white"
          style={{ minWidth: "300px", maxWidth: "600px",
            borderWidth: "20px", borderStyle: "solid", borderColor: "white" }}>
          data science theory/practice
        </Col>`
  2. The footer has a padding-top and a fixed height that makes spaces around the nav which does not look good and makes the links not visible there.
  
 
landing-page-mobiles

lokijuhy (dismissed):
Looks fantastic! Thank you!

gavin-k-lee:
Some feedback:
- Would we be able to have the use cases not aligned left, right and then centre? 
- Apologies - I didn't add a full stop to this sentence: 'Renku gives you tools and functionality for each stage of the data science lifecycle: from datasets to workflow execution.' Can we add this?
- I mentioned this on slack - but maybe we should remove the old components with the out of date icons, as well as the "Learn more" tab which only scrolls down.

ciyer:
> Some feedback:
> 
> * Would we be able to have the use cases not aligned left, right and then centre? It doesn't look intentional on first glance.
> * Apologies - I didn't add a full stop to this sentence: 'Renku gives you tools and functionality for each stage of the data science lifecycle: from datasets to workflow execution.' Can we add this?

@gavin-k-lee These changes have been made (see screenshots above).

lokijuhy (dismissed):

lorenzo-cavazzi:
Tiny detail: does it make sense to add a margin (`mb-2`) on the bottom of the EPFL/ETH logos?
In the screen, EPFL is with no margin (current solution), while ETH is with margin (suggested mb-2)
Screenshot_20220524_134551

andre-code:
> > @ciyer I like very much how it looks now. There are a couple of pending changes to improve how it looks on mobile devices.
> 
> Great suggestions! I have fixed on mobile both the the illustration
> 
> image
> 
> and the footer
> 
> image

Much better 🤩.
Don't we want to support those links (help, forum, gitter) on mobile devices or is the design of the menus part of another issue? because if not you could include them in the main menu or enlarge the footer height and make it visible.

lorenzo-cavazzi:
Sorry for being pedantic, but there is still something off when resizing the browser. The logos seem to overflow.
They should either wrap below a certain width (but un-wrap on `xs` :facepalm: ) or just reduce the elements per row to 1 already on larger screens (`col-md-6` --> `col-lg-6`?)

Screenshot_20220524_162042

lokijuhy (dismissed):

andre-code (approved):
🚀

PR #1860, opened 2022-05-20
fix(client): adjust checking project lock status

### LINKED ISSUE #1858: Fix how we verify the project status
## Description

When visiting some of the project tabs (e.g. status overview or datasets) we invoke the `/renku/project.lock_status` API to verify whether the user is already modifying the project. If that is the case, we show a label next to the title.
This currently has 2 problems.

* We occasionally invoke the API multiple times in a row (E.G. when using a URL leading directly to the `/projects/<id>/datasets` page). This triggers the project clone on the backend and locks it for some time, depending on the project size. The last API invocation returns a locked status even if that is a transitory state.
Screenshot_20220520_090140

* The anonymous users can't write to the project, and the API returns an error. We show a confusing message. image (5)

## Fix

We should:

* [ ] invoke the API only once
* [ ] do not invoke it (nor show any warning) for anonymous users
* [ ] _extra (maybe follow-up issue?): use `setTimeout` to check the status again after a few seconds when the project is locked_

lorenzo-cavazzi:
The lock status should not show errors anymore for anonymous users, and we don't fetch it when non-necessary.

fix #1858

### DIFF
--- a/client/src/api-client/project.js
+++ b/client/src/api-client/project.js
@@ -462,7 +462,6 @@ function addProjectMethods(client) {
    * @param {string} projectRepositoryUrl - external repository full url.
    * @param {string} [versionUrl] - project version url.
    */
-  /* eslint-enable max-len */
   client.getProjectLockStatus = async (projectRepositoryUrl, versionUrl = null) => {
     const url = client.versionedCoreUrl("project.lock_status", versionUrl);
     const queryParams = { git_url: projectRepositoryUrl };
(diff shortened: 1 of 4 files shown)


ciyer (approved):
Nice job -- very good surgical fixes to resolve the problem. Thanks for the quick response!

ciyer (changes requested):
I see that the cypress tests seem to be failing.

andre-code (approved):
👏

ciyer (approved):
🎉

PR #1864, opened 2022-05-23
1858 test

### LINKED ISSUE #1858: Fix how we verify the project status

lorenzo-cavazzi:
_Testing this instead of #1860_ 

fix #1858


lorenzo-cavazzi:
closing, duplicate of #1860

PR #1868, opened 2022-05-30
Add Docs and GitHub links to initial section of anonymous home

### LINKED ISSUE #1852: Add GitHub and Docs buttons to the landing page top menu bar
Add links to Renku GitHub and Docs to the top of the landing page, next to "Login" (you can ignore everything else in the Figma). You can make it a pretty GitHub logo as in the Figma, or just text `GitHub` - doesn't matter for now!

* GitHub should link to https://github.com/SwissDataScienceCenter/renku
* Docs should link to https://renku.readthedocs.io/en/latest/index.html

[Figma](https://www.figma.com/file/AoqDhVEei3xApz8dMYYA9g/Landing-page-alternatives?node-id=0%3A1)

Image

ciyer:
Also, fix rendering problem of menu. Both "GitHub" as text and the GitHub icon are widely used, I chose the first, since I think it is easier to incorporate with what is already there from a design perspective.

image

image

Fix #1852


lorenzo-cavazzi (changes requested):
Looks good on a big screen, but resizing works weird.
E.G. first `Login` wraps, then unwrap and `GitHub` wraps.
Screenshot_20220530_141437


Screenshot_20220530_141426

Also, on small screens, they misalign the content

image

I would suggest either hiding them on small screens (and add a simple entry in the dropdown menu) or wrapping them in a container so that they float together under the logo -- this seems the best option but I'm not sure about the effort

andre-code:
I agree with @lorenzo-cavazzi and suggest to include those links in the dropdown menu for small devices to not have several navigations at once. In fact for large screens it makes me noise to have a hamburger menu and a menu bar together. That part would be interesting to design it... in another story.

ciyer:
Thanks for the reviews, and that is a good catch!

I have fixed the display on smaller devices, and it now looks like this:

image

lorenzo-cavazzi:
Nice solution!
Tiny detail: the size of the new links has changed in the last commit, they are now smaller than the Login button (compare screen with the previous)

Screenshot_20220530_172503

ciyer:
> Nice solution! Tiny detail: the size of the new links has changed in the last commit, they are now smaller than the Login button (compare screen with the previous)
> 
> Screenshot_20220530_172503

Oops -- I was able to reproduce this problem using Firefox and fix it. Please test again.

lorenzo-cavazzi (approved):
Lgtm!

PR #1872, opened 2022-05-30
build(client): pin react-error-overlay version

lorenzo-cavazzi:
Prevent this problem while developing with telepresence

Screenshot_20220530_105110

Reference for the fix: https://stackoverflow.com/questions/70368760/react-uncaught-referenceerror-process-is-not-defined

/deploy


ciyer (approved):
This problem has been a nuisance for a while now. I tested, and this solution works well. Great that you found a fix!

PR #1875, opened 2022-05-31
1804 UI server saving user last search queries

### LINKED ISSUE #1804: UX (search): backend for saving the user's last search queries (keywords)
## Motivation
Reduce user effort to remember recent queries

## Proposal

Store all recent searches, in order to provide this data to the user the next time they conduct a search.
TBD if use REDIS as https://github.com/swissdatasciencecenter/renku-ui/issues/1627

## Context
Screenshot 2022-04-26 at 12 29 03

andre-code:
A middleware has been added in the ui server to log the user's last search queries (keywords). And the endpoint /ui-server/api/last-searches to request it. By default it logs the last 10 queries.

### to test

1. Authenticate in https://renku-ci-ui-1875.dev.renku.ch/
2. Request entities including the keyword in the query parameter e.g. https://renku-ci-ui-1875.dev.renku.ch/ui-server/api/kg/entities?query=*test*
3.  Request list of latest search queries https://renku-ci-ui-1875.dev.renku.ch/ui-server/api/last-searches/4

Closes #1804 

/deploy #persist

### DIFF
--- a/server/src/utils/middlewares/lastSearchQueriesMiddleware.ts
+++ b/server/src/utils/middlewares/lastSearchQueriesMiddleware.ts
@@ -25,7 +25,7 @@ const lastSearchQueriesMiddleware = (storage: Storage) =>
   (req: express.Request, res: express.Response, next: express.NextFunction): void => {
     const token = req.headers[config.auth.authHeaderField] as string;
     const query = req.query["query"];
-    const phrase = query ? (query as string).trim().replace(/\*/g, "") : "";
+    const phrase = query ? (query as string).trim() : "";
 
     if (req.query?.doNotTrack !== "true" && phrase) {
       res.on("finish", function() {


server/src/routes/helperFunctions.ts
```diff
@@ -0,0 +1,35 @@
...
+import { getUserIdFromToken } from "../authentication";
+import { Storage, StorageGetOptions, TypeData } from "../storage";
+
+export async function getUserData(prefix: string, token: string, storage: Storage, length?: number): Promise<string[]> {
+  const userId = getUserIdFromToken(token);
+  let data: string[] = [];
+  const options: StorageGetOptions = {
+    type: TypeData.Collections,
+    start: 0,
+    stop: (length || 0 ) - 1
```
lorenzo-cavazzi, line 28: instead of `length?` in the argument and then `(length || 0 )` later, I would assign a default value to the argument `length = 0` to make it explicit.
Also, the default case would be start=0 and stop=-1. Is that intended?

server/src/routes/apis.ts
```diff
@@ -124,18 +126,19 @@ function registerApiRoutes(app: express.Application,
       res.json({ error: "User not authenticated" });
       return;
     }
+    const data = await getUserData(config.data.projectsStoragePrefix, token, storage, parseFloat(req.params["length"]));
```
lorenzo-cavazzi, line 129:
```suggestion
    const data = await getUserData(config.data.projectsStoragePrefix, token, storage, parseInt(req.params["length"]));
```

server/src/utils/middlewares/lastSearchQueriesMiddleware.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const lastSearchQueriesMiddleware = (storage: Storage) =>
+  (req: express.Request, res: express.Response, next: express.NextFunction): void => {
+    const token = req.headers[config.auth.authHeaderField] as string;
+    const query = req.query["query"];
+    const phrase = query ? (query as string).trim().replace(/\*/g, "") : "";
+
+    if (req.query?.doNotTrack !== "true" && phrase) {
+      res.on("finish", function() {
+        if (![304, 200].includes(res.statusCode) || !token) {
```
lorenzo-cavazzi, line 32: This seems very specific. I would relax it to `res.statusCode >= 400`

lorenzo-cavazzi: It probably makes sense to change it also in the other middleware `lastProjectsMiddleware`

lorenzo-cavazzi (changes requested):
The code looks good, I added a few comments to improve a bit the readability.

I'll add `#persist` to the deployment to these it "live"

server/src/routes/helperFunctions.ts
andre-code: Thanks Lorenzo, I updated to make it clear
yes, -1 would bring all records, in case of a positive number > 0 it would bring that amount of records. Also note that the number of records is limited by the user. For projects 20, for search queries 10.

lorenzo-cavazzi:
This looks great! It's basically ready to merge, but I tried it on the fly and I'm wondering if we should really strip the `*` from the query or not.
E.G: If I search for `query=flight*data`, then I get `{"queries":["flightdata"]}` in the list of the last queries. Looking for `flightdata` won't probably return the same result.
Maybe this is not really relevant since my understanding is that users won't need to add `*` -- and it doesn't even seem to work correctly atm when used in the middle of the sentence. Nevertheless `.replace(/\*/g, "")` doesn't seem necessary.
P.S: we can quickly discuss this during tomorrow's standup.

andre-code:
> This looks great! It's basically ready to merge, but I tried it on the fly and I'm wondering if we should really strip the `*` from the query or not. E.G: If I search for `query=flight*data`, then I get `{"queries":["flightdata"]}` in the list of the last queries. Looking for `flightdata` won't probably return the same result. Maybe this is not really relevant since my understanding is that users won't need to add `*` -- and it doesn't even seem to work correctly atm when used in the middle of the sentence. Nevertheless `.replace(/\*/g, "")` doesn't seem necessary. P.S: we can quickly discuss this during tomorrow's standup.

@lorenzo-cavazzi  In the UI is being attached to the query * at the beginning and the end is why it was being removed from here to save the query without *, but according to what you mentioned I have opted to leave the logic of removing and discarding what is not needed for the frontend. Thanks for your suggestion 🙌

lorenzo-cavazzi (approved):
Looks great!

P.S: we really need to add (at least) unit tests at some point. I will create an issue. Not strictly related to this PR, I just thought things like `lastSearchQueriesMiddleware` should be tested

Issue #1830, opened 2022-05-10
Chunked file upload in the UI


ciyer:
The PR in renku-python https://github.com/SwissDataScienceCenter/renku-python/pull/2892 implements chunked file upload using [dropzone](https://docs.dropzone.dev/getting-started/installation/npm-or-yarn). We should adapt incorporate it into the UI.

PR #1882, opened 2022-06-03
feat: add dropzone component for large files (#1830)

andre-code:
Tmp PR to implement dropzone to upload large files.

### Cases to test
- [x] Add single or multiple files, should send 5 paralle files
- [x] Remove File
- [x]  Cancel Upload
- [x] Retry Upload
- [x] Upload compressed file, should request if uncompress or not before send
- [x] The url parameter `unpack_archive=true` should be sent when the user clicks Yes to compress the file
- [x] The url parameter `unpack_archive=false` should be sent when the user clicks No to compress the file
- [x] Show file tree when sending .zip files for unzip
- [x] Show error if try to upload an existing file or url file in the list
- [x] Load url file, status pending until submit
- [x] Run queue after complete current files, in case of large list
- [x] Validate file are full upload before send
- [x] Show existing dataset files when modify dataset
- [x] Show files in draft if exist by project when navigate and return to the form
- [x] Save when submit

Closes https://github.com/SwissDataScienceCenter/renku-ui/issues/1830

/deploy renku-core=develop extra-values=gateway.allowOrigin={“*”} renku=tests-ui-2.5.0 #persist


ciyer (changes requested):
This is going to be great once it gets merged!

As discussed, I have encountered some problems testing, and it looks like I have not been testing what I thought I was testing. But, while that gets resolved, I have some cosmetic suggestions:

1. Tighten up the status display a little

image

2. Update feedback

image

andre-code:
@ciyer thanks for your suggestions, I include them with some changes.
1. As retry is an action I have decided to include it as well as the delete option and add a description tooltip.
2. I included calculating the file error when the user retry and/or deletes a file. Also the x(delete) link instead of the disappearance after n seconds so we make sure the user reads the error to take action. 

The case _Show file tree when sending .zip files for unzip_ should be working now.

ciyer (changes requested):
Most of the features are working well now! I was able to upload individual files and they were done in parallel; I could cancel, delete, and so on -- everything worked perfectly.

However, I still get an error in some cases. I think it is in the following case:

- Create a dataset
- Add a file
- Set an image

Then I get this error:

image

And I get this error from the server
```
{"error": {"code": 3130, "userMessage": "The dataset creation failed because it already exists. Please refresh the page.", "devMessage": "Unexpected error creating a dataset, possibly caused by concurrent actions.", "devReference": "https://renku-python.readthedocs.io/en/latest/service_errors.html#renku.ui.service.errors.IntermittentDatasetExistsError"}}
```

andre-code:
@ciyer yes, there was definitely an update in the api, but easy to fix in the UI.
Thanks, that case should be covered by now.

ciyer (dismissed):
🎉 Working great!

When you merge, please add a `!` to the type (e.g., `feat!`) and a `BREAKING CHANGE:` footer that explains that this commit requires a certain minimum version of renku-python.

ciyer (approved):

ciyer (approved):

PR #1884, opened 2022-06-07
feat: UI show the user last search queries

### LINKED ISSUE #1855: UX (search): frontend to show the user's last search queries

andre-code:
PR to add the last queries searched by the user as suggestions in the search field.
Displays the last 5 searches, only for logged users.

last searches

Closes #1855 

/deploy #persist


client/src/utils/components/quicknav/KgLastQueriesApi.ts
```diff
@@ -0,0 +1,49 @@
...
+import { createApi, fetchBaseQuery } from "@reduxjs/toolkit/query/react";
+
+interface KgLastSearchResult {
+  queries?: string[],
+  error?: string,
+}
+export const TOTAL_QUERIES = 5;
+
+
+export const kgLastQueriesApi = createApi({
```
ciyer, line 27: I like this approach a lot! The question is, though, what is the right scope for this RTK-Query API?

We also have the "recent-projects" API that is quite similar, and I think it makes sense to combine that (and other future endpoints) into a `recentUserActivityApi`. 

I don't think the other query needs to integrated as part of this PR, but we can prepare for it by using the name and setting the `baseQuery` to `/ui-server/api/`.

client/src/utils/components/quicknav/KgLastQueriesApi.ts
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { createApi, fetchBaseQuery } from "@reduxjs/toolkit/query/react";
+
+interface KgLastSearchResult {
+  queries?: string[],
+  error?: string,
+}
+export const TOTAL_QUERIES = 5;
```
ciyer, line 24: Do we store a total of 4 or 5 queries? I never see more than 4:

image

ciyer (changes requested):
This is a cool feature! 🎉  

I have a comment about how the API is structured and one small thing about the number of search results that are shown, but they should both be easy to address.

client/src/utils/components/quicknav/KgLastQueriesApi.ts
andre-code: Thanks, great suggestion. I moved the kgLastQueries to recentUserActivityApi.

client/src/utils/components/quicknav/KgLastQueriesApi.ts
andre-code: The reason was that the last searches query also returns `*` and in the frontend we removed that option reducing the amount of results. I have added a change to always have the last 5 words if those exist.

ciyer (approved):
This is very nice to use!

PR #1887, opened 2022-06-07
fix: send dataset current image url when updating the dataset (#1861)

### LINKED ISSUE #1861: The dataset image should not be deleted when updating the dataset
## Description

Dataset image is removed if the dataset is updated
bug update dataset

andre-code:
PR to send the url of the current image when updating the dataset so as not to delete the existing image.
Currently the endpoint for updating a dataset requests the url of the current image from the dataset to keep that image, otherwise it deletes the image. This is a workaround for this case, but the API should not delete the image if it is not included in the dataset object when updating a dataset.

closes #1861 

/deploy #persist

### DIFF
--- a/client/src/project/datasets/change/DatasetChange.container.js
+++ b/client/src/project/datasets/change/DatasetChange.container.js
@@ -236,18 +236,22 @@ function ChangeDataset(props) {
     dataset.keywords = mappedInputs.keywords;
     dataset.creators = mappedInputs.creators.map(creator => getCreator(creator));
 
-    if (mappedInputs.image.options.length && mappedInputs.image.selected) {
-      const images = await uploadDatasetImages(mappedInputs.image, handlers);
-      if (images)
-        dataset.images = images;
+    if (mappedInputs.image.options.length) {
+      const imageSelected = mappedInputs.image.options[mappedInputs.image.selected];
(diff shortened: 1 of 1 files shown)


ciyer (changes requested):
The code changes look good, but when testing I encountered some problems, and maybe we can fix them as part of this issue, since it is so far not too big.

1. I get an error when I try to **create** a dataset that includes an image
image

(on dev, I can create a dataset and supply an image in the form -- create succeeds, but the image is not is not actually added 🤷 )

2. Once I have added an image to the form, there is no way to remove the image

andre-code:
> You can access the deployment of this PR at https://renku-ci-ui-1887.dev.renku.ch

Thanks for the review and testing, indeed there was something to fix in the creation of the dataset. 
about the delete option we can include it when this is ready https://github.com/SwissDataScienceCenter/renku-python/issues/2938

andre-code:
@ciyer  Thanks for the review and testing, indeed there was something to fix in the creation of the dataset. Fixed in the last commit.
about the delete option we can include it when this is ready https://github.com/SwissDataScienceCenter/renku-python/issues/2938

ciyer (approved):
Works great now! 🙏 Do not worry about the acceptance tests, those are fixed in another PR. You can merge even if they do not pass.

PR #1893, opened 2022-06-09
fix(client): hide download button for folders in lineage view

lorenzo-cavazzi:
As reported by @gavin-k-lee , the lineage preview shows a download button also for folders (see below). Clicking on it brings the user to a 404 page on GitLab.

This PR removes it -- that is already the case for the file preview.

Peek 2022-06-09 08-50

/deploy #persist


ciyer (approved):
Thanks for the quick fix!

PR #1899, opened 2022-06-13
fix: editing datasets without images

ciyer:
#1887 introduced an error in editing datasets that do not contain images. This PR makes dataset actions work in the following scenarios:

- Add a dataset without an image
- Add a dataset with an image
- Modify a dataset without an image
- Modify a dataset with an image

/deploy #persist


andre-code (approved):
Great, thanks for fix it.

Issue #1903, opened 2022-06-14
WebSocket client side


lorenzo-cavazzi:
## Motivation

As a follow-up to #1338 , we should implement the client-side of the WebSocket channel.

## Mock-ups / Design

_TO DO_
_See temporary PR for now_

## Progress

- [ ] API Design (sprint-??)
- [ ] UI Implementation

lorenzo-cavazzi:
this was done in #2013

PR #1888, opened 2022-06-07
feat(server): websocket

lorenzo-cavazzi:
_Still a draft_  :warning:
_For the documentation, check the README file in the `/server/src/websocket` folder._

Websocket channel available at `wss://<URL>/ui-server/ws`

/deploy #persist
re #1338

Issue #1338, opened 2021-05-03
Set up a WebSocket channel between client and server


lorenzo-cavazzi:
The UI client should connect to a WebSocket channel with the abilities to:
- [x] _[design]_ define the payload, a strategy to store the data both server and client side
- [x] get push notifications for generic events (e.g. started/stopped sessions)
- [x] handle tokens expiration --> this requires notifying the client when the credentials expire
~~- [ ] get push notifications for specific events for which the UI can subscribe/unsubscribe (e.g. new commits pushed to a specific project the user is visualizing)~~

The last 2 points require a sort of communication dictionary in both directions.
Expired credentials could be notified by the server with:
```
{
    scope: "authentication",
    data: "expired"
}
```
Requests to subscribe to `target-namespace/target-project` events could be notified by the client with:
```
{
    scope: "project",
    target: "target-namespace/target-project",
    data: "subscribe:all"
}
```
This requires to be expanded to cover more cases.

A stub of the first point is available in #1331

ciyer:
We decided to just make minimal changes to prove that this is working on the client, we will later make the more extensive changes when there is a good opportunity.

PR #1904, opened 2022-06-14
feat(client): websocket

lorenzo-cavazzi:
This is a preliminary stub for the client-side WebSocket code. Mind that most of the logic is missing.

To simulate the client connection, you can use Postman.
The address for the connection is `wss://<deployment>.dev.renku.ch/ui-server/ws`.
Remember to provide the cookies in the headers for `ui-server-session`

Refer to #1338 and #1903 for further explanations. It contains the server-side code from #1888

/deploy #persist


lorenzo-cavazzi:
We decided to limit the client to the UI version info -- it will still receive messages when sessions change, but we don't need to query APIs to refresh the local status. Move that to a new issue

lorenzo-cavazzi:
superseded by #2013

PR #1911, opened 2022-06-16
chore(chart): modify security context

olevski:
This does a few things:
- Changes the security context defaults in the values file for the server so that the "global" pod-level security context is not set but rather the container-level security context is used. We cannot set the global security context for the whole pod because it interferes with the certificates container which needs elevated permissions
- Adds a security context to the nginx server in the ui client component, this already ran as non root which is great but it still does not hurt to make the security context modifiable through the helm chart and not have it hardcoded.

/deploy #persist


ciyer (approved):
Thanks for providing these improvements.

PR #1919, opened 2022-06-21
1830 e2e tests

andre-code:
PR e2e tests added:
- when create a New Dataset (all fields)

Upload dataset files
- when is a .zip file
- when multiple files
- when delete file 
- when error uploading file

/deploy renku=tests-ui-2.5.0


ciyer (approved):
Very nice!

PR #1921, opened 2022-06-23
fix: check if autosave commit exists in project commit list before use it

andre-code:
PR to handle error where autosave commit id of autosave is used which does not exist in project commit list.

/deploy #persist


ciyer (approved):
👍

PR #1926, opened 2022-06-24
fix: error when notebook doesnt have value

### LINKED ISSUE #1906: Crash on view file
## Description

This notebook causes the UI to crash

https://dev.renku.ch/projects/cramakri/inflation/files/blob/qtm/QTM-Part1.ipynb

## Fix

We should do two things:

- Catch this error in the file renderer and show a message the the content cannot be displayed
- Fix the problem with the file rendering

andre-code:
PR to fix error when open notebook.
Solution: Process Markdown only if it contain valid props.

#### How to  test:
Open https://renku-ci-ui-1926.dev.renku.ch/projects/cramakri/inflation/files/blob/qtm/QTM-Part1.ipynb and its equivalent in gitlab https://dev.renku.ch/gitlab/cramakri/inflation/blob/master/qtm/QTM-Part1.ipynb the content displayed should be the same

closes #1906 

/deploy renku=tests-ui-2.5.0 #persist


ciyer (approved):
🤩

PR #1929, opened 2022-06-24
feat: show status message in notebook view

### LINKED ISSUE #1879: Show meaningful information on the session page while starting
## Motivation

Sessions take time to start, 1 minute (or more) is not uncommon.
Especially when clicking on the quickstart button, users may be confused since they need to wait for such a long time with no feedback apart from the bouncing dots. The status line on top saying `Started x seconds ago` adds to the confusion.

## Mock-ups / Design

We could improve the feedback by:
* Changing the status line to `Starting`.
* Adding better feedback in the main window.
    - A short message saying the sessions are starting and it may take a minute or two
    - A progress bar instead of the bounding dots
    - Some feedback from the `status.message` content. It currently mentions the pods with a non-ready state, maybe that should be changed a bit. The idea is to give the user the perception that something is going on in the background. E.G. showing `starting x of 5 components`, or come up with a short descriptive sentence based on what still needs to start.

## Progress

- [ ] API Design (sprint-??)
- [ ] UI Implementation

andre-code:
PR to display better session status in notebook view

Closes #1879 

/deploy #persist

### DIFF
--- a/client/src/notebooks/Notebooks.present.js
+++ b/client/src/notebooks/Notebooks.present.js
@@ -52,6 +52,7 @@ import { EnvironmentLogs, LogDownloadButton, LogTabs, useDownloadLogs } from "..
 import { SessionStatus } from "../utils/constants/Notebooks";
 
 import "./Notebooks.css";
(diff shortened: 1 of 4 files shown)


andre-code:
Duplicated #2054

PR #1941, opened 2022-07-01
feat: add support for mermaid format in markdown view

andre-code:
This PR is to add support for mermaid code.

Screenshot 2022-07-04 at 17 19 50


ciyer (approved):
This is very nice! I need to start using more mermaid in my readme files.

PR #1945, opened 2022-07-07
feat: add pending and done label when change a value in project settings

andre-code:
PR to add confirmation when update a value in Project Settings.

/deploy #persist


ciyer (changes requested):
There are many usability improvements here! 

I found a few things that I think should be fixed.

1. Settings in a project without edit rights

image

In the case of a project for which the user does not have write privileges, I think we can hide the _Project Tags, Project Description, and Project Avatar_. The user cannot change them, and those fields are visible elsewhere. All we need to show are the _Clone Commands and Repository URL_.

(The sessions tab in a read-only project is fine as is).

2. Edit feedback

image

I think the update button should look differently if the user has not made any changes to the _Tags or Description_ fields. Right now, I cannot tell if I have made changes that need to be saved or not, and changing the button appearance would be a nice way to provide that feedback. One idea is to use an outline appearance and make the button reads-only if there are no changes to save (we could also add a tooltip to the update button ).

image

3. Conventions for the _Project Avatar_ action buttons

image

The colors of the _Submit_ and _Cancel_ buttons should be reversed, to be consistent with the conventions elsewhere (the "expected" action should be in green, the secondary action in blue)

andre-code:
@ciyer thanks for your review and suggestions. I have included a tooltip when the update button is disabled if there is nothing to update, styling for the cancel button and proper handling when the user does not have the right to modify the project.

ciyer (approved):

PR #1950, opened 2022-07-11
feat: Add kg search dates filter

### LINKED ISSUE #1936: UI/UX (search): Include filter by date
Include filter by date on the entity search page.

Options to filter updated/created entity:
- last week
- last month
- last 90 days
- oldest
- exact date (from date - to date)

Related https://github.com/SwissDataScienceCenter/renku-graph/issues/952

andre-code:
PR to add date filter in kg-search form

Closes #1936 

/deploy #persist

### DIFF
--- a/client/src/utils/components/dateFilter/DateFilter.tsx
+++ b/client/src/utils/components/dateFilter/DateFilter.tsx
@@ -1,0 +1,146 @@
+/*!
+ * Copyright 2022 - Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as React from "react";
+import { Input } from "../../ts-wrappers";
+import { ChangeEvent, useEffect, useState } from "react";
+import Time from "../../helpers/Time";
+
+/**
+ *  renku-ui
+ *
+ *  DateFilter.tsx
+ *  Date filter component
+ */
+
+export interface DateFilterProps {
+  handler: Function;
+  values: DatesFilter;
+}
+
+/* eslint-disable no-unused-vars */
+export enum dateFilterTypes {
+  all = "all",
+  custom = "custom",
+  last90days = "last90days",
+  lastMonth = "lastMonth",
+  lastWeek = "lastWeek",
+  older = "older" // before 90 days
+}
+
+export interface DatesFilter {
+  since?: string;
+  until?: string;
+  type?: dateFilterTypes;
+}
+/* eslint-enable no-unused-vars */
+
+const DateFilter = ({ handler, values }: DateFilterProps) => {
+  const [dates, setDates] = useState<DatesFilter>({});
+
+  useEffect(() => {
+    setDates(values);
+  }, []); // eslint-disable-line
+
+  useEffect(() => {
+    if (handler)
+      handler(dates);
+  }, [dates]); // eslint-disable-line
+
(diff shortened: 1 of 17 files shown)


client/src/utils/components/dateFilter/DateFilter.tsx
```diff
@@ -0,0 +1,144 @@
...
+          className="author-input"
+          checked={selectedType === item.value}
+          data-cy={nameInput}/>
+        <label className="px-2 author-label">{item.title}</label>
+      </div>
+    );
+  });
+  return (
+    <>
+      <h3 className="filter-label">Date</h3>
```
ciyer, line 137: I think we should be more explicit about what we are filtering on.

```suggestion
      <h3 className="filter-label">Creation Date</h3>
```
```suggestion
      <h3 className="filter-label">Date</h3>
```

ciyer (changes requested):
This new functionality works very well for the most part! 🎉 

I found few small things that should be improved. Some are mentioned in the code comments. Also

1. The label ("All", "Last Week", etc.) should be clickable, not just the radio button.
image

2. The state can get out of sync.

* Search for projects in the last week
image

* Then go to some page in the UI and return to the search page
image

Now the filter says "All," but it is actually applying the "Last Week" filter.



image

lorenzo-cavazzi (changes requested):
This looks great! A few things to adjust:

* I see you included a `package-lock.json` file version 1, but I believe we want to keep version 2 #1963
Could you remove it? You may want to upgrade your local npm/node version to prevent this in the future.

* When clicking on the button "Yes, remove all filters" I would expect the search term to be preserved and the filters on the left to be reset. Instead, it shows me again all the projects. Is that intended?
Screenshot_20220719_094548

These 2 following points might be problems on the `kg-search` branch. If you don't mind adding a commit here, it would be great to fix them. Otherwise, we could have a separate issue/PR

* Please give a look at the top bar in Firefox, the user icon wraps to the next line (upper part in the screenshot)
Screenshot_20220719_092735

* Sekhar mentioned that the label ("All", "Last Week", etc.) should be clickable, not just the radio button.
This should also be the case for the other options (visibility, type, ...) so that the UX is the same.

The code looks good! I quickly went through it, and I will give it a second look when the above details are fixed.
P.S: This new search looks great :star_struck:

andre-code:
Thanks @lorenzo-cavazzi  for your review and suggestions.
I have implemented the selection by label/icon, fixed the package-lock.json version and the functionality to remove filters.
I couldn't reproduce the error in firefox, I'm using firefox v101.0.1, in safari it works fine too. According to coteries design, the menu is going to disappear, there is no story yet but we can add tests to make it compatible with the version of firefox you are testing with.

lorenzo-cavazzi:
Great! I would say it's almost ready to be merged. A couple more details:
* Should the selectable label/icon filters show the same cursor as a button or link? It's the default cursor now but I feel it should be the other one (unless we got different hints from Coteries)
* The wrapping problem seems to depend on the environment where the browser is running -- if you try to increase/decrease the zoom (CTRL `+`) you should be able to reproduce it (also on Chrome, but in my case it's easier on Firefox). The culprit seems to be the class `menu-right` that imposes `max-width: 260px`. A strategy to fix it would be to:
  - remove the max-width
  - remove the `ms-lg-auto` on the child element that increases the margin on the left
  - make the search-box element expand the margins automatically to take up all the unoccupied space

P.S. let me know if you still can't reproduce it or if you don't find my solution feasible (understandable?). I'm happy to give it a try :)

andre-code:
> Great! I would say it's almost ready to be merged. A couple more details:
> 
> * Should the selectable label/icon filters show the same cursor as a button or link? It's the default cursor now but I feel it should be the other one (unless we got different hints from Coteries)

yes, it should definitely be cursor-pointer. good  👀. thx.

> P.S. let me know if you still can't reproduce it or if you don't find my solution feasible (understandable?). I'm happy to give it a try :)

I have tried (CTRL +) and it works fine. If you have time, go ahead otherwise we will create an Issue

lorenzo-cavazzi (approved):
I've added a commit to prevent the elements from wrapping when the zoom is too big.
Notice that I extended the max search bar size to 900px (from 800) to simplify the CSS -- otherwise I would have needed a few extra `@media` rules.

Please give it a try before merging (it might be easier to run it through telepresence).
After merging, it would be nice to open a PR to merge master back to `kg-search` and be sure the layouts work well together -- we will also be able to run the acceptance tests.

P.S. sorry for keeping this pending for that long. The new search bar works great! Can't wait to see it deployed :star_struck:

ciyer (approved):
This works very well, looks good, and I really like the use of RTK-Query here. 🎉

PR #1962, opened 2022-07-15
chore: run ca cert injection as non-root

olevski:
The latest certificates image and chart can run as non-root.

This is safer than what we had before.

Also the chart and image tags are not tracking each other. This is on purpose.

Lastly, the `.lock` file is not needed to be tracked in git. We always pin the requirements versions so this is not necessary to check into git and keep updating.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,6 @@ virtualenv
 
 # All traces of the GitLab server running locally
 /gitlab
+
+# Ignore helm charts requirements lock file
+requirements.lock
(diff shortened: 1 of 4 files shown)


lorenzo-cavazzi (approved):
Lgtm! Thanks!

P.S: I added a tiny commit and changed the deploy string cause the UI currently requires a specific version of the renku repo

olevski:
The acceptance tests failed on uploading a dataset which is completely unrelated to this change.

PR #1968, opened 2022-07-21
chore(client): update readthedocs links

### LINKED ISSUE #1967: Link to upgrading renku project info is broken
## Description

The link on the instructions for upgrading is broken - see attachment.

image

This is what I see when I click on it:

image

The link in the page is this one:

https://renku.readthedocs.io/en/latest/how-to-guides/upgrading-renku.html#upgrading-your-image-to-use-the-latest-renku-cli-version

## Reproduce

Find out of date project, click on information for upgrading and click on this link.

lorenzo-cavazzi:
update readthedocs links

fix #1967

### DIFF
--- a/client/src/project/status/RenkuVersionStatus.present.js
+++ b/client/src/project/status/RenkuVersionStatus.present.js
@@ -112,7 +112,7 @@ function AutoUpdateButton({ externalUrl, maintainer, migration_status, onMigrate
   </p>;
 }
 
-const docUrl = Docs.rtdHowToGuide("upgrading-renku.html#upgrading-your-image-to-use-the-latest-renku-cli-version");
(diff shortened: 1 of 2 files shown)


andre-code (commented):
Thanks for the update.
Question, if the instruction is to update the template version, it should go to https://renku.readthedocs.io/en/latest/renku-python/docs/reference/commands.html#update-a-project-s-template instead of the migrate instruction, no?

lorenzo-cavazzi:
I think it makes sense to keep the link to `#module-renku.ui.cli.migrate` since it's a bit more generic.
It suggests running `renku migrate -c` that actually includes info for both the renku migration and the template update

> If you created your repository from a project template and the template has changed since you created the project, it will also update files with their newest version from the template, without overwriting local changes if there are any.

PR #1969, opened 2022-07-21
build: keep the values.client structure

lorenzo-cavazzi:
Changes to keep the `ui` values under `ui` instead of moving to `ui.client`
I personally would keep the current structure cause it's cleaner, but this will prevent introducing a breaking change for admins.

/deploy renku=ui1901-unify-ui-charts #persist


andre-code (approved):
LGTM 🚀

lorenzo-cavazzi:
This is not needed anymore, we are going for a different strategy

PR #1972, opened 2022-07-22
build: set resources

lorenzo-cavazzi:


olevski:
@lorenzo-cavazzi why are we setting the requests now?

I think a nicer solution to this is to add a horizontal pod autoscaler and set resources and at least cpu limits. And the hpa can also scale at 80% of the limits.

And the values files can just ask for the max number of replicas the hpa can go to.

lorenzo-cavazzi:
That is something we are setting anyway with values at a later point, I thought it would make sense to set them here.
I agree setting also autoscaling would be nice -- isn't that complementary to these changes? I see a similar pattern in the UI server and it seems to work well

lorenzo-cavazzi:
Let's simplify -- #1973

PR #1973, opened 2022-07-25
build(client): change security context user

lorenzo-cavazzi:
Change security context user -- should allow telepresence v1 to run properly again


andre-code (approved):
It worked for me! 🙌👏

lorenzo-cavazzi:
Telepresence works, but there are problems with getting the JSON values (i.e. nginx reading files). This requires more fine-tuning :unamused:

ciyer (changes requested):
I think this can be closed since #1987 allows us to switch to telepresence 2 for the UI client.

lorenzo-cavazzi:
Yes, this can be closed

Issue #1964, opened 2022-07-19
Implement [Project card in landing page and project lists, Dataset cards, Forms] from Coteries design system


gavin-k-lee:
What remains to be implemented from the design system apart from [fonts, colours, buttons and cards]?

andre-code:
Posible design changes to do:

- Project card in landing page and project lists **1** .
![Screenshot 2022-07-21 at 09 47 38](https://user-images.githubusercontent.com/43388408/180159205-df67255c-efce-4bcb-b474-d89383187f37.png)


- Dataset card in dataset list **1** 
Main button action to be defined
![Screenshot 2022-07-21 at 09 47 53](https://user-images.githubusercontent.com/43388408/180159242-306b7be8-65cc-4222-9b02-f012c3dd3066.png)

- Form styles (checkbox, dropdown, input, input with buttons) **1.**
https://zeroheight.com

- kg-search filters and search field **2**
![Screenshot 2022-07-21 at 09 50 39](https://user-images.githubusercontent.com/43388408/180159790-10775484-85af-4dd7-8f21-8b9aa314ea09.png)


- Project view header **1**
No include mail-icon
![Screenshot 2022-07-21 at 09 51 10](https://user-images.githubusercontent.com/43388408/180159904-0795c987-74cb-451c-af13-dd6956a0e648.png)

- Project view overview (it is posible to display: Creator, Created on, n datasets, picture) 0.5
![Screenshot 2022-07-21 at 09 52 20](https://user-images.githubusercontent.com/43388408/180160176-7a012c2b-5946-48de-8d72-9e92659bc7b2.png)

andre-code:
To be implemented in a feature branch:
- Project card in landing page and project lists
- Dataset cards
- Forms

gavin-k-lee:
Hi @andre-code - I just checked with the Projects - mail icon - I believe we have that metadata in the KG:
```
{
 ...
  "identifier": 19373,
  "path": "lee.gavin.k/test",
  "name": "test",
  "visibility": "public",
  "created": {
    "creator": {
      "email": "lee.gavin.k@gmail.com", <-------------- We could use mailto:lee.gavin.k@gmail.com?
      "name": "Gavin Lee"
    },
    "dateCreated": "2022-02-14T04:28:43Z"
  },
```
Is this something we can use?

andre-code:
Yes, we can use that email. I will include it like so.

PR #1978, opened 2022-07-29
feat: add styles for project and dataset cards (#1964)

andre-code:
This PR add the styles for project and dataset cards and the button to start a session from the project card.

card-styles

Ref #1964 

#### Pending
- [ ] Fix acceptance tests

/deploy renku=1978-styles-cards-projects-and-datasets #persist


andre-code:
Note: we agreed with Ordes not to merge it into master but into a feature branch while making the internal card for projects and datasets.

ciyer (changes requested):
This looks good!

For the most part, I think the code is good, I think it could benefit from introducing a few more components in `client/src/utils/components/List.js`. I made a commit that starts this, but I think it could go a bit further.

The other thing that would be nice is to use SVG for the background images. I think the PNG files look grainy.

andre-code:
> This looks good!
> 
> For the most part, I think the code is good, I think it could benefit from introducing a few more components in `client/src/utils/components/List.js`. I made a commit that starts this, but I think it could go a bit further.


Thanks for the feedback, I have also added new components in the List.js. Note that in kg-search that component is refactored, so in this PR I only focus on the styles and don't make this PR so extensive.


> 
> The other thing that would be nice is to use SVG for the background images. I think the PNG files look grainy.

That grainy effect is intentional in the design, those resources in Figma and zeroheight are only available in png, and I think it makes sense because they are 2 images combined.

ciyer (approved):
👍

PR #1979, opened 2022-07-29
feat: add styles for form elements (#1964)

andre-code:
This PR apply new design system for form elements (input, input-groups, checkbox, radio, select*, button-groups).
_Note: To customize the display of the selection options it is necessary to create a custom component, it will not be included here._

Design System: https://zeroheight.com/7c23c970c/p/90b4ba-renku---design-system

form styles

Rel #1964 
/deploy #persist


ciyer (approved):
This looks very good! 🎉 I made some small scss refactorings to remove duplication, and I think this can be merged.

PR #1987, opened 2022-08-04
telepresence 2 take 2

ciyer:
A second attempt to set up telepresence 2, replacing #1799.

Telepresence (2) works for the client, but the server seems to encountering this issue: https://github.com/telepresenceio/telepresence/issues/2654

/deploy #persist

### DIFF
--- a/client/run-telepresence.sh
+++ b/client/run-telepresence.sh
@@ -185,20 +185,9 @@ tee > "./public/sitemap.xml" << EOF
 </urlset>
 EOF
 
-# The following is necessary if we start telepresence with --run-shell
-# echo "================================================================================================================="
-# echo "Once telepresence has started, type the following command to start the development server:"
-# echo "BROWSER=none npm start"
-# echo "================================================================================================================="
-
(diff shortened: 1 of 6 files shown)


lorenzo-cavazzi (approved):
Thanks for fixing this!
I added a commit with a couple of changes:
- Security context: there were some conflicts when setting both `securityContext` and `podSecurityContext` that led to the `tel-agent-init` pod to fail with a permission denied error (`Fatal: can't open lock file /run/xtables.lock: Permission denied`)
- the `automountServiceAccountToken` setting prevented the token to be avilable in the `traffic-agent` pod resulting in another error (`unable to get the Kubernetes InClusterConfig: open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory`)

With these changes, it works fine on both client and server. Please, give it a try before merging.

ciyer:
> With these changes, it works fine on both client and server. Please, give it a try before merging.

Yes! Everything works now!

> Thanks for fixing this! 

My only contribution was to keep returning to this annoying task and not giving up. The actual fixes came from you and @olevski. ;)

PR #1989, opened 2022-08-08
feat: use new notebook server error response format

andre-code:
Use new notebook server error format
rel https://github.com/SwissDataScienceCenter/renku-notebooks/pull/849

### testing
Start a session in which the branch does not exist, you should be able to see the error message in the alert and in the error notification
error notebook

/deploy renku-notebooks=feat-standardize-error-messages #persist

### DIFF
--- a/client/src/api-client/notebook-servers.js
+++ b/client/src/api-client/notebook-servers.js
@@ -118,8 +118,8 @@ function addNotebookServersMethods(client) {
     }).then(resp => {
       return resp.data;
     }).catch(error => {
-      if (error.errorData && error.errorData.messages && error.errorData.messages.error) {
-        const err = new Error(error.errorData.messages.error);
+      if (error.errorData && error.errorData.error && error.errorData.error.message) {
+        const err = new Error(error.errorData.error.message);
         err.cause = error;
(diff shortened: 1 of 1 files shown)


ciyer (commented):
I tested, and this works well. I just have one question for clarification to make sure we are using the right fields. 

When I try to start a session on a branch that does not exist, I get this error:

image

This error message is not very helpful, but the `detail` field in the response tells me what the problem is:

image

```
{
  "error": {
    "code": 1404, 
    "detail": "Missing resources and errors: branch testing does not exist", 
    "message": "Cannot start the session because the required Git or Docker resources are missing."
  }
}
```

Is `message` the right field to use, or should me file an issue against `renku-notebook` that the message should be more informative?

andre-code:
@ciyer thanks for pointing that out and yes, the message is the correct field to use. Tasko will update that message to make it more informative.

ciyer:
> @ciyer thanks for pointing that out and yes, the message is the correct field to use. Tasko will update that message to make it more informative.

Then I will approve.

ciyer (approved):
👍

Issue #1981, opened 2022-08-02
Host the Inter typeface on Renku


ciyer:
## Description

Renku needs to support air-gapped deployments, so we need to host all resources, including typefaces ourselves.

https://github.com/SwissDataScienceCenter/renku-ui/blob/master/client/src/styles/fonts/_fonts.scss

PR #1996, opened 2022-08-11
feat: host Inter font (#1981)

andre-code:
This RP is to host the font "Inter".

/deploy #persist


client/src/styles/fonts/_fonts.scss
```diff
@@ -3,8 +3,21 @@
...
 */
 
-/* Webfont: Calcutta-Bold */
-@import url('https://fonts.googleapis.com/css2?family=Inter:wght@400;600&display=swap');
+@font-face {
+  font-family: 'Inter';
+  font-style: normal;
+  font-weight: 400;
+  src: local('Inter Regular'), local('Inter-Regular'),
+  url('../fonts/Inter/static/Inter-Regular.ttf') format('ttf');
```
ciyer, line 11: I suspect this path is not correct.

ciyer (changes requested):
I think the path to the font is not correct. This is what it looks like on my Mac

image

That is in **Helvetica**.

It should like like this:

image

(E.g., look at the dot on the _lowercase-i_. It is square in Helvetica, round in Inter).

client/src/styles/fonts/_fonts.scss
andre-code: yes, was missing `font-display: swap;`

Should work now in all browsers

ciyer (approved):
👍

PR #1998, opened 2022-08-12
build(chart): remove duplicates in client/server charts

ciyer:
Remove duplicate values from chart

### DIFF
--- a/helm-chart/renku-ui/templates/ui-client-deployment-template.yaml
+++ b/helm-chart/renku-ui/templates/ui-client-deployment-template.yaml
@@ -44,11 +44,11 @@ spec:
           {{- end }}
           env:
             - name: BASE_URL
-              value: {{ .Values.baseUrl | default (printf "%s://%s" (include "ui.protocol" .) .Values.global.renku.domain) | quote }}
+              value: {{ .Values.client.url | default (printf "%s://%s" (include "ui.protocol" .) .Values.global.renku.domain) | quote }}
             - name: GATEWAY_URL
-              value: {{ .Values.gatewayUrl | default (printf "%s://%s/api" (include "ui.protocol" .) .Values.global.renku.domain) | quote }}
+              value: {{ .Values.gateway.url | default (printf "%s://%s/api" (include "ui.protocol" .) .Values.global.renku.domain) | quote }}
             - name: UISERVER_URL
-              value: {{ .Values.client.uiserverUrl | default (printf "%s://%s/ui-server" (include "ui.protocol" .) .Values.global.renku.domain) | quote }}
+              value: {{ .Values.server.url | default (printf "%s://%s" (include "ui.protocol" .) .Values.global.renku.domain) | printf "%s/ui-server" | quote }}
             - name: WELCOME_PAGE
               value: {{ .Values.client.welcomePage.text | b64enc | quote }}
             {{- if .Values.client.statuspage }}
--- a/helm-chart/renku-ui/templates/ui-server-deployment.yaml
+++ b/helm-chart/renku-ui/templates/ui-server-deployment.yaml
@@ -43,8 +43,6 @@ spec:
               value: {{ .Values.server.url | default (printf "%s://%s" (include "ui.protocol" .) .Values.global.renku.domain) | quote }}
             - name: UI_SERVER_VERSION
               value: {{ .Chart.Version | quote }}
-            - name: SERVER_PORT
-              value: {{ .Values.server.port | default (printf "8080") | quote }}
             - name: GATEWAY_URL
               value: {{ .Values.gateway.url | default (printf "%s://%s/api" (include "ui.protocol" .) .Values.global.renku.domain) | quote }}
             - name: GATEWAY_LOGIN_PATH
--- a/helm-chart/renku-ui/templates/ui-server-deployment.yaml
+++ b/helm-chart/renku-ui/templates/ui-server-deployment.yaml
@@ -40,21 +40,19 @@ spec:
               protocol: TCP
           env:
             - name: SERVER_URL
-              value: {{ .Values.server.serverData.url | default (printf "https://%s" .Values.global.renku.domain) | quote }}
+              value: {{ .Values.server.url | default (printf "%s://%s" (include "ui.protocol" .) .Values.global.renku.domain) | quote }}
             - name: UI_SERVER_VERSION
               value: {{ .Chart.Version | quote }}
             - name: SERVER_PORT
-              value: {{ .Values.server.serverData.port | default (printf "8080") | quote }}
-            - name: SERVER_PREFIX
-              value: {{ .Values.server.serverData.prefix | default (printf "/ui-server") | quote }}
+              value: {{ .Values.server.port | default (printf "8080") | quote }}
             - name: GATEWAY_URL
-              value: {{ .Values.server.gateway.url | default (printf "https://%s/api" .Values.global.renku.domain) | quote }}
+              value: {{ .Values.gateway.url | default (printf "%s://%s/api" (include "ui.protocol" .) .Values.global.renku.domain) | quote }}
             - name: GATEWAY_LOGIN_PATH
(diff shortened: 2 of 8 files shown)


lorenzo-cavazzi (changes requested):
I think we might want to merge the URLs in the values so that we don't have something like `server.serverData.url` + `server.serverData.prefix` anymore, but just `uiServerUrl` or even better `server.url`. The same applies to similar values like `gatewayUrl` that currently comes from different sources: `gatewayUrl` for the client and `server.gateway.url` for the server.

In the PR, we define `ui.uiserverUrl` in the helpers taking it from `ui.server.*`, which is confusing and not coherent with what we wrote here https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/values.yaml.changelog.md#upgrading-to-renku-0150 

I suggest aiming for the following structure:
```
ui:
  gatewayUrl:
  client:
    url:
    *
  server:
    url:
    *
  global:
    *
```

I would remove all the unnecessary `global.*` values, an probably also `ui.ingress` (is it actually used? If yes, then `ui.client.ingress` and its server counterpart are probably redundant). And give it a look to other values not used anymore, so that we have breaking changes only once and then we resume working with a clean values structure.

P.S: switching to `ui.gatewayUrl` and to `ui.server.url` will likely require changes to the server and/or client code. I don't mind if we keep them separate as `server.url.base` + `server.url.prefix` (easier to adapt the server code) or we just merge to `server.url` (easier to adapt for the client code) as long as we define them only once in the values.

helm-chart/renku-ui/templates/ui-client-deployment-template.yaml
```diff
@@ -46,9 +46,9 @@ spec:
             - name: BASE_URL
               value: {{ .Values.baseUrl | default (printf "%s://%s" (include "ui.protocol" .) .Values.global.renku.domain) | quote }}
```
lorenzo-cavazzi, line 47: I would change also the `baseUrl` to be `client.url`

helm-chart/renku-ui/values.yaml
```diff
@@ -228,16 +233,11 @@ server:
 
   affinity: {}
 
-  serverData:
+  route:
     url:
```
lorenzo-cavazzi, line 237: Should we add a comment to explain that this URL should be the "origin" (i.e. schema + domain) only?

helm-chart/renku-ui/values.yaml
```diff
@@ -228,16 +233,11 @@ server:
...
     port: 8080
     prefix: /ui-server
```
lorenzo-cavazzi, line 239: Should we add a comment to explain this is the "pathname"?

lorenzo-cavazzi (changes requested):
Thank you, this goes in the right direction!
I would change the `baseUrl` to `client.url` to follow the same structure of the gateway.
I'm also tempted to suggest dropping the `url`+`prefix` structure for the server. In that case, we would have just a full URL as `server.url`. The reason is that I find the naming a bit confusing since it's different than what we do with gateway and client; also, `url` is in fact what is usually defined as `origin`, and `prefix` is the `pathname`.
We might actually just limit to adding a comment to clarify that in the values, or maybe change the actual names so that they are `server.origin` and `server.pathname` (+ `server.port`).

Do you have any preference for how to handle that? I would probably go with the last suggestion (i.e. server.pathname + server.origin + server.port) or the first one (server.url, intended as full URL, and server.port -- that will require a little change to the config file to separate pathanme and origin).

ciyer:
> Thank you, this goes in the right direction! I would change the `baseUrl` to `client.url` to follow the same structure of the gateway. I'm also tempted to suggest dropping the `url`+`prefix` structure for the server. In that case, we would have just a full URL as `server.url`. The reason is that I find the naming a bit confusing since it's different than what we do with gateway and client; also, `url` is in fact what is usually defined as `origin`, and `prefix` is the `pathname`. We might actually just limit to adding a comment to clarify that in the values, or maybe change the actual names so that they are `server.origin` and `server.pathname` (+ `server.port`).
> 
> Do you have any preference for how to handle that? I would probably go with the last suggestion (i.e. server.pathname + server.origin + server.port) or the first one (server.url, intended as full URL, and server.port -- that will require a little change to the config file to separate pathanme and origin).

I like your suggestions. I myself stumbled on the fact that the `url` was not actually the `url`, which is annoying. I think it should be possible to further clean up the terminology so that the structures have enough flexibility but with sensible names.

helm-chart/renku-ui/values.yaml
ciyer: As discussed, this can just be dropped from the values file. We do not need to give admins control over the pathname.

helm-chart/renku-ui/templates/ui-server-deployment.yaml
```diff
@@ -40,21 +40,21 @@ spec:
...
-              value: {{ .Values.server.serverData.url | default (printf "https://%s" .Values.global.renku.domain) | quote }}
+              value: {{ .Values.server.url | default (printf "%s://%s" (include "ui.protocol" .) .Values.global.renku.domain) | quote }}
             - name: UI_SERVER_VERSION
               value: {{ .Chart.Version | quote }}
             - name: SERVER_PORT
-              value: {{ .Values.server.serverData.port | default (printf "8080") | quote }}
+              value: {{ .Values.server.port | default (printf "8080") | quote }}
             - name: SERVER_PREFIX
-              value: {{ .Values.server.serverData.prefix | default (printf "/ui-server") | quote }}
+              value: "/ui-server"
```
lorenzo-cavazzi, line 49: Keeping the env variable here and using a hard-coded string seems a bit counter-intuitive.
I would completely remove `SERVER_PREFIX` from here, and just hardcode it in server/src/config.ts in `SERVER.prefix`.

lorenzo-cavazzi (changes requested):
Thank you! This looks cleaner to me. Just a minor change on the hard-coded "/ui-server" value and it's ready to get merged

lorenzo-cavazzi (approved):
Looks good now, thanks!

helm-chart/renku-ui/values.yaml
```diff
@@ -166,6 +173,9 @@ client:
       tag: "0.11.16"
 
 server:
+  ## The URL for the server service; the URL used by the client is the server.url + /ui-server endpoint.
+  url:
+  port: 8080
```
olevski, line 178: There is really no point in making this a parameter in the values file. Just hardcode the value and remove the option to set it. I see that the value is not even properly templated in all places. I.e. it is hardcoded in the service and in the list of ports in the deployment. Also you do not even use this value to access the server, you use the service port which is different. You can leave the service port as is in the values file below.

server/src/config.ts
```diff
@@ -18,22 +18,23 @@
...
 
 const DEPLOYMENT = {
   gatewayUrl: process.env.GATEWAY_URL || SERVER.url + "/api",
   gatewayLoginUrl:
-    (process.env.GATEWAY_URL || SERVER.url + "/api") + (process.env.GATEWAY_LOGIN_PATH || "/auth/login"),
+    (process.env.GATEWAY_URL || SERVER.url + "/api") +
+    (process.env.GATEWAY_LOGIN_PATH || "/auth/login"),
   gatewayLogoutUrl:
-    (process.env.GATEWAY_URL || SERVER.url + "/api") + (process.env.GATEWAY_LOGOUT_PATH || "/auth/logout"),
+    (process.env.GATEWAY_URL || SERVER.url + "/api") +
```
olevski, line 36: I know this was the same before you made this change. But does javascript have a `urljoin` or equivalent function that is smart about joining urls and avoiding double slashes in the final url. Not sure how much of a problem double slashes are in the first place.

olevski (changes requested):
Sekhar, as we discussed I took a quick look at this helm chart PR. I think these changes look really good. 

Just two minor things.

olevski (approved):
Thank you!

Issue #1781, opened 2022-04-08
Autostart URLs should allow to directly open a notebook


lorenzo-cavazzi:
## Motivation

We already open JupyterHub sessions directly into a notebook from the file viewer.
Some users expressed interest in having quickstart links for the sessions doing the same thing (@rokroskar ).

## Proposal

Technically it is not too difficult, although this may need some coordination with renku-notebooks since the current mechanism is based on JupyterLab, and that may not be there anymore for R-based images.

This probably belongs to the "Create shareable link" modal. An idea could be adding an expandable menu that allows selecting (through a radio button) a target notebook to open.
When clicking on `[set default notebook]`, the UI should scan the files in the project and identify the notebooks, listing them (verify the feasibility here for projects containing thousands of files). The selection should always include also `none` to reset a previous choice.

## Context

_[this is the current modal, not a mockup]_

![Screenshot_20220407_080954](https://user-images.githubusercontent.com/43481553/162386294-f816213a-6432-480a-b932-f5cec3c6291c.png)

rokroskar:
Thanks @lorenzo-cavazzi for writing this up! I think the proposal makes sense. cc @volodymyrss

PR #2000, opened 2022-08-15
show the correct dialog when starting a session on an autosave branch 

### LINKED ISSUE #1947: Autosaves occasionally mess up with the "Start button" session
## Description

Occasionally, having an autosave in a project mess up with the "Start button" when trying to start a session.
In the following GIF, I show how clicking on the "Start session" button actually triggers the "create sharable link" modal.

Deleting the autosaves _and after refreshing the list of autosaves_ (in the video by moving out an it the sessions page again) solves it.

starting with autosaves

## Reproduce

For some reason, I can't always reproduce it and I'm not sure why. In the video, I used a pet project on limited.
It seems plausible that the problem is in the `checkServer` function here https://github.com/SwissDataScienceCenter/renku-ui/blob/master/client/src/notebooks/NotebookStart.present.js#L1137-L1210

Not sure if something changed there (maybe check with renku-notebooks for the autosaves endpoint?), but the `.filter` function not matching the commit could cause the `toggleShareLinkModal` function to be triggered on the Start button instead of the `handlers.startServer`

## Actions

* [ ] Investiage the issue (and open issues in other repos if necessary)
* [ ] Fix the issue

ciyer:
Undo a change that was introduced in #1781, which opens the wrong dialog.

Fix #1947

## Test

To test, fork this project:

https://sekhar.dev.renku.ch/projects/cramakri/test-accessing-gpus

Then, in your fork, create a branch based on `renku/autosave/cramakri/master/2f88472/2f88472` with the name `renku/autosave/{your username}/master/2f88472/2f88472`

/deploy #persist

### DIFF
--- a/client/src/notebooks/NotebookStart.present.js
+++ b/client/src/notebooks/NotebookStart.present.js
@@ -1138,11 +1138,13 @@ class ServerOptionLaunch extends Component {
   checkServer() {
     const { filters } = this.props;
     const { autosaved } = this.props.data;
+    const selectedBranchName = filters.branch.name;
+    const selectedCommitShort = filters.commit.id.substr(0, 7);
     const current = autosaved.filter(c =>
-      c.autosave.branch === filters.branch.name && c.autosave.commit === filters.commit.id.substr(0, 7));
+      c.autosave.branch === selectedBranchName && c.autosave.commit === selectedCommitShort);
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi (approved):
Lgtm :rocket:

PR #2002, opened 2022-08-16
feat(server): handle anon-id cookies and headers

### LINKED ISSUE #1601: Generate the `anon-id` header
## Motivation

The gateway still generates a `session` cookie and an `anon-id` cookie to identify the user.
This should now be handled only by the UI server.

## Proposal

* Always generate a session ID even when users don't hit the login endpoint, and preserve it permanently (tip: use `getOrCreateSessionId` instead of `getSessionId`, and verify we never remove it)
* Provide this id to the requests forwarded to the gateway as `anon-id` cookie (tip: either add the proper cookie in `renkuAuth`, or create a support function to invoke in `proxyMiddleware` so that `onProxyReq` still removes unnecessary cookies but adds this one when needed)
* extra: clean up the unnecessary code in the gateway (e.g. https://github.com/SwissDataScienceCenter/renku-gateway/blob/master/app/auth/notebook_auth.py#L55)

lorenzo-cavazzi:
This PR handles the generation of anonymous users' ids.
To keep it as a non-breaking change, `set-cookie` for the `anon-id` values is temporarily removed from responses. The logic can be dropped once the gateway doesn't set it anymore.

/deploy #persist
fix #1601


server/src/routes/apis.ts
```diff
@@ -66,6 +71,24 @@ const proxyMiddleware = createProxyMiddleware({
         res.json({ error: "Invalid authentication tokens" });
       }
     }
+
+    // remove setting anon-id cookies from the gateway -- should be temporary until gateway gets updated
+    const setCookie = null ?? clientRes.headers["set-cookie"];
```
ciyer, line 76: As discussed, this logic here is a little confusing, and it will go away once https://github.com/SwissDataScienceCenter/renku-gateway/pull/584 is merged... let us remove this and just make this PR dependent on the changes to the gateway.

ciyer (changes requested):
Looks good -- I think we can make this PR dependent on https://github.com/SwissDataScienceCenter/renku-gateway/pull/584. I don't see much benefit in de-coupling these two changes.

server/src/routes/apis.ts
lorenzo-cavazzi: I thought the outcome was to keep the logic for a while to prevent the breaking change even if it's a bit confusing and clarify it with a comment.
Anyway, we can try to remove it and add the constraint if you prefer.

ciyer (approved):
This works well and is the right place for this functionality, I think.

PR #2003, opened 2022-08-18
feat: new session start flow

### LINKED ISSUE #1990: Move the session start form and launch info out of the project page
In preparation for #1988, we should
- hide tabs and other information from session start form 
- hide tabs and other information from the launch info page
- remove the session listing in the project page

In the end, the start flow should look something like the following:

Image

ciyer:
Develop a new session start flow with more emphasis on the session and less on the project.

image

# Testing

Try starting sessions with and without options and see that the interaction is smooth.

I'm still not sure about the behavior after closing a session -- currently, the user gets taken back to the main project page. I think that feels better than going to watch the session shut down, since you usually do not care, but we can discuss this.

/deploy renku=ui-tests-2.8.0 renku-gateway=master #persist
Fix #1990

### DIFF
--- a/client/src/notebooks/NotebookStart.present.js
+++ b/client/src/notebooks/NotebookStart.present.js
@@ -1194,19 +1194,21 @@ class ServerOptionLaunch extends Component {
       </Button>;
 
 
-    return [
-      imageStatusAlert,
-      startBaseButton,
-      " ",
-      startButtonWithMenu,
+    return <div className="d-flex flex-row-reverse">
+      <div>
+        {imageStatusAlert}
+        {startBaseButton}
+        {" "}
+        {startButtonWithMenu}
+      </div>
       <AutosavedDataModal key="modal"
         toggleModal={this.toggleModal.bind(this)}
         showModal={this.state.showModal}
         currentBranch={this.state.current}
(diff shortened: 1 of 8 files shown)


lorenzo-cavazzi (changes requested):
The code looks great!

A couple of details I noticed with the layout changes:

- Now that we have a 2 columns layout, I would take the full space on the right side so that the content looks centered -- it feels misaligned at large/xl screen size.
  Screenshot_20220831_111357

- There is some issue with text overflow when the screen size goes down to medium -- it should be easy to reproduce by picking a long project title
  Screenshot_20220831_105820

andre-code:
@ciyer Great to see the session with the new navigation flow so the user has full attention on the form.

Regarding the session form, when the user click to start the session there are some fields that are still available to change, is it for any particular reason? maybe not leaving them active or hiding them would be the best option?

session start

ciyer:
Thank you @lorenzo-cavazzi and @andre-code for your reviews! The latest commits should fix the problems you identified.

lorenzo-cavazzi (approved):
All works well now :tada:

andre-code (approved):
🚀

Issue #1984, opened 2022-08-04
[Sprint 5] Implement Project + Dataset headers design system


lokijuhy:
+ Release Cards + Headers + Forms feature branch

[Coteries Iteration 6 Figma Prototype](https://www.figma.com/proto/cmVH2YlFGz6uAv2EEH1sBZ/Renku---User-interface?page-id=1%3A2&node-id=29%3A756&viewport=1613%2C-2648%2C0.6&scaling=scale-down&starting-point-node-id=29%3A756)

![Image](https://user-images.githubusercontent.com/7387902/183636810-2519ec30-f60a-4ebc-97b5-e49c8b6f0039.png)

ciyer:
View changes are easy, but the interaction with the settings page where user can update some of these values could be a bit more complicated, since we need to take care to refresh the information in the right way.

PR #2005, opened 2022-08-22
feat: refresh project header after update tags and description (#1984)

andre-code:
PR to fresh project header after update tagList and description values.

refresh header after update taglist and description

/deploy renku=1978-styles-cards-projects-and-datasets  #persist


ciyer (approved):
Works very well!

PR #2006, opened 2022-08-22
feat: hide project header in dataset view (#1984)

andre-code:
PR to hide the project header in the view of a project dataset and modify a dataset.

project and dataset header

/deploy renku-gateway=master renku=1978-styles-cards-projects-and-datasets #persist


ciyer (approved):
Very nice! 🎉

PR #2010, opened 2022-08-24
build: changes to support redux-toolkit use

ciyer:
Internal changes necessary to support redux-toolkit.

/deploy #persist


lorenzo-cavazzi (approved):
This works well, thank you!

PR #2012, opened 2022-08-26
fix(server): handle websocket auth exceptions

lorenzo-cavazzi:
This PR fixes a couple of auth-relate issues

- Prevent unnecessary errors for anonymous users.
- Send an error message when session cookies are missing (unlikely to happen in the UI, but possible when using an external tool like  Postman).


ciyer (approved):
This is good -- though these problems should not happen in normal usage, it's good to get messages that should explain what is happening in debugging situations.

PR #2015, opened 2022-08-29
fix: wrong file tree display when changing to original project from fork (#1907)

### LINKED ISSUE #1907: Wrong file tree display when changing to original project from fork
Go to https://renkulab.io/projects/rok.roskar/atmospheric-vertical-profiles/files/blob/.renku (.renku folder opened)
image

Click on the link to get to the original project from the fork:
image

Click "Files"
image

See 
image

Hit F5, see actual files:
image

The original project does not have `.renku/metadata`, but instead `.renku/datasets`, but when switching to it this way, the old file tree is shown

andre-code:
PR to fix wrong file tree display when changing to original project from fork.

Closes #1907 

### DIFF
--- a/client/src/project/Project.state.js
+++ b/client/src/project/Project.state.js
@@ -127,7 +127,7 @@ const FileTreeMixin = {
     if (oldTree.loaded === false)
       return this.initialFetchProjectFilesTree(client, openFilePath, openFolder);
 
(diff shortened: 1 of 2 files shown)


ciyer (approved):
🎉 Tested and the bug is fixed.

PR #2017, opened 2022-08-30
Fix a few UI glitches

### LINKED ISSUE #2014: Issues browsing projects as an anonymous user
There are a couple of glitches when browsing the UI as an anonymous user:
- On the project's overview page, the number of stars is updated correctly but there is an endless spinning wheel.
  Screenshot_20220829_101207
- When clicking on the global Projects page, it occasionally shows no projects. This seems to be cause by a few extra queries to the projects endpoint including the `membership=true` flag (this seems to be caused by a redirect).
  Screenshot_20220829_101357
- After going to the Projects page, the back button of the browser doesn't behave correctly anymore, the user can go further back (likely to be related to the above auto-redirect)

lorenzo-cavazzi:
This PR contains a few fixes. I split them in different commits so that it's easier to review the code changes separately:
1. The star button displayed an endless spinning wheel for anonymous users. Not it shows the start and it restores the behavior on click (will redirect to the login to continue)
  Screenshot_20220830_161208
2. Adjust the margins on search containers by reducing margin in equal parts on both left and right sides, instead of left only (in the screenshot, the left screen shows balanced margins, while on the right screen there is a smaller space on the left and a bigger one on the right).
  Screenshot_20220830_161041
3. Prevents invoking multiple times the projects API with the same parameters when reaching the projects page. This was worse in the case of anonymous users since they were also redirected, adding an extra query for their own projects that always returns empty since they can't have any own projects (in the screenshot, there are 3 queries on the right side, while only 1 on the left).
  Screenshot_20220830_164725
4. The projects page redirected anonymous users to the public projects page using `history.push`, therefore preventing any backward navigation.

fix #2014

### DIFF
--- a/client/src/project/Project.present.js
+++ b/client/src/project/Project.present.js
@@ -405,7 +405,8 @@ class ProjectViewHeaderOverview extends Component {
   star(event) {
     event.preventDefault();
     // do nothing if the star state is not yet loaded
-    if (this.props.starred == null) return;
+    if (this.props.starred == null && this.props.user?.logged)
+      return;
     this.setState({ updating_star: true });
     this.props.onStar().then(
       result => {
@@ -415,6 +416,7 @@ class ProjectViewHeaderOverview extends Component {
 
   render() {
     const metadata = this.props.metadata;
+    const logged = this.props.user?.logged;
 
     let starElement;
     let starText;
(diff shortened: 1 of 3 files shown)


andre-code (approved):
The code looks good and works as expected!
Thanks for the corrections :)

PR #2022, opened 2022-09-02
chore: update package-lock.json

andre-code:
PR to expose metadata from Renku projects as json-ld structured data.


client/src/features/projects/ProjectKgApi.ts
```diff
@@ -0,0 +1,39 @@
...
+};
+
+export const projectKgApi = createApi({
+  reducerPath: "projectKgApi",
+  baseQuery: fetchBaseQuery({ baseUrl: "/ui-server/api/kg" }),
+  endpoints: (builder) => ({
+    projectJsonLd: builder.query<any, string>({
+      query: (projectPath) => ({ url: `projects/${projectPath}`, headers }),
+      transformResponse: (response: any) => {
+        return response;
```
ciyer, line 31: Is this necessary? Looks like it was added for debugging, but the response isn't modified anymore.

andre-code: > This works well, and I like how you implemented it! 🤩
> 
> I found one thing in the code that is unnecessary, I think, but otherwise this looks great.

you are right, it is not necessary. I removed it.
Thanks for your review 🙌

andre-code:
> This works well, and I like how you implemented it! 🤩
> 
> I found one thing in the code that is unnecessary, I think, but otherwise this looks great.


@ciyer  you are right, it is not necessary. I removed it.
Thanks for your review 🙌

ciyer (dismissed):
👍

ciyer (dismissed):
👍

ciyer (approved):

Issue #2018, opened 2022-08-31
Do not strip certain cookies in ui-server


olevski:
Needed to implement core service scaling.

The sticky session cookie should not be stripped.

Followup from https://github.com/SwissDataScienceCenter/renku-python/pull/3094

lorenzo-cavazzi:
A few info: the relevant code is here https://github.com/SwissDataScienceCenter/renku-ui/blob/master/server/src/routes/apis.ts#L45-L57
We can move the `proxyMiddleware` component to its own file. The name of the cookies _not_ to be stripped should possibly come from the values.yaml file (loaded in the `config` as usual). The function `getCookieValueByName` might come in handy here.

PR #2029, opened 2022-09-09
feat(server): keep some cookies when proxying

olevski:
Hello friends :)

Hopefully this is acceptable. We need it so that we can start to implement sticky sessions in the core service.

closes https://github.com/SwissDataScienceCenter/renku-ui/issues/2018

/deploy #persist


ciyer (commented):
This looks good, but I want to delay merging until 2.9.0 is released.

ciyer (approved):
Thanks for contributing this improvement! 

Although the JS community typically is ok with adding packages that have just one straightforward 10-line function as a dependency, we try to avoid it in this project, so I have modified your initial PR to remove a dependency, but it otherwise looks good! Hopefully it is sufficient to satisfy your use case.

PR #2031, opened 2022-09-09
fix(server): anonymous id set to fixed value

### LINKED ISSUE #2030: All anonymous sessions have the same user id in Renku 0.17.0
This line sets the user id to a constant value in all cases:
https://github.com/SwissDataScienceCenter/renku-ui/blob/master/server/src/routes/apis.ts#L54

So the anonymous id in the cookie is fixed. Both parts are hardcoded in the config.

And it is always `anon-anon-id`.

This means that every anonymous user has the same id and can see absolutely all anonymous sessions.

olevski:
fixes #2030 

Someone should take a closer look to make sure we do not need any other changes.

/deploy renku=1978-styles-cards-projects-and-datasets renku-gateway=master #persist


lorenzo-cavazzi (approved):
Thank you for the fix! This accidentally slipped in :pray:

PR #2032, opened 2022-09-12
feat(server): prevent gateway from setting anon users cookies

lorenzo-cavazzi:
This should prevent propagating the headers to set the anon-id related cookies from the gateway to the UI since that is managed by the UI server now.

/deploy #persist

### DIFF
--- a/server/src/routes/apis.ts
+++ b/server/src/routes/apis.ts
@@ -74,21 +74,17 @@ const proxyMiddleware = createProxyMiddleware({
 
     // Prevent gateway from setting anon-id cookies. That's not needed in the UI anymore
     const setCookie = null ?? clientRes.headers["set-cookie"];
-    if (setCookie != null && setCookie.length) {
-      if (setCookie.length === 1) {
-        if (setCookie[0].startsWith(config.auth.cookiesAnonymousKey))
-          clientRes.headers["set-cookie"] = null;
-      }
-      else {
-        const allowedSetCookie = [];
-        for (const cookie of setCookie) {
-          if (!cookie.startsWith(config.auth.cookiesAnonymousKey))
-            allowedSetCookie.push(cookie);
-        }
-        if (allowedSetCookie.length < setCookie.length)
-          clientRes.headers["set-cookie"] = allowedSetCookie;
-      }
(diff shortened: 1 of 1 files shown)


server/src/routes/apis.ts
```diff
@@ -71,6 +71,24 @@ const proxyMiddleware = createProxyMiddleware({
...
     }
+
+    // Prevent gateway from setting anon-id cookies. That's not needed in the UI anymore
+    const setCookie = null ?? clientRes.headers["set-cookie"];
+    if (setCookie != null && setCookie.length) {
+      if (setCookie.length === 1) {
+        if (setCookie[0].startsWith(config.auth.cookiesAnonymousKey))
+          clientRes.headers["set-cookie"] = null;
+      }
+      else {
```
olevski, line 82: @lorenzo-cavazzi will the part in the `else` clause not handle cases where `set-cookie` is only setting a single cookie? I think it will. Why do you have two separate cases for when there is one `set-cookie` vs many.

lorenzo-cavazzi: I guess I can merge the 2 parts to make it less confusing if you think it's for the good.
My idea was that, most of the time, this will match the first set of `if` hitting line 80. If we have another case having multiple cookies (this shouldn't happen atm), then the `for` cycle would handle it better but slower -- ok I know, the performance difference is negligible here.

ciyer, line 82: Can't this be simplified to 
```
    // Prevent gateway from setting anon-id cookies. That's not needed in the UI anymore
    const setCookie = null ?? clientRes.headers["set-cookie"];
    if (setCookie == null || setCookie.length == 0) return;
    const allowedSetCookie = [];
    for (const cookie of setCookie) {
      if (!cookie.startsWith(config.auth.cookiesAnonymousKey))
        allowedSetCookie.push(cookie);
    }
    if (allowedSetCookie.length < 1)
      clientRes.headers["set-cookie"] = null;
    else
      clientRes.headers["set-cookie"] = allowedSetCookie;
```

ciyer (changes requested):

olevski (approved):

PR #2033, opened 2022-09-13
feat: add page for projects that are not in kg (#2020)

### LINKED ISSUE #2020: UI/UX (search): Show user projects that are not in KG
## Motivation

Now that we have the cross-entity search that returns the entities that exist in kg, the user must have an alternative to find those projects that do not have active kg so it is not possible to find them in the cross-entity search.

The user must be able to:
- list projects that are not in kg
- have the option to activate kg to several projects in the list
- view the progress of activation

Rel [#1038](https://github.com/SwissDataScienceCenter/renku-graph/issues/1038)

andre-code:
PR to add page of projects that are not in KG

Closes #2020 


server/src/websocket/handlers/activationKgStatus.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+  return statuses;
+}
+
+async function handlerRequestActivationKgStatus(
+  data: Record<string, unknown>, channel: Channel, socket: ws): Promise<void> {
+  // save the request enabler
+  if (data.projects) {
+    const projectsIds = data.projects as number[];
+    channel.data.set("projectsIds", data.projects);
```
lorenzo-cavazzi, line 53: Wouldn't this overwrite other ids already there? E.G. the user activates 2 projects, then another one just one second later. We would lose the data about the first 2 projects.

Should we instead check if there are already ids in the list (something like `channel.data.get("projectsIds")`) and add the new ones?

server/src/websocket/handlers/activationKgStatus.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (!projectsIds.length)
+      return;
+
+    const statuses = await getActivationStatus(projectsIds as number[]);
+    const dataResponse = JSON.stringify(statuses);
+    channel.data.set("activationProjects", statuses);
+    const response = { start: true, message: dataResponse };
+    socket.send((new WsMessage(response, "user", "activation")).toString());
```
lorenzo-cavazzi, line 61: I feel this shouldn't be here, but part of the heartbeat logic. This seems to be potentially time-consuming; the setup might just save the ids and let the heartbeat handle getting/updating the current status.

server/src/websocket/handlers/activationKgStatus.ts
```diff
@@ -0,0 +1,85 @@
...
+  if (projectsIds?.length && previousStatuses) {
+    // remove ids are complete
+    const ids: number[] = (projectsIds).filter( (id: number) => previousStatuses[id] != 100);
+
+    channel.data.set("projectsIds", ids.length ? ids : []);
+    const statuses = await getActivationStatus(ids);
+
+    if (!util.isDeepStrictEqual(previousStatuses, statuses)) {
+      channel.data.set("activationProjects", statuses);
+      const response = { start: true, message: JSON.stringify(statuses) };
```
lorenzo-cavazzi, line 78: `start: true`  seems a leftover from the above handler function.

server/src/websocket/handlers/activationKgStatus.ts
```diff
@@ -0,0 +1,85 @@
...
+
+async function getActivationStatus(projectIds: number[]): Promise<ActivationStatus> {
+  const { gatewayUrl } = config.deployment;
+  let statuses = {};
+  for (let i = 0; i < projectIds.length; i++) {
+    const id = projectIds[i];
+    const activationStatusURL = `${gatewayUrl}/projects/${id}/graph/status`;
+    logger.info(`Fetching activation from ${id} projects`);
+
+    const response = await fetch(activationStatusURL);
```
lorenzo-cavazzi, line 40: `for` + `await` seems a dangerous combination. I see this potentially taking more than the 5 seconds heartbeat.

I would consider:
- parallilizing it: you do not `await` but send out multiple requests at the same time.
- using a semaphore logic: you save the info about "I started a heartbeat" with a boolean variable, and in the next iteration you check if it's still ongoing -- in that case, you "skip a turn"

A further optimization would be considering just 5 projects per time. That would be a bigger change since you would need to keep track of all the projects the user would like to index, then take the first 5, start indexing and wait until all is done. Then you move on to the next 5.
This would prevent sending tens of queries, but since indexing projects is a once-in-a-lifetime event we don't want to waste too much time on optimizing it (I mentioned the optimization just as a general suggestion for cases like this :slightly_smiling_face: ).

lorenzo-cavazzi (changes requested):
The logic here is already very solid!
The one thing to fix is how received ids are saved since previous ids might get lost in the process. It's an easy fix (see inline).

Furthermore, I have 2 suggestions to simplify the logic and make it more robust:
- do _not_ perform a heartbeat action in the handler -- or in other words, just save the ids in the handler. This simplifies following the logic and prevents annoying cases (E.G. the following heartbeat might happen just a few milliseconds after the setup, and you don't want to duplicate the queries -- in general, just save the necessary data in the handler, and delegate all the business logic to the heartbeat function)
- I'm a bit concerned about the loop that goes through all the ids at once and `await` for the responses one by one. We should either parallelize the queries or use a semaphore logic

client/src/utils/components/Tooltip.js
```diff
@@ -64,7 +64,7 @@ function ThrottledTooltip(props) {
 
   const toggle = standardToggler(tooltipOpen, setTooltipOpen, lastToggleTime, setLastToggleTime);
 
-  return <Tooltip placement="top" target={props.target} isOpen={tooltipOpen} toggle={toggle}
+  return <Tooltip placement="top" target={props.target} isOpen={tooltipOpen} toggle={toggle} autohide={false}
```
ciyer, line 67: Is there a reason to set `autohide={false}` for all tooltips? Would it make sense to accept an override to `autohide` as a property and only change in the cases where it is necessary?

client/src/landing/Landing.present.js
```diff
@@ -145,6 +149,35 @@ class Welcome extends Component {
   }
 }
 
+export function ProjectsInactiveKG() {
```
ciyer, line 152:
```suggestion
export function ProjectsInactiveKGWarning() {
```

client/src/features/inactiveKgProjects/components/KgActivationHeader.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <>
+      <h1 className="activationHeader d-flex gap-2 my-3 align-items-center">
+        Projects Inactive in the Knowledge Graph
+        <FontAwesomeIcon id="activation-question" className="cursor-pointer" size="sm" icon={faQuestionCircle} />
+      </h1>
+      <ThrottledTooltip
+        target="activation-question"
+        tooltip={tooltip} />
```
ciyer, line 43: See the suggestion below regarding the change to ThrottledTooltip

```suggestion
      <ThrottledTooltip
       autohide={false}
        target="activation-question"
        tooltip={tooltip} 
        />
```

client/src/features/inactiveKgProjects/InactiveKgProjects.tsx
```diff
@@ -0,0 +1,220 @@
...
+            ))}
+          </tbody>
+        </Table>
+
+      </div>
+    ) :
+    (
+      <div className="col-md-12 p-4 border-radius-8 bg-white">
+        <div className="d-flex justify-content-center align-items-center gap-2">
+          <Balloon size="30" /> Great. There are no pending projects to activate Knowledge Graph.</div></div>
```
ciyer, line 205:
```suggestion
          <Balloon size="30" /> All projects are in the Knowledge Graph.</div></div>
```

ciyer (changes requested):
This works very well! 

I only tested the success cases, since all my attempts to activate worked, and I think it worked very well. I like how I can navigate away from the page and still get notified and see the outcome once a project is activated -- that works very well! 🎉 

I have a few minor comments on the code. Also, if possible, I think it would be good to have 2 cypress tests (one for the success case and one for a failure case), but I do not know how much effort there is in realize such a test. If it is very complicated, then we can discuss this.

I have a

server/src/websocket/handlers/activationKgStatus.ts
andre-code: great suggestion, I changed to added the new ids instead of overwrite it

server/src/websocket/handlers/activationKgStatus.ts
andre-code: true, removed!

server/src/websocket/handlers/activationKgStatus.ts
andre-code: thanks for pointing this out, this part definitely needed to be improved, I have chosen to parallelize all the requests

client/src/utils/components/Tooltip.js
andre-code: sure, I added the autoHide property to our tooltip component

server/src/websocket/handlers/activationKgStatus.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    fetch(activationStatusURL).then(async (response) => {
+      const status = await response.json();
+      const previousStatuses = channel.data.get("activationProjects") as ActivationStatus;
+      const currentStatus = status?.progress ?? -1;
+      const previousStatus = previousStatuses ? previousStatuses[id] : null;
+      if (!util.isDeepStrictEqual(previousStatus, currentStatus)) {
+        const currentStatuses = { ...previousStatuses, [id]: currentStatus };
+        sendMessage(JSON.stringify(currentStatuses), channel);
+        channel.data.set("activationProjects", currentStatuses);
```
lorenzo-cavazzi, line 50:
```suggestion
        channel.data.set("activationProjects", currentStatuses);
        // eliminate possible duplicates
        channel.data.set("projectsIds", [...new Set(ids)]);
```
Duplicates may happen (not very likely, but maybe some latency and a repeated click from the user..).
Passing the array to a Set will prevent it.

server/src/websocket/handlers/activationKgStatus.ts
```diff
@@ -0,0 +1,94 @@
...
+
+function getActivationStatus(projectIds: number[], channel: Channel): void {
+  const { gatewayUrl } = config.deployment;
+  for (let i = 0; i < projectIds.length; i++) {
+    const id = projectIds[i];
+    const activationStatusURL = `${gatewayUrl}/projects/${id}/graph/status`;
+    logger.info(`Fetching activation from ${id} projects`);
+
+    fetch(activationStatusURL).then(async (response) => {
+      const status = await response.json();
```
lorenzo-cavazzi, line 43: Both `fetch()` and `.json()` usually have a non-negligible chance to throwing an error. It would be nice to include them in a `try .. catch` block.

lorenzo-cavazzi: More in general, having all the fetch functions in a single place (like in the client) would be nice. Would you mind creating a new folder `server/src/api-client` and moving the function there? It would include computing the URL, fetching and parsing to json -- a pattern we already use in the client.

lorenzo-cavazzi (changes requested):
This looks really good! The WebSockets part works well, I added a suggestion to prevent duplicates.

A nice-to-have would be moving the `fetch` logic into its own function and handling possible errors with a `try --catch` block, since fetching and parsing the response may throw errors. The pattern would be similar to what we already do in the client.
After that, it's ready to be merged :slightly_smiling_face:

ciyer (approved):
All my requests have been addressed. I think this works very well!

lorenzo-cavazzi (approved):
This feature is a great addition! Can't wait to see the feature branch in action!

P.S. I added a tiny PR to fix a couple of lining warnings -- it was quicker than adding the comments

PR #2038, opened 2022-09-15
FEATURE BRANCH - Shape Up #1, Workflows UI

lorenzo-cavazzi:
This is a feature branch for bringing the workflows into the UI. It implements the solution proposed in the ([Workflows UI ShapeUp document](https://github.com/SwissDataScienceCenter/renku-rfc/blob/main/feature-pitches/001-project-workflows-ui-view/001-project-workflows-ui-view.md)).

Missing pieces:
- Using RTK instead of the coordinators
- prevent concurrent operations when the core service is already busy

renku-python refrence: SwissDataScienceCenter/renku-python/pull/3135
/deploy renku-core=develop extra-values=gateway.gitlabUrl=https://gitlab.dev.renku.ch/ #persist


client/src/utils/components/entityHeader/EntityHeader.tsx
```diff
@@ -73,45 +73,43 @@ function EntityHeader(
       <div className={`card-header-entity--large card-header-entity--${itemType}-large`}>
         <div className="card-bg-title card-bg-title--large" data-cy={`${itemType}-title`}>{title}</div>
         <div className="d-flex justify-content-between align-items-center m-3">
-          <EntityLabel type={itemType} />
-          <VisibilityIcon visibility={visibility} className="" />
+          <EntityLabel type={itemType} workflowType={null} />
+          { visibility ? (<VisibilityIcon visibility={visibility} className="" />) : null }
```
ciyer, line 77: Why `className=""`? Couldn't that just be left out?

client/src/utils/components/entities/Duration.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <>
+      <p id={elemId}>
+        <span className="text-rk-text small">Runs in</span> {Time.getDuration(duration)}
+      </p>
+      <UncontrolledTooltip key={`duration-elem-${elemId}`} placement="top" target={elemId}>
+        <span>Esitmated runtime</span>
+      </UncontrolledTooltip>
+    </>
+  );
```
ciyer, line 46: I think it makes sense to make the description for the duration configurable. It is not necessary to add it to the props, but this way, it will be easy to make that change later.
```suggestion
  if (duration == null)
    return null;
  const elemId = `duration-${workflowId}`;
  const durationDesc = "Estimated runtime";

  return (
    <>
      <p id={elemId}>
        <span className="text-rk-text small">Runs in</span> {Time.getDuration(duration)}
      </p>
      <UncontrolledTooltip key={`duration-elem-${elemId}`} placement="top" target={elemId}>
        <span>{durationDesc}</span>
      </UncontrolledTooltip>
    </>
  );
```

client/src/workflows/Workflows.present.tsx
```diff
@@ -0,0 +1,910 @@
...
+  const filtered = !showInactive ? workflows.filter(w => w.active) : workflows;
+
+  // ? Pre-sort by a unique prop to guarantee consistency
+  const preSorted = filtered.sort((a, b) => (a["name"] > b["name"]) ? 1 : ((b["name"] > a["name"]) ? -1 : 0));
+  // ? Then second-sort by last execution
+  const secondSorted = preSorted.sort((a, b) => (a["lastExecuted"] < b["lastExecuted"]) ? 1 :
+    ((b["lastExecuted"] < a["lastExecuted"]) ? -1 : 0));
+
+  // ? Final sorting
+  const sorted = secondSorted.sort((a, b) => (a[orderBy] > b[orderBy]) ? 1 : ((b[orderBy] > a[orderBy]) ? -1 : 0));
```
ciyer, line 229: I do not follow the logic here. I think the last invocation of sort will determine how the collection is sorted. Why not do it like this?:

```
function workflowCompare(orderBy, a, b) {
  const orderByCompare = (a[orderBy] > b[orderBy]) ? 1 : ((b[orderBy] > a[orderBy]) ? -1 : 0);
  const lastExecutedCompare = a["lastExecuted"].localeCompare(b["lastExecuted"]);
  const nameCompare = a["name"].localeCompare(b["name"]);
  if (orderByCompare !== 0) return orderByCompare;
  if (lastExecutedCompare !== 0) return lastExecutedCompare;
  return nameCompare;
}
[...]
 const sorted = filtered.sort((a, b) => workflowCompare(orderBy, a, b);
```

ciyer (changes requested):
Nice functionality in here! All of this code has already been reviewed before, but I found a few small things to improve in the code. The changes should be small, and then we can get this feature merged! 🎉

client/src/utils/components/entityHeader/EntityHeader.tsx
lorenzo-cavazzi: Ops, that's an oversight -- I'll remove the `className` property

client/src/workflows/Workflows.present.tsx
lorenzo-cavazzi: The first pre-sorting is necessary because many properties used for `orderBy` might have the same value (E.G: author, runtime, ...) and the workflows API returns items in a non-deterministic order. Since we occasionally refresh the list of workflows, the order of the elements seems to change at random when clicking on a new workflow.

The second ordering is a request from the shape-up group: always order by lastest execution on ties -- sadly, that's not enough to remove the above pre-sorting since we still might have ties.

lorenzo-cavazzi: I tried the proposed solution, but it requires a few adjustments because
* `localeCompare` doesn't seem to work on non-string elements (like dates) and
* it doesn't take into account that some elements might be missing.

lorenzo-cavazzi:
Thank you for the review!
I implemented all the suggestions apart from the new solution for sorting. I added a verbose explanation of the reasons behind adopting the overcomplicated multi-stage sorting, and what is not working in the suggested alternative -- I'm still open to update it if you think we should simplify it.

ciyer (approved):
Looks great! I think there still could be some improvements to that workflow sorting code, but I see that it is not as straightforward as I had hoped, and I think we can live with it as is for the time being.

ciyer (approved):
👍

PR #2039, opened 2022-09-16
chore(client): save and propagate socket object

lorenzo-cavazzi:
Propagate WebSocket object so that the `.send()` function can be invoked everywhere at any point.

/deploy #persist


ciyer (approved):
👍

PR #2041, opened 2022-09-16
Shape Up #1 - First prorotype: workflows browser

lorenzo-cavazzi:
PR for the first prototype. This will include a workflows browser

/deploy renku-python=activities-list-service #persist


client/src/workflows/Workflows.present.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  return (
+    <Row className="my-3">
+      <Col xs={12} sm="auto" className="my-auto">
+        <div className="form-check form-switch d-inline-block">
+          <Label className="text-rk-text me-2">Exclude inactive</Label>
+          <Input type="switch"
+            id="wfExcludeInactive" label="label here" className="form-check-input rounded-pill"
+            checked={excludeInactive} onChange={() => toggleExcludeInactive()}
+          />
```
ciyer, line 63: I need to think a bit to understand what I am viewing here and what will be shown if I toggle this switch:

image


I think it would be better to flip it and make the switch `Show inactive` (turned off by default).

ciyer (changes requested):
The new code looks good, and there are some nice refactorings of the existing components here! 🎉 

I have one comment to make an interaction a bit clearer. Also, I tested on a few different projects and encountered one that fails to work:

https://renku-ci-ui-2041.dev.renku.ch/projects/cramakri/inflation/workflows. Here I get:

image

`workflow_plans.list` returns
```
{
    "error": {
        "code": -32404,
        "reason": "Not Found"
    }
}
```

client/src/workflows/Workflows.present.tsx
lorenzo-cavazzi: You are right, this is unnecessarily confusing. I changed it to follow your suggestion

lorenzo-cavazzi:
Thanks for the review.
I'm re-deploying to include a backend fix, let's see if the error you mentioned above goes away :crossed_fingers:

lorenzo-cavazzi:
@ciyer thanks for sharing the error. That is an older project (v8) and I haven't handled that case.
Since workflows will be available only for projects ">= v9", I will add a warning suggesting to upgrade the project to interact with workflows in the UI.

lorenzo-cavazzi:
I added the logic to check the project version before querying the backend. Please mind that the warning message is pretty minimal and I will add more info and the link to the project status in the 2nd prototype (I would need to port back a few changes to the `Url` object that I've already done there -- it's faster if I address this in the other branch directly)

image

ciyer (approved):
🎉 I like how this is "shaping up"!

PR #2043, opened 2022-09-19
feat: styles kg search (#2021)

### LINKED ISSUE #2021: UI/UX (search): Apply card design in cross entity search form
Apply the new card layout to the cross entity search.

In a previous case we applied all the new card layout to the project list and the current dataset list, this issue is to apply that layout to the components created for the cross entity search which will have the same style and keep the kg-search branch updated.

https://zeroheight.com/7c23c970c/p/40ab03-search-card/b/391560

andre-code:
PR to update kg-search styles.

kg search styles

Figma: https://www.figma.com/file/cmVH2YlFGz6uAv2EEH1sBZ/Renku---User-interface?node-id=3%3A91
Closes #2021

/deploy #persist


ciyer (changes requested):
Code looks very good, but while using, I noticed an interaction that I think should be a bit different.

I think the results should scroll and the filters should stay fixed.

image

If you scroll into a position like this

image


And then change the filter, everything suddenly jumps around.

Otherwise, everything looks great!

ciyer (approved):
Works very well! Excited to see this progressing. 🎉

PR #2048, opened 2022-09-22
build: release 2.9.0

### LINKED ISSUE #2047: Release 2.9.0
Tag and release 2.9.0

ciyer:
Release 2.9.0

Fix #2047 

/deploy #persist

### DIFF
--- a/helm-chart/renku-ui/values.yaml
+++ b/helm-chart/renku-ui/values.yaml
@@ -51,7 +51,7 @@ client:
 
   image:
     repository: renku/renku-ui
-    tag: "2.8.2"
+    tag: "2.9.0"
     pullPolicy: IfNotPresent
 
     ## Optionally specify an array of imagePullSecrets.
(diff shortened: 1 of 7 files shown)


helm-chart/renku-ui/values.yaml
```diff
@@ -51,7 +51,7 @@ client:
 
   image:
     repository: renku/renku-ui
-    tag: "2.8.2"
+    tag: "2.9.0"
```
olevski, line 54: I see that the server (further below in the values file) is not updated. Are they synced? Is that supposed to be updated or not?

olevski (changes requested):

helm-chart/renku-ui/values.yaml
ciyer: Good catch; thanks a lot! You are right that they should be in sync and I have just fixed it.

PR #2049, opened 2022-09-23
fix: use existing anon. user id if present in cookie

olevski:
This will make it so that we check if there is the old cookie the gateway used to set for anonymous users. If the cookie is present we use it for anonymous user id. If not then we set our own - ie we use the session id then.

With this users will not lose access to their anonymous sessions.

**Note though that** if someone used 2.9.0 already and has anonymous sessions with 2.9.0 already then using this would result in their pre-2.9.0 anonymous sessions showing up (if they still have the anonymous id cookie) but their 2.9.0 anonymous sessions not showing up.

So I am not really sure it is worth merging this tbh. I just wanted to understand what was going on in the ui-server better and decided to do this as a learning experiment.


olevski:
Yeah I will close this because it means we break/interrupt users potentially twice instead of once.

PR #2050, opened 2022-09-26
Shape Up #1 - Second prototype: workflow page and improved workflow browser

lorenzo-cavazzi:
This prototype includes:
- A revamped workflows browser with a tree-like view (only the full-size browser for now)
- A separate workflow detail page. That will likely be removed in the next prototype, and the details will be shown directly in the file browser.

Testing could include trying to visualize workflows with a couple of different projects (and verifying outdated projects show a warning) and playing with the filter to verify they work correctly.
I removed the link to workflow details from the list, but I left the logic to visualize a single workflow for testing purposes -- that will be adjusted in the 3rd prototype (one needs to manually change the URL `<project_id>/workflows/<workflow_id> removing the `s` in workflows).

image

renku-python reference: SwissDataScienceCenter/renku-python/pull/3152


client/src/workflows/Workflows.container.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const [orderByAscending, setOrderByAscending] = useState(false);
+  const [excludeInactive, setExcludeInactive] = useState(true);
+  const toggleAscending = () => setOrderByAscending(!orderByAscending);
+  const toggleExcludeInactive = () => setExcludeInactive(!excludeInactive);
+
+  const minVersion = 9;
+  const projectVersion = versionUrl != null && versionUrl.length ?
+    parseInt(versionUrl.replace(/^\/+|\/+$/g, "")) :
+    0;
+  const unsupported = projectVersion && projectVersion < minVersion ? true : false;
```
ciyer, line 64: I think this should be refactored into a function, with`MIN_WORKFLOW_SUPPORTED_VERSION` made more clearly into a global constant.

client/src/utils/components/Tree.tsx
```diff
@@ -0,0 +1,135 @@
...
+    });
+  }
+
+  let currentIndentation = "indentation-";
+  if (!indentation)
+    currentIndentation += 0;
+  else if (indentation >= 4)
+    currentIndentation += 4;
+  else
+    currentIndentation += indentation;
```
ciyer, line 110: This is interesting, but I'm not sure it is the best approach for implementing this. I think it would be better to just set the margin-left property of the style, or, if you wanted to do this with CSS, then setting the style to `{"--renku-tree-indent": indentation}` and using a var in the style (see above).

client/src/styles/components/_renku_tree.scss
```diff
@@ -0,0 +1,68 @@
...
+    background-color: $rk-background-0;
+  }
+  
+  &.selected {
+    background-color: $rk-light !important;
+  }
+
+  &.indentation-1 {
+    margin-left: 1em;
+  }
```
ciyer, line 21: I think this way might be too brittle. How about something like :
```suggestion
  &.indentation {
   --renku-tree-indent: 1;
    margin-left: calc(1em * var(--renku-tree-indent));
  }
```

client/src/utils/components/Tree.tsx
```diff
@@ -0,0 +1,135 @@
...
+            <h5>{title}</h5>
+            <EntityCreators display="tree" creators={creators} itemType={itemType} />
+          </Col>
+          <Col xs={12} sm={7} md={4} className="title center-vertically">
+            <EntityExecutions display="tree" executions={executions} itemType={itemType} lastExecuted={lastExecuted} />
+          </Col>
+          <Col xs={12} sm={5} md={3} className="title center-vertically">
+            <TimeCaption caption="Created" className="text-rk-text-light" time={timeCaption} endPunctuation="" />
+            {/* <Link to={urlSingle}><FontAwesomeIcon className="text-rk-yellow float-end" icon={faLink} /></Link> */}
+          </Col>
```
ciyer, line 127: I think it would be easier/better to lay this out using flexbox layout than using the bootstrap grid.

client/src/workflows/Workflows.state.ts
```diff
@@ -66,11 +92,157 @@ class WorkflowsCoordinator {
...
+  }
+}
+
+
+/**
+ * Temporary function to simulate a response
+ * @param workflowId - real id to keep the simulated response data coherent (also for the cache comparison)
+ * @returns fake response
+ */
+// ! TODO: move this to a cypress test
```
ciyer, line 159: Was this forgotten or will this be done in a later PR?

client/src/utils/components/Tree.tsx
lorenzo-cavazzi: Good point, I'll go with the margin-left property of the style; it makes this more readable and raises the priority when considering the marking-left prop.

client/src/styles/components/_renku_tree.scss
lorenzo-cavazzi: I ended up removing this and optioning for the "element style" solution

client/src/workflows/Workflows.state.ts
lorenzo-cavazzi: It's there for a later PR -- I left a couple of `// ! TODO` comments so that I _hopefully_ won't forget to address the points before merging the feature PR

client/src/utils/components/Tree.tsx
lorenzo-cavazzi: I actually tried keeping elements flexible, but the final result was somehow worse -- I couldn't keep the information aligned between different rows.
I'm happy to give it another try, I don't really like the current output. I feel it would be better to revise this as soon as I finish adding the shrunk version of the elements.

PR #2051, opened 2022-09-28
fix: show namespace name when there is no owner data

andre-code:
PR to display the namespace name when the owner data is not available. This occurs when the namespace type is a group.
/deploy #persist


ciyer (approved):
LGTM!


**before**
image


**after**
image

PR #2052, opened 2022-09-28
fix: load preview notebook

andre-code:
PR that update @nteract/commutable library and change NotebookRender component as functional component.

This error appears when opening the project https://renkulab.io/projects/renku-insights/session-load-times/files/blob/notebooks/CaseStudy.ipynb

/deploy #persist


ciyer (approved):
Nice!

PR #2053, opened 2022-09-28
fix(server): use the correct redis namespace

olevski:
The ui server was defaulting to using the `0` redis namespace which is reserved for the gateway.

This has not caused issues as far as I know because the likelihood of collisions is low.

But it is better to use a dedicated redis namespace for the ui-server and not put keys in the gateway's namespace.

/deploy #persist

### DIFF
--- a/server/src/config.ts
+++ b/server/src/config.ts
@@ -71,7 +71,7 @@ const AUTHENTICATION = {
 const REDIS = {
   host: process.env.REDIS_HOST || "localhost",
   port: convertType(process.env.REDIS_PORT) || 6379,
-  database: parseInt(process.env.REDIS_DATABASE || "0"),
+  database: convertType(process.env.REDIS_DATABASE) || 0,
   password: process.env.REDIS_PASSWORD || null,
   isSentinel: convertType(process.env.REDIS_IS_SENTINEL) || false,
   masterSet: process.env.REDIS_MASTER_SET || "mymaster",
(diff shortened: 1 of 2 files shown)


server/src/config.ts
```diff
@@ -71,7 +71,7 @@ const AUTHENTICATION = {
 const REDIS = {
   host: process.env.REDIS_HOST || "localhost",
   port: convertType(process.env.REDIS_PORT) || 6379,
-  database: process.env.REDIS_DATABASE || "0",
+  database: parseInt(process.env.REDIS_DATABASE || "0"),
```
lorenzo-cavazzi, line 74:
```suggestion
  database: convertType(process.env.REDIS_DATABASE) || 0),
```

lorenzo-cavazzi (changes requested):
Thank you @olevski for the fix! :pray: 
Just a tiny change to keep it coherent with the other fields.

olevski:
@lorenzo-cavazzi the change you suggested does not work because `convertType` returns `string | number | true` but typescript expects only `number` for that field. 

So what is the proper "typescript" way to resolve this? I am not really sure.

lorenzo-cavazzi:
I pushed the change to pass the `db` variable through the constructor where we can specify the expected type.

Your solution was the most reasonable way to do it in Typescript, but we introduced a custom convert function that doesn't throw errors so that a wrong config value wouldn't crash the server at the startup (ofc, depending on the variable, the server might crash just after :grin: )

lorenzo-cavazzi (approved):

PR #2054, opened 2022-10-03
feat: add progress indicator when start a session (#1879)

### LINKED ISSUE #1879: Show meaningful information on the session page while starting

andre-code:
PR to add progress indicator when start a session.

### Testing
- Start a new session
start a session

- Reload a session
reload session

- Unavailable session
Screenshot 2022-10-05 at 15 38 30

Closes #1879 

/deploy #persist renku-notebooks=feat-detailed-status-message extra-values=notebooks.amalthea.image.repository=registry.dev.renku.ch/tasko.olevski/renku-images/amalthea,notebooks.amalthea.image.tag=0.5.2-n014.h50ddcaa

### DIFF
--- a/client/src/notebooks/Notebooks.container.js
+++ b/client/src/notebooks/Notebooks.container.js
@@ -50,9 +50,7 @@ import ShowSessionFullscreen from "./components/SessionFullScreen";
 
 function mapSessionStateToProps(state, ownProps) {
   const notebooks = state.stateModel.notebooks.notebooks;
-  const available = notebooks.all[ownProps.target] ?
-    true :
-    false;
+  const available = !!notebooks.all[ownProps.target];
   const notebook = {
     available,
     fetched: notebooks.fetched,
--- a/client/src/notebooks/Notebooks.container.js
+++ b/client/src/notebooks/Notebooks.container.js
@@ -27,7 +27,6 @@ import {
   CheckNotebookIcon,
 } from "./Notebooks.present";
 import { StatusHelper } from "../model/Model";
-import { Loader } from "../utils/components/Loader";
 import { Url } from "../utils/helpers/url";
 import { sleep } from "../utils/helpers/HelperFunctions";
 import ShowSessionFullscreen from "./components/SessionFullScreen";
@@ -127,10 +126,6 @@ class ShowSession extends Component {
   render() {
     if (this.props.blockAnonymous && !this.userLogged)
       return <NotebooksDisabled logged={this.userLogged} />;
-
-    if (!this.model.get("notebooks.fetched"))
-      return <Loader />;
-
     return (
       <ShowSessionMapped
         target={this.target}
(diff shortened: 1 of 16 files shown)


client/src/notebooks/Notebooks.container.js
```diff
@@ -129,7 +129,7 @@ class ShowSession extends Component {
       return <NotebooksDisabled logged={this.userLogged} />;
 
     if (!this.model.get("notebooks.fetched"))
-      return <Loader />;
+      return <div className="d-flex flex-column align-items-center fst-italic"><Loader />Fetching session data</div>;
```
ciyer, line 132: As discussed in the call, this can be removed.

client/src/utils/components/progress/ProgressSteps.tsx
```diff
@@ -0,0 +1,179 @@
...
+function ProgressStep({ step }: progressStepProps) {
+  let content;
+  switch (step.status) {
+    case StatusStepProgressBar.EXECUTING:
+      content = <><Loader size="14" inline="true" />{step.step}</>;
+      break;
+    case StatusStepProgressBar.READY:
+      content = <><CheckCircleFill className="text-rk-green" />{step.step}</>;
+      break;
+    case StatusStepProgressBar.WAITING:
```
ciyer, line 168: An idea for realizing what we discussed in the meeting this morning: if **all** steps are in 'ready' or 'waiting' state, then display the first step in waiting state with a loader as well (so the same as executing).

This makes the logic here a bit more complex, but I think it is an ok tradeoff.

ciyer (changes requested):
Thanks for showing the progress this morning -- I think this is look very good! I have added a couple of suggestions from our discussion, maybe they help, otherwise, just implement things as we discussed in the meeting.

ciyer (approved):
Wow, this looks great and works very well! Very nice job done here! 🎉

rokroskar:
This is awesome! My only comment is that it's a bit verbose and showing too many steps.  - but we can leave it for now imho. Also, is there a reason to use a different style for "preparing session" vs "starting session"?

ciyer:
BTW, there is a comment in the discussion thread of #1879 that there should be an indication that the user is going to be put into an autosave branch in the startup information sequence. This was not included in this PR, so a new issue should be created for this part.

andre-code:
> This is awesome! My only comment is that it's a bit verbose and showing too many steps. - but we can leave it for now imho. Also, is there a reason to use a different style for "preparing session" vs "starting session"?

Yes, the idea is to unify the style or screens, but we defined to do it in another Issue.
https://github.com/SwissDataScienceCenter/renku-ui/issues/2064

PR #2065, opened 2022-10-10
Shape Up #1 - Third prototype: workflow browser with workflow details

lorenzo-cavazzi:
This prototype includes a fully working workflow browser and a first draft of the workflow details page.
Mind that the details components will be simplified and streamlined.

**What should be tested**
Using a couple of projects, we should verify that:
- The workflows browser works well, including the filters and the ordering
- The information should be consistent and up-to-date: this means the details components on the right are always in sync with the browser

Screenshot_20221012_154517

renku-python reference: https://github.com/SwissDataScienceCenter/renku-python/pull/3135 
/deploy renku-core=workflow-ui-feature-branch #persist extra-values=core.sentry.dsn=https://0cc02f3a02d3474d9f7e90915873ab4d@sentry.dev.renku.ch/2,core.sentry.environment=renku-dev,core.sentry.enabled=true

### DIFF
--- a/client/src/utils/components/Tree.tsx
+++ b/client/src/utils/components/Tree.tsx
@@ -34,17 +34,18 @@ interface TreeBrowserProps {
   expanded: string[];
   items: Record<string, any>;
   selected: string;
+  shrunk: boolean;
   toggleExpanded: Function;
 }
 
 function TreeBrowser({
-  expanded, items = [], selected, toggleExpanded
+  expanded, items = [], selected, shrunk, toggleExpanded
 }: TreeBrowserProps) {
   if (!items.length)
     return (<p>No elements to display</p>);
 
   const treeElements = items.map((item: any) => {
-    let newProps: Record<string, any> = { selected, expanded, toggleExpanded, items };
+    let newProps: Record<string, any> = { expanded, items, selected, shrunk, toggleExpanded };
     return (<TreeElement key={item.workflowId} {...item} {...newProps} />);
   });
 
@@ -72,7 +73,7 @@ interface TreeElementProps extends TreeBrowserProps {
 
 function TreeElement({
   active, children, creators, expanded, executions, indentation, itemType, items, lastExecuted, selected,
-  timeCaption, title, toggleExpanded, url, urlSingle, workflowId, workflowType
+  shrunk, timeCaption, title, toggleExpanded, url, urlSingle, workflowId, workflowType
 }: TreeElementProps) {
   const newClasses = workflowId === selected ? "selected" : "";
   const isComposite = workflowType === "CompositePlan" ? true : false;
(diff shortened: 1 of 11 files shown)


client/src/workflows/Workflows.state.ts
```diff
@@ -65,27 +66,41 @@ class WorkflowsCoordinator {
...
 
     // pre-fetching state changes
     let newWorkflowsState: Record<string, any> = { fetching: true };
-    if (this.workflowsModel.get("error"))
+    if (oldModel.error)
       newWorkflowsState.error = null;
     this.workflowsModel.setObject(newWorkflowsState);
 
-    const workflowsList = await this.client.fetchWorkflowsList(repositoryUrl, reference, versionUrl);
+    let workflowsList: any;
```
andre-code, line 90: Is it possible to define interfaces for these variables?

andre-code (changes requested):
This works very well 👏 
It's nice to see this functionality in action.
Do you have plans to include unit tests or e2e test in other PRs?
I left some minor requests otherwise this looks ready.

client/src/utils/components/Tree.tsx
```diff
@@ -139,4 +165,117 @@ function TreeElement({
...
-export { TreeBrowser, TreeElement };
+
+interface TreeDetailsProps {
+  backElement: React.ReactNode;
+  waiting: boolean;
+  workflow: Record<string, any>;
+  workflowId: string;
+}
+
+function TreeDetails({ backElement, waiting, workflow, workflowId }: TreeDetailsProps) {
```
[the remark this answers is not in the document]
lorenzo-cavazzi: That would make sense. I initially planned to make all the Tree elements more generic so they could be reused.
For the details component, maybe that doesn't make sense. Or, I could make a very generic `TreeDetails` container, and then render whatever `{children}` is passed. I would then make the `WorkflowTreeDetails` element as you suggested and move it to a Workflows file. Would that be a good solution?

client/src/workflows/Workflows.state.ts
lorenzo-cavazzi: That would contain either `error` or `result`. I swapped `any` with a generic `Record<string, any>`. I can make an interface to specify the 2 values if you think it's a valuable solution, but they will both be optional.

client/src/utils/components/Tree.tsx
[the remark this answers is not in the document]
andre-code: Yes, that would be a good solution, so far we don't have a tree component in our design system, but if we have the component we can apply in the future the definitions and styles easily to all components using that component.

client/src/workflows/Workflows.state.ts
andre-code: Or you can define the interface for a valid result and an error result and assign both to the variable.
```
interface WorkflowsList {
...
}
interface WorkflowsError {
...
}

let workflowssList: WorkflowsList | WorkflowsError;
```

lorenzo-cavazzi: As discussed offline, I went for a single interface with optional values `error?` and `result?`. Feel free to suggest otherwise if you think there is a better solution

client/src/utils/components/Tree.tsx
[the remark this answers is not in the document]
lorenzo-cavazzi: I went for the `{children}` solution. As you said, this makes the component more reusable :+1: 
P.S. I temporarily moved almost everything to the new component. In the next prototype, I'll add back some content to the generic `TreeDetails` component so that it handles the alignment/positioning.

lorenzo-cavazzi:
Thank you @andre-code for the review! I addressed your comments.

About this question:
> Do you have plans to include unit tests or e2e test in other PRs?

Yes, definitely! I prefer to postpone that to the last prototype since we kept changing the components a few times during the first phases and I didn't want to spend time on something temporary. It's on my list, together with using the RTK for storing the state instead of the 2 Coordinators

andre-code (approved):
lgtm 🚀

PR #2066, opened 2022-10-10
feat: add env variables in start session form

### LINKED ISSUE #2058: add support for passing environment variables in session start link
renku-notebooks in https://github.com/SwissDataScienceCenter/renku-notebooks/pull/1131 added support for custom environment variables for sessions. This feature is useful in general, but is explicitly required by the Smartsky project. They want to be able to pass environment variables in a session autostart link as URl parameters.

The UI should add support to pass these along to the notebook server and maybe also expose it in the session start screen or have it configurable as a project setting. This is mostly useful for storing secrets, but then that has some security implications on how/where the Ui would store those, and is maybe better left for a follow-up issue.

andre-code:
This PR add the env variables fields in start session form.

### Testing
- User should be able to start a session with environment variables
start session link with env variables

- Verify environment variable is available
verify sesssio environments variables

- Create a session link to share with environment variables
open a session with env variables

Closes #2058 

*Note: Added renku-notebooks dependency just to make available other functionalities merged before in master
/deploy #persist renku-notebooks=master extra-values=notebooks.amalthea.image.repository=registry.dev.renku.ch/tasko.olevski/renku-images/amalthea,notebooks.amalthea.image.tag=0.5.2-n014.h50ddcaa

### DIFF
--- a/client/src/model/RenkuModels.js
+++ b/client/src/model/RenkuModels.js
@@ -364,6 +364,8 @@ const projectSchema = new Schema({
   },
 });
 
+const DEFAULT_ENV_VARIABLES = [{ key: "", value: "" }];
+
 const notebooksSchema = new Schema({
   autosaves: {
     schema: {
@@ -391,7 +393,7 @@ const notebooksSchema = new Schema({
       commit: { initial: {} },
       discard: { initial: false },
       options: { initial: {} },
-      environment_variables: { initial: [{ key: "", value: "" }] },
+      environment_variables: { initial: DEFAULT_ENV_VARIABLES },
       objectStoresConfiguration: { initial: [] },
       includeMergedBranches: { initial: false },
       displayedCommits: { initial: 25 },
@@ -723,5 +725,5 @@ const formGeneratorSchema = new Schema({
 export {
   datasetFormSchema, datasetSchema, datasetImportFormSchema, environmentSchema,
   formGeneratorSchema, issueFormSchema, newProjectSchema, notebooksSchema, notificationsSchema,
-  projectSchema, projectsSchema, statuspageSchema, userSchema, webSocketSchema
+  projectSchema, projectsSchema, statuspageSchema, userSchema, webSocketSchema, DEFAULT_ENV_VARIABLES
 };
--- a/client/src/model/RenkuModels.js
+++ b/client/src/model/RenkuModels.js
@@ -391,6 +391,7 @@ const notebooksSchema = new Schema({
       commit: { initial: {} },
       discard: { initial: false },
       options: { initial: {} },
+      environment_variables: { initial: [{ key: "", value: "" }] },
       objectStoresConfiguration: { initial: [] },
       includeMergedBranches: { initial: false },
       displayedCommits: { initial: 25 },
--- a/client/src/model/RenkuModels.js
+++ b/client/src/model/RenkuModels.js
@@ -364,8 +364,6 @@ const projectSchema = new Schema({
   },
 });
 
-const DEFAULT_ENV_VARIABLES = [{ key: "", value: "" }];
-
 const notebooksSchema = new Schema({
   autosaves: {
     schema: {
@@ -393,7 +391,7 @@ const notebooksSchema = new Schema({
       commit: { initial: {} },
       discard: { initial: false },
       options: { initial: {} },
-      environment_variables: { initial: DEFAULT_ENV_VARIABLES },
+      environment_variables: { initial: [] },
       objectStoresConfiguration: { initial: [] },
       includeMergedBranches: { initial: false },
       displayedCommits: { initial: 25 },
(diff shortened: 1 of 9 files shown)


client/src/notebooks/NotebookStart.present.js
```diff
@@ -123,6 +124,18 @@ function StartNotebookServer(props) {
...
+  const setNotebookEnvVariables = (variables) => {
+    props.handlers.setNotebookEnvVariables(variables);
+    setEnvironmentVariables(variables);
+  };
+
+  useEffect(() => {
+    if (props.envVariablesQueryParams)
+      setEnvironmentVariables(props.envVariablesQueryParams);
+  }, []); // eslint-disable-line
+
```
lorenzo-cavazzi, line 138: This logic seems redundant to me.
If you are already save the status in the redux store, you can get it from `props.filters.environment_variables`.
Why saving it locally with a different format? Wouldn't it be easier to use only the redux store?

I feel the easiest thing would be adopting the pattern `[ { key: "", value: "" }]` also for the redux store, although that requires a few extra changes.

client/src/model/RenkuModels.js
```diff
@@ -391,6 +391,7 @@ const notebooksSchema = new Schema({
       commit: { initial: {} },
       discard: { initial: false },
       options: { initial: {} },
+      environment_variables: { initial: [{ key: "", value: "" }] },
```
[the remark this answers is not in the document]
lorenzo-cavazzi: P.S: after finishing the review, I now understand this better.
I feel going for `[{ key: "", value: "" }]` is the best alternative since it's easier to add empty fields.

lorenzo-cavazzi (changes requested):
Nice addition! It works well, but I feel the logic for storing the environment variables in the presentation component is redundant since you already store them in the redux store.

I would remove `const [environmentVariables, setEnvironmentVariables] = useState(` and use the variables coming from the redux model direcly.

andre-code:
> Nice addition! It works well, but I feel the logic for storing the environment variables in the presentation component is redundant since you already store them in the redux store.
> 
> I would remove `const [environmentVariables, setEnvironmentVariables] = useState(` and use the variables coming from the redux model direcly.

Thanks @lorenzo-cavazzi for your review.
I agree to use the redux state in the environmentVariables component, that would be cleaner and remove some of the complexity.
Find the suggestion implemented in the last commit.

client/src/model/RenkuModels.js
```diff
@@ -364,6 +364,8 @@ const projectSchema = new Schema({
   },
 });
 
+const DEFAULT_ENV_VARIABLES = [{ key: "", value: "" }];
+
```
lorenzo-cavazzi, line 368:
```suggestion
```

client/src/model/RenkuModels.js
```diff
@@ -391,6 +393,7 @@ const notebooksSchema = new Schema({
       commit: { initial: {} },
       discard: { initial: false },
       options: { initial: {} },
+      environment_variables: { initial: DEFAULT_ENV_VARIABLES },
```
lorenzo-cavazzi, line 396:
```suggestion
      environment_variables: { initial: [] },
```

client/src/model/RenkuModels.js
```diff
@@ -722,5 +725,5 @@ const formGeneratorSchema = new Schema({
 export {
   datasetFormSchema, datasetSchema, datasetImportFormSchema, environmentSchema,
   formGeneratorSchema, issueFormSchema, newProjectSchema, notebooksSchema, notificationsSchema,
-  projectSchema, projectsSchema, statuspageSchema, userSchema, webSocketSchema
+  projectSchema, projectsSchema, statuspageSchema, userSchema, webSocketSchema, DEFAULT_ENV_VARIABLES
```
lorenzo-cavazzi, line 728:
```suggestion
  projectSchema, projectsSchema, statuspageSchema, userSchema, webSocketSchema
```

client/src/notebooks/Notebooks.container.js
```diff
@@ -30,6 +30,7 @@ import { StatusHelper } from "../model/Model";
 import { Url } from "../utils/helpers/url";
 import { sleep } from "../utils/helpers/HelperFunctions";
 import ShowSessionFullscreen from "./components/SessionFullScreen";
+import { DEFAULT_ENV_VARIABLES } from "../model/RenkuModels";
```
lorenzo-cavazzi, line 33:
```suggestion
```

client/src/notebooks/Notebooks.container.js
```diff
@@ -300,6 +301,11 @@ class StartNotebookServer extends Component {
     this.customNotebookFilePath = currentSearch && this.autostart && currentSearch["notebook"] ?
       currentSearch["notebook"] :
       null;
+
+    const environmentVariables = this.getEnvironmentVariablesFromSearch(currentSearch);
+    this.customEnvVariables = currentSearch && this.autostart && environmentVariables.length ?
+      environmentVariables : DEFAULT_ENV_VARIABLES;
```
lorenzo-cavazzi, line 307:
```suggestion
      environmentVariables : [];
```

lorenzo-cavazzi (changes requested):
This looks great! A tiny detail before merging.

I personally think we can start with an empty array for the variables and let the user click on "Add Variable" instead of showing an empty one since it's optional and probably only very few people will use them. The initial state would be as in the following screenshot.

If you agree, I think I added all the necessary changes as suggestions. If you prefer to keep an initial empty entry, I would then move `const DEFAULT_ENV_VARIABLES` away from the RenkuModel file to a "NotebooksSomething" file since it looks like a session-related constant.

Screenshot_20221012_151822

lorenzo-cavazzi (dismissed):
Lgtm! :rocket:

lorenzo-cavazzi (approved):

PR #2072, opened 2022-10-14
fix(client): prevent repeated queries to gitlab graphql

lorenzo-cavazzi:
This PR prevents re-fetching the starred/member projects too often. By mistake, we were checking `!featured.featured` instead of `!featured.fetching`. By using `refreshIfNecessary` we still refresh the projects every now and then.

/deploy #persist


andre-code (approved):
lgtm 🚀

PR #2075, opened 2022-10-17
feat: add get namespaces graphql

### LINKED ISSUE #2027: when creating renku project, groups where user is "Direct Member" are shown, and only those
## Description

when creating renku project, groups where user is "Direct Member" are shown, and only those. This misses indirect groups, and includes  groups where user has no right to create as a Guest.

## Reproduce

1. try to create renku project
2. see list of available namespaces
3. notice absence of non-direct gitlab user groups

## Context

https://renkulab.io/

andre-code:
This PR is to fetch groups by minimum access level, it must include subgroups with inherited rights from the parent group.

Closes #2027 

/deploy extra-values=notebooks.amalthea.resourceUsageCheck.enabled=true   #persist

### DIFF
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -237,6 +237,7 @@
           "telepresence",
           "textarea",
           "thead",
+          "typename",
           "toastify",
           "toggler",
           "toml",
--- a/client/src/api-client/index.js
+++ b/client/src/api-client/index.js
@@ -159,7 +159,8 @@ class APIClient {
    *
    * @param {string} url - API url
    * @param {object} [parameters] - Optional parameters object to be provided to clientFetch.
-   * @param {number} [parameters.maxIterations] - Maximum iterations before throwing an error. Used to prevent
+   * @param {object} [parameters.options] - Fetch options, like method, headers... Default only include basic headers.
+   * @param {number} [options.maxIterations] - Maximum iterations before throwing an error. Used to prevent
    *   long/endless loops that would trigger the gateway rate limit. Set 0 for unlimited.
    * @example "for await" syntax that consumes the async iterator
    * for await (const partialData of clientIterableFetch("myApiUrl")) { console.log(partialData) }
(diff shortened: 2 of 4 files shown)


ciyer:
The current flow is:
1. UI asks core-service to create a project
2. Core-service creates a new project locally and pushes to the specified location in GitLab
3. UI renames the project to match the expected name

Step 2 requires `Maintainer` or greater access to the group, Step 3 requires `Owner` access to the group.

Fixing the problem would require changing the way projects are created. The options are:
1. The UI first creates an empty project in GitLab before the core-service pushes
2. The core service creates the project using GitLab more intelligently, introducing a dependency to GitLab
3. The UI renames the project in the same way GitLab, lowering the necessary privilege level

lorenzo-cavazzi:
As a reference, here are the different outcomes based on the permissions:
* owner: project is created correctly and remained correctly: https://dev.renku.ch/projects/lor-owner/renkulab-group-owner
* maintainer: project is created correctly but not renamed: https://dev.renku.ch/projects/lor-maintainer/renkulab-group-maintainer
    * Fixing requires some changes to UI and CORE so that projects are created by the UI providing the correct name, then CORE init and pushed the changes
* developer: project is created without any content and not renamed https://dev.renku.ch/projects/lor-developer/renkulab-group-developer
    * Fixing requires and ad-hoc solution on CORE just for GitLab

lorenzo-cavazzi:
We will go for an easy solution that doesn't bind us more the GitLab: limiting project creation to namespace owners.
Here is the update on the reference issue https://github.com/SwissDataScienceCenter/renku-ui/issues/2027#issuecomment-1314068180
And here is a suggestion on how to quickly get there https://github.com/SwissDataScienceCenter/renku-ui/issues/2027#issuecomment-1321803418

This PR can either be reworked on closed for another one -- feel free to go for the more convenient solution

andre-code:
The scope and solution for the original Isuue #2027  changed and is implemented in the PR #2187

PR #2076, opened 2022-10-17
chore: prevent service account from being automounted

Panaetius:
addresses https://github.com/SwissDataScienceCenter/renku/issues/2683 for the UI server and client

/deploy #persist


lorenzo-cavazzi (approved):
Thank you! :pray:

lorenzo-cavazzi (approved):

PR #2080, opened 2022-10-19
chore: make service account token mounting configurable

Panaetius:
Goes with https://github.com/SwissDataScienceCenter/renku/pull/2743
/deploy #persist


lorenzo-cavazzi (approved):
Thank you!

PR #2088, opened 2022-10-20
sessions: handle corrupted autosaves and improve warning messages positioning

### LINKED ISSUE #2086: Certain autosave branch names crash the ui
A user reported the following:
> Good morning Renku team,
since yesterday I'm having some problems starting a new session on Renku on [one project](https://renkulab.io/projects/omnibenchmark/omni_clustering/filter-expression): a warning message tells me that there is no image available but whenever I want to choose a specific commit, the page goes blank.  If I don't try to choose a commit, the page is stuck at 'checking sessions status'. I have tried:
to load a session from an other browser
restarting a new image build with a new commit
logout, login
Maybe related: yesterday I had this problem for any project  (from my namespace and from 'omnibenchmark') but it disappeared after some time (except this one project).
Any idea what is the problem and how I could solve it ?

Here is what their browser console looks like:
image

The problem was resolved after the user deleted the autosave branch. The name of the branch was like this `renku/autosave/fname.lname@uzh.ch/master/08007f6/134cf16`. I removed the actual first and last name for privacy but there was nothing weird there.

I will check how the notebook service handles this branch. But I think it should be fine because the format is as expected and the name does not even have special character or anything else that could cause trouble.

lorenzo-cavazzi:
This fixes the issue reported in #2086 

The problem seems to happen when the commit specified in the autosaves is not available in the reference branch. There might be a few reasons for that, and it's very difficult to guess the correct one in every case.

Since it doesn't seem to happen frequently, I decided to handle it by showing a warning to the user (see next screen) and removing the corrupted autosave to prevent the problem from happening again. We could also decide to leave the branch and point the user in that direction, but that seems more fragile (PVS can't allow that, so it might not always work)

image

**How to test**

Create a corrupted branch: pick any repo, create a new branch with this pattern:
```
renku/autosave/<user>/<exising_branch>/<fake_commit_sha-7-chars>/fake_commit_sha-7-chars
```
Here is an example with my user, swap it with yours
```
renku/autosave/lorenzo.cavazzi.tech@myemail.com/master/1111111/1111111
```

------------

I added another fix here since the warning for failed autostarts was moved to a potentially (very) narrow side column. I moved it to the top. First screenshot is the previous solution, second is the newest in this PR

Screenshot_20221020_151317

Screenshot_20221020_151445

/deploy #persist
fix #2086

### DIFF
--- a/client/src/notebooks/NotebookStart.present.js
+++ b/client/src/notebooks/NotebookStart.present.js
@@ -93,7 +93,9 @@ function StartNotebookAdvancedOptions(props) {
     </div>
     <div className="field-group">
       <AutosavesInfoAlert autosaves={autosaves} autosavesId={props.autosavesCommit}
-        currentId={props.filters.commit?.id} deleteAutosave={deleteAutosave} setCommit={setCommit} />
+        autosavesWrong={props.autosavesWrong} currentId={props.filters.commit?.id}
+        deleteAutosave={deleteAutosave} setCommit={setCommit}
+      />
     </div>
     <div className="field-group">
       <StartNotebookBranches {...props} disabled={props.disabled} />
@@ -196,7 +198,7 @@ function StartNotebookServer(props) {
   );
 }
 
-function AutosavesInfoAlert({ autosaves, autosavesId, currentId, deleteAutosave, setCommit }) {
+function AutosavesInfoAlert({ autosaves, autosavesId, autosavesWrong, currentId, deleteAutosave, setCommit }) {
   const [deleteOngoing, setDeleteOngoing] = useState(false);
   const [deleteResult, setDeleteResult] = useState(null);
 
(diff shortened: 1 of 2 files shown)


lorenzo-cavazzi:
P.S: the second error can be tested here once you log in (it's an internal project):
[New solution - full width](https://renku-ci-ui-2088.dev.renku.ch/projects/lorenzo.cavazzi.tech/swiss-energy/sessions/new?autostart=1&branch=faulty_image)

Compare with the previous solution:
[Old solution - column](https://dev.renku.ch/projects/lorenzo.cavazzi.tech/swiss-energy/sessions/new?autostart=1&branch=faulty_image)

ciyer (approved):
This fix works for me! 

As far as displaying the LaunchErrorAlert above the form content goes, an alternative could be leaving the display as it was, and increasing the breakpoint to where the view switchings into single-column mode, but this change is also acceptable.

PR #2089, opened 2022-10-20
fix: pre-fill all values in the new project form (#2026)(#2089)

### LINKED ISSUE #2026: links for creating renku prokect with prefilled template, namespace, visibility do not work
## Description

I am unable to create sharable links for creating renku projects, as appartent in the interface and advised by @rokroskar 

image

## Reproduce

https://renkulab.io/projects/new?data=eyJuYW1lc3BhY2UiOiJhc3Ryb25vbXkvbW1vZGEiLCJ2aXNpYmlsaXR5IjoicHVibGljIiwidXJsIjoiaHR0cHM6Ly9naXRodWIuY29tL29kYS1odWIvY29udHJpYnV0ZWQtcHJvamVjdC10ZW1wbGF0ZXMiLCJyZWYiOiJtbW9kYS1zbWFydHNreSJ9

it decodes to
```json
{"namespace":"astronomy/mmoda","visibility":"public","url":"https://github.com/oda-hub/contributed-project-templates","ref":"mmoda-smartsky"}
```
which makes sense.

Clicking on this leads to perpetual this:

image

<details><summary>Browser Output</summary>
<pre>
Uncaught (in promise) TypeError: this.getVisibilities is not a function
    e ProjectNew.state.js:221
    Babel 17
    value main.ab1ae2be5ab478f8bd77-8340587.chunk.js:2
    value ProjectNew.state.js:155
    n ProjectNew.container.js:458
    React 8
    unstable_runWithPriority scheduler.production.min.js:18
    React 4
    Redux 5
    value Model.js:158
    value Model.js:246
    value Model.js:232
    value Model.js:288
    value User.state.js:40
ProjectNew.state.js:221:13
    Babel 14
    value main.ab1ae2be5ab478f8bd77-8340587.chunk.js:2
    value ProjectNew.state.js:155
    n ProjectNew.container.js:458
    React 7
    _l self-hosted:1198
    Qi React
    unstable_runWithPriority scheduler.production.min.js:18
    React 4
    Redux 5
    value Model.js:158
    value Model.js:246
    value Model.js:232
    value Model.js:288
    value User.state.js:40

</pre>
</details>

I saw some issue about visibility in console so switched off visibility.

New link:

https://renkulab.io/projects/new?data=eyJ1cmwiOiJodHRwczovL2dpdGh1Yi5jb20vb2RhLWh1Yi9jb250cmlidXRlZC1wcm9qZWN0LXRlbXBsYXRlcyIsInJlZiI6Im1tb2RhLXNtYXJ0c2t5In0

{"url":"https://github.com/oda-hub/contributed-project-templates","ref":"mmoda-smartsky"}

Does not get stuck but claims to fill something, but nothing is actually filled:

image

I tried a couple of other combinations but no luck.

## Context

https://renkulab.io
Firefox 104.0

andre-code:
This PR is to pre-fill all values in the new project form.

- Fixed interminal spin when selecting a namespace.
- Show url and ref of custom templates if included, previously it was only if the template was also included.
- added e2e test of all cases

Closes #2026 

/deploy #persist

### DIFF
--- a/client/src/project/new/ProjectNew.container.js
+++ b/client/src/project/new/ProjectNew.container.js
@@ -456,7 +456,7 @@ class NewProject extends Component {
     try {
       const data = getDataFromParams(searchParams);
       if (data)
-        this.coordinator.setAutomated(data);
+        this.coordinator.setAutomated(data, undefined, this.projectsCoordinator);
 
       // do not update url if is importing a dataset
       if (!props.importingDataset) {
--- a/client/src/project/new/components/ShareLinkModal.js
+++ b/client/src/project/new/components/ShareLinkModal.js
@@ -121,7 +121,14 @@ function ShareLinkModal(props) {
   }, [createUrl, include, input, userTemplates]);
 
   const handleCheckbox = (target, event) => {
-    setInclude({ ...include, [target]: event.target.checked });
+    // select the template source if is selected a custom template
+    if (target === "template" && event.target.checked && available.templateRepo)
+      setInclude({ ...include, templateRepo: event.target.checked, [target]: event.target.checked });
+    // deselect template if deselect templateRepo
+    else if (target === "templateRepo" && !event.target.checked && available.template)
+      setInclude({ ...include, template: false, [target]: event.target.checked });
+    else
+      setInclude({ ...include, [target]: event.target.checked });
   };
 
   const labels = Object.keys(include).map(item => (
@@ -129,7 +136,7 @@ function ShareLinkModal(props) {
       <Label check className={`text-capitalize${available[item] ? " cursor-pointer" : " text-muted cursor-pointer"}`}>
         <Input
           type="checkbox"
-          disabled={available[item] ? false : true}
+          disabled={!available[item]}
           checked={include[item]}
           onChange={e => handleCheckbox(item, e)}
         /> {item === "templateRepo" ? "template source" : item}
--- a/client/src/project/new/ProjectNew.state.js
+++ b/client/src/project/new/ProjectNew.state.js
@@ -167,6 +167,7 @@ class NewProjectCoordinator {
       automatedObject :
       newProjectSchema.createInitialized().automated;
     let availableVariables = [];
+    let visibilities;
 
     // Step 1: wait for templates and namespaces to be fetched
     automated.step = 1;
@@ -220,7 +221,9 @@ class NewProjectCoordinator {
       else {
         this.setProperty("namespace", data.namespace); // full path
         newInput.namespace = data.namespace;
-        projectsCoordinator.getVisibilities(namespaceAvailable);
+        const visibilitiesByNamespace = await projectsCoordinator.getVisibilities(namespaceAvailable);
+        visibilities = visibilitiesByNamespace.visibilities;
+        this.setVisibilities(visibilitiesByNamespace, namespaceAvailable);
       }
     }
     if (data.template && !data.url) {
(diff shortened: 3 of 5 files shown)


client/src/project/new/ProjectNew.container.js
```diff
@@ -456,7 +456,7 @@ class NewProject extends Component {
     try {
       const data = getDataFromParams(searchParams);
       if (data)
-        this.coordinator.setAutomated(data);
+        this.coordinator.setAutomated(data, undefined, this.projectsCoordinator);
```
lorenzo-cavazzi, line 459: This partially works because the visibilities get fetched, but the user can still select a wrong one -- try with a private namespace, and `private` will be selected but you will still be able to click on `public`.

A better solution would be to pass the `calculateVisibilities` function. A dirty trick for doing that would be passing a function where you save `this`, something like the following:
```
this.coordinator.setAutomated(data, undefined, (namespace) => {
    const self = this;
    self.calculateVisibilities(namespace);
});
```
And you change the code in the NewProjectCoordinator to invoke that function instead of invoking `projectsCoordinator.getVisibilities(namespaceAvailable);`

Ofc feel free to spend a bit more time on this and avoid the ugly `const self = this;` if you wish :smile:

client/src/project/new/components/ShareLinkModal.js
```diff
@@ -121,15 +121,22 @@ function ShareLinkModal(props) {
...
   const handleCheckbox = (target, event) => {
-    setInclude({ ...include, [target]: event.target.checked });
+    // select the template source if is selected a custom template
+    if (target === "template" && event.target.checked && available.templateRepo)
+      setInclude({ ...include, templateRepo: event.target.checked, [target]: event.target.checked });
+    // deselect template if deselect templateRepo
+    else if (target === "templateRepo" && !event.target.checked && available.template)
+      setInclude({ ...include, template: false, [target]: event.target.checked });
+    else
+      setInclude({ ...include, [target]: event.target.checked });
```
lorenzo-cavazzi, line 131: I'm not sure I get the logic here... What unexpected situation is this trying to cover? I quickly tried with the previous solution and I always get the same result as with this change.

The first `if (target === "template" &&` seems redundant since we don't need to set `templateRepo` in that case. And for the second case `else if (target === "templateRepo"` I couldn't find an example where the output is different. Do you have one?

lorenzo-cavazzi (changes requested):
Nice fix! There is one problem with visibilities since our code is a bit messy and we need to cover the selectable items too -- I suggested a dirty trick but feel free to explore other solutions if you wish.

I didn't get the change on `handleCheckbox`; could you provide an example link where the new code results in better output for the user? I tried a few but I'm probably missing something

client/src/project/new/ProjectNew.container.js
andre-code: yes, thanks for the review. I found the option to use the function `setVisibilities` to update the visibilities state. Check it in the last commit

client/src/project/new/components/ShareLinkModal.js
andre-code: Yes, the case it covers is when the user select a custom template but do not select the template source that is needed to load the template.
In that case, when it load the link params the user will get the error:
Screenshot 2022-10-24 at 16 32 04
With this check we make sure that all the necessary data is included in the creation of the link.

andre-code:
> Nice fix! There is one problem with visibilities since our code is a bit messy and we need to cover the selectable items too -- I suggested a dirty trick but feel free to explore other solutions if you wish.
> 
> I didn't get the change on `handleCheckbox`; could you provide an example link where the new code results in better output for the user? I tried a few but I'm probably missing something

Thanks for the review and reporting the visibilities status issue. Let me know if you have any further questions.

lorenzo-cavazzi (approved):

Issue #1956, opened 2022-07-14
Define flows for "Refresh" button


gavin-k-lee:
As part of targetting no/low code solutions and their users, we should implement clickable buttons to run certain commands, like `git pull`. Possible breakdown of cases (TBC with devs):
https://www.figma.com/file/5WnX17bEKxRN46XWVyDKWk/Refresh-button?node-id=0%3A1

Simplify flows for initial implementation, set scope for future rounds.

Could we lean into a Git GUI for this?

## Acceptance criteria
- [ ] confirm all possible cases for this flow

ciyer:
The logic needed for this is described in https://github.com/SwissDataScienceCenter/renku-notebooks/issues/1196.

PR #2091, opened 2022-10-24
feat: add refresh button (#1986)

### LINKED ISSUE #1986: Implement no-code git Refresh button as defined
Is this ready to go after https://github.com/SwissDataScienceCenter/renku-ui/issues/1956 is done?

andre-code:
As decided in #1956, the plan is to handle only the simplest case where the session is behind the orgin, but is not dirty and has not diverged.

1. Session is not behind origin (could be ahead)

image

2. Session is strictly behind origin

image

3. Session is dirty

image

4. Session is behind and ahead of origin

image

6. No sidecar or anonymous

image

Closes #1986 
/deploy renku=000-fix-acceptance-tests-after-changing-gitlab-url extra-values=gateway.gitlabUrl=https://gitlab.dev.renku.ch/ #persist

### DIFF
--- a/client/src/notebooks/components/PullSession.tsx
+++ b/client/src/notebooks/components/PullSession.tsx
@@ -1,0 +1,184 @@
+/*!
+ * Copyright 2022 - Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React, { useState } from "react";
+
+import { Button, Col, Modal, ModalBody, ModalHeader, Row } from "../../utils/ts-wrappers";
+import { Loader } from "../../utils/components/Loader";
+import { Notebook } from "./Session";
+import { useGitStatusQuery, useHealthQuery, useRenkuPullMutation } from "../../features/session/sidecarApi";
+import type { GitStatusResult } from "../../features/session/sidecarApi";
+import { commitsPhrasing } from "./SaveSession";
+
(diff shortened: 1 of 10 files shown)


ciyer:
As decided in #1956, the plan is to handle only the simplest case where the session is behind the orgin, but is not dirty and has not diverged.

1. Session is not behind origin (could be ahead)

image

2. Session is strictly behind origin

image

3. Session is dirty

image

4. Session is behind and ahead of origin

image


6. No sidecar or anonymous

image

client/src/notebooks/components/PullSession.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { Button, Col, Modal, ModalBody, ModalHeader, Row } from "../../utils/ts-wrappers";
+import { useGitStatusQuery, useHealthQuery, useRenkuPullMutation } from "../../features/session/sidecarApi";
+import type { GitStatusResult } from "../../features/session/sidecarApi";
+
+import { Notebook } from "./Session";
+import { CenteredLoader, InformationalBody, commitsPhrasing } from "./Sidecar";
+import type { CloseModalProps, ModalProps } from "./Sidecar";
+
+interface PullSessionProps extends ModalProps {
+  hasPullAccess: boolean;
```
andre-code, line 30: I think in this iteration isLogged and hasPullAccess are not needed. If that is the case could you please remove it.

andre-code (commented):
It works very well 👏, also thanks for the refactoring of the components, it looks great.
I have found only 2 small suggestions.

e2e/cypress/integration/local/session.spec.ts
```diff
@@ -72,4 +72,81 @@ describe("display a session", () => {
...
+    cy.gui_open_session();
+    // save session
+    cy.get_cy("save-session-button").click();
+    cy.get(".modal-dialog").should("exist");
+    cy.get(".modal-dialog").get("h5").contains("Save Session").should("be.visible");
+    cy.get(".modal-dialog").get("div")
+      .contains("You have work in this session that has not yet been saved to the server.").should("be.visible");
+  });
+
+  it.only("pull changes button -- no sidecar", () => {
```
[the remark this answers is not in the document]
ciyer: Oops!

ciyer:
> While the session is starting, the buttons are available and return the same message as for anonymous sessions. I would either disable them with a popover saying "Session isn't ready", or leave them enabled and add a similar text to the bigger modal !

Good point! This is exactly what was done in #2103.

client/src/notebooks/components/PullSession.tsx
ciyer: Good point, done!

andre-code (commented):
It works as expected and all changes were implemented.
Thanks 🙌

🚀

ciyer (approved):

PR #2092, opened 2022-10-25
Shape Up #1 - Apply Design

andre-code:
This PR is to apply design in workflows.

wireframe workflows

It includes also:
- Display in multiple lines when the slug or project title in the entity card exceeds the width of the card for mobiles.

  image

- Fix show the right arrow-indicator when expand a row in a nested row
image

### DIFF
--- a/client/src/utils/components/Tree.tsx
+++ b/client/src/utils/components/Tree.tsx
@@ -19,13 +19,11 @@
 
 import React from "react";
 import { Link } from "react-router-dom";
-import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import { faChevronDown, faChevronUp } from "@fortawesome/free-solid-svg-icons";
 
 import EntityCreators, { EntityCreator } from "./entities/Creators";
 import EntityExecutions from "./entities/Executions";
 import EntityDuration from "./entities/Duration";
-import { Col } from "../ts-wrappers";
+import { CaretDownFill, CaretRightFill, Col, Diagram2 } from "../ts-wrappers";
 import { EntityChildrenDot } from "./entities/Children";
 import { EntityType, WorkflowType } from "./entities/Entities";
 import { simpleHash } from "../helpers/HelperFunctions";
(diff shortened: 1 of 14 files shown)


client/src/utils/components/Tree.tsx
```diff
@@ -115,15 +119,7 @@ function TreeElement({
...
-
-  let currentIndentation = 0;
-  if (indentation >= 4)
-    currentIndentation = 4;
-  else if (indentation)
-    currentIndentation = indentation;
-  if (shrunk)
-    currentIndentation = currentIndentation / 2;
-  const elementStyle = { marginLeft: `${currentIndentation}em` };
+  const elementStyle = { marginLeft: `${indentation * 30}px` };
```
lorenzo-cavazzi, line 122: Would it make sense to add 30px up to the 2nd level and then go 5 by 5 (or 10 by 10)? I feel that would preserve readability for the (few) cases with more nested workflows

lorenzo-cavazzi: So the total margin would be
L1 --> 30
L2 ----> 60
L3 ------> 70
L4 --------> 80

lorenzo-cavazzi (changes requested):
This looks great! I added a suggestion in line; let me know if ye or if you prefer to keep it as is.
Apart from that, the only thing that bothers me a little bit is the vertical alignment when composite workflows are open. Exeuctions and run-time are not aligned on the same column. Also, the number of children workflows seems a bit too much on the right
image

andre-code:
> This looks great! I added a suggestion in line; let me know if ye or if you prefer to keep it as is. Apart from that, the only thing that bothers me a little bit is the vertical alignment when composite workflows are open. Exeuctions and run-time are not aligned on the same column. Also, the number of children workflows seems a bit too much on the right.


Thanks @lorenzo-cavazzi for the review and suggestions, definitely aligning columns is a great improvement for data readability, and for margin in nested workflows I don't have workflows with more than 3 levels to check but the logic is in the code.
Something new is that I included also some improvements for small screens for the table and detail workflows. 

As the base branch changed I found it easier to put everything in one commit, I hope it's ok for the revision.

andre-code:
Thank you @ciyer for your review.
I have used the border box for filters (toggle element) as it is the design used in kg-search filters. The idea is to keep unified those elements that have in common their functioning as in this case where they are filters of the content shown below.
Regarding your second suggestion, I have included it.
Thanks again 🙂
Let me know if you have more questions

lorenzo-cavazzi (changes requested):
It looks great!
I personally feel the dot showing the number of composite workflows looks better when aligned on the left, but this is a tiny detail and we can confront on that on Monday.

I played a bit with windows sizes and I noticed 2 details. I tested with Firefox.

* At any window size, after opening the details page, the left browser shows an unnecessary horizontal scroll bar -- maybe there is an extra margin somewhere?
  Screenshot_20221027_143213
* When reducing the width of the browser, components mess up a little bit: on the left screenshot they start compacting, on the right (smaller width) they expand again and overflow -- maybe they are no wrapping correctly?
  Screenshot_20221027_143019

ciyer (approved):
🎉

lorenzo-cavazzi (approved):
This looks amazing!
As discussed offline, we will bring up the question on the small-width browser again at a later point. Let merge this :rocket:

PR #2093, opened 2022-10-26
Shape Up #1 - Fifth prototype fixes

lorenzo-cavazzi:
This PR includes the following:

- automatically select the first input or mapping on composite workflow (when available)
	* any workflow is ok for testing
- visualize links on composite workflows
	* example: https://renku-ci-ui-2093.dev.renku.ch/projects/workflows-pet-projects/workflows-parameters/workflows/952c938055a041168f6e795ae33b0d22
- improve auto-scroll behavior when workflow details become available
	* any project with workflow is ok for testing -- compare with an older version like https://renku-ci-ui-2038.dev.renku.ch
- (fix) visualize outdated workflows
	* example: https://renku-ci-ui-2093.dev.renku.ch/projects/workflows-pet-projects/nested-workflows/workflows/998ff6b7853d4889bbec55dfb31b4ba6

### DIFF
--- a/client/src/workflows/Workflows.present.tsx
+++ b/client/src/workflows/Workflows.present.tsx
@@ -20,33 +20,14 @@ import React, { useState } from "react";
 import { Link } from "react-router-dom";
 import { FontAwesomeIcon, } from "@fortawesome/react-fontawesome";
 import {
-  faCheck, faExclamationTriangle, faInfoCircle, faLink, faSortAmountDown, faSortAmountUp,
+  faArrowRight, faCheck, faExclamationTriangle, faInfoCircle, faLink, faSortAmountDown, faSortAmountUp
 } from "@fortawesome/free-solid-svg-icons";
 
 import {
-  Button,
-  ButtonDropdown,
-  Card,
-  CardBody,
-  Col,
-  CardHeader,
-  DropdownItem,
-  DropdownMenu,
-  DropdownToggle,
-  Input,
-  Label,
-  PopoverBody,
-  Row,
-  UncontrolledTooltip,
-  UncontrolledPopover,
-  Table,
-  XLg,
-  People,
-  Journals,
-  Bookmarks,
-  Calendar4
+  Bookmarks, Button, ButtonDropdown, Calendar4, Card, CardBody, CardHeader, Col, DropdownItem,
+  DropdownMenu, DropdownToggle, Input, Journals, Label, People, PopoverBody, Row, Table,
+  UncontrolledPopover, UncontrolledTooltip, XLg
 } from "../utils/ts-wrappers";
-
 import EntityCreators from "../utils/components/entities/Creators";
 import Time from "../utils/helpers/Time";
 import { Clipboard } from "../utils/components/Clipboard";
--- a/client/src/workflows/Workflows.present.tsx
+++ b/client/src/workflows/Workflows.present.tsx
@@ -619,11 +619,9 @@ function VisualizerLinks({
   if (!data?.length)
     return (<p className="m-2"><UnavailableDetail /></p>);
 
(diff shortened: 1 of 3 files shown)


client/src/workflows/Workflows.present.tsx
```diff
@@ -632,6 +607,45 @@ function VisualizerMappings({
...
+}
+
+function VisualizerLinks({
+  data, workflows
+}: VisualizerLinksProps) {
+  if (!data?.length)
+    return (<p className="m-2"><UnavailableDetail /></p>);
+
+  let num = 0;
+  let links = data.map((t: any) => {
```
ciyer, line 623: If you provide a second argument, you can get the index in map.
```suggestion
  let links = data.map((t: any, num: number) => {
  
  }
```

client/src/workflows/Workflows.present.tsx
```diff
@@ -632,6 +607,45 @@ function VisualizerMappings({
...
+    num++;
+    let targetCounter = 0;
+    const targets = t.sinks.map((target: any) => {
```
ciyer, line 626:
```suggestion
    const targets = t.sinks.map((target: any, targetCounter: number) => {
```

client/src/workflows/Workflows.present.tsx
```diff
@@ -760,6 +774,30 @@ function VisualizerDetailExpanded({ data, fullPath }: VisualizerDetailExpandedPr
...
+  workflows: Record<string, any>;
+}
+
+function VisualizerLocalResource({ data, workflows }: VisualizerLocalResourceProps) {
+  try {
+    const targetWorkflow = workflows?.list?.length ?
+      workflows.list.find((w: any) => (w.id === data.plan_id)) :
+      null;
+    const subItem = data.id.replace(targetWorkflow.id + "/", "");
+    const newName = `[WF: ${targetWorkflow.name}] @ ${subItem.replace("/", " #")}`;
```
ciyer, line 788: What does WF stand for? "Workflow"? I think it's ok to leave it as is ATM, but something that should be fixed before the feature is finalized.

ciyer (changes requested):
Overall, this looks good, and the things mentioned in the description work as explained.

The only thing I think should be changed in the code is that I think you can use `map` better to avoid having to introduce a counter. 

A minor comment: I find the "Loading" page a bit distracting

image

I think it could be improved by just removing the "Loading details" header and leaving just the spinner.

client/src/workflows/Workflows.present.tsx
lorenzo-cavazzi: You are right, this is definitely not clear. We should come up with a better output
image

client/src/workflows/Workflows.present.tsx
lorenzo-cavazzi: Ah, thanks! I added it in a separate commit since I will also initialize it

lorenzo-cavazzi:
Thanks for the review! I addressed the comments and removed the header above the spinner: that wasn't looking good. Here is the new output

image

client/src/workflows/Workflows.present.tsx
```diff
@@ -619,11 +619,9 @@ function VisualizerLinks({
   if (!data?.length)
     return (<p className="m-2"><UnavailableDetail /></p>);
 
-  let num = 0;
-  let links = data.map((t: any) => {
+  let links = data.map((t: any, num: number) => {
     num++;
```
ciyer, line 623: This looks a bug!

ciyer (changes requested):
I like the improvements! I think one variable increment slipped through -- I think the `num++` should be removed now.

ciyer (approved):

PR #2097, opened 2022-10-27
fix: correctly set up the projectKgApi

### LINKED ISSUE #2096: Unnecessary queries to KG
## Description

When looking at a project, every time the user moves from a tab to another, an API to `/api/kg/projects/<project_id>` is invoked again. This is unnecessary, we should not query again after the first time unless the user re-indexes the project.

The responsible component seems to be `projectKgApi`.

## Reproduce

Open any project, keep open the Network tab of the Developer Tools, and try to click on different tabs (E.G: stats, commits, status, settings, ...).

ciyer:
The middleware for the projectKgApi was not correctly registered.

## Calls from clicking through tabs

**before**

image

**after**

image

Fix #2096

/deploy #persist

### DIFF
--- a/client/src/utils/helpers/EnhancedState.js
+++ b/client/src/utils/helpers/EnhancedState.js
@@ -38,7 +38,8 @@ function createStore(renkuStateModelReducer, name = "renku") {
       getDefaultMiddleware({
         immutableCheck: false,
         serializableCheck: false,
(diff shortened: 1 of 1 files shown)


andre-code (approved):
Thanks for fixing it 🚀

PR #2098, opened 2022-10-27
fix: ui state tracking the session status (#2095)

### LINKED ISSUE #2095: ui state is no tracking well the session status
## Description
🐞 When a user starts a session, the progress of each step is not shown, it goes from all waiting to all ready even when from the notebook service the progress of the steps is being received.

andre-code:
Closes #2095 

/deploy extra-values=notebooks.amalthea.resourceUsageCheck.enabled=true #persist


ciyer (approved):
Tested this, and the progress is displayed properly here (compared to dev, where progress happens all at once).

PR #2102, opened 2022-10-28
Shapeup workflows design v2

andre-code:
PR to apply design to workflows details (steps, links, mappings, outputs, inputs and parameters).

workflows design 2

### DIFF
--- a/client/src/styles/components/_renku_tree.scss
+++ b/client/src/styles/components/_renku_tree.scss
@@ -96,7 +96,7 @@
     }
 
     &.selected {
-      background-color: $rk-light !important;
+      border-bottom: 2px solid $rk-yellow;
     }
   }
 }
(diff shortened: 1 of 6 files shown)


lorenzo-cavazzi (approved):
This works well, thank you!

I added a commit with 3 simple changes (couldn't easily use "suggest" cause some code was older). If you agree with the changes feel free to merge. Otherwise, we can find a different solution.

* The Steps inside a workflow had a big vertical spacing between elements depending on the window size (see screenshot for an example). This is a side effect of the `gap-3` class. I changed to `gap-2` to prevent it, Feel free to add a right-margin to the workflow name if you think the horizontal spacing is not enough anymore.
  Screenshot_20221031_134141

* `.internal-steps` class: the content was not vertically aligned (see screenshot), the quickest fix I found was adding left/right margins
  Screenshot_20221031_134856


* The first element inside some `card-body` cards looked very compressed to the top (see screenshot) because all the padding was removed with `p-0`. I switched to `px-0 py-2`
  image

PR #2103, opened 2022-10-28
Do not request sidecar health too early

### LINKED ISSUE #2100: Session save button works only in new projects
When logged in, the newly added save button in the sessions iframe view does not work for older projects.

I tested this on dev.renku.ch and on renkulab.io. The same behaviour is present in both places.

The only way I could get this to work is to create a brand new project. When I launch a session from such a project things work.

In the cases where this not work I do not see a request sent from the browser to the sessions jsonrpc endpoint sent at all. So somehow and for some reason the code on the client is deciding that the old projects cannot do the save. But this should not matter the project version does not matter for the save.

The projects I most recently tested were just the regular new project templates with no changes:
- RENKU_BASE_IMAGE=renku/renkulab-py:3.9-0.13.1, worked
- RENKU_BASE_IMAGE=renku/renkulab-py:3.9-0.12.0, did not work

Is this limitation perhaps enforced on purpose?

ciyer:
The save/pull buttons were not always working because the UI was requesting session sidecar health before it could possibly be available.

Fix #2100

### DIFF
--- a/client/src/notebooks/components/PullSession.tsx
+++ b/client/src/notebooks/components/PullSession.tsx
@@ -19,6 +19,7 @@
 import React from "react";
 
 import { Button, Col, Modal, ModalBody, ModalHeader, Row } from "../../utils/ts-wrappers";
+import { Loader } from "../../utils/components/Loader";
 import { useGitStatusQuery, useHealthQuery, useRenkuPullMutation } from "../../features/session/sidecarApi";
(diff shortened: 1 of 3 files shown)


andre-code (approved):
The code looks good and works as expected ✨.
🚀

lorenzo-cavazzi (approved):

PR #2104, opened 2022-10-31
feat(client): use progress step indicator for prepare session loader (#2064)

### LINKED ISSUE #2064: UX: Unify views when preparing a session and when starting a new session
## Motivation
Provide a better UX where the process to start a new session does not require more than one screen or those screens are similar.

Current status:
1. Screen preparing a session
Screenshot 2022-10-10 at 09 21 35
2. Screen starting a session
Screenshot 2022-10-10 at 09 17 08

## Proposal

Using step-by-step progress loading in the session preparation screen.

andre-code:
PR to unify preparing a session and starting a session loaders.

unify session loading

Closes #2064 


client/src/project/Project.present.js
```diff
@@ -1056,8 +1056,8 @@ const ProjectSessions = (props) => {
         <Route path={props.launchNotebookUrl}
           render={p => (
             <Fragment>
-              {backButton}
-              <ProjectStartNotebookServer key="startNotebookForm" {...props} />
+              <ProjectStartNotebookServer
+                key="startNotebookForm" {...props} backUrl={backUrl} defaultBackButton={backButton} />
             </Fragment>
```
ciyer, line 1061: Can remove the `<Fragment>` now that there is only one element.

client/src/notebooks/NotebookStart.present.js
```diff
@@ -322,23 +359,22 @@ function StartNotebookAutostart(props) {
...
-        description=""
-        title="Preparing Session"
-        style={ProgressStyle.Light}
-        type={ProgressType.Determinate}
-        currentStatus={`${message}...`}
-        percentage={progress} />
-    </div>
+    <>
+      {backButtonSessions}
+      <div className="progress-box-small">
```
ciyer, line 368: As mentioned in the review, it would be good to indicate that this is step 1 on 2. Something like:

`Step 1 of 2: Checking if launch is possible`

And then add a 

`Step 2 of 2: Starting session`

To the starting page.

client/src/project/Project.present.js
andre-code: sure, included in the last commit

client/src/notebooks/NotebookStart.present.js
andre-code: Thanks for reminding me of this change, it was included in the last commit.

PR #2105, opened 2022-11-01
feat(client): update tooltip texts

andre-code:
This PR is for modifying the tooltip texts in the workflow view.


lorenzo-cavazzi (approved):
I think "open" is fine because this looks a bit like a file browser where I would say "open" the file.
If you feel that "go to" is a better option I'm also fine with that -- I'll approve, feel free to either merge the PR or just close it.

PR #2111, opened 2022-11-02
fix: tests after gitlab url changed

andre-code:
This PR is to fix e2e and acceptance tests after gitlab url changed.

List of changes needed to fix the tests:
- the gitlabUrl set in gateway.gitlaburl should be `https://gitlab.dev.renku.ch/`.
- The CSP at https://gitlab.dev.renku.ch/ should include the frame-ancestors `https://*.renku.ch`. 
- The necessary changes to the acceptance tests are in the `000-fix-acceptance-tests-after-changing-gitlab-url `branch in the renku repository

/deploy renku=000-fix-acceptance-tests-after-changing-gitlab-url extra-values=gateway.gitlabUrl=https://gitlab.dev.renku.ch/ #persist

### DIFF
--- a/e2e/cypress/support/authentication/gui_commands.ts
+++ b/e2e/cypress/support/authentication/gui_commands.ts
@@ -17,7 +17,6 @@
  */
 
 import { User, userToUsername } from "./user.interfaces";
-const GITLAB_PROVIDER = Cypress.env("GITLAB_PROVIDER");
 
 Cypress.Commands.add("gui_kc_login", (user: User, startFromHome = false) => {
   if (startFromHome) {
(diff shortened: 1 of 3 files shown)


e2e/cypress/support/authentication/gui_commands.ts
```diff
@@ -35,11 +35,11 @@ Cypress.Commands.add("gui_kc_login", (user: User, startFromHome = false) => {
     if (url.includes("auth/realms/Renku/login-actions/authenticate")) {
       cy.gui_kc_register(user);
     }
-    else if (url.includes(`${GITLAB_PROVIDER}/auth/realms/Renku/protocol/openid-connect/auth`)) {
+    else if (url.includes(`/auth/realms/Renku/protocol/openid-connect/auth`)) {
```
ciyer, line 38: The assignment to GITLAB_PROVIDER can be removed since it is not used anymore.

ciyer (changes requested):
Thanks for investigating this! Your change made a variable unreferenced, so we can remove it.

lorenzo-cavazzi (approved):

ciyer (approved):
Thanks for sorting this out!

PR #2112, opened 2022-11-02
feat: add cheatsheet based on renku-python json file

andre-code:
PR to create cheatsheet based on renku-python json file


client/src/notebooks/components/SessionCheatSheet.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const [groups, setGroups] = useState<GroupCommand[]>([]);
+  useEffect(() => {
+    const data = JSON.parse(JSON.stringify(cheatsheetJson));
+    const allGroups: GroupCommand[] = data?.default?.groups;
+    const uiGroups = allGroups.filter((group: GroupCommand) => {
+      const uiCommands = group?.commands?.filter( (command: Command) => {
+        return command.target?.includes("ui");
+      });
+      return uiCommands.length;
+    });
```
ciyer, line 141: This would make sense to extract to a helper function.

client/src/notebooks/components/SessionCheatSheet.tsx
```diff
@@ -0,0 +1,153 @@
...
+  target: string[];
+}
+interface GroupCommand {
+  name: string;
+  commands: Command[];
+}
+function SessionCheatSheetGenerated() {
+  const [groups, setGroups] = useState<GroupCommand[]>([]);
+  useEffect(() => {
+    const data = JSON.parse(JSON.stringify(cheatsheetJson));
```
ciyer, line 134: We discussed being able to adapt the cheatsheet to different version of renku. I think it might make sense to make the the version a property.

Clients could do something like:

```
<SessionCheatSheetGenerated version={session.renkuVersion} />
```

Then, it wouldn't be necessary to use state in the component. It could be simplified to:

```suggestion
function SessionCheatSheetGenerated() {
  const groups = (version === 1) ? 
    cheatsheetJsonToGroups(cheatsheetJsonV1) : 
    cheatsheetJsonToGroups(cheetsheetJsonV2);
```

ciyer (changes requested):
This works very well, and I like that it is now much easier to ensure that the UI cheat sheet is in sync with the renku-python cheat sheet.

I have some suggestions around the code to prepare for supporting multiple versions of the cheat sheet.

client/src/notebooks/components/SessionCheatSheet.tsx
andre-code: sure

ciyer (approved):
Fantastic! This should help us keep the cheat sheets in sync! 🎉

PR #2119, opened 2022-11-04
Show statuspage info on maintenance page.

### LINKED ISSUE #1954: Show maintenance message from statuspage (if present) on "renkulab down" page
It would be nice to show maintenance messages when a maintenance is active in the "renkulab down" page.

Currently there is a scheduled maintenance on renkulab. But because components are not available (due to the scheduled maintenance) going to renkulab.io shows only the generic "renkulab down" page. There is nothing wrong with this page but it would be nice to conditionally insert a section that shows the maintenance message (if one is active) inside the renkulab down page.

Because for a user that hasn't seen the prior notifications and just comes to renkulab.io it seems like there is an unexpected outage. And users would have to know / remember where the status page for renkulab (or other deployment is) so that they navigate to it and see whether the outage is planned or not.

image

ciyer:
If RenkuLab is unreachable, show the statuspage info on the "RenkuLab Down" page.

image

# Testing

The easiest way to test is to run the UI locally using `npm run start`. That will bring up the "RenkuLab Down" page. You can modify the config.json and change the `STATUSPAGE_ID` field to try out different scenarios, including:

* `"STATUSPAGE_ID": null` (status page not configured)
* `"STATUSPAGE_ID": "r3j2c84ftq49"` (statuspage set to the renku dev environment

Fix #1954
/deploy extra-values=gateway.gitlabUrl=https://gitlab.dev.renku.ch #persist


lorenzo-cavazzi (dismissed):
This works really well! :tada: 
I added the `/deploy` string so that I could try it in a full deployment.

It would be nice to add e2e tests before merging, but feel free to go on if you feel it would require too much additional time.

ciyer:
> It would be nice to add e2e tests before merging, but feel free to go on if you feel it would require too much additional time.

Good suggestion! I was able to easily add a test to the PR.

lorenzo-cavazzi (approved):
Thank you for adding the tests!

PR #2122, opened 2022-11-07
WIP feat: use REST endpoint for lineage

andre-code:
PR to switch from GraphQL endpoint to REST endpoint for lineage


client/src/api-client/graph.js
```diff
@@ -16,8 +16,6 @@
  * limitations under the License.
  */
 
-import { gql } from "apollo-boost";
```
lorenzo-cavazzi, line 19: Do we use `apollo` somewhere else in the application? I think there other place was in the API client index where we define `graphqlClient`.

 Would you mind giving it a look and removing it from the dependencies if it's not necessary anymore?

client/src/file/Lineage.present.js
```diff
@@ -234,7 +234,7 @@ class FileLineage extends Component {
         lineagesUrl={this.props.lineagesUrl}
         history={this.props.history} /> :
       (this.props.error) ?
-        <p>{this.props.error}</p> :
+        <div>No lineage information.</div> :
```
lorenzo-cavazzi, line 237: Should we set this message in the `handleFileLineageError` function and then we keep visualizing the `props.error` here? Otherwise, we lose the other error cases (the possible non 404 cases), and we won't even need to save error text in  `handleFileLineageError` if we don't show it anywhere

lorenzo-cavazzi (changes requested):
Very nice! I will test it with a couple more cases before the final approval.
I noticed a couple of things we could change in this PR:

* I think we can get rid of the apollo library with this change, could you give it a look? (more info inline).
* I would either visualize the error message set from the `handleFileLineageError` function, or just change that to save a boolean when the error is thrown instead of storing text we don't show (also here more info inline).

andre-code:
Thank you @lorenzo-cavazzi for your review. Sure enough, that was the last use of the `apollo-boots` library, so I removed it from the app. I've also implemented the suggestion to display the error in the .container instead of a static message for all cases.

lorenzo-cavazzi:
Thanks for removing the apollo library!

I tested it with a few different projects. It works well when there is some actual content for the lineage, and also when opening files with no lineage.
When I try to open a folder tho, it fleshes a `404` error message for a moment. I hope it's visible from the following GIF taken from this project https://renku-ci-ui-2122.dev.renku.ch/projects/workflows-pet-projects/nested-workflows/files/lineage/.renku/metadata/01
Could you give it a look?

Peek 2022-11-08 16-24 workflows 404 lineage

andre-code:
Thanks @lorenzo-cavazzi for your review, that case is because it tries to retrieve the file to get the file size, but for folders it always gets a 404 error, so it doesn't make sense to show it in the lineage component. I have removed to show the error in that case.

lorenzo-cavazzi:
I was trying to test but I keep getting the KG message about a recently started re-indexing on all projects.
Do you get the same?

image

lorenzo-cavazzi:
I managed to index a project, but I always get "no lineage".
Here is an example https://renku-ci-ui-2122.dev.renku.ch/projects/demo/flights-tutorial/files/lineage/src/count_flights.py
image

This is surprising since the workflows tab shows it was used at some point https://renku-ci-ui-2122.dev.renku.ch/projects/demo/flights-tutorial/workflows/5df350711d24487986d484b55b449ac6

And it actually does work if I try it on dev https://dev.renku.ch/projects/demo/flights-tutorial/files/lineage/src/count_flights.py
image

This is surprising since it was working on an earlier version :thinking:

lorenzo-cavazzi (approved):
This solution works well, thanks for handling all the details around it!

lorenzo-cavazzi (approved):

lorenzo-cavazzi (approved):

lorenzo-cavazzi (approved):
:rocket:

PR #2123, opened 2022-11-07
feat: upload artifact cypress

andre-code:
PR to upload cypress artifacts.

/deploy #persist

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -123,6 +123,21 @@ jobs:
         with:
           working-directory: e2e
           record: true
+        # NOTE: cypress screenshots will be generated only if the test failed
+        # thus we store screenshots only on failures
+      - uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: cypress-screenshots
+          path: e2e/cypress/screenshots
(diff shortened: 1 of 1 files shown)


olevski:
Btw upgrading to the latest cypress v10 is probably worth it. The videos are nicer from it IMO. The videos here seem to get stuck after the first minute or so.

andre-code:
> Btw upgrading to the latest cypress v10 is probably worth it. The videos are nicer from it IMO. The videos here seem to get stuck after the first minute or so.
> 

Thanks for the review @olevski , it would definitely be great to upgrade it to 10, we have created this Issue to make that part https://github.com/SwissDataScienceCenter/renku-ui/issues/2128 so it will not be covered in this PR.

> Alternatively you can try switching from using the electron browser to chrome. That may work better when it comes to getting clearer videos.


thanks, for the suggestion. I have modified it to use Chrome as the browser.

lorenzo-cavazzi (approved):
This will be very useful!
The videos are kinda poor at the moment: I see very few frames. I checked on the other PR on renku where we use cypress 10 and the quality seems to be good there. Still, there are no drawbacks in merging the PR as is already even though it will be more useful when we update Cypress

PR #2126, opened 2022-11-08
feat: add static files for renku-gateway

andre-code:
PR to add static files (css, fonts, images) for renku-gateway.

gateway styles

/deploy #persist


ciyer (approved):
👍

lorenzo-cavazzi (changes requested):
Sorry to be pedantic here: could we remove one of the `public` from the URL and make it only `client/public/static`?

Screenshot_20221114_092556

lorenzo-cavazzi (approved):

PR #2133, opened 2022-11-10
build: update server node version to 18.12

### LINKED ISSUE #2037: Update server to newer node version
Node 14.x is in maintenance which will end sometime in 2023.

Also we should use the exact same image for building and running. Currently we use slightly different images in the dockerfile.

https://nodejs.org/en/about/releases/

ciyer:
Fix #2037

/deploy #persist

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:14.18-alpine as builder
+FROM node:18.12.1-alpine3.16 as builder
 
 WORKDIR /app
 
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi:
Thanks for updating this! Should we update the client as well, so that we keep both aligned?

lorenzo-cavazzi (approved):

lorenzo-cavazzi (approved):

PR #2134, opened 2022-11-10
chore: modify renku version check url

olevski:
The core service was serving this data at both /api/renku/versions and /api/renku/version.

It is better to standardize and just have one endpoint.

Since the endpoint returns information about all supported renku versions - I propose we use the one with the plural in its name. And the core service will no longer support the url with the singular variant.

/deploy #persist #cypress renku-core=develop renku=core-svc-horizontal-scaling renku-gateway=master


lorenzo-cavazzi (dismissed):
Lgtm!
Since it's a breaking change for the UI, if you don't mind I'll wait until just before the release to merge it, so we won't need to enrich the `deploy` string with all the extra values in all the other UI PRs

lorenzo-cavazzi:
This is waiting for https://github.com/SwissDataScienceCenter/renku-python/pull/3178

PR #2144, opened 2022-11-16
fix(client): prevent resetting user inputs on the Fork component

lorenzo-cavazzi:
_WIP_

/deploy #persist


andre-code (approved):
Lgtm, as a quick fix it works very well.

ciyer (approved):
This works! With this fix in place, we should be able to start working through the backlog, and we can return to this some time later.

PR #2145, opened 2022-11-16
feat(server): Add pull session status websocket handler (#2016)

### LINKED ISSUE #2016: Move the sessions polling to the server and notify using the websocket channel
## Motivation

As a follow-up to #1338, we should stop looping for the sessions in the UI client and move the logic to the UI server.

## Plan

This is broadly how this should work:

**SERVER**
- Query sessions in the short loop, hash the whole content, and store it locally (we might want to remove data we expect to change often like resources usage and timestamps, or we might even just take a slice of the whole response as far as it detects all the relevant changes)
- Compare the hashed content to verify whether something changed. If that's the case, store the new content and send a "sessions changed" notification to the client.

**CLIENT**
- When getting a "sessions changed" notification, refretch the sessions _when relevant_ . This might either be _always_ or depending on the current page (i.e. relevant when look at any project page or any session page).

**INFO**
A stub of the code is already available here https://github.com/SwissDataScienceCenter/renku-ui/blob/master/server/src/websocket/handlers/sessions.ts
Please mind that this was made for a previous POC and the logic has changed a bit in the meanwhile (E.G. we might not want "session-id" or "storage" anymore, we can cache on the channel `data` property; we also want to notify all the channels directly from the function).

andre-code:
PR to add session status websocket handler.

- [x] (ui) Always check if there are changes in the session status using the websocket channel.
- [x] (ui) Ignore changes if the pathname is a sessionUrl (`*/sessions`,`*/sessions/new`, `*/sessions/show/*` ).
- [x] (ui) if there is a change and the pathname is not a session page pull the session using the NotebooksCoordinator.
- [x] (server-ui) Add websocket handler that check the sessions status object and send a message ("true") when there is a change.
- [ ] Add Tests

Closes #2016 

/deploy #persist


server/src/websocket/handlers/sessions.ts
```diff
@@ -16,51 +16,61 @@
...
-  const oldHash = await storage.get(storageKey) as string;
-  if (oldHash && oldHash !== hashedSessions)
-    changed = true;
-  if (!oldHash || oldHash !== hashedSessions)
-    await storage.save(storageKey, hashedSessions);
-  return changed;
+      // only send message when something change
+      if (!util.isDeepStrictEqual(previousStatuses, cleanStatus)) {
+        sendMessage("true", channel);
+        channel.data.set("sessionStatus", cleanStatus);
```
lorenzo-cavazzi, line 67: Since we use `cleanStatus` just to compare with the previous one, can't we just store a hash of the stringified object?

server/src/api-client/index.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  constructor() {
+    this.gatewayUrl = config.deployment.gatewayUrl;
+  }
+
+  /**
+   * Fetch session status
+   *
+   */
+  async sessionStatus(authHeathers: Record<string, string>): Promise<Response> {
```
lorenzo-cavazzi, line 41: This is just a cosmetic detail: can we call it `getSessionStatus` to make it clear it's a function to get some data?

server/src/api-client/index.ts
```diff
@@ -0,0 +1,125 @@
...
+    const newCookies: Array<string> = [];
+    if (anonId) {
+      // ? the anon-id MUST start with a letter to prevent k8s limitations
+      const fullAnonId = config.auth.anonPrefix + anonId;
+      newCookies.push(
+        serializeCookie(config.auth.cookiesAnonymousKey, fullAnonId)
+      );
+    }
+    if (newCookies.length > 0)
+      options.headers.set("cookie", newCookies.join("; "));
```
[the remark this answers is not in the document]
andre-code: thanks, I have moved to add the cookie to the wsRenkuAuth function

client/src/App.js
```diff
@@ -212,7 +214,11 @@ class App extends Component {
       webSocketUrl = "ws" + webSocketUrl.substring(4);
     // ? adding a small delay to allow session cookie to be saved to local browser before sending requests
     setTimeout(
-      () => { this.webSocket = setupWebSocket(webSocketUrl, this.props.model); return; },
+      () => {
+        this.webSocket = setupWebSocket(webSocketUrl, this.props.model, getLocation, this.props.client);
+        this.webSocket?.send(JSON.stringify(new WsMessage({}, "pullSessionStatus")));
```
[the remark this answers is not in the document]
andre-code: true 😉, added in the onopen function

client/src/utils/helpers/url/Url.js
```diff
@@ -460,7 +460,20 @@ function getSearchParams(expectedParams = null, convertParams = null, convertTyp
...
+ *
+ * @param {string} [url] - url
+ * @returns {boolean} if match with a session url
+ */
+function isSessionUrl(url) {
+  const isNewSessionUrl = /\/sessions\/new/g.exec(url);
+  const isShowSessionUrl = /\/sessions\/show\//g.exec(url);
+  const endUrl = /\w+$/g.exec(url);
+  return isNewSessionUrl?.length || isShowSessionUrl?.length || (endUrl && endUrl[0] === "sessions");
+}
```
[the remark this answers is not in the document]
andre-code: added unit tests👌

server/src/websocket/handlers/sessions.ts
andre-code: if we make a hash of the object to save it in the channel then the comparison would be hash vs hash of the new state and it is not recommended since there is no guarantee that the order of the properties in the object will always be the same, this also applies after the cleanup.

client/src/websocket/index.ts
```diff
@@ -95,13 +107,21 @@ function setupWebSocket(webSocketUrl: string, fullModel: any) {
     }
   }
 
+  function startPullingSessionStatus(targetWebSocket: WebSocket) {
+    if (model.get("open") && targetWebSocket.readyState === targetWebSocket.OPEN)
+      targetWebSocket.send(JSON.stringify(new WsMessage({}, "pullSessionStatus")));
+  }
+
```
lorenzo-cavazzi, line 114: The `model.setObject` function gets invoked just before `startPullingSessionStatus`. In the past, it happened that the effect wasn't immediate and it took a moment before `model.get` can return the correct value. I suggest removing the section
```
if (model.get("open") && targetWebSocket.readyState === targetWebSocket.OPEN)
```
and moving the `startPullingSessionStatus` invocation (see the other comment below)

client/src/websocket/index.ts
```diff
@@ -95,13 +107,21 @@ function setupWebSocket(webSocketUrl: string, fullModel: any) {
...
   webSocket.onopen = (status) => {
     // start pinging regularly when the connection is open
     const target = status.target as WebSocket;
     const webSocketOpen = target && target["readyState"] ? true : false;
     if (webSocketOpen)
       model.setObject({ open: true, error: false, lastReceived: null });
 
+    // request session status
+    startPullingSessionStatus(webSocket);
```
lorenzo-cavazzi, line 123: I would move this in the `if (webSocketOpen) {` block above so that we don't need to check for the `model` status in the function. Also, a `model.get` is occasionally unreliable if invoked just after the corresponding `model.set`

client/src/utils/helpers/url/Url.test.js
```diff
@@ -133,6 +133,17 @@ describe("UrlRule private class", () => {
...
 
+describe("Url session validation", () => {
+  it("Test valid and invalid session url", () => {
+    expect(isSessionUrl("projects/namespaceProject/projectName/sessions")).toBe(true);
+    expect(isSessionUrl("projects/namespaceProject/projectName")).toBe(false);
+    expect(isSessionUrl("projects/namespaceProject/projectName/sessions/new")).toBe(true);
+    expect(isSessionUrl("projects/namespaceProject/projectName/sessions/show/sessionName")).toBe(true);
+    expect(isSessionUrl("projects/datasets")).toBe(false);
+    expect(isSessionUrl("projects/sessions")).toBe(true);
+  });
```
lorenzo-cavazzi, line 144: Could we add a test to verify that `"projects/namespaceProject/session"` tests `false`? Just in case someone calls the project `session` :grin: 
P.S: better to verify first that other parts of the UI work well in that case. If not, we can ignore this case

server/src/authentication/middleware.ts
```diff
@@ -103,8 +106,20 @@ async function wsRenkuAuth(authenticator: Authenticator, sessionId: string): Pro
...
+  const fullAnonId = config.auth.anonPrefix + sessionId;
+  const authorizationHeader = { [config.auth.cookiesAnonymousKey]: sessionId };
+  const headers = new Headers(authorizationHeader);
+  const newCookies: Array<string> = [];
+  newCookies.push(
+    serializeCookie(config.auth.cookiesAnonymousKey, fullAnonId)
+  );
+  if (newCookies.length > 0)
+    headers.set("cookie", newCookies.join("; "));
+  return headers;
```
lorenzo-cavazzi, line 122: This is the right place where to handle the anonymous users' case, but the change seems inconsistent with what we did in the `kg-search` PR.
What about keeping the `Record<string, string>` approach and we re-work things so that we get something like:

```
 if (tokens) {
    const value = config.auth.authHeaderPrefix + tokens.access_token;
  return { [config.auth.authHeaderField]: value };
}
else {
    const value = sessionId;
return {[config.auth.cookiesAnonymousKey]: value}  
}
```

Then the API would be similar to this: https://github.com/SwissDataScienceCenter/renku-ui/pull/2170/files#diff-05130c6c244b058adb42428a05517effbe3d9f809cc76224e7552b3cce1b4d89

server/src/websocket/handlers/sessions.ts
lorenzo-cavazzi: Can't use alphabetically order servers so that we won't have inconsistencies? Altough I remember trying this is the past and the ordering was consistent with the start order of the servers, so it might not be necessary. We are saving a lot of data just to compare 2 strings.
I suggest either giving it a try now or, if you prefer, we can make a follow-up issue time-limited to 1 day to try that change and see if it works.

lorenzo-cavazzi (changes requested):
Great, almost there! I have just another detail on how we handle server-side headers so that we can keep that coherent with the changes already done in the `kg-search` branch.

server/src/authentication/middleware.ts
andre-code: 👌, I went back to using Record instead of Header to make it easier to rebase with the kg-search branch.

lorenzo-cavazzi (changes requested):
Great! This works well!
I tested a bit more with anonymous users and I submitted a change to clean the headers (it should work but give it a quick test) along with a couple of other cosmetic changes. Nothing big, the PR is basically ready.

Also, give a run to `npm run lint` on the "server" folder. There are a couple of lining errors but somehow the tests are not failing on that
image

lorenzo-cavazzi (approved):
This works well! Let's merge it :rocket:

PR #2150, opened 2022-11-17
fix: limit re-rendering of fork project modal

ciyer:
Another attempt to fix the fork modal.

/deploy #persist


ciyer:
We have opted for #2144

PR #2159, opened 2022-11-21
minor: improve styling of tables in markdown cells of notebooks

ciyer:
**New appearance**
image

**Old appearance**
image

**(post framework upgrade appearance)**
image

/deploy renku=2115-unify-start-session-flow #persist


lorenzo-cavazzi (changes requested):
I feel having the borders around all the cells would be better -- keeping it a bit more similar to the GitLab preview.
In the following example, on the left (GitLab) it looks immediately clear it's a table, while on the right (our current solution) it took me a moment to understand where the table content starts and ends, especially considering that spacing on other blocks is a bit off.

image

ciyer:
Updated image
image

lorenzo-cavazzi (approved):
The last solution works well, thank you!

PR #2160, opened 2022-11-21
Unify session log styling

### LINKED ISSUE #2139: Streamline sessions logs
## Motivation

We show session logs in 2 locations: the "Resources" panel in the running session, and the "Show logs" panel from the list of sessions.

We should unify and improve the output.

## Proposal

1. Unify the layout: let's use the "Resources" layout (on the right in the screenshot) also for the other logs modal.  That means:
    * Enclose the logs on a white background panel
    * Move the "Refresh" and "Download" buttons to the top left with the same style.
    * Use the same element for the "Logs" title as the one in "Resources". We can leave the details about the project,  branch and commit but move that below the title.
    * We could use a shared the component that includes even more common code (only if helpful/feasible)
    Screenshot_20221114_113942
2. We can use a larger model (fluid modal?), or improve the braking point; at the moment, it's very narrow when the browser width is too small.
Peek 2022-11-14 11-33
3. Verify what's going on with the vertical scroll bar: we sometimes have 2 bars, one in the log component with the black background, one in the modal. We should always have only 1 (ideally in the code component, unless the other way is easier). P.S. it's probably more visible in the GIF above than in the following screenshot -- Firefox has very minimal scrollbars
Screenshot_20221114_113406

ciyer:
Fix #2139 

/deploy renku-notebooks=fix-logs-maxlines-parameter #persist

### DIFF
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -146,6 +146,7 @@
           "dockerfile",
           "dockerignore",
           "enum",
+          "fetchable",
           "formgenerator",
           "fortran",
           "frameborder",
--- a/client/src/notebooks/Notebooks.css
+++ b/client/src/notebooks/Notebooks.css
@@ -34,34 +34,8 @@
   }
 }
 
-
 .log-container {
   white-space: pre-line;
-}
-
-/* Modals body personalization for logs */
-.logs-modal.log-container {
-  white-space: pre-line;
-  height: 600px;
-}
-
-@media (max-height: 1024px) {
-  .logs-modal .log-container {
-    height: 400px;
-  }
-}
-
-@media (max-height: 700px) {
-  .logs-modal .log-container {
-    height: auto;
-  }
-}
-
-/* Logs personalization for session-view */
-@media (min-height: 1200px) {
-  .log-container {
-    height: 600px;
-  }
 }
 
 /* sessions page borders */
--- a/client/src/notebooks/Notebooks.present.js
+++ b/client/src/notebooks/Notebooks.present.js
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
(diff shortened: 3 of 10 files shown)


ciyer:
# Screenshots

**Sessions list**
image

**Notebook**
image

**Scrolling**
image

**Narrow viewport**
image

lorenzo-cavazzi (changes requested):
This works very well in most cases, but I identified a problem.
On the full-screen session page, the modal seems to break when the screen is small (not sure if it's because it expands to the whole window):
* there is no space between the tabs and the title "Resources"
* I see the 2 scroll bars on the right

image

Otherwise, all the points are fixed and the code looks good!
It also requires updating the 2e2 tests for the logs, I've noticed they failed.

ciyer:
The cosmetic issues have been fixed (spacing of the header and removed the extra scroll bars on narrow screens). The tests still remain, though.

image

lorenzo-cavazzi:
Thank you, that works well!
Another detail I noticed: downloading the logs from the session full-page works, but trying it from the sessions list results in a bunch of files containing just single letters.

image

ciyer:
> Thank you, that works well! Another detail I noticed: downloading the logs from the session full-page works, but trying it from the sessions list results in a bunch of files containing just single letters.

So the problem downloading the logs has renku-notebooks at its source. A fix is in the works: https://github.com/SwissDataScienceCenter/renku-notebooks/pull/1336

The fact that downloading from the session full-page worked, was actually a bug (it did not download the full logs, just the excerpt that is shown in the UI)! But that has been fixed, and so have the tests.

lorenzo-cavazzi (approved):
This works well! Thank you for taking care of other logs-related problems here :pray:

Issue #2128, opened 2022-11-08
Upgrade to cypress 11


andre-code:
## Motivation

Cypress has release a new version in June 2022 with new functionalities like Component testing that would be great to start using.
More information here: ~~https://www.cypress.io/blog/2022/06/01/cypress-10-release/~~.

**Update** Cypress 11 is already out: https://www.cypress.io/blog/2022/11/08/announcing-cypress-11/

PR #2173, opened 2022-11-29
tests: migrate to cypress 11

ciyer:
Update the e2e tests to the latest version of cypress

/deploy #persist
re #2128


lorenzo-cavazzi (commented):
Looks good! In the end, not many changes were required :tada: 

P.S: I find the folder naming slightly confusing with the repeated `e2e`. I assume that's used for the `--e2e` flag? Should we consider renaming the first `e2e` to `tests` so that we get `/tests/cypress/e2e`?
Actually, this is just a detail. If you prefer the current structure I'm fine with that, and I'm gonna approve after the temp file is removed

Screenshot_20221129_162116

ciyer:
> P.S: I find the folder naming slightly confusing with the repeated `e2e`. I assume that's used for the `--e2e` flag? Should we consider renaming the first `e2e` to `tests` so that we get `/tests/cypress/e2e`? Actually, this is just a detail. If you prefer the current structure I'm fine with that, and I'm gonna approve after the temp file is removed

That is a good point. I have made this change.

lorenzo-cavazzi (approved):
Thank you! The folder structure looks cleaner now.
I re-ran the acceptance tests

PR #2176, opened 2022-11-30
refactor(client): Fork component to use redux toolkit (#2157)

### LINKED ISSUE #2157: Fix the Fork components
## Motivation

We recently made a quickfix to prevent the Modal component from resetting every time the Redux store updates (`connect` was triggering `mapStateToProps` for _any_  change to `state` -- see here for the changes  https://github.com/SwissDataScienceCenter/renku-ui/pull/2144 ).
Now projects and namespaces are not connected properly to the final component.

## Proposal

We should fix the `<ForkProjectMapper>` component and its descendants.
Ideally, we can move away from the `ProjectsCoordinator` and leverage RTK's `createApi` to fetch featured projects and available namespaces. In the process, we will probably remove also `ForkProjectMapper` leaving just `ForkProject` where we import the new `use(xyz)Query` functions created through `createApi`.
 
## Follow-up ⚠️

After closing this issue, we should create ore or more follow-up issues to swap the other `ProjectsCoordinator` functions with the new RTK Query functions.

andre-code:
PR to refactor fork component to use redux toolkit to fetch namespaces, groups and member projects.

Closes #2157 

/deploy #persist

### DIFF
--- a/client/src/features/projects/ProjectApi.ts
+++ b/client/src/features/projects/ProjectApi.ts
@@ -35,16 +35,16 @@ export const projectApi = createApi({
   reducerPath: "projects",
   baseQuery: fetchBaseQuery({ baseUrl: "/ui-server/api" }),
   endpoints: (builder) => ({
-    namespaces: builder.query({
+    getNamespaces: builder.query({
       query: (arg) => ({ url: "/namespaces", method: "GET" }),
     }),
-    groupByPath: builder.query<any, string>({
+    getGroupByPath: builder.query<any, string>({
       query: (projectPath) => {
         const urlEncodedPath = encodeURIComponent(projectPath);
         return { url: `/groups/${urlEncodedPath}`, method: "GET" };
       }
     }),
-    memberProjects: builder.query<any, QueryParams>({
+    getMemberProjects: builder.query<any, QueryParams>({
       query: (queryParams: QueryParams ) => {
         const params = { "variables": null, "operationName": null };
         let query = `{
@@ -108,4 +108,4 @@ export const projectApi = createApi({
 
 // Export hooks for usage in function components, which are
 // auto-generated based on the defined endpoints
-export const { useNamespacesQuery, useGroupByPathQuery, useMemberProjectsQuery } = projectApi;
+export const { useGetNamespacesQuery, useGetGroupByPathQuery, useGetMemberProjectsQuery } = projectApi;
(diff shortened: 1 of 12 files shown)


ciyer (approved):
Code looks great! Good to see these modernizations getting implemented. I tried forking a few projects, and everything worked as before as far as I can tell.

client/src/features/projects/ProjectApi.ts
```diff
@@ -0,0 +1,111 @@
...
+  endCursor: string;
+  hasNextPage: boolean;
+}
+
+
+export const projectApi = createApi({
+  reducerPath: "projects",
+  baseQuery: fetchBaseQuery({ baseUrl: "/ui-server/api" }),
+  endpoints: (builder) => ({
+    namespaces: builder.query({
```
lorenzo-cavazzi, line 38: I suggest trying to start the RTK Query functions with a verb that explains the action.
Here it would be `getNamespaces` instead of `namespaces`. Similar pattern for the other endpoints.
Following the current naming convention, the exported API would become something like `useGetNamespacesQuery` that makes it clear what the action is.

client/src/utils/customHooks/UseGetProjects.ts
```diff
@@ -0,0 +1,60 @@
...
+import { useEffect, useState } from "react";
+import { useMemberProjectsQuery } from "../../features/projects/ProjectApi";
+
+/**
+ *  useGetProjects custom hook
+ *
+ *  UseGetProjects.ts
+ *  hook to fetch member projects
+ */
+function useGetProjects() {
```
lorenzo-cavazzi, line 28: Instead of `GetProjects`, should that be `GetUserProjects`?

lorenzo-cavazzi (changes requested):
All works very well!
With RTK, we have a few extra names for functions. How do you feel about always starting with a verb that describes the action (e.g. `get`, `set`, `delete`...)?

client/src/features/projects/ProjectApi.ts
andre-code: > All works very well! With RTK, we have a few extra names for functions. How do you feel about always starting with a verb that describes the action (e.g. `get`, `set`, `delete`...)?

Thanks for your suggestions, I have implemented them in the last commit. 🙌

andre-code:
> All works very well! With RTK, we have a few extra names for functions. How do you feel about always starting with a verb that describes the action (e.g. `get`, `set`, `delete`...)?

Thanks for your suggestions, I have implemented them in the last commit. 🙌

lorenzo-cavazzi (approved):
Lgtm :rocket: 
P.S. Do you mind opening another issue to migrate all the other APIs from the `ProjectsCoordinator` (I saw it's mentioned here https://github.com/SwissDataScienceCenter/renku-ui/issues/2157 )? Since you worked on this, you know better than me which parts are left :wink:

PR #2182, opened 2022-12-01
Integrate test-renku-cypress action

ciyer:
/deploy #persist #notest #cypress

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -35,6 +35,7 @@ jobs:
       renku-graph: ${{ steps.deploy-comment.outputs.renku-graph}}
       renku-notebooks: ${{ steps.deploy-comment.outputs.renku-notebooks}}
       test-enabled: ${{ steps.deploy-comment.outputs.test-enabled}}
+      test-cypress-enabled: ${{ steps.deploy-comment.outputs.test-cypress-enabled}}
       persist: ${{ steps.deploy-comment.outputs.persist}}
       extra-values: ${{ steps.deploy-comment.outputs.extra-values}}
     steps:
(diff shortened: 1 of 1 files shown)


ciyer:
# Testing

Try commands like 
```
/deploy #persist #notest #cypress renku=update-renku-to-use-new-gitlab-chart
```

.github/workflows/acceptance-tests.yml
```diff
@@ -105,7 +106,20 @@ jobs:
...
+    steps:
+      - name: Extract renku repo ref
+        run: echo "RENKU=`echo '${{ needs.check-deploy.outputs.renku }}' | cut -d'@' -f2`" >> $GITHUB_ENV
+      - uses: SwissDataScienceCenter/renku-actions/test-renku-cypress@000-cypress-tests
+        with:
+          gitlab-token: ${{ secrets.DEV_GITLAB_TOKEN }}
+          renku-release: renku-ci-ui-${{ github.event.number }}
+          test-user-password: ${{ secrets.RENKU_BOT_DEV_PASSWORD }}
+          renku: ${{ env.RENKU }}
+          test-timeout-mins: "20"
```
lorenzo-cavazzi, line 122: Should we try to leave this out for now and rely on the default timeout of the action? It would be nice to set that to a proper value so that we don't need to adjust it in every repo where the actions are used (P.S. default is 60 mins iirc, but it's fine since the Selenium tests last longer anyway -- we can re-adjust once we add more tests and have a better estimation of the expected runtime)

lorenzo-cavazzi (changes requested):
I would remove the timeout for now and rely on the default (that we can better adjust later).
Otherwise, all good! Thanks for going through the renku-actions repo and implementing all the necessary logic!

As you suggested, let's wait for the referenced action to be merged before merging this

lorenzo-cavazzi (approved):
Lgtm! It's great to have the cypress test available in the UI!

lorenzo-cavazzi (approved):

PR #2187, opened 2022-12-05
feat: get namespaces owned only when fork a project

### LINKED ISSUE #2027: when creating renku project, groups where user is "Direct Member" are shown, and only those

andre-code:
PR to fetch namespaces owned only for fork and create projects.
Pending:
- [x] Fix e2e tests

/deploy #persist #cypress renku=renku-ui-2.15.0-tests

closes #2027

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -113,6 +113,7 @@ function CentralContentContainer(props) {
             key="newProject"
             model={props.model}
             user={props.user}
+            client={props.client}
             {...p}
           /></ContainerWrap>}
         />
--- a/client/src/project/new/ProjectNew.container.js
+++ b/client/src/project/new/ProjectNew.container.js
@@ -23,15 +23,13 @@
  *  Container components for new project
  */
 
-import React, { Component, useEffect, useState, useRef } from "react";
-// TODO: switch to useSelector
-import { connect, useSelector } from "react-redux";
-import { useLocation, withRouter } from "react-router-dom";
+import React, { useEffect, useState, useRef, useContext, Component } from "react";
+import { useSelector } from "react-redux";
+import { useLocation } from "react-router-dom";
 import { useHistory } from "react-router";
 
 import { NewProject as NewProjectPresent, ForkProject as ForkProjectPresent } from "./ProjectNew.present";
-import { NewProjectCoordinator, validateTitle, checkTitleDuplicates } from "./ProjectNew.state";
-import { ProjectsCoordinator } from "../shared";
+import { validateTitle, checkTitleDuplicates, NewProjectCoordinator } from "./ProjectNew.state";
 import { gitLabUrlFromProfileUrl, slugFromTitle } from "../../utils/helpers/HelperFunctions";
 import { Url, getSearchParams } from "../../utils/helpers/url";
 import { atobUTF8, btoaUTF8 } from "../../utils/helpers/Encoding";
@@ -40,6 +38,7 @@ import AppContext from "../../utils/context/appContext";
 import useGetNamespaces from "../../utils/customHooks/UseGetNamespaces";
 import useGetUserProjects from "../../utils/customHooks/UseGetProjects";
 import useGetVisibilities from "../../utils/customHooks/UseGetVisibilities";
+import { Loader } from "../../utils/components/Loader";
 
 const CUSTOM_REPO_NAME = "Custom";
 
@@ -92,7 +91,7 @@ function addForkNotification(notifications, url, info, startingLocation, success
 
 function ForkProject(props) {
   const { client, forkedId, forkedTitle, toggleModal } = props;
-  const namespaces = useGetNamespaces();
+  const namespaces = useGetNamespaces(true);
   const { projectsMember, isFetchingProjects } = useGetUserProjects();
 
   const [title, setTitle] = useState(forkedTitle);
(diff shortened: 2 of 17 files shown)


client/src/App.js
```diff
@@ -113,6 +114,9 @@ function CentralContentContainer(props) {
             key="newProject"
             model={props.model}
             user={props.user}
+            coordinator={
+              new NewProjectCoordinator(
+                props.client, props.model.subModel("newProject"), props.model.subModel("projects"))}
```
lorenzo-cavazzi, line 119: I understand the reason for this, but it feels a bit weird importing and creating the `NewProjectCoordinator` here in the App (or in the `AddDatasetNewProject` section).
What about adding a temporary wrapper?
This would mean creating a `class NewProjectWrapper` in the "ProjectNew.coordinator.js" file that takes the props, create the coordinator and passes it as a prop to the functional component. And that can be exported as `export { NewProjectWrapper as NewProject}`
That way, once we remove the wrapper, we don't need to modify the code wherever we use `NewProject` but just clean up the  "ProjectNew.coordinator.js" content.
P.S: I realize this is harder to explain than to implement. I'm happy to push a commit with this change.

client/src/project/new/ProjectNew.state.js
```diff
@@ -701,10 +689,13 @@ class NewProjectCoordinator {
 
     // check warnings (temporary problems)
     let warnings = {};
-    if (projects && projects.namespaces.fetching)
+    if (namespaces && namespaces.fetching)
       warnings["namespace"] = "Fetching namespaces...";
 
-    if (meta.namespace.fetching)
+    if (projects && projects.fetching)
+      warnings["title"] = "Fetching projects to avoid duplicated projects...";
```
lorenzo-cavazzi, line 696:
```suggestion
      warnings["title"] = "Fetching projects to prevent duplicates...";
```

lorenzo-cavazzi (dismissed):
This looks good!
I suggest using a wrapper instead of creating the old "NewProjectCoordinator" in other contexts for the time being -- I added a description inline but I'm also happy to implement it if the comment is not clear.

I've noticed a detail with visibilities: when creating a new project, while re-computing the available visibilities, it doesn't show the temporary loader. The behavior is correct on the Fork tho. 
I added a gif to explain it better (the comparison is PR vs DEV since I didn't notice it was already working well on forks also in the PR)

Peek 2022-12-09 11-29

P.S. the local e2e tests are failing; it's probably necessary to update them.

ciyer (changes requested):
This looks very good! When reviewing the code, I noticed a couple of small things. The first one was there before your code, but since we are making changes, I think it would make sense to improve.

lorenzo-cavazzi:
Another detail: could you incorporate the fix to the `getSlugAndReset` function from here? https://github.com/SwissDataScienceCenter/renku-ui/pull/2209

It's just a tiny change, from this line
> return `${creation.newNamespace}/${creation.newName}`;

to this line
> return `${creation.newNamespace}/${creation.newNameSlug}`;

client/src/App.js
andre-code: sure, added the wrapper

client/src/project/new/ProjectNew.state.js
```diff
@@ -153,47 +155,41 @@ class NewProjectCoordinator {
...
       if (!templates) {
         const templatesStatus = this.model.get("templates");
         if (templatesStatus.fetched && !templatesStatus.fetching)
           templates = templatesStatus.all;
       }
-
       await sleep(intervalLength);
-    } while (!namespaces || !templates);
-    const { data } = automated;
+    } while (!templates);
```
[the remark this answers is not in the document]
andre-code: cool, I added a counter to stop the loop if the resource has not been stopped/find after ˜60sec.

client/src/project/new/ProjectNew.container.js
```diff
@@ -348,205 +347,212 @@ function getDataFromParams(params) {
...
+   * Validate form when projects/namespace are ready or the auto fill form finished
+   */
+  useEffect(() => {
+    if (namespaces.fetching || isFetchingProjects || (newProject.automated.received && !newProject.automated.finished))
+      return;
+    validateForm(null, null, true);
+  }, [ //eslint-disable-line
+    namespaces.list,
+    namespaces.fetching,
+    projectsMember,
```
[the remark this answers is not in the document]
andre-code: because it is used in the function called in the useEffect, if the data changes, it must call the function again.

andre-code:
thanks team for your review, I added all your suggestions 💪

lorenzo-cavazzi (commented):
This works very well!

The acceptance tests seem to break at project creation. Is it possible that a change in this PR breaks them?

andre-code:
> This works very well!
> 
> The acceptance tests seem to break at project creation. Is it possible that a change in this PR breaks them?

Thanks @lorenzo-cavazzi I re-deployed everything because I can't reproduce the error locally 🤞.

ciyer (approved):
The changes look good! There were some problems with the tests because the test user had thousands of projects and some of the actions were taking longer now, but those problems have now been fixed.

PR #2188, opened 2022-12-05
test: fail tests on linting warnings

lorenzo-cavazzi:
Fix linting
/deploy #persist #notest


ciyer (approved):
Good idea! Good to see that there were no listing issues that needed to be fixed, but it's good to add this to our CI pipeline.

PR #2201, opened 2022-12-08
000 testing kg search

andre-code:
PR to test cross entity search.

Rel #1894 
/deploy #persist


andre-code:
Used for testing purposes only

PR #2203, opened 2022-12-08
build: fix client dependencies

lorenzo-cavazzi:
Fix an issue with react-script that prevents starting the client in dev mode
/deploy #persist


ciyer (dismissed):
I was able to recreate the original problem and can confirm that downgrading the storybook dependency fixes it.

andre-code (approved):
lgtm 🚀

PR #2207, opened 2022-12-09
chore(client): update S3 sources descriptions

### LINKED ISSUE #2198: We cannot mount azure blob storage in sessions
Currently we only support mounting s3 compatible storage in sessions.

But in the ui we mention that we support azure blob storage. We should remove this.

image

image

lorenzo-cavazzi:
Update description for S3 services

/deploy #persist
fix #2198


lorenzo-cavazzi:
Thank you for the addition! :pray:

PR #2208, opened 2022-12-09
fix: UI kg-search glitches

andre-code:
This PR is to fix the following cases:
- The Visibility options aren’t checked and still we get results: all should be checked
    
- The “Hide filters” section on the top disappears when there are no results
  
- Should we order by newest by default instead of A-Z? Or by relevance (but KG should improve `relevance` to order by `latest` as second ordering item). Also A-Z sort should be case-insensitiveAlso A-Z sort should be case-insensitiveAlso A-Z sort should be case-insensitive
      
- Should we move the `Author` above visibility to simplify filtering for “own projects”? And add a “user” emoji before the “Owned by me” option, and a world before the “All” option
    
- Clicking on the last search term (after cleaning/changing the search input but not hitting enter) doesn’t work
    
- The quickstart session button doesn’t work: it brings to a 404 page (in the URL, there is only the namespace but not the project name)
    
- Some characters break the search — *we should URL encode (E.G. the query should look like this* `curl --location --request GET '[https://renku-ci-ui-2201.dev.renku.ch/knowledge-graph/entities?query=cramakri%20kg%20test&sort=matchingScore:desc](https://renku-ci-ui-2201.dev.renku.ch/knowledge-graph/entities?query=cramakri%20kg%20test&sort=matchingScore:desc)'`)
    - `:` colon in search string --> 500 error from endpoint, but no error shown in UI
    - `^` or ``` or `%` in search --> 400 error because it breaks the search url (not URL encoded)
    - `#` in search --> everything after the hash is dropped in the query URL. Breaks the page (no more project icons)
    - `&` is also not escaped, breaking search
    - `</` also gives a 500
- Add a “clear” icon to close the dropdown search menu

Pending:
- [ ]  Icon to use in Author filters not to be confused with visibility icons

/deploy #persist


ciyer (changes requested):
Some good fixes here! The following all work for me as expected:

* The “Hide filters” section on the top disappears when there are no results
* Sort by relevance and make it default
* Add a “clear” icon to close the dropdown search menu
* Quickstart session button works
* Author is earlier in the filtering list

I am still encountering problems with the following:

**The Visibility options aren’t checked and still we get results: all should be checked**

The default is that all options are checked, but I can still _uncheck_ all options, and the search results are the same as if all were _checked_. Would it make sense to disable the last checked visibility option to prevent unchecking all options? The same applies to the entity (project/dataset).

**Sort alphabetically**

This is still case sensitive ("Wave" comes before "angel").

**Clearing search input**
Clicking on the last search term (after cleaning/changing the search input but not hitting enter) works generally, but the display is a bit confusing if you select the last tried search term (see video).

https://user-images.githubusercontent.com/1196411/207018849-05eefe65-90ea-4be7-a34b-7133d104e1e5.mov

**characters breaking search**

I do not see errors, but the results are still sometimes confusing when I input these non-alpha-numeric characters.

For example, see this video where I do one search for "^project" which returns 5 results, then I search again and it returns 0 results.

https://user-images.githubusercontent.com/1196411/207020715-8928c675-561f-4af4-b958-c0c722d02319.mov

: colon in search string --> 500 error from endpoint, but no error shown in UI
^ or ``` or % in search --> 400 error because it breaks the search url (not URL encoded)
# in search --> everything after the hash is dropped in the query URL. Breaks the page (no more project icons)
& is also not escaped, breaking search
</ also gives a 500

andre-code:
Thanks @ciyer  for your review, here a description for each of your points.
1. The Visibility options aren’t checked and still we get results: I have added validation so that it does not allow unchecking all options, same for the entity filter type. Here I also included not to show the visibility filters for anonymous users.
2. Sort alphabetically:  There is nothing to do in the UI, here is the issue in the KG https://github.com/SwissDataScienceCenter/renku-graph/issues/1233
3. Clearing search input: Fixed in the last commits
4. characters breaking search: I added an error handler. Thus, if the query returns an error, the user will be informed.
5. I do not see errors, but the results are still sometimes confusing when I input these non-alpha-numeric characters.
For example, see this video where I do one search for "^project" which returns 5 results, then I search again and it returns 0 results. : I have not been able to reproduce this error, could you check if it persists?

ciyer (approved):
Works very well now! 🎉

PR #2209, opened 2022-12-09
fix(client): use project slug for links

lorenzo-cavazzi:
When project creation succeeds but one of the following APIs fails (E.G. the webhook creation) the user gets a warning and can click on a "go to the project" button. The links was broken for projects with different name and slug.
This PR fixes it.

image

/deploy #persist #cypress


ciyer (approved):
👍

PR #2211, opened 2022-12-13
feat: add stop session button by default when session fails (#1892)

### LINKED ISSUE #1892: Sessions: use "stop" as the defaul action when the session is in error state
## Motivation

The default action for sessions in an error state is currently "show logs". This wasn't critical in the past since failed sessions were culled immediately, but now they stay around for some time.

## Proposal

Switch the default action to "stop" and move "show logs" to an item in the button dropdown menu.

andre-code:
PR to add stop button by default when session fails.

Closes #1892 

/deploy #persist

### DIFF
--- a/client/src/notebooks/Notebooks.present.js
+++ b/client/src/notebooks/Notebooks.present.js
@@ -684,7 +684,9 @@ const NotebookServerRowAction = memo((props) => {
   }
   else {
     const classes = { className: "text-nowrap btn-outline-rk-green" };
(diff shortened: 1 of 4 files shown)


lorenzo-cavazzi (approved):
This works well :+1: 

image

P.S. here is a test project -- any session will error https://gitlab.dev.renku.ch/lorenzo.cavazzi.tech/corrupted-project-session-error

PR #2223, opened 2022-12-15
fix: load all inactive projects

andre-code:
PR to test changes in kg-search repo.


andre-code:
I have included some fixes needed after the last kg update.
1. Include a custom hook to retrieve all projects if there are more projects than records per page which was set to 100 records.
2. Remove the default "*" query when there is no phrase, also remove the hasSpecialCharacters validation because so far it only expects strings so it is not necessary to add "*" in the phrase. I keep the hasSpecialCharacters as it is a helper function that we can use later when the advanced search is available.
3. Add show error message if the endpoint to retrieve inactive projects returns an error.
4. Fix typo in the error message when the search returns an error.

andre-code:
duplicated #2229

PR #2222, opened 2022-12-15
[Snyk] Upgrade ws from 7.5.3 to 7.5.9

ciyer:
<p>This PR was automatically created by Snyk using the credentials of a real user.</p><br /><h3>Snyk has created this PR to upgrade ws from 7.5.3 to 7.5.9.</h3>

:information_source: Keep your dependencies up-to-date. This makes it easier to fix existing vulnerabilities and to more quickly identify and fix newly disclosed vulnerabilities when they affect your project.
<hr/>

- The recommended version is **6 versions** ahead of your current version.
- The recommended version was released **5 months ago**, on 2022-07-15.

<details>
<summary><b>Release notes</b></summary>
<br/>
  <details>
    <summary>Package name: <b>ws</b></summary>
    <ul>
      <li>
        <b>7.5.9</b> - <a href="https://snyk.io/redirect/github/websockets/ws/releases/tag/7.5.9">2022-07-15</a></br><h1>Bug fixes</h1>
<ul>
<li>Backported <a class="commit-link" data-hovercard-type="commit" data-hovercard-url="https://github.com/websockets/ws/commit/bc8bd34e4125a5edc37211fb384c17a4703e9000/hovercard" href="https://snyk.io/redirect/github/websockets/ws/commit/bc8bd34e4125a5edc37211fb384c17a4703e9000"><tt>bc8bd34</tt></a> to the 7.x release line (<a class="commit-link" data-hovercard-type="commit" data-hovercard-url="https://github.com/websockets/ws/commit/0435e6e12b8d38992cf0651cb8605dde2294bd25/hovercard" href="https://snyk.io/redirect/github/websockets/ws/commit/0435e6e12b8d38992cf0651cb8605dde2294bd25"><tt>0435e6e</tt></a>).</li>
</ul>
      </li>
      <li>
        <b>7.5.8</b> - <a href="https://snyk.io/redirect/github/websockets/ws/releases/tag/7.5.8">2022-05-26</a></br><h1>Bug fixes</h1>
<ul>
[...]

PR #2221, opened 2022-12-15
[Snyk] Upgrade ioredis from 4.27.9 to 4.28.5

ciyer:
<p>This PR was automatically created by Snyk using the credentials of a real user.</p><br /><h3>Snyk has created this PR to upgrade ioredis from 4.27.9 to 4.28.5.</h3>

:information_source: Keep your dependencies up-to-date. This makes it easier to fix existing vulnerabilities and to more quickly identify and fix newly disclosed vulnerabilities when they affect your project.
<hr/>

- The recommended version is **8 versions** ahead of your current version.
- The recommended version was released **10 months ago**, on 2022-02-06.

<details>
<summary><b>Release notes</b></summary>
<br/>
  <details>
    <summary>Package name: <b>ioredis</b></summary>
    <ul>
      <li>
        <b>4.28.5</b> - <a href="https://snyk.io/redirect/github/luin/ioredis/releases/tag/v4.28.5">2022-02-06</a></br><h2><a href="https://snyk.io/redirect/github/luin/ioredis/compare/v4.28.4...v4.28.5">4.28.5</a> (2022-02-06)</h2>
<h3>Bug Fixes</h3>
<ul>
<li>Reset loaded script hashes to force a reload of scripts after reconnect of redis (<a href="https://snyk.io/redirect/github/luin/ioredis/issues/1497" data-hovercard-type="pull_request" data-hovercard-url="/luin/ioredis/pull/1497/hovercard">#1497</a>) (<a href="https://snyk.io/redirect/github/luin/ioredis/commit/f357a3145694250bf361827403252b31435afabd">f357a31</a>)</li>
</ul>
      </li>
      <li>
        <b>4.28.4</b> - <a href="https://snyk.io/redirect/github/luin/ioredis/releases/tag/v4.28.4">2022-02-02</a></br><h2><a href="https://snyk.io/redirect/github/luin/ioredis/compare/v4.28.3...v4.28.4">4.28.4</a> (2022-02-02)</h2>
<h3>Bug Fixes</h3>
<ul>
<li>make sure timers is cleared on exit (<a href="https://snyk.io/redirect/github/luin/ioredis/issues/1502" data-hovercard-type="pull_request" data-hovercard-url="/luin/ioredis/pull/1502/hovercard">#1502</a>) (<a href="https://snyk.io/redirect/github/luin/ioredis/commit/cfb04a062b380bad5655b6f97b4259f328f1ee4a">cfb04a0</a>)</li>
</ul>
      </li>
      <li>
[...]

PR #2220, opened 2022-12-15
[Snyk] Upgrade winston from 3.3.3 to 3.8.1

ciyer:
<p>This PR was automatically created by Snyk using the credentials of a real user.</p><br /><h3>Snyk has created this PR to upgrade winston from 3.3.3 to 3.8.1.</h3>

:information_source: Keep your dependencies up-to-date. This makes it easier to fix existing vulnerabilities and to more quickly identify and fix newly disclosed vulnerabilities when they affect your project.
<hr/>

- The recommended version is **9 versions** ahead of your current version.
- The recommended version was released **5 months ago**, on 2022-06-30.

<details>
<summary><b>Release notes</b></summary>
<br/>
  <details>
    <summary>Package name: <b>winston</b></summary>
    <ul>
      <li>
        <b>3.8.1</b> - <a href="https://snyk.io/redirect/github/winstonjs/winston/releases/tag/v3.8.1">2022-06-30</a></br><h3>Patch-level changes</h3>
<ul>
<li>Update types to match in-code definitions in <a class="issue-link js-issue-link" data-error-text="Failed to load title" data-id="1286718049" data-permission-text="Title is private" data-url="https://github.com/winstonjs/winston/issues/2157" data-hovercard-type="pull_request" data-hovercard-url="/winstonjs/winston/pull/2157/hovercard" href="https://snyk.io/redirect/github/winstonjs/winston/pull/2157">#2157</a>; thanks to new contributor <a class="user-mention notranslate" data-hovercard-type="user" data-hovercard-url="/users/flappyBug/hovercard" data-octo-click="hovercard-link-click" data-octo-dimensions="link_type:self" href="https://snyk.io/redirect/github/flappyBug">@ flappyBug</a></li>
</ul>
<h3>Dependency updates by <a class="user-mention notranslate" data-hovercard-type="organization" data-hovercard-url="/orgs/dependabot/hovercard" data-octo-click="hovercard-link-click" data-octo-dimensions="link_type:self" href="https://snyk.io/redirect/github/dependabot">@ dependabot</a> + CI autotesting</h3>
<ul>
[...]

PR #2219, opened 2022-12-15
[Snyk] Upgrade @sentry/node from 6.16.1 to 6.19.7

ciyer:
<p>This PR was automatically created by Snyk using the credentials of a real user.</p><br /><h3>Snyk has created this PR to upgrade @sentry/node from 6.16.1 to 6.19.7.</h3>

:information_source: Keep your dependencies up-to-date. This makes it easier to fix existing vulnerabilities and to more quickly identify and fix newly disclosed vulnerabilities when they affect your project.
<hr/>

- The recommended version is **24 versions** ahead of your current version.
- The recommended version was released **8 months ago**, on 2022-04-26.

<details>
<summary><b>Release notes</b></summary>
<br/>
  <details>
    <summary>Package name: <b>@sentry/node</b></summary>
    <ul>
      <li>
        <b>6.19.7</b> - 2022-04-26
      </li>
      <li>
        <b>6.19.6</b> - 2022-04-07
      </li>
      <li>
        <b>6.19.5</b> - 2022-04-07
      </li>
      <li>
        <b>6.19.4</b> - 2022-04-05
      </li>
      <li>
        <b>6.19.3</b> - 2022-03-30
      </li>
      <li>
        <b>6.19.2</b> - 2022-03-23
      </li>
      <li>
        <b>6.19.1</b> - 2022-03-21
      </li>
      <li>
        <b>6.19.0</b> - 2022-03-21
      </li>
      <li>
        <b>6.18.2</b> - 2022-03-08
      </li>
      <li>
        <b>6.18.1</b> - 2022-02-28
      </li>
      <li>
        <b>6.18.1-beta.0</b> - 2022-03-01
      </li>
      <li>
        <b>6.18.0</b> - 2022-02-24
      </li>
      <li>
        <b>6.17.9</b> - 2022-02-16
      </li>
      <li>
        <b>6.17.8</b> - 2022-02-15
      </li>
      <li>
        <b>6.17.8-beta.0</b> - 2022-02-11
      </li>
      <li>
        <b>6.17.7</b> - 2022-02-11
      </li>
      <li>
        <b>6.17.6</b> - 2022-02-08
      </li>
      <li>
        <b>6.17.5</b> - 2022-02-07
      </li>
      <li>
        <b>6.17.4</b> - 2022-02-02
      </li>
      <li>
        <b>6.17.3</b> - 2022-01-28
      </li>
      <li>
        <b>6.17.2</b> - 2022-01-25
      </li>
      <li>
[...]

PR #2226, opened 2022-12-16
build(server): upgrade packages

lorenzo-cavazzi:
This includes running `npm audit fix` and the following upgrades from snyk: #2218 #2219 #2220 #2221 #2222

/deploy #persist #cypress


ciyer (approved):
👍

PR #2229, opened 2022-12-16
fix: load all inactive projects in kg

andre-code:
PR to test changes in kg-search repo. v2


lorenzo-cavazzi (changes requested):
The changes work well and the code looks good! :rocket: 

Since this could be the last PR for the kg-search (:crossed_fingers: ), I have 2 minor details to adjust:
* Users can now sort by "Relevance Asc" and "Relevance Desc". I suggest leaving only "Relevance" and sorting by highest relevance. I find "asc" and "desc" confusing when speaking about relevance (also, searching for the resources that are as different as possible from my search query sounds unusual 😁)
* When there is no match, we show a text and then an inline action button. I  suggest modifying it a bit as follows
  Screenshot_20221216_180408
  In this way, we will:
    * include the button inline in the sentence as we do in other places (here an example)
      Screenshot_20221216_180340
    * Move the sad face to a separate line to prevent misaligning the code on smaller windows

lorenzo-cavazzi (approved):
Lgtm!

PR #2231, opened 2022-12-16
feat: restore projects and datasets images (#2179)

### LINKED ISSUE #2179: Restore projects and datasets images
## Motivation

We used to show images in projects and datasets -- both in the listings and in the descriptions.

image

After the last design iteration, we removed them but the users want them back.
We now have a new design for that: we should re-add the images back.

## Mock-ups / Design

Figma files: 

* Project and Dataset Cards for listings -- including new position for visibility and type icons
    * [Figma](https://www.figma.com/file/abXsWrYhxeJ0GlkS7ypg7j/Project-and-Dataset-Cards?node-id=43%3A76&t=AJI3lGbO0Fd8CyJd-1)
    * [PDF](https://github.com/SwissDataScienceCenter/renku-ui/files/10152157/Project.and.Dataset.Cards.pdf)
* Project details with changes in the headers
    * [Figma](https://www.figma.com/file/AyZh0mKF2CY5nA48CuleRr/Project-Page-(Header-prioritization)-FINAL?node-id=25%3A447&t=DvCVd0l1FJAgrmr1-1)
    * [PDF](https://github.com/SwissDataScienceCenter/renku-ui/files/10152150/Project.Page.Header.prioritization._FINAL.pdf)

andre-code:
PR to add new entity header styles and restore entity images.

Screenshot 2022-12-19 at 16 17 49

Screenshot 2022-12-19 at 16 18 06

Screenshot 2022-12-19 at 16 18 23

Screenshot 2022-12-19 at 16 18 50

Screenshot 2022-12-19 at 16 19 13
0349b088.png)

Screenshot 2022-12-19 at 16 23 10

dataset-project images

Closes #2179 

/deploy #persist #cypress

### DIFF
--- a/client/src/dataset/Dataset.present.js
+++ b/client/src/dataset/Dataset.present.js
@@ -179,12 +179,6 @@ function DisplayMetadata({ dataset, sameAs, insideProject }) {
     <CardBody className="p-4 pt-3 pb-3 lh-lg pb-2">
       {
         <div className="d-flex">
-          {dataset.mediaContent ?
-            <div className="flex-shrink-0 pe-3" style={{ width: "120px" }}>
-              <img src={dataset.mediaContent} className=" rounded" alt=""
-                style={{ objectFit: "cover", width: "100%", height: "90px" }} />
-            </div>
-            : null}
           <div className="flex-grow-1">
             <DisplayInfoTable
               dataset={dataset}
@@ -289,7 +283,7 @@ function AddToProjectButton({ insideKg, locked, logged, identifier }) {
 function EditDatasetButton({ dataset, insideProject, locked, maintainer }) {
   if (!insideProject || !maintainer) return null;
   if (locked) {
-    return <span className="float-right mb-1 me-1" id="editDatasetTooltip">
+    return <span className="float-right mb-1" id="editDatasetTooltip">
       <Button className="btn-rk-white text-rk-pink icon-button"
         data-cy="edit-dataset-button" disabled={true} size="sm">
         <FontAwesomeIcon icon={faPen} color="dark" />
@@ -299,7 +293,7 @@ function EditDatasetButton({ dataset, insideProject, locked, maintainer }) {
         tooltip="Cannot edit dataset until project modification finishes." />
     </span>;
   }
-  return <Link className="float-right mb-1 me-1" id="editDatasetTooltip"
+  return <Link className="float-right mb-1" id="editDatasetTooltip"
     data-cy="edit-dataset-button"
     to={{ pathname: "modify", state: { dataset: dataset } }} >
     <Button className="btn-rk-white text-rk-pink icon-button" size="sm"
@@ -321,7 +315,7 @@ function getLinksDatasetHeader(projects) {
   };
   if (linksHeader.status !== "is_updating" && projects?.length > 0) {
     linksHeader.total = projects.length;
-    projects.slice(0, 5).map( project => {
+    projects.slice(0, 3).map( project => {
       linksHeader.data.push({
         title: project.name,
         url: `/projects/${project.path}`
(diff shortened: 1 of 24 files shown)


client/src/dataset/Dataset.present.js
```diff
@@ -380,6 +374,10 @@ export default function DatasetView(props) {
     new Date(datasetDate.replace(/ /g, "T")) :
     "";
 
+  const imageUrl = dataset.mediaContent ? dataset.mediaContent :
+    dataset.images?.length > 0 ?
+      dataset.images[0]["_links"][0].href : undefined;
+
```
ciyer, line 380: I see this code in at least one other place. I think it would be good to extract a shared utility function.

client/src/dataset/list/DatasetList.present.js
```diff
@@ -200,6 +200,7 @@ class DatasetsRows extends Component {
         labelCaption: projectsCount + ". Created",
         creators: dataset.published !== undefined && dataset.published.creator !== undefined ?
           dataset.published.creator : null,
+        imageUrl: dataset.images?.length > 0 ? dataset.images[0]["_links"][0].href : undefined,
```
ciyer, line 203: See comment about extracting this to a function.

client/src/dataset/Dataset.present.js
```diff
@@ -380,6 +374,10 @@ export default function DatasetView(props) {
     new Date(datasetDate.replace(/ /g, "T")) :
     "";
 
+  const imageUrl = dataset.mediaContent ? dataset.mediaContent :
+    dataset.images?.length > 0 ?
+      dataset.images[0]["_links"][0].href : undefined;
```
lorenzo-cavazzi, line 379: If `dataset.images` is an array but contains no links, the application will crash.
Should we do a more conservative check here and also verify that `_links` exists, is an array, and contains elements?

client/src/dataset/list/DatasetList.present.js
lorenzo-cavazzi, line 203: Same comment on testing `images[0]["_links"][0].href` before using it

ciyer (changes requested):
This looks really, really great! 🎉  I like the code and the use of grid layout. I just have a small comment about extracting some shared code to a function.

lorenzo-cavazzi (changes requested):
This looks great!
I have an inline comment, but I noticed it is almost the same comment you already got from @ciyer 

Two other details:

* The project warning icon doesn't show the cursor for links (see screenshot). It also looks vertically misaligned, could you check that? Maybe it's because the text is in a `div` element instead of a text element.
  Screenshot_20221220_091134

* On Firefox, the start button wrap word
  Screenshot_20221220_090326

andre-code:
@ciyer @lorenzo-cavazzi thanks for your review.
I have added a new function to extract the image url from the dataset and some unit tests for that function.

I have also included the cursor pointer for the project status indicator, as for the start session button there is another PR #2243  to implement the new styles for that button and there that problem does not exist. So in this case we are not going to include a solution for that.

client/src/dataset/DatasetFunctions.js
```diff
@@ -45,4 +45,19 @@ function getDatasetAuthors(dataset) {
...
+ * @param {Object[]} images - Images link arrays
+ */
+function getDatasetImageUrl(images) {
+  try {
+    return images?.length > 0 ?
+      images[0]["_links"][0].href : undefined;
+  }
+  catch {
+    return undefined;
+  }
```
lorenzo-cavazzi, line 60:
```suggestion
  try {
    return images[0]["_links"][0].href;
  }
  catch {
    return undefined;
  }
```
Testing for the length of the first array seems unnecessary since we rely on `try/catch` for all the other unexpected cases (and the `getDatasetImageUrl` function is always invoked after a `images?.length > 0 ?` in the code).

lorenzo-cavazzi (changes requested):
Great, the changes fix the icon!
I suggested a tiny code change. Otherwise, ready to go from my side :rocket:

lorenzo-cavazzi (approved):
Lgtm! :rocket:

ciyer (approved):
Looks great -- both the new design and the code!

PR #2246, opened 2022-12-20
feat: add image editor in dataset form (#2180)

### LINKED ISSUE #2180: Add a simple image editor when adding datasets avatar
## Motivation

Users can add an image to datasets at creation time, or change it later by editing the dataset.
Since having proportions too far from the shape we use in listings and details pages would lead to poor visualization, we should implement a very simple editor to crop/resize on-the-fly (check [this issue](https://github.com/SwissDataScienceCenter/renku-ui/issues/2059) for suggestions -- [react-avatar-editor](https://github.com/mosch/react-avatar-editor)).

Since we have a limitation on the image size, the library should handle optimizing and scaling down the image when it's too big.

_ℹ️ Wait for #2179 to be completed before starting the implementation_

## Mock-ups / Design

Figma files: 

* Uploader
    * [Figma](https://www.figma.com/file/sJS1EDIsYWe7MtFQ4G3eeM/I-want-my-project-images-back_Final?node-id=164%3A749&t=xFo0RQPZ3cews97X-1)
    * [PDF](https://github.com/SwissDataScienceCenter/renku-ui/files/10151995/I.want.my.project.images.back_Final.pdf)

andre-code:
PR to add image editor.

editor

Note: Also included the correction of the title in the project header when an image already exists.

Closes #2180 


client/src/model/RenkuModels.js
```diff
@@ -618,9 +618,9 @@ const datasetFormSchema = new Schema({
     label: "Image",
     edit: true,
     type: FormGenerator.FieldTypes.IMAGE,
-    maxSize: 200 * 1024,
+    maxSize: 10000 * 1024,
```
ciyer, line 621: This seems like a *very* high size limit to me. I can understand wanting to permit images larger than 200KB, but 10MB seems really big. I think 2MB would be a good _big_ image size.

ciyer (changes requested):
Very nice implementation. The edit feature looks good and works well!

I have a few minor things that could be improved.

* It was not immediately obvious that I could reposition the center of the image. I think either adding some text to say that you can reposition the image or some controls (arrows) to move the center would make this clearer

* It would be nice to have a "Reset" button to go back to how the image looked in the beginning.

* The image "moves" when switching to "Edit," even though I did not reposition the image

image

image

client/src/model/RenkuModels.js
lorenzo-cavazzi: This is just for selecting the image, the uploaded one should be resized/downgraded to <200kb.
We can allow for bigger images here. Maybe users wish to take just a tiny part of a bigger image (E.G. I often take a full-screen screenshot and use just a tiny part from it -- that would be bigger than 2MB)

ciyer (changes requested):
I like the new changes, I think this works very well!

I found one small problem, which is that the old images seem to be cached in the dataset. Try the following:
- create a dataset (with an avatar image)
- change the avatar

Then the UI will still show the old image. If you *refresh* the page, the new image will appear, but not until then.

andre-code:
> I like the new changes, I think this works very well!
> 
> I found one small problem, which is that the old images seem to be cached in the dataset. Try the following:
> 
> * create a dataset (with an avatar image)
> * change the avatar
> 
> Then the UI will still show the old image. If you _refresh_ the page, the new image will appear, but not until then.

Thanks for reporting this, I have added a workaround to display the latest image available. 
FYI, Gitlab returns the same url after editing so it was necessary to add an extra value in the url to force the browser to fetch the latest image.

ciyer (approved):
👍

PR #2261, opened 2022-12-23
Bundle together multiple dependency updates

ciyer:
Upgrades to
* bootstrap
* reactstrap
* dompurify

/deploy #persist #cypress renku=000-cypres-tests-new-search

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -70,7 +70,7 @@
         "rehype-stringify": "^9.0.3",
         "remark-math": "^5.1.1",
         "remark-rehype": "^10.1.0",
-        "sass": "^1.51.0",
+        "sass": "^1.56.1",
         "showdown": "^1.9.1",
         "showdown-highlight": "^2.1.8",
         "showdown-katex": "0.6.0",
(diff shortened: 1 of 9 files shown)


ciyer:
The navbar elements are too dark in color. 

image

ciyer:
Navbar elements have been corrected, but project header does not have the right padding between elements.

image

andre-code (changes requested):
Overall the application looks good, I only found a change in the font color for the buttons (primary and secondary) in active state.


Screenshot 2023-01-06 at 11 22 12
Screenshot 2023-01-06 at 11 21 24
Screenshot 2023-01-06 at 11 20 25


Also I just noticed that on the resource page the background color changed so the text is now no longer readable, but it's like that in master, maybe you can fix it here.
Screenshot 2023-01-06 at 11 19 57

lorenzo-cavazzi (approved):
Lgtm! :tada:

andre-code (approved):
The styles were corrected and it works as expected. 🚀

lorenzo-cavazzi (approved):

PR #2263, opened 2023-01-03
Revert  2c8d5f4

ciyer:
The update to storybook breaks telepresence.

/deploy #cypress #persist


ciyer:
See also #2264

lorenzo-cavazzi (approved):
Tried with telepresence, all fine after the rollback :+1:

PR #2265, opened 2023-01-03
fix(client): improve headers layout

lorenzo-cavazzi:
This PR addresses a few layout details in the new headers.

* Restore the missing title in the default avatar on the projects and datasets detail pages.
* Add minimal margin to entity tags to prevent collapsing into the description (in the image, the previous version with the collapsed content)
    Screenshot_20230103_144933
* Remove extra space on the detail pages when tags are missing. This happened only on small and medium window sizes (in the image, the previous version with the extra space on the bottom)
    Screenshot_20230103_145107

/deploy renku=renku-ui-2.15.0-tests #persist #cypress

### DIFF
--- a/client/src/dataset/Dataset.present.js
+++ b/client/src/dataset/Dataset.present.js
@@ -389,18 +389,19 @@ export default function DatasetView(props) {
       }
       <div className="mb-4">
         <EntityHeader
-          title={dataset.title}
-          description=""
-          itemType="dataset"
-          tagList={dataset.keywords}
           creators={dataset?.published?.creator}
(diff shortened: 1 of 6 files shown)


andre-code (dismissed):
It works as expected and the code looks good🚀.

lorenzo-cavazzi:
Tests failed because the default avatar's text overflowed the parent container (see image).
I fixed it and added the same fix to the entities lists -- it wasn't creating issues there cause it was overflowing only vertically, but it's better to keep it clean.

**Before**
Screenshot_20230104_085452

**After**
Screenshot_20230104_085806

ciyer (approved):
Code and appearance look good!

PR #1831, opened 2022-05-10
Fix CSS and other styling problems post 1823

ciyer:
Correct the CSS generation and fix minor styling problems identified in #1823 

# Screenshots

image

image

image

/deploy renku=1811-main-menu-changes #persist

PR #2268, opened 2023-01-03
feat: show session button as connect when there is a session running (#1831)

### LINKED ISSUE #1381: Session start/connect button
# Description

As a follow-up to #1152, the "Start" button should become a "Connect" button is a session is already running.

This should be easy to implement after #2016.

Expected result:
- if any session is running, then change the label to "connect"
- clicking connect should connect to the running server
- If there is more than one server, then show a list of running sessions and let the user pick

andre-code:
PR to add connect button when there is a session running.

Screenshot 2023-01-03 at 17 45 17

Screenshot 2023-01-10 at 16 45 46

connect

Closes #1381 


client/src/project/Project.present.js
```diff
@@ -474,11 +475,29 @@ class ProjectViewHeaderOverview extends Component {
...
+      const session = currentSessions[sessionName];
+      const annotations = NotebooksHelper.cleanAnnotations(session.annotations, "renku.io");
+      const autoStartUrl = Url.get(Url.pages.project.session.autostart, {
+        namespace: annotations["namespace"],
+        path: annotations["projectName"],
+      });
+      if (autoStartUrl === sessionAutostartUrl)
+        isRunningSession = true;
+    });
+  }
```
ciyer, line 491: I think it would be nice to extract this to a function
```
function isSessionRunnning(currentSessions)
...
```

ciyer:
See this movie for context about my comments.

https://user-images.githubusercontent.com/1196411/210526212-dbb0c20f-1f15-4fa0-bc9f-c410d5b58e8f.mov

client/src/project/Project.present.js
andre-code: Sure, I refactored this part to also include the suggestion to open the session instead of trying to initiate one

ciyer (dismissed):
Thanks for making the improvements! I have approved. If you want to resize the "connect" icon, I can re-approve.

ciyer (approved):
🎉

PR #2269, opened 2023-01-04
fix: handle error sidecar API

### LINKED ISSUE #2169: The UI crashes when invoking any sidecar API fails
## Description

When a sidecar's API fails (any command invoked on a full-page session, like "Pull" os "Save") the UI crashes.
Here is an example of the response we get from ``

```
{
   "error":{
      "code":-32000,
      "message":"Running a git command failed with the error: fatal: could not read Username for 'https://gitlab.dev.renku.ch': No such device or address\n"
   },
   "id":0,
   "jsonrpc":"2.0"
}
```

## Reproduce

To simulate the "real" error, start a session and... wait for 24h :grin: 
Otherwise, the sidecar API `dummy/get_error` will help a lot to simulate errors (the structure is always similar)

andre-code:
PR to fix error sidecar.

Closes #2169 


ciyer (dismissed):
Nice fix! I added a commit with a cypress test for this case.

ciyer (dismissed):

ciyer (approved):
👍

PR #2275, opened 2023-01-05
docs: developer documentation

### LINKED ISSUE #2165: Start documenting UI best practices
We should document UI best practices:
- RTK (point to good examples and docs)
- Testing strategy (include a couple of examples)
- How to develop/run tests (refresh telepresence scripts if necessary)

ciyer:
Developer documentation

Fix #2165

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,19 @@ For the default mode of the script, you will need to attach a debugger to finish
 
 With this configuration, after running `run-telepresence.sh` you should invoke the _Run > Start Debugging_ in VS Code to attach the debugger.
 
+# Troubleshooting
+
+## Telepresence
+
+Telepresence 2 has greatly improved robustness to network problems and is much more reliable than the earlier version. Still, sometimes situations arise that cause problems for telepresence. When that happens, the best course of action is to quit and restart telepresence:
+
+```
+$ telepresence quit -ur
+$ telepresence connect
+```
+
+If that is not sufficient, you may need use the activity monitor to find and kill zombie telepresence processes before restarting.
+
 # Site navigation map
 Note: stroke-dasharray when the link is only for anonymous users
 
--- a/README.md
+++ b/README.md
@@ -23,16 +23,31 @@ Below, we explain each of those components in greater detail and guidelines for 
 
 To develop on this codebase, you will want the following tools installed.
 
+|                                                                      |                                                                     |
+|----------------------------------------------------------------------|---------------------------------------------------------------------|
+| [node](https://nodejs.org/en/)                                       | Node JavaScript execution environment                               |
+| [npm](https://github.com/npm/cli)                                    | Node package manager                                                |
+| [nvm](https://github.com/nvm-sh/nvm)                                 | NVM or some similar tool for managing node versions                 |
+
+Node is a requirement; when you install node, you will probably also get npm automatically. Though not absolutely necessary, we recommend using a **node version manager** like nvm to manage node/npm installations.
+
+However you get it, you need the node version specified in the [Dockerfile](https://github.com/SwissDataScienceCenter/renku-ui/blob/master/client/Dockerfile).
+
+## Recommended tools
+
+If you are content with testing your changes using Cypress, all you need is node. If you want to develop and test your changes against a real RenkuLab deployment, then you will need a few more tools.
+
+|                                                                      |                                                                     |
+|----------------------------------------------------------------------|---------------------------------------------------------------------|
+| [kubectl](https://kubernetes.io/docs/tasks/tools/)                   | K8s command-line tool                                               |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,77 +2,237 @@
...
+
+### **Group shared code into features**
+
+[Features](https://redux.js.org/faq/code-structure) should be used to group together UI components and state manipulation code.
+
+It commonly happens that there are multiple components and multiple state interactions that center around one shared slice and/or one backend API. These should be grouped into a feature to make this structure easier to work with.
+
+# Server
+
+The server is the [Express-based](https://expressjs.com) back-end for the RenkuLab UI. At first, the front-end would interact directly with back-end services, but this structure imposed some limitations and complications, so we introduced a server component with the sole responsibility of serving the UI and simplifying interactions with the other backend services.
```
lorenzo-cavazzi, line 159: I would remove the historical reasons for creating the server, and mention that it also handles requesting, storing, and renewing user tokens because it's safer than doing it in the client and it allows us to poll some resources on behalf of the user

README.md
```diff
@@ -2,77 +2,237 @@
...
+The server folder includes a `run-telepresence.sh` script that is tailored for the SDSC development cluster.
+
+    $ cd server
+    $ ./run-telepresence.sh
+
+Telepresence replaces the UI server pod in the target Kubernetes instance. All the traffic is then redirected to a local process, making changes to files almost immediately available in your development RenkuLab instance.
+
+The instructions for deploying the renku application into K8s are the same as in the client, so see that section for those details. There are some small differences in how the server `run-telepresence.sh` script works compared to the client.
+
+* In the server, `run-telepresence.sh` will prompt you to provide or override the deployment you want to intercept. You can enter the id of a PR to intercept the deployment associated with a GitHub PR.
```
lorenzo-cavazzi, line 211: I'm tempted to adjust the 2 telepresence scripts to unify this behavior.
If it's ok with you, once the PR is ready I would add a commit to align it (and update this section accordingly) .

lorenzo-cavazzi (changes requested):
This documentation is great! Sections are on point, easy to follow, and links to tutorials are helpful for new people.

I would modify the prerequisites because client-side development with Cypress would require only npm.
I also added a comment to the server section to mention user tokens -- you said it might require more extensive changes, but it seems already good to me!

README.md
ciyer: In general, I'm ok with that change, but I think it would be nicer if the prompting only happened if no values were provided. That is, if there is no DEV_NAMESPACE or PR environment variable set, then prompt for it, but one of them is set, do not prompt and just use what was provided.

andre-code (commented):
Great documentation 👏 I find the added information very clear and useful.

In addition, have you considered including a Support/Troubleshooting section for common errors for example when we are working with telepresence and have to deal with dropping connections.
Should the Administrator's Guide work for all external contributions or do they need additional permissions?

README.md
lorenzo-cavazzi: Good suggestion! I will go that way :+1:

README.md
```diff
@@ -2,77 +2,237 @@
...
+
+|                                                                      |                                                                     |
+|----------------------------------------------------------------------|---------------------------------------------------------------------|
+| [docker](https://www.docker.com)                                     | For building containers                                             |
+| [helm](https://helm.sh)                                              | For packaging things for K8s                                        |
+| [kubectl](https://kubernetes.io/docs/tasks/tools/)                   | K8s command-line tool                                               |
+| [nvm](https://github.com/nvm-sh/nvm)                                 | NVM or some similar tool for managing node environments             |
+| [telepresence](https://www.telepresence.io/docs/latest/quick-start/) | Tool for redirecting requests to your local development environment |
+
+You should use your **node version manager** to install the node version specified in the [Dockerfile](https://github.com/SwissDataScienceCenter/renku-ui/blob/master/client/Dockerfile).
```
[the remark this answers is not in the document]
ciyer: Good suggestion, I have split up the prerequisites into three sections -- always needed, needed for telepresence, and needed to deploy to K8s.

README.md
ciyer: Good idea. I have made these changes.

ciyer:
> In addition, have you considered including a Support/Troubleshooting section for common errors for example when we are working with telepresence and have to deal with dropping connections.

Good point! I added a section about fixing telepresence problems. Is there anything else that you think should be in there?

> Should the Administrator's Guide work for all external contributions or do they need additional permissions?

My understanding is that if you have access to a K8s cluster and have sufficient permissions, you should be able to deploy the renku app using the instructions there. External contributors will need to organize their own K8s cluster, though, since they will not have access to ours.

lorenzo-cavazzi (approved):
Lgtm! :rocket: 

I will open another PR later to unify the behavior for telepresence

PR #2281, opened 2023-01-09
test(e2e): re-arrange global tests

lorenzo-cavazzi:
This PR introduces minor changes to the Cypress e2e tests:
* move global tests to a separate file
* add fake endpoint and response for Statuspage
* prevent opening localhost:3000 in a browser when starting e2e tests

Since some users observed a "RenkuLab down" page flashing in the past few days, we should add tests to handle failures from the Statuspage endpoint API as a follow-up, and adapt the component accordingly.
That _might_ be the cause of the problems.

/deploy renku=auto-update/renku-ui-3.0.0 #persist #cypress #notest


ciyer (approved):
This should be useful for fixing some problems with the initial load.

andre-code (approved):
lgtm 🚀

PR #2282, opened 2023-01-09
fix: Remove the delay when starting a session (#2279)

### LINKED ISSUE #2279: UI-UX (sessions) Delay to move to page from step 1 of 2 to step 2 of 2 after all steps have been completed
## Description

Step 1 of 2 when starting a session has a delay to move to the next page after all steps have been completed.

## Reproduce
step 1 to 2

andre-code:
PR to fix the delay when starting a session.

Closes #2279 


ciyer (approved):
Works as advertised!

PR #2287, opened 2023-01-11
Validate object store inputs

### LINKED ISSUE #2210: Validate Endpoint field in the Object Store Configuration (aka S3)
When adding a "Cloude Storage" in the "start session with options" page, we don't validate inputs. We should add a basic validation for some of them so we don't need to wait until clicking on "Start session" to get the error:
* [x] `Endpoint`: assess the string is a is well-formatted URL.
* [x] `Bucket name`: check the length is at least 3 chars and name consists only of lowercase letters, numbers, dots (.), and hyphens (-); also, the name must start with a letter or a number.
* [x] Change the title from `Object Store Configuration` to `Cloud Sotrage configuration` and verify we consistently use "Cloud Storage" instead of other definitions

Image

P.S. Here is the reference for the bucket names: https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html
There are more restrictions than the ones listed above, but different implementations might implement slightly different variations and we don't need to be 100% precise here.

ciyer:
# Screenshots

image

image

image

image

# Testing

* The endpoint must be a valid URL
* The bucket name must be at least 3 characters, contain only lowercase characters, numbers, '.', or '-', and must start and end with a letter or number

Fix #2210

/deploy #persist #cypress renku=000-cypres-tests-new-search

### DIFF
--- a/client/src/notebooks/ObjectStoresConfig.present.js
+++ b/client/src/notebooks/ObjectStoresConfig.present.js
@@ -26,6 +26,9 @@ import { faTrash } from "@fortawesome/free-solid-svg-icons";
 import { Map, List } from "immutable";
 
 import { ExternalLink } from "../utils/components/ExternalLinks";
+import { isURL } from "../utils/helpers/HelperFunctions";
+
+import "./ObjectStoresConfig.scss";
 
 function S3ExplanationLink({ title }) {
   const theTitle = title ?? "S3-compatible storage";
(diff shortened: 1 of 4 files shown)


lorenzo-cavazzi (approved):
:+1: 
I fixed a typo and made the modal full-screen on smaller screens. Feel free to change it if you don't like it.

PR #2292, opened 2023-01-11
test(e2e): avoid loading the URL early

lorenzo-cavazzi:
This should prevent failures when waiting on APIs

/deploy renku=000-cypres-tests-new-search #persist #notest #cypress


andre-code (approved):
lgtm 🚀

PR #2293, opened 2023-01-11
test: eliminate console errors and warnings in unit tests (#2228)

### LINKED ISSUE #2228: Fix unit tests errors on the client
There are many errors on the console when running `npm test` oin the UI client.
They are all treated as non-fatal, but we should:
- [ ] verify that they _really_ are non-critical
- [ ] get rid of them, including unnecessary `console.log`

Image

andre-code:
PR to eliminate console errors and warning in unit tests.

Screenshot 2023-01-12 at 16 31 13

Close #2228 

/deploy renku=000-cypres-tests-new-search #cypress #persist

### DIFF
--- a/client/src/file/FilePreview.tsx
+++ b/client/src/file/FilePreview.tsx
@@ -20,6 +20,8 @@ import hljs from "highlight.js";
 import React from "react";
 import { CardBody } from "reactstrap";
 import { Document, Page } from "react-pdf/dist/esm/entry.webpack";
+import DOMPurify from "dompurify";
+
 import "react-pdf/dist/esm/Page/AnnotationLayer.css";
 
 import { FileNoPreview, StyledNotebook } from "./File.present";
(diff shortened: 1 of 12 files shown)


client/src/file/FilePreview.tsx
```diff
@@ -227,11 +227,29 @@ type CodePreviewProps = {
...
+      case "<":
+        return "&lt;";
+      case '"':
+        return "&quot;";
+      default:
+        return "&#039;";
+    }
+  });
+}
+
```
lorenzo-cavazzi, line 244: Should we consider using a sanitizing function from a library instead of writing our own? I see we are already using `DOMPurify` somewhere else.
Maybe `DOMPurify.sanitize` already does what we need

lorenzo-cavazzi (changes requested):
The errors are all gone :tada: 

I'm not sure about manually implementing a sanitizing function. That doesn't seem to work since I see extra code blocks elements . I added a suggestion inline.
In general, we might want to sanitize the pure text before it's wrapped by HTML elements.

Screenshot_20230113_133633

client/src/file/FilePreview.tsx
andre-code: thanks @lorenzo-cavazzi  for your review, unfortunately `DOMPurify.sanitize` does not escape the HTML which is what the unit test warning refers to.

I've removed the error from the displayed `<code>`, so now we don't have the warning and the code displays just fine as before.

lorenzo-cavazzi (approved):
I swapped the custom sanitize function with the one from `DOMPurify`.

If this looks ok to you. let's merge it. Otherwise, we can look for a different solution.

PR #2297, opened 2023-01-13
Revive storybook

### LINKED ISSUE #2264: Migrate Storybook to be compatible with CRA 5
We updated to create-react-app 5. We need to update storybook to be compatible with this version of CRA

  https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#cra5-upgrade

Make sure that telepresence works afterwards.

ciyer:
Changes that allow storybook to work with the react-scripts v5 environment. 

Problems remain with the following components

- FileUploaderInput
- ListBar
- ListCard
- LoginAlert
- SortingEntities

But the others should work.

Fix #2264

/deploy #persist #cypress renku=auto-update/renku-ui-3.0.0


ciyer:
Please check that after installing the dependencies, you can still use telepresence.

client/.storybook/preview.js
```diff
@@ -1,10 +1,33 @@
+import { enhancer as withReduxEnhancer } from 'addon-redux'
+
+
 import "../src/styles/index.scss";
+
+import AppContext from '../src/utils/context/appContext';
+import {createStoreWithEnhancers} from "../src/utils/helpers/EnhancedState.js";
+
+const store = createStoreWithEnhancers({}, [withReduxEnhancer]);
```
lorenzo-cavazzi, line 9: This is not used. Was it an attempt to inject the store? If that's the case, should we add a comment to clarify why it's not used?

lorenzo-cavazzi (changes requested):
Nice to have Storybook back and working!
Can we temporarily remove the broken components (and create a follow-up issue)? When I load one of them, I can't use Storybook anymore unless I move to another component and reload the whole page

image

client/.storybook/preview.js
ciyer: You are right, this is a little confusing! It is how the documentation says you should introduce the redux store https://storybook.js.org/addons/@dreamworld/addon-redux/, but I added an export so that the value does not go unused.

ciyer:
> Nice to have Storybook back and working! Can we temporarily remove the broken components (and create a follow-up issue)? When I load one of them, I can't use Storybook anymore unless I move to another component and reload the whole page
> 
> image

For info, you can just click the "x close" button there and then continue to use storybook, but I agree that disabling the broken storybooks is better, and I have added a commit that does this.

andre-code (dismissed):
Thanks @ciyer , happy to see this working again! 
I have tested telepresence and the storybook components that should work and those work perfect.
Like @lorenzo-cavazzi my only suggestion is to create the following issue for the broken components.

lorenzo-cavazzi (approved):
Lgtm :+1:

lorenzo-cavazzi (approved):

PR #2299, opened 2023-01-16
Apply correct visibility limitations on fork

### LINKED ISSUE #2296: Forking a project with restricted visibility lets me choose public visibility
## Description

There is a regression in the available visibilities when forking a project. In the screenshot, the project is `Internal` (see on the top-right, left to the start button) and the forking modal lets me select visibility `Public`. That triggers a non-fatal failure later.

Screenshot_20230113_095723

## Reproduce

The easiest way would be creating an internal (or private) project in the user's own namespace, and try to fork that.

## Possible fix

We should double-check that the function to filter visibilities checks first the target project, and then the namespace.

ciyer:
Fix #2296

/deploy #persist #cypress renku=auto-update/renku-ui-3.0.0


ciyer:
# Testing

Navigate to a project and fork it. Try out a few combinations of project visibility / namespace visibility (e.g., public project / private namespace, private project / internal namespace): the visibility options for the forked project should be limited to those allowed by more restrictive of project and namespace.

andre-code (approved):
It works as expected 🚀

lorenzo-cavazzi (approved):

PR #2314, opened 2023-01-18
Upgrade to react 18

### LINKED ISSUE #2164: Upgrade to react 18 and remove overrides
We should update to React 18 (the switch from 17 shouldn't include many breaking changes) and remove the unnecessary libraries from the `overrides` section in the package.json file

ciyer:
Upgrade to react 18

Fix #2164 


ciyer:
Same as #2326

PR #2318, opened 2023-01-19
fix(client): align elements in the datasets file browser

lorenzo-cavazzi:
Elements rendered on the file browser were not properly aligned. This PR fixes it

image

/deploy #persist #cypress


ciyer (dismissed):
Yes, it looks much better after these changes!

ciyer (approved):
👍

PR #2319, opened 2023-01-19
build(server): remove default autoscaling settings

lorenzo-cavazzi:
Disable UI server default autoscaling settings since it might collide with settings in some deployments (E.G. in the CI deployments where the reserved resources are _very_ low).

### DIFF
--- a/helm-chart/renku-ui/values.yaml
+++ b/helm-chart/renku-ui/values.yaml
@@ -224,8 +224,8 @@ server:
 
   ## Configure autoscaling
   # configure autoscale only where needed
-  # autoscaling:
-  #   enabled: true
+  autoscaling:
+    enabled: false
   #   minReplicas: 2
   #   maxReplicas: 5
   #   cpuUtilization: 70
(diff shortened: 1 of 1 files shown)


helm-chart/renku-ui/values.yaml
```diff
@@ -223,11 +223,12 @@ server:
...
+  # configure autoscale only where needed
   autoscaling:
-    enabled: true
-    minReplicas: 1
-    maxReplicas: 5
-    cpuUtilization: 95
+    enabled: false
+  #   minReplicas: 2
+  #   maxReplicas: 5
+  #   cpuUtilization: 70
```
seanrmurphy, line 231: Can we fix indenting/spacing here? In the rest of the file, the `#` relates to where in the hierarchy the keys should appear...here, it's a little confusing...

seanrmurphy (changes requested):

helm-chart/renku-ui/values.yaml
lorenzo-cavazzi: Sure! Let's keep it coherent to make it obvious where the properties belong :+1:

seanrmurphy (approved):

PR #2326, opened 2023-01-24
build: upgrade to react 18.2 (#2164) (#2314)

ciyer:
* adapt tests for react 18
* add README to explain overridden dependencies


lorenzo-cavazzi (approved):
Lgtm! The failure of the acceptance test is really strange. I would just try to re-run it

PR #2327, opened 2023-01-24
chore(client): add data-cy refresh-commits

lorenzo-cavazzi:
Changes for cypress tests. I'll add here a bunch of `data-cy` properties.


lorenzo-cavazzi:
Added to #2328

Issue #2181, opened 2022-12-01
Update image editor in projects


lorenzo-cavazzi:
## Motivation

We have an updated image editor for datasets ([see 2180](https://github.com/SwissDataScienceCenter/renku-ui/issues/2180)) that allows users to crop/resize images; we want to use that also for projects.
Currently, the backend is different because we use GitLab APIs to store the project's avatars instead of renku-core. This is not a big deal when editing the avatar _after_ the project creation.

We would like to add the image picker also on project creation, and that requires adding the component and invoking the API _after_ the project creation. We already do something similar for setting up the visibility and adjusting the name (when necessary). We should follow a similar pattern.

⚠️ users mentioned that it's unclear when the image will be cropped. We should also add a simple sentence mentioning "Image will be cropped"  when that's the case.

Adding drag&drop would also be nice. If it's not easy, make a follow-up issue.

## Mock-ups / Design

Figma files: see #2180. That issue should already implement/refresh the uploader component that can be reused here.

PR #2331, opened 2023-01-26
feat: allow creating a new project with an avatar (#2181)

ciyer:
# Testing

- [x] Create a project without an avatar
- [x] Create a project with an avatar
- [x] Upload a file and then upload a new one -- the second upload should be the avatar

# Screenshots

image

/deploy #persist #cypress renku=update-renku-ui-3.1.0


client/src/project/new/components/NewProjectAvatar.tsx
```diff
@@ -0,0 +1,81 @@
+/*!
+ * Copyright 2021 - Swiss Data Science Center (SDSC)
```
andre-code, line 2: Could you please update it to 2023.

andre-code (commented):
The avatar component works as expected and the code looks fine to me, I just found something to review.
 
 Could you take a look at why that comma is there and include the label about the image being cropped?

Screenshot 2023-01-30 at 10 30 05

The comma also appear in the new project form, and project setting form.
image

lorenzo-cavazzi (approved):
This works great! And the code looks good :+1: 
I noticed just one tiny detail. When updating an image on an already existing project, the avatar on the project detail section on the top doesn't change. Would it be easy to re-fetch it once the upload API returns a response?

_In the screenshot, notice the "uploaded" message on the bottom right and the new image, but the image on top is still the old one._

Screenshot_20230130_180252

andre-code (dismissed):
🚀

ciyer:
> Lgtm and tests passed 🎉
> 
> P.S. Ideally, I would keep the max size aligned between projects and datasets instead of having 3mb and 10mb. But I understand we still have the limitation from GitLab for projects, so we can live with this for now instead of optimizing the resizing function 👍
> 
> image

Yes, I think we can standardize the limit once we move to storing the avatar in the repo with renku-core.

PR #2307, opened 2023-01-17
feat: add entity list view styles (#2302)

andre-code:
PR to add list view styles in entity component

Closes #2302 

### How to test
- You can use storybook ListBar story
- Or can use the link https://renku-ci-ui-2307.dev.renku.ch/projects and https://renku-ci-ui-2307.dev.renku.ch/datasets

/deploy renku=000-connect-btn #persist #cypress

PR #2332, opened 2023-01-27
feat: add entity list view (#2302)(#2307)

### LINKED ISSUE #2302: Dashboard: Add entity component list view
Add entity component list view.
Note: Should be included in the storybook for testing.
Figma file: https://www.figma.com/file/vVxTg14Nho2A3DhvF5iT4k/Final-Renkulab-Dashboard?node-id=438%3A747&t=R5VHxckcw2OKWZqg-0

**Project entity:**

Screenshot 2023-01-17 at 10 40 39

Screenshot 2023-01-17 at 10 40 48

Screenshot 2023-01-17 at 10 40 59

**Dataset Entity:**

Screenshot 2023-01-17 at 10 41 12

Screenshot 2023-01-17 at 10 41 21

Screenshot 2023-01-17 at 10 41 30

Closes #2307

andre-code:
Feature branch for the Dashboard
Fix #2302

/deploy #cypress


client/src/features/dashboard/components/DatasetsDashboard.tsx
```diff
@@ -0,0 +1,137 @@
...
+            <img src="/frame-pink.svg" className="rk-icon rk-icon-md" />View all my Datasets</div>
+        </Button></div>
+    ) :
+    (<div className="d-flex justify-content-center">
+      <Button data-cy="explore-other-datasets-btn" className="btn btn-outline-rk-pink"
+        onClick={(e: React.MouseEvent<HTMLElement>) => handleOnClick(e, "all")}>
+        <div className="d-flex gap-2 text-rk-pink">
+          <img src="/explore-pink.svg" className="rk-icon rk-icon-md" />Explore other Datasets</div>
+      </Button></div>);
+}
```
lorenzo-cavazzi, line 61: Should we use a `Link` here instead of `onClick` and then `e.preventDefault();`?
That would improve the UX because it would add all the interactions typical for the Anchor elements (E.G. middle mouse click, or all the options to "Open in a new tab" when using the right click)

client/src/utils/components/list/ListBar.scss
```diff
@@ -0,0 +1,382 @@
...
+}
+
+.listBar-title {
+    overflow: hidden;
+    text-overflow: ellipsis;
+    display: inline-block;
+    white-space: nowrap;
+    margin: 0;
+    min-height: 30px;
+    height: 30px;
```
lorenzo-cavazzi, line 72: I would remove the height because it cuts out text when zooming in (see screenshot) and we show a single line anyway.
We can leave `min-height: 30px;` even though there will always be a non-empty title
image

client/src/utils/components/list/ListBar.tsx
```diff
@@ -16,64 +16,99 @@
...
 
-  return <Link className="d-flex flex-row rk-search-result" to={url}>
-    <div className="me-3 mt-2 d-flex flex-column align-items-center">
-      <div>
-        <span className={"circle " + itemType}> </span>
+  const history = useHistory();
+  const handleClick = useCallback((e: any) => {
+    e.preventDefault();
+    history.push(url);
+  }, [url]); //eslint-disable-line
```
lorenzo-cavazzi, line 64: Same comment as above: can we use a `Link` instead of manually handling the click event?
I think we do something similar in the cross-entity search Cards because I can right-click and select "Open in a new tab" there.

lorenzo-cavazzi (changes requested):
Code looks generally great, can't wait to see this feature merged :raised_hands: 

I have a request before merging. I noticed you sometimes use the following pattern: add an `onClick` event, prevent the default event, compose the URL, and redirect using `history.push()`. That works, but the browser doesn't show the typical interactions for links (e.g. middle mouse click to open in a new tab, or the right-click menu options like "open in a new tab"). Would it make sense to use an Anchor component like `Link`? The UX is better and it simplifies the logic. I think we already use that approach in dropdowns, buttons, or more complex elements like the cards on the entity search page.

ciyer (approved):
This looks great! Most of the code was already reviewed before, and I see that all of @lorenzo-cavazzi's suggestions have been incorporated. I think we can merge this! 🎉

Issue #2306, opened 2023-01-17
Dashboard: Include sessions in the project section


andre-code:
If there are sessions in progress, include them at the top of the project list.
Note:
- Include all existing sessions + 3 recently visited projects (do not duplicate projects)
- Include recently visited projects to complete at least 5 cards 

![Screenshot 2023-01-17 at 11 14 26](https://user-images.githubusercontent.com/43388408/212871344-d22d3786-0680-444d-a59c-8bd38dd132f6.png)

lorenzo-cavazzi:
Finished

PR #2333, opened 2023-01-27
feat: add sessions in dashboard (#2306)

### LINKED ISSUE #2305: Dashboard: Add the session component list view
Add session component list view.
Note 1: Should be included in the storybook for testing.
Figma file: https://www.figma.com/file/vVxTg14Nho2A3DhvF5iT4k/Final-Renkulab-Dashboard?node-id=438%3A747&t=R5VHxckcw2OKWZqg-0

Note 2: Saved and unsaved work icon is not included
Screenshot 2023-01-17 at 11 09 15

Screenshot 2023-01-17 at 11 10 11

andre-code:
PR to add session view.

Closes #2305 #2306

/deploy renku=000-ui-dashboard #persist #cypress


ciyer (changes requested):
This looks generally very good! 🎉 

I have a suggestion about. merging some CSS because I think there is a danger for things to get out of sync if we make changes and the SCSS isn't in one file.

Also, I noticed that the extra actions on a running session are wrong:

image

No need to show "Start with Options" (the session is already running), and we probably want to offer viewing logs or at least stopping the session.

One comment, I noticed that there is a lot of use of `any` as a type. Was this an oversight, or was it tedious to provide more specific types.

ciyer (approved):
This looks very good!

PR #2350, opened 2023-02-03
feat: add session actions in dashboard (#2305)

andre-code:
PR to add get logs, stop session and link to Dashboard.

dashboard sessions

### testing
- It should show all existing sessions
- If a session is in Error it should show the error icon, the error message and the main action to stop the session.
- If a session is starting or running, it should show "Connect" as the main action and the corresponding icon.
- If the session is being stopped, it should show "Stop" as the main action, but disabled.
- The user should be able to get Logs in all cases

/deploy renku=000-ui-dashboard #persist #cypress


ciyer (approved):
Code looks good and feature works as advertised! I added a commit to remove two `@ts-ignore` comments that are no longer necessary.

PR #2351, opened 2023-02-03
feat: store search parameters in URL (#2257)

### LINKED ISSUE #2257: Add back search parameters to the search page
## Description

In the previous search pages for projects and datasets, we used to add the filters and the search sentence in the parameters as soon as users selected them. This simplified sharing a specific search page and navigating to previous searches using the back buttons from the browser.
With the new unified search this went lost.

We should restore the behavior

ciyer:
Store search parameters in the URL

## Testing

- [x] All search parameters are correctly handled
- [x] Pasting a URL works in a new browser window
- [x] Pasting a URL works in a anonymous mode

Fix #2257

/deploy #persist #cypress


client/src/utils/components/dateFilter/DateFilter.tsx
```diff
@@ -33,21 +33,62 @@ export interface DateFilterProps {
...
+export function stringToDateFilter(str: string) {
+  switch (str) {
+    case "all": return DateFilterTypes.all;
+    case "custom": return DateFilterTypes.custom;
+    case "last90days": return DateFilterTypes.last90days;
+    case "lastMonth": return DateFilterTypes.lastMonth;
+    case "lastWeek": return DateFilterTypes.lastWeek;
+    case "older": return DateFilterTypes.older;
+    default: return undefined;
+  }
```
lorenzo-cavazzi, line 55: Is it necessary to repeat all the strings here? I'm wondering if we should use `Object.values` instead, something like the following:

```suggestion
  return Object.values(DateFilterTypes).includes(str as DateFilterTypes) ?
    str as DateFilterTypes :
    undefined;
```

client/src/utils/components/sortingEntities/SortingEntities.tsx
```diff
@@ -40,6 +40,18 @@ export enum SortingOptions {
...
+export function stringToSortingOption(str: string) {
+  switch (str) {
+    case "name:asc": return SortingOptions.AscTitle;
+    case "name:desc": return SortingOptions.DescTitle;
+    case "date:asc": return SortingOptions.AscDate;
+    case "date:desc": return SortingOptions.DescDate;
+    case "matchingScore:asc": return SortingOptions.AscMatchingScore;
+    case "matchingScore:desc": return SortingOptions.DescMatchingScore;
+    default: return undefined;
+  }
```
lorenzo-cavazzi, line 52:
```suggestion
  return Object.values(SortingOptions).includes(str as SortingOptions) ?
    str as SortingOptions :
    undefined;
```

lorenzo-cavazzi (changes requested):
That's a very elegant solution, and it works very well!
I added a couple of inline comments on functions converting string to `enum`. Otherwise, good to go :+1:

lorenzo-cavazzi (approved):
Great addition to the entity search :rocket:

PR #2353, opened 2023-02-03
refactor: provide type information for appContext

ciyer:
This provides type information for appContext so it is not necessary to do `//@ts-ignore` everywhere that the appContext is used.


lorenzo-cavazzi (approved):
Thanks for fixing this! Less `ts-ignore` around :smiley:

PR #2357, opened 2023-02-06
fix(client): add trailing slash to dataset links

### LINKED ISSUE #2355: Dataset view broken through certain navigation path
## Description

If you go to a dataset by clicking on it from the "Linked datasets" header section

image

Then you end up in a view of the dataset which does not work correctly:

image

(The Modify Dataset" button does not work).

lorenzo-cavazzi:
This is a quick fix to #2355 

/deploy #persist #cypress
fix #2355


ciyer (approved):
This fixes the problem!

PR #2358, opened 2023-02-06
Fix errors in tests

lorenzo-cavazzi:
Remove console errors from the unit tests

### DIFF
--- a/.github/workflows/test-and-ci.yml
+++ b/.github/workflows/test-and-ci.yml
@@ -19,10 +19,11 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: '16.13.2'
-    - name: Install client dependencies
+    - name: Install client dependencies and build
       run: |
         cd client
         npm install
+        npm run-script build
       env:
         NODE_OPTIONS: '--max-old-space-size=4096'
     - name: Run test client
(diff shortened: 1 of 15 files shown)


.github/workflows/test-and-ci.yml
```diff
@@ -19,10 +19,11 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: '16.13.2'
-    - name: Install client dependencies
+    - name: Install client dependencies and build
```
ciyer, line 22: This was split into two steps as part of a recent PR #2187. The reason was that debugging install problems was much more time consuming and painful when the two steps were merged into one. 

What is the advantage to returning to just one install/build step rather than keeping it as two steps?

ciyer (changes requested):
Overall, this looks good. The updates to the tests are necessary, and fixing the memory leak is a great improvement!

I only wonder about the change to the `test_and_ci` workflow file. That change was introduced recently to make debugging problems with installing packages easier to diagnose and fix. What are the advantages to returning to the previous configuration with one step that installs and builds (and therefore takes much longer) together?

.github/workflows/test-and-ci.yml
lorenzo-cavazzi: Fair point! I thought about keeping it aligned with the UI server action, I forgot about the reasons for introducing that change.
I reverted it in the last commit.

ciyer (approved):
Looks good! I think we can merge this once the release is out.

PR #2362, opened 2023-02-07
feat: Add commit selector (#2240)

### LINKED ISSUE #2240: Show commit message when selecting commits on "start session" page
## Final mockup
_Updated_

This is the final mockup we should implement https://github.com/SwissDataScienceCenter/renku-ui/issues/2240#issuecomment-1405128483
We need to switch from a standard `select` element to a richer dropdown.

### Motivation

When starting a new session "with options", it's challenging to identify a specific commit in the "Commit" list, especially when there are many from the same user at close intervals.

image

Some users suggested somehow showing the commit message, to at least part of it when it's too long.

### Mock-ups / Design

An idea could be switching the `select` element with a more complex dropdown input (as in the search) and try to fit in (the beginning of) the commit message.
⚠️ the element size should still be small enough.

Screenshot_20221219_164940

andre-code:
PR to add commit selector.

commit selector

Closes #2240 

/deploy renku=renku-ui-3.2.0-tests #persist #cypress


client/src/notebooks/NotebookStart.present.js
```diff
@@ -723,7 +723,7 @@ class StartNotebookPipelinesContent extends Component {
 class StartNotebookCommits extends Component {
   render() {
     const { commits, fetching, autosaved } = this.props.data;
-    const { delays, disabled, filters } = this.props;
+    const { delays, filters } = this.props;
```
lorenzo-cavazzi, line 726: Was the `disabled` property useful to prevent showing the dropdown menu when a branch is still not selected (or while refreshing them)?

client/src/utils/components/commitSelector/CommitSelector.tsx
```diff
@@ -0,0 +1,163 @@
...
+  };
+
+  const selector = selectedCommit ? (
+    <div className={`selected-commit cursor-pointer ${isSelectorOpened ? "selected-commit--open" : ""}`}>
+      <div onClick={clickOnSelector}>
+        <div className="commit-row">
+          <div className="commit-row-id">
+            {selectedCommit.short_id}
+            <div className="commit-row-icon">
+              {isSelectorOpened ? <ChevronDown size="20" /> : <ChevronUp size="20" /> }
```
lorenzo-cavazzi, line 122: These seem to be inverted -- I see the arrow pointing up when the element is collapsed and down when it's expanded. Usually it's the opposite.

client/src/utils/components/commitSelector/CommitSelector.tsx
```diff
@@ -0,0 +1,163 @@
...
+          </div>
+          <div className="commit-row-author text-truncate">{selectedCommit.author_name}</div>
+          <div className="commit-row-date">
+            <TimeCaption key="time-caption" caption="authored" className="text-truncate"
+              time={selectedCommit.committed_date} endPunctuation=""/>
+          </div>
+          <div className="commit-row-box-message text-truncate">
+            <div className="commit-row-message text-truncate">{selectedCommit.message}</div>
+            <div className="commit-row-icon">
+              {isSelectorOpened ? <ChevronDown size="20" /> : <ChevronUp size="20" /> }
```
lorenzo-cavazzi, line 133: Same here

lorenzo-cavazzi (changes requested):
This looks nice! :star_struck:  The arrows seem to be inverted tho, could you swap them? (see inline)

A more generic detail: I noticed that commits are shown a bit too early (E.G. when branches are still loading). Would it make sense to keep showing a temporary message instead? I believe we were used a `disabled` property to handle that (see inline), perhaps you could reuse that as an on/off switch for the new component. P.S. the previous implementation was showing a disabled element in that case, not sure that's great. Probably keep showing a temporary text is better or... any other idea is welcome :smile: 

Peek 2023-02-10 17-43

ciyer (changes requested):
This looks very good, and for the most part works well. I did notice a few issues and thought it would be worth checking how easy they are to fix.

1. The suggestions background does not go all the way to the border, leaving a weird region where content from below peaks through.

image


2. Clicking on the top option in the commit list causes the list to flicker (see video). I think it should minimize the options.

https://user-images.githubusercontent.com/1196411/218477518-4dee4e91-0893-4ae0-98df-e961cc6476c7.mov


3. The different columns of content do not line up, making it a bit more effort to processes what you are looking at. It would very nice if it could be shown more like a table.
image


I think 1. and 2. should be fixed/improved. I realize that 3 might not be easy to improve, but if something could be done, it would be nice.

andre-code:
Thanks @ciyer for your review, I have addressed and corrected all reported cases in the last commit.

ciyer (approved):
Everything I suggested has now been fixed with latest changes. Thanks, looks great! 🚀

lorenzo-cavazzi (approved):
Lgtm! :rocket:

PR #2365, opened 2023-02-08
Remove collaboration tab

### LINKED ISSUE #2256: Remove Collaboration tab from project pages
## Motivation

The collaboration tab has a long history that started with `Ku` (missing reference on purpose). It was transformed a few times, and in its current form it's rarely used. Therefore, we are planning to remove it.
We still wish to provide users with an easy way to interact and collaborate, but we need to re-work it.

In short, we will remove the Collaboration tab entirely and modify the "View in GitLab" dropdown menu in the Overview (the one in the screenshot below). More precisely, we need to:

- [ ] Rename `View in GitLab` to `Open in GitLab`.
- [ ] Rename `View in Web IDT` to `Web IDE`.
- [ ] Add a `Issues` link that will link to the same destination as the current "Collaboration --> Issues --> Open in tab" button.
- [ ] Add a `Merge Requests` link that will link to the same destination as the current "Collaboration --> Merge Requests --> Open in tab" button.
- [ ] Remove unused components from the Collaboration tab -- there should be quite a few and possibly some leftovers from the past. Ming that the Fork component is already available in a modal from the Overview tab.

image

ciyer:
# Screenshots

image

# Testing

- [ ] Check that Open in GitLab, Issues, Merge Requests, and Web IDE links work

Fix #2256


ciyer:
# More Testing

- [ ] Check that creating a dataset in a project where the default branch is protected shows a message with a correct link

client/src/project/components/GitLabConnect.tsx
```diff
@@ -0,0 +1,64 @@
...
+  url: string;
+};
+
+function GitLabLinkItem({ size, text, url }: GitLabLinkItemProps) {
+  const onClick = () => window.open(url, "_blank");
+  return (
+    <DropdownItem onClick={onClick} size={size}>
+      {text}
+    </DropdownItem>
+  );
```
lorenzo-cavazzi, line 28: `window.open` doesn't allow the typical interactions offered by an anchor element.
In other similar scenarios, we use the `ExternalLink` component with the `nav-link` class. I suggest adopting the same solution here

```suggestion
  return (
    <DropdownItem onClick={onClick} size={size}>
      <ExternalLink className="nav-link" role="text" title={text} url={url} />
    </DropdownItem>
  );
```

lorenzo-cavazzi (changes requested):
This works well!

I added a code suggestion to use standard components on the dropdown instead of linking the `onClick` action to `window.open`

client/src/project/components/GitLabConnect.tsx
ciyer: Good suggestion, I have incorporated it.

lorenzo-cavazzi (approved):
:rocket:

PR #2368, opened 2023-02-10
Show session table without overflowing

### LINKED ISSUE #2345: Button sticks out of session card (only when session is stopping)
Saw this on https://renku-ci-gw-623.dev.renku.ch/ when using this for the ui: renku/renku-ui:3.0.0

So this will happen on any deployment that uses the same version of the ui.

To reproduce start session, go to the session listing, stop it. When the session is stopping the button will poke out of the div it should be in. I saw this when my browser was smaller because I had the network tab open to look at requests. So it may only appear if the browser window is smaller.

image

ciyer:
# Screenshots

image

image

image

image

Fix #2345

/deploy #persist #cypress

### DIFF
--- a/client/src/notebooks/Notebooks.present.js
+++ b/client/src/notebooks/Notebooks.present.js
@@ -425,36 +425,38 @@ class NotebookServerRowFull extends Component {
     </span>) : null;
 
     return (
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi (approved):
Great fix! I tested with different window sizes :+1:

PR #2370, opened 2023-02-10
Update react-ipynb-renderer 

### LINKED ISSUE #2224: Upgrade react-ipynb-renderer
There is a new release of [react-ipynb-renderer](https://github.com/righ/react-ipynb-renderer) and we should upgrade to it.

ciyer:
# Testing

Check that notebooks render correctly and the same as before. Here are some examples

## Graphs, etc. are rendered correctly

* https://renku-ci-ui-2370.dev.renku.ch/projects/cramakri/migration-time-prediction/files/blob/notebooks/Model.ipynb
* https://dev.renku.ch/projects/cramakri/migration-time-prediction/files/blob/notebooks/Model.ipynb

## LaTeX is rendered correctly

* https://renku-ci-ui-2370.dev.renku.ch/projects/cramakri/mathjax/files/blob/notebooks/latex-notebook.ipynb
* https://dev.renku.ch/projects/cramakri/mathjax/files/blob/notebooks/latex-notebook.ipynb

## Cannot perform HTML Injection

* https://renku-ci-ui-2370.dev.renku.ch/projects/cramakri/js-injection-test/files/blob/notebooks/InjectionTest.ipynb
* https://dev.renku.ch/projects/cramakri/js-injection-test/files/blob/notebooks/InjectionTest.ipynb

Fix #2224 


lorenzo-cavazzi (approved):
It works well in most scenarios!
The one negative change I noticed is with Math formulas. Sometimes the components renders as inline (smaller and floating to the left) instead of being more prominent. I assume it's related to the change on `MathJax`.
Otoh Gitlab renders the components similar to what we get on this PR so... I'm fine with both options -- I'll approve the PR and let you decide.

I was comparing the 2 following scenarios:
* https://renku-ci-ui-2370.dev.renku.ch/projects/lorenzo.cavazzi.tech/essm_egu_2020/files/blob/notebooks/Priestley-Taylor_equation.ipynb
* https://dev.renku.ch/projects/lorenzo.cavazzi.tech/essm_egu_2020/files/blob/notebooks/Priestley-Taylor_equation.ipynb

image

ciyer:
> The one negative change I noticed is with Math formulas. Sometimes the components renders as inline (smaller and floating to the left) instead of being more prominent. I assume it's related to the change on `MathJax`.

I have investigated this. If the math markdown is on the same line as text, the math is rendered in "inline" mode. If it is on its own line, it is rendered in "display"mode.

image

Unfortunately, 
1. JupyterLab always displays TeX in display mode
2. It does not seem to be possible to control when Inline mode is invoked in our renderer https://github.com/righ/react-ipynb-renderer/blob/master/src/markdown.ts.

I agree that this isn't optimal, but I think leaving it as is is the best solution, since there is an easy workaround if you want the math rendered in block mode.

Issue #2308, opened 2023-01-18
Links with underscores are not rendered correctly


ciyer:
## Description

In certain circumstances, a link that includes a underscores gets treated as markdown to italicize text.  

## Reproduce

This has been seen in the dataset description field, it may appear in other places as well.
Input the following as the dataset description
```
[https://kedro.readthedocs.io/en/stable/tutorial/set_up_data.html](https://kedro.readthedocs.io/en/stable/tutorial/set_up_data.html)

.renku/dataset_images/5b5008c5a6f2415393cec4951f64e226/0.png
```

With this the `up` will be rendered as _up_.

## Task

* Find out if there are places other than dataset description where this is happening
* Fix the problem

The most likely candidate for the source of the problem is the markdown renderer we use. There is a new major version of https://www.npmjs.com/package/showdown, so that would be a good place to start.

tomashco:
Adding literalMidWordUnderscores in showdown config solves the issue.

I've opened the following PR [https://github.com/SwissDataScienceCenter/renku-ui/pull/2371](https://github.com/SwissDataScienceCenter/renku-ui/pull/2371)

lorenzo-cavazzi:
Thank you @tomashco for your contribution to this issue and #1719. We have merged your commit through #2374, and it's included in the latest UI version freshly tagged; that will hit renkulab.io in the next few days.

tomashco:
@lorenzo-cavazzi @ciyer,
thank you for the opportunity to help!

PR #2371, opened 2023-02-11
fix: Addition of literalMidWordUnderscores in showdown config to prevent sanitizedHTMLFromMarkdown from breaking

tomashco:
Testing the RenkuMarkdown component with $`\zeta_t \sim N(0,\Sigma_z(t))`$ caused the sanitizedHTMLFromMarkdown to break. 

Actually the inline markdown conversion of formulas was already working (i.e. testing it with $`E=mc^2`$ works).
The error was that the underscore inside the formula was translated into <em>, breaking the conversion.
Adding literalMidWordUnderscores: true in showDownOptions solves the issue.

As side note, I've updated the showdown-katex and showdown packages and added the RenkuMarkdown.stories.tsx in order to conveniently test the component inside storybook.

This PR solves also the [2308](https://github.com/SwissDataScienceCenter/renku-ui/issues/2308)


ciyer (approved):
Thanks a lot for this contribution! It looks good to me.

Due to the way our CI pipeline is set up, it is necessary to to first merge your changes into a branch that is part of this repo, but your fix makes sense, and I will take it from here.

tomashco:
Hi @ciyer,

thank you very much for accepting the PR, as I wrote also in the comment, this should fix also the bug 2308.

PR #2373, opened 2023-02-13
feat: add sorting for date:desc in all search

andre-code:
PR to include sorting by date desc in kg-search.


ciyer (dismissed):
Very nice! This works exactly as I would expect:

* results are sorted in the desired sort order + descending date order if there are multiple matches with the same sort order value (I tested this in several cases)
* results are stored in desired date sort order if a date sort order is specified

lorenzo-cavazzi:
This requires updating the acceptance tests from the KG team to support their current `development` branch

lorenzo-cavazzi (approved):

PR #2374, opened 2023-02-13
fix: markdown rendering bugs

### LINKED ISSUE #1719: Rendering inline math formulas
## Description

It seems there are still issues with rendering inline math, even after #1622

## Reproduce

This is a sample with a couple of formulas in the readme
https://renkulab.io/projects/lorenzo.cavazzi.tech/math-formulas

They are not rendered when using the inline notation, but they are when put inside a code block.
GitLab seems to always render them correctly (on the right in the Screenshot)

Screenshot_20220303_173313

### LINKED ISSUE #2308: Links with underscores are not rendered correctly

ciyer:
Contribution by @tomashco that fixes some rendering issues with markdown (links and forumlas).

Fix #1719 
Fix #2308

Follow up to #2371


ciyer:
# Testing

Compare the readme in these two projects.

* https://renku-ci-ui-2374.dev.renku.ch/projects/cramakri/markdown-test
* https://dev.renku.ch/projects/cramakri/markdown-test

lorenzo-cavazzi (approved):
Lgtm :rocket:

PR #2376, opened 2023-02-15
feat(client): add SSH support on the UI

### LINKED ISSUE #2324: 🔶 SSH: verify ssh support on project pages
We need to check whether a project supports SSH and show appropriate feedback.

We will get the information from the `GET /renku/check_migration` endpoint. We already do that when loading the project page to verify renku-core compatibility, so the change should be minimal.
[Here is the renku-python PR](https://github.com/SwissDataScienceCenter/renku-python/pull/3294) implementing the new response for the endpoint -- the new flag is `template_status.ssh_supported`. It can be tested [using this project](https://gitlab.dev.renku.ch/ralf.grubenmann/test-ssh-support).
Here is the link to the API on swagger:
https://renku-ci-rp-3294.dev.renku.ch/swagger/#/cache/get_renku_cache_migrations_check
Here is the renku-notebooks PR with the changes to the `GET /notebooks/version` endpoint:
https://renku-ci-nb-1389.dev.renku.ch

**EDIT**: some deployments will **not** support SSH. We can verify that by querying `GET /notebooks/version`. We should adjust the UI accordingly to prevent showing wrong information or commands that cannon be run.

### Showing the info
There are a few possible places:
* The project status page
* Somewhere on the project session page
* In the "start with options" page
  * Maybe just a label, ssh enabled or not? That info should come from the "renku.ini" file.
  * If it's more complex than this, or we can't just readf "renku.ini", we should create a separate issue for this

lorenzo-cavazzi:
Add SSH support in the UI

/deploy #persist #notest renku-core=feature/ssh-into-sessions renku-notebooks=feat-ssh-access extra-values=notebooks.ssh.enabled=true
fix #2324


client/src/features/display/displaySlice.ts
```diff
@@ -0,0 +1,62 @@
...
+  reducers: {
+    showSshModal: (state, action: PayloadAction<ProjectConfig>) => {
+      return {
+        ...state, modals: {
+          ...state.modals, ssh: {
+            ...state.modals.ssh,
+            show: true,
+            projectPath: action.payload.projectPath,
+            gitUrl: action.payload.gitUrl,
+          }
```
ciyer, line 45: This kind of thing isn't necessary. The redux-toolkit reducer arguments are [wrapped using Immer](https://redux-toolkit.js.org/usage/immer-reducers) so you can just mutate the values.

```suggestion
      state.modals.show = true;
      state.modals.projectPath = action.payload.projectPath;
      state.modals.gitUrl = action.payload.gitUrl;
```

client/src/features/display/displaySlice.ts
```diff
@@ -0,0 +1,62 @@
...
+        }
+      };
+    },
+    hideSshModal: (state) => {
+      return { ...state, modals: { ...state.modals, ssh: { ...state.modals.ssh, show: false } } };
```
ciyer, line 50: See above comment.

```suggestion
     state.modals.show = false;
```

client/src/features/display/displaySlice.ts
```diff
@@ -0,0 +1,62 @@
...
+    },
+    toggleSshModal: (state) => {
+      return { ...state, modals: { ...state.modals, ssh: { ...state.modals.ssh, show: !state.modals.ssh.show } } };
```
ciyer, line 53: See above comment.

```suggestion
      state.modals.ssh.show: !state.modals.ssh.show;
```

client/src/features/versions/versions.d.ts
```diff
@@ -0,0 +1,44 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+interface NotebookComponent {
+  data: {
+    anonymousSessionsEnabled: boolean;
+    cloudstorageEnabled: {
```
ciyer, line 22: I think it would be better of "Cloud Storage" were two word:

```suggestion
    cloudStorageEnabled: {
```

client/src/features/versions/versions.d.ts
```diff
@@ -0,0 +1,44 @@
...
+  name: string;
+  versions: NotebookComponent[];
+}
+
+interface NotebooksVersion {
+  name: string;
+  version: string;
+  anonymousSessionsEnabled: boolean;
+  sshEnabled: boolean;
+  cloudstorageEnabled: boolean;
```
ciyer, line 41: See above. 

```suggestion
  cloudStorageEnabled: boolean;
```

client/src/styles/bootstrap_ext/_button.scss
```diff
@@ -39,6 +39,10 @@ $borderRadius: 1000px;
   margin-right: 12px;
 }
 
+.btn-with-menu-options .btn-with-menu-margin {
+  margin-right: 12px;
+}
+
```
ciyer, line 45: I'm guessing that the idea is that the margin should be the same for both `svg` and elements with the specified class. I think in that case, it would be better to combine the definitions:

```suggestion
.btn-with-menu-options {
    svg, .btn-with-menu-margin {
      margin-right: 12px;
     }
}
```
(Not 100% sure on that syntax, but should be something similar.)

ciyer (changes requested):
Still working on testing the UI, but I have reviewed the code. The main thing I noticed is that the reducers can be simplified. Otherwise, it looks good!

client/src/utils/components/ssh/ssh.tsx
```diff
@@ -0,0 +1,223 @@
...
+    modalBody = (
+      <>
+        <p>
+          <img src={rkIconSshCross} className="rk-icon rk-icon-md filter-dark me-2" />
+          Your project needs to be updated to support SSH.
+        </p>
+        <p>
+          To update your project, go to{" "}
+          <Link to={updateUrl} onClick={() => closeModal()}>Overview Status</Link>
+          {" "}then come back to this window to view the commands to connect remotely to sessions.
```
ciyer, line 163: I think "come back to this window" is a bit misleading, since the window disappears when you go to the status page. Might better to say:

```suggestion
          {" "}start an ssh session again to view the commands to connect remotely to sessions.
```

client/src/utils/components/ssh/ssh.tsx
```diff
@@ -0,0 +1,223 @@
...
+      </>
+    );
+  }
+  else {
+    const command = "renku session ssh-setup";
+    modalBody = (
+      <>
+        <p>
+          <img src={rkIconSshTicked} className="rk-icon rk-icon-md filter-green me-2" />
+          Your project supports SSH.
```
ciyer, line 179: I think this text here is very small / easy to overlook.

image

I think we should make it a bit more prominent and set it off from the rest of the body text somehow.

client/src/utils/components/ssh/ssh.tsx
```diff
@@ -0,0 +1,223 @@
...
+        </p>
+        <p>
+          You can now switch to your local CLI and run the following command to set up the project
+          correctly. Then follow the instructions to connect.
```
ciyer, line 183: Is it necessary to have the project checked-out locally? If so, we should mention that.

ciyer, line 183: I do not see any instructions to connect. After running the specified commands, I only see:
```
$ renku session ssh-setup
Info: Generating keys
Info: Writing SSH config
```

client/src/features/display/displaySlice.ts
lorenzo-cavazzi: Oh nice! That simplifies reading the code. I'll add this and the following changes (I'll add another commit since I need to add `.ssh` and I'll use a single object to further simplify)

client/src/features/versions/versions.d.ts
lorenzo-cavazzi: I agree that `cloudStorageEnabled` would read better. I kept that aligned with the response from the service.

I would either keep this as is now, or change also `azure_blob` to `azure` or `azureBlob` (if we don't care about consistency with the response, then let's use camel case everywhere):

lorenzo-cavazzi: Ah nevermind, let's stick to camelCase for coherence -- I'll change it

client/src/styles/bootstrap_ext/_button.scss
lorenzo-cavazzi: Ah yes, good point. I'll harmonize that CSS rules with the previous one on `svg` elements

client/src/utils/components/ssh/ssh.tsx
lorenzo-cavazzi: I further rephrased it based on both yours and other offline comments.
Please let me know if it works well

client/src/utils/components/ssh/ssh.tsx
lorenzo-cavazzi: Good point, I made it bold. Let me know if it works well with the new content.

client/src/utils/components/ssh/ssh.tsx
lorenzo-cavazzi: Good point. Based on this and the previous comment, the page has been reworked to contain more information. Please let me know if it's an improvement in the right direction.
P.S. I mixed in other offline suggestions too.

lorenzo-cavazzi:
Thank you for the review @ciyer :pray: 
I addressed all your points (mind that it takes ~40 mins to re-deploy due to referencing custom branches for renku-core and renku-notebooks).

Please double-check the changes in the text for the SSH instructions. Inspired by your and other offline comments, I ended up adding more information.

ciyer (approved):
This looks great! 🎉  As discussed in the stand up, I added a small commit with minor cosmetic changes for visual consistency.

PR #2388, opened 2023-02-22
fix(client): show and use timezone aware dates

lorenzo-cavazzi:
We use ISO timezones strings where we should not, resulting in confusing outputs.
Here is an example where we convert to an ISO string, strip the timezone, and show the output. GitLab shows the proper time (timezone aware) while RenkuLab shows the UTC time.

image

This PR fixes it where necessary.
/deploy #persist #cypress


ciyer (commented):
This looks ok to me, and is certainly an improvement. I noticed, though, that GitLab includes the time zone, which makes it less ambiguous. Do we want to consider including that too, or was that idea already rejected?

lorenzo-cavazzi:
@ciyer That's a good idea! This is mostly a bugfix since the dates shown were wrong.
Can I create a follow-up issue to also show the timezone?

ciyer:
> @ciyer That's a good idea! This is mostly a bugfix since the dates shown were wrong. Can I create a follow-up issue to also show the timezone?

I will approve then.

ciyer (approved):
👍

PR #2398, opened 2023-02-24
feat(client): add basic support for SSH

### LINKED ISSUE #2324: 🔶 SSH: verify ssh support on project pages

lorenzo-cavazzi:
Add SSH support in the UI, showing a "Connect to session with SSH" entry on the "start/connect" dropdown button on the dashboard and on the project header.

/deploy #persist #cypress renku-core=v2.3.0 renku-notebooks=0.15.0 extra-values=notebooks.ssh.enabled=true

fix https://github.com/SwissDataScienceCenter/renku-ui/issues/2324
fix #2324


lorenzo-cavazzi:
#2376 will be merged instead

PR #2400, opened 2023-02-24
fix(client): fetch sessions early when loading the dashboard

lorenzo-cavazzi:
Fetches the sessions when rendering the Dashboard to have up-to-date data immediately when loading the app in the UI

/deploy #cypress #persist


ciyer (approved):
Much smoother looking on the initial load now! 🚀

PR #2405, opened 2023-03-02
refactor(client): convert workflows coordinator to RTK Query

### LINKED ISSUE #2113: Convert WorkflowsCoordinator to RTK
## Motivation

We are transitioning to RTK, we should avoid implementing new `Coordinator` classes.
Yet, to speed up the development of the workflows feature, a new `WorkflowsCoordinator` was added. We should convert that RTK before we add more functions and queries.

lorenzo-cavazzi:
Refactor the workflows coordinator to use RTK instead of the old Model approach.

fix #2113


client/src/features/workflows/WorkflowsSlice.ts
```diff
@@ -0,0 +1,74 @@
...
+interface WorkflowsSetDetailPayload {
+  targetDetails: Record<string, any>; // ? There are many types, not sure it makes sense to add specific types
+}
+
+export const workflowsSlice = createSlice({
+  name: "workflowsDisplay",
+  initialState,
+  reducers: {
+    toggleExpanded(state, action: PayloadAction<WorkflowsTogglePayload>) {
+      const expandedCurrent = [...state.expanded];
```
ciyer, line 51: I do not think this copy is necessary.

client/src/workflows/Workflows.container.tsx
```diff
@@ -45,45 +37,67 @@ const WorkflowsSorting = {
...
+    list: workflowsQuery.data,
+    error: workflowsQuery.error,
+    fetched: workflowsQuery.data !== null && !workflowsQuery.isLoading,
+    fetching: workflowsQuery.isFetching || workflowsQuery.isLoading,
+    expanded: workflowsDisplay.expanded,
+    showInactive: workflowsDisplay.showInactive,
+    orderAscending: workflowsDisplay.orderAscending,
+    orderProperty: workflowsDisplay.orderProperty,
+  };
+  const waiting = !versionUrlAvailable || workflowsQuery.isLoading;
```
ciyer, line 82: Is this correct, or would it be better to use `workflows.fetching`?

client/src/workflows/Workflows.container.tsx
```diff
@@ -45,45 +37,67 @@ const WorkflowsSorting = {
...
-    workflowsCoordinator.fetchWorkflowDetails(selected, repositoryUrl, reference, versionUrl, fullPath).
-      then((result: any) => {
-        // ? not sure if this should be here, but this scrolls nicely to the workflow detail
-        setTimeout(() => {
-          document.getElementById("workflowsDetailsContent")?.scrollIntoView(true);
-        }, 100);
-      });
-  }, [selected, repositoryUrl, reference, versionUrl]); // eslint-disable-line react-hooks/exhaustive-deps
+  // Fetch workflow details
+  const skipDetails = (!versionUrlAvailable || !repositoryUrl || !selected) ? true : false;
```
ciyer, line 85:
```suggestion
  const skipDetails = (skipList || !selected) ? true : false;
```

ciyer (changes requested):
Nice refactoring! I tested the UI, and it looks like the behavior is the same as before (good!).

I have some minor suggestions for code improvements, but I think this is overall very good.

client/src/features/workflows/Workflows.d.ts
```diff
@@ -0,0 +1,111 @@
...
+  creators: Creator[],
+  description: string,
+  duration: number,
+  id: string,
+  keywords: string[],
+  last_executed: Date,
+  name: string,
+  number_of_executions: number,
+  touches_existing_files: boolean,
+  type: "Plan" | "CompositePlan"
```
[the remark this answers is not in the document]
lorenzo-cavazzi: Fair point! Actually `WorkflowType` already exists in the components entities. I imported that here too -- and re-exported it since it seems more natural to look for that in the Workflows definitions

client/src/features/workflows/WorkflowsApi.ts
```diff
@@ -0,0 +1,133 @@
...
+  } as WorkflowDetails;
+}
+
+export const workflowsApi = createApi({
+  reducerPath: "workflowsApi",
+  baseQuery: fetchBaseQuery({ baseUrl: "" }),
+  endpoints: (builder) => ({
+    getWorkflowList: builder.query<WorkflowPlan[], {
+      coreUrl: string, gitUrl: string, reference: string, fullPath: string
+    }>({
```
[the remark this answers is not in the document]
lorenzo-cavazzi: Good idea! I added that to the workflows definitions file

client/src/workflows/Workflows.container.tsx
lorenzo-cavazzi: This is correct: the idea is to show the spinning wheel on the first fetch, but we don't want to flash it later.
Still, your comment made be realize we lost the mechanism to "refetch when stale" so I added `refetchOnMountOrArgChange`

ciyer (dismissed):
Looks great!

ciyer (approved):
The updates to handle errors look good. No need to make it overly complicated.

Issue #2386, opened 2023-02-21
Test the new gateway and adjust the UI server (and client?) where necessary


lorenzo-cavazzi:
The UI server will "move" behind the Gateway and delegate all the Authentication tasks.
In the long term, it will essentially become a WebSocket service. For now, the tasks are:
- [ ] Adapt the UI server to the new Gateway.
- [ ] Change the routing in the UI client wherever needed.

leafty:
This is probably not relevant anymore.

PR #2407, opened 2023-03-07
feat: use gateway for authentication

olevski:
The goal here is to remove the authentication and session management from the ui-server in the least intrusive way.

The reason for this is that we can switch back easily if things go poorly with the gateway or there are issues.

Please ignore for now - still WIP and being tested with the gateway.


lorenzo-cavazzi:
This is basically fixing #2386

lorenzo-cavazzi:
Leftover

PR #2408, opened 2023-03-08
Do not remember projects that are just ids (#2316)

### LINKED ISSUE #2316: Last projects showed twice
## Description

A user reported seeing the same project listed twice on the home page where we show the last visited projects.
Apparently, the `GET /last-projects/<length>` API on the UI server might return duplicates.

After inspecting that specific case, it seems that we save both project slugs and numerical IDs, so it's possible we save both version for the same project ad in 
```
{
  projects: {
    "pamela.delgado/ga-intro",
    "24457"
  }
}
```

The easiest way to address this is to prevent the UI server from saving the numerical ID -- that's converted to the full slug anyway when visiting the project in the UI.

image

## Reproduce

Visit a project using the numerical ID and then go back to the landing page
https://renkulab.io/projects/24457

ciyer:
Fix #2316

### DIFF
--- a/server/src/utils/middlewares/lastProjectsMiddleware.ts
+++ b/server/src/utils/middlewares/lastProjectsMiddleware.ts
@@ -21,10 +21,19 @@ import express from "express";
 import config from "../../config";
 import { getUserIdFromToken } from "../../authentication";
 
+function projectNameIsId(projectName: string): boolean {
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi (approved):
The fix works well! I tested with a few projects and it constantly drops the numeric IDs (on the left) where before it used to store and return them (on the right).

image

PR #2410, opened 2023-03-08
fix(client): start sessions with base images on broken pinned images

### LINKED ISSUE #2395: Starting with base image button does not work properly when image is pinned
To reproduce:
- create a project (or pick an existing one)
- pin the image to something that does not exist - I just type a bunch of gibberish in the image name field
- go to start with options page
- the page will correctly recognize the image does not exist but it will allow you to use the "start with base image" button
- pressing the "start with base image" button results in a failure to launch the image because the POST request to the notebook service still uses the pinned non-existing image

The logic on the notebooks around this is as follows:
- if the image field in the POST request to launch a session is specified then there is no fallback to a default, if the image requested does not exist then we fail with a 404 and an error message stating this (the error message properly shows up in the ui too actually which is great)
- if the image is not specified we derive an image name based on the project and commit sha, if that image does not exist we use the default

So one way to fix this is to simply not request the pinned image when the pinned image does not exist and the user pressed the "start with base image" button.

lorenzo-cavazzi:
Restores the "start with base image" button for pinned images with a broken image.

@olevski Using an example like the project `lorenzo.cavazzi.tech/pinned-image-broken`, everything works well but I noticed that the `GET /servers` endpoint returns `false` on the annotation `default_image_used`.

1. Should I pass some specific value to the `POST /servers` API to properly set the `default_image_used` annotation?
2. [_very_ minor] Any chance we can get a real `true/false` boolean for that field instead of a stringified Pythonish boolean `"False"/"True"`?

```
"annotations": {
        ...
        "renku.io/default_image_used": "False",
},
```

fix #2395

### DIFF
--- a/client/src/notebooks/NotebookStart.present.js
+++ b/client/src/notebooks/NotebookStart.present.js
@@ -1099,7 +1099,7 @@ class ServerOptionLaunch extends Component {
     this.setState({ showModal: !this.state.showModal });
   }
 
-  checkServer() {
+  checkServer(forceBaseImage = false) {
     const { filters } = this.props;
     const { autosaved } = this.props.data;
     const selectedBranchName = filters.branch.name;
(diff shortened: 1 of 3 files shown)


ciyer (approved):
The fix here is clear!

PR #2411, opened 2023-03-09
feat: improve display of session error information (#2396)

### LINKED ISSUE #2396: Error messages from failing sessions are hard to find
I have seen this with multiple people using Renku. Most recently Till had this problem.

When a session fails. The status of the session is properly shown in the ui. However the error message that explains the failure is hidden in this small red icon. And it was not clear to Till (and several other people I have talked to) that you can click this small icon to see the message.

I think it would be better to display the error message more prominently or to make it clear that the little icon is clickable and also make it clear what contents it has.

image

ciyer:
image

image

Fix #2396


lorenzo-cavazzi (approved):
Lgtm!

PR #2414, opened 2023-03-10
Query KG on project access

### LINKED ISSUE #2337: Fetch KG project data and consider switching to that for getting descriptions and avatars
## Description

KG has already implemented a solution for getting the project data and metadata from a "project" endpoint.
We should invoke that when we load a project and store the data into a new slice ([API reference](https://github.com/SwissDataScienceCenter/renku-graph/tree/development/knowledge-graph#get-knowledge-graphprojectsnamespacename)).
ℹ️ This is the only mandatory step for this issue, and we want to add that because KG will soon start keeping track of the last accesses to projects to delete obsolete metadata.

After this step, we should explore the possibility to use those data. The 2 most interesting parts are:

- Description: this is already added at creation time through renku-core. We should:
  - use the `POST /renku/project.edit` endpoint to edit the project `description` also from the project settings page ([API reference](https://dev.renku.ch/swagger/#/project/post_renku_project_edit))
  - verify how quickly that information is updated on KG (do we get the new description if we invoke the KG project API immediately after modifying it? how long does it take re-indexing in this specific case?)
  - If that's slow, verify using the renku-core API to fetch that property ([API reference](https://dev.renku.ch/swagger/#/project/post_renku_project_show))
- Avatars: we _should_ have support soon for that on renku core. The pattern would be similar to the one described above. Verify it's feasible and consider creating a follow-up issue.
- Other properties: do they come from the KG or are those taken directly from GitLab? In the second case, we can consume them right away and start phasing out GitLab for that... Verify with KG team this info, and create follow-up issue where necessary

ciyer:
Make a request to the KG on project access, so the KG can keep track of when projects are viewed.

Fix #2337

/deploy #persist #cypress


ciyer:
Thanks for the review! The requested changes have been made:
- Remove the code related to fetching description from the KG. That will be part of a later PR
- Skip the KG requests if the project is **not** in the KG

client/src/features/project/ProjectPageTitle.tsx
```diff
@@ -0,0 +1,53 @@
...
+  projectTitle: string;
+};
+
+function ProjectPageTitle({ isInKg, projectPathWithNamespace, projectTitle }: ProjectJsonLdProps) {
+  const kgProjectQueryParams = {
+    projectPath: projectPathWithNamespace,
+  };
+  const options = { skip: isInKg != true };
+  const { data, isFetching, isLoading } = useProjectJsonLdQuery(kgProjectQueryParams, options);
+  // eslint-disable-next-line no-unused-vars
```
lorenzo-cavazzi, line 36: I believe this slipped in

lorenzo-cavazzi (dismissed):
This looks good!
There would be a comment to remove but it's not important

lorenzo-cavazzi (approved):

Issue #2402, opened 2023-02-27
Show timezone when showing times


lorenzo-cavazzi:
## Motivation

We addressed an issue with times not considering timezones (see #2388 ) but the UX might still be quite confusing since we still do not show timezones at all.

## Proposal

[As suggested here](https://github.com/SwissDataScienceCenter/renku-ui/pull/2388#pullrequestreview-1315511953), we should use an approach similar to GitLab, actually showing the time and the timezone. Looking at the following screenshot, we should display the timezone in a new line in a similar fashion, without changing other details about the styling of out popovers/tooltips.

### Suggestions
* Adding more structure to the Tooltips could be nice, but we can probably limit the effort by providing a single string for the timezone without whitespaces; this should guarantee decent word wrapping (from the screenshot, I assume GitLab is using a similar approach :detective: )
* We have a common component for showing times with a tooltip (`TimeCaption` iirc), but I fear we don't consistently use it. We should verify whether that's the case and either uses the common component or update the local code accordingly to show a similar text (:bulb: tip: check every place where we invoke `Time.toIsoTimezoneString` )

![image](https://user-images.githubusercontent.com/43481553/220574980-2c1ac439-adc6-4218-b158-8c447cfc81a5.png)

StevyCoding:
Hello @lorenzo-cavazzi  can I take care of this issue ?

StevyCoding:
@lorenzo-cavazzi Thank you for the update and the tips on how to properly address the problem. I appreciate it. Could you please let me know what the deadline for this task is?

lorenzo-cavazzi:
It's not a high-priority issue so there is no internal deadline yet. Issues scheduled for a specific sprint are marked in the GitHub Project "UI sprint" and assigned to a specific sprint (you can see it in the right column under Projects).
We welcome your contribution whenever you have time :slightly_smiling_face:

lorenzo-cavazzi:
closed in #2527

PR #2417, opened 2023-03-12
Show timezone when showing times #2402

StevyCoding:
Hello, I have updated the display format for the date on the TimeCaption component. Please let me know if you are satisfied with the changes.
image

image

### DIFF
--- a/client/src/dataset/DatasetFunctions.js
+++ b/client/src/dataset/DatasetFunctions.js
@@ -65,7 +65,7 @@ function getDatasetImageUrl(images) {
  * Returns image url with extra parameter to avoid outdated cached url
  *
  * @param {string} imageUrl - Url image
- * @param {string} lastUpdateDate - last url update
+ * @param {string | Date} lastUpdateDate - last url update
  */
 function getUpdatedDatasetImage(imageUrl, lastUpdateDate) {
   if (!imageUrl)
--- a/client/src/features/project/Project.d.ts
+++ b/client/src/features/project/Project.d.ts
@@ -1,0 +1,50 @@
+type IDataset = {
+  created_at: string; // could be created?
+  creators: string[];
+  description: string;
+  exists?: boolean;
+  hasPart?: boolean;
+  identifier: string;
+  insideKg: boolean;
+  keywords: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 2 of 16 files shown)


client/src/utils/components/TimeCaption.js
```diff
@@ -28,32 +28,55 @@ import Time from "../helpers/Time";
...
+  // format the time string 2023-02-26T17:11:00.000Z in Feb 26, 2023, 5:11 PM GMT+1 format
+  const formattedDate = new Intl.DateTimeFormat("en-US", {
+    month: "short",
+    day: "numeric",
+    year: "numeric",
+    hour: "numeric",
+    minute: "numeric",
+    timeZoneName: "short",
+    timeZone: userTimeZone, // set your desired timezone here
+  }).format(date);
```
lorenzo-cavazzi, line 61: This includes the timezone but also prints the date with the "US" format, which is an undesired change. We currently don't handle browser's language settings, so we brodly stick to a long ISO format as in the screenshot

image

What about re-using the `timeZoneName: "short"` approach to get just the timezone string and keeping the previous datetime string? 
I suggest modifying the function `Time.toIsoTimezoneString` and adding a new type `"datetime-timezone"` to print a full ISO date plus the timezone at the end.
In that way, we can easily re-use the same function in other components.

Here are a few examples of ISO Time offset representations: https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC
We should stick to that and show something like `"+02:00"`, `"+05:30"`, `"−05:00"`. It's fine to add `UTC` or `GMT` in front to make it less confusing, as in `"UTC+02:00"`,  `"UTC+05:30"`

P.S. I _think_ you don't need `Intl.DateTimeFormat().resolvedOptions().timeZone`, you could probably just invoke `toLocaleTimeString` directly on the date. Something like `date.toLocaleTimeString("en-us",{timeZoneName:"short"}).split(" ")[2]` might work, but I haven't tried it.

lorenzo-cavazzi (changes requested):
Thank you for your contribution! I like the approach you used but I suggest changing a couple of things:
* Stick to the ISO format instead of switching to the `us` format.
* Integrate the conversion into the `Time` utility we also use in other components.

I added a more exhaustive comment inline. Don't esitate to ask if anything isn't clear.

Issue #2415, opened 2023-03-10
Refactoring datasets components


lorenzo-cavazzi:
This is an umbrella issue for refactoring datasets components.

### Tasks

Tasks include converting all the datasets components to:
- Use React functional components.
- Invoke APIs through RTK Query.
    - :information_source: **UPDATE**: we should phase out the `GET  /knowledge-graph/datasets` API (this: https://github.com/SwissDataScienceCenter/renku-graph/tree/2.28.3/knowledge-graph#get-knowledge-graphdatasets) and use the `GET /knowledge-graph/entities` to  fetch all the datasets data
    - We can be cautious and create a separate API for that, and swap them at the end; if something doesn't work, we don't need to hold on but we should notify KG to add the missing data and then create an issue to swap them in the (near) future
- Save local status to Redux slices using RTK.

### Changes

In the process, we should:
- Avoid storing data in the Redux store through the old Model approach, nor storing them locally (remember that even simple preferences stored locally might be reset by ancestor components refreshing at any time since many Datasets components are descendants of old Project components).
- Avoid using the previous From-generator approach; consider using a different library for that (maybe [react-final-form](https://github.com/final-form/react-final-form), or [redux-form](https://github.com/redux-form/redux-form) if we need to keep the state in Redux for the reasons suggested above), or otherwise just use a very simplified way of preserving transient user inputs in Redux.
- Remove all the datasets components passed as properties from the Project page (in the `subComponents` function).
- Remember these minor changes from other issues when re-implementing the components
    - #2356 
    - #1930 
    - #1912 
    - #1270

### Caveats

Here are a few things to keep in mind.

Datasets can be accessed either through a project or alone. There are some differences:
- As an independent entity they:
    - Have a unique ID, similar to a SHA, that identifies them (there might be new versions, and older versions of these ids might contain `-` that _must_ be stripped out). There is no user-friendly name. E.G: https://dev.renku.ch/datasets/7492c98815a544418762fcb5fae2d6ba
    - Data is available only from the KG, and it might be temporarily not up-to-date after modifying a dataset. Currently, there is no way around it since the KG takes some time to process new commits.
    - They are usually contained in 1 or more projects; even when contained in just 1 project, they are a separate entity from the project's counterpart.
- Inside a project: 
[...]

PR #2428, opened 2023-03-20
Stop passing components as properties in the Project view

ciyer:
The is preparation for further work on #2415. 

/deploy #persist #cypress renku=renku-ui-3.4.1-tests

### DIFF
--- a/client/src/project/datasets/ProjectDatasets.test.js
+++ b/client/src/project/datasets/ProjectDatasets.test.js
@@ -27,7 +27,6 @@
 import React from "react";
 import { createRoot } from "react-dom/client";
 import { createMemoryHistory } from "history";
-import { Provider } from "react-redux";
 import { act } from "react-test-renderer";
 
 import { MemoryRouter } from "react-router-dom";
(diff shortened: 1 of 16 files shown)


client/src/project/datasets/ProjectDatasets.test.js
```diff
@@ -72,44 +62,6 @@ describe("rendering", () => {
     spy.mockRestore();
   });
 
-  it("renders datasets list without crashing", async () => {
```
lorenzo-cavazzi, line 75: Should we replace this with an e2e test?
Or would you like to do it later when updating the dataset components? I assume the other sub-components should have e2e tests as well

lorenzo-cavazzi (approved):
This looks very good! I added an inline comment; feel free to address it either here or in the follow-up datasets refactoring issue.

client/src/project/datasets/ProjectDatasets.test.js
ciyer: > Should we replace this with an e2e test? Or would you like to do it later when updating the dataset components? I assume the other sub-components should have e2e tests as well

Yes, this is already covered by the `datasets.spec.ts` test case.

lorenzo-cavazzi (dismissed):
Lgtm! :rocket:

lorenzo-cavazzi (approved):

PR #2429, opened 2023-03-20
chore(client): fix linter to support tsx files

lorenzo-cavazzi:
Sadly, out linter was misconfigured and the `.tsx` files were not linted. We have quite a few errors.
This PR fixes the linter configuration and ignores most of the past errors (fixing them would require quite some time)

/deploy #persist #cypress #renku=2337-project-kg


ciyer (approved):
Very important improvement here!

PR #2431, opened 2023-03-20
test(e2e): remove live test

lorenzo-cavazzi:
Remove e2e live tests -- we now have the Cypress acceptance tests to cover that, and the current implementation doesn't test the deployment availability before trying to log in.


.github/workflows/test-and-ci.yml
```diff
@@ -80,13 +80,45 @@ jobs:
...
+          working-directory: tests
+      - name: Cypress run
+        uses: cypress-io/github-action@v4
+        with:
+          browser: chrome
+          start: npm run start
+          wait-on: http://localhost:3000
+          wait-on-timeout: 180
+          working-directory: tests
+      - uses: actions/upload-artifact@v3
```
ciyer, line 109: I do not follow why the `build client` step was removed and merged into here.

Given the way the cypress test is set up here (with `npm run start`) it does not use the output of `npm run build` (unless I'm missing something).

I think the better solution would be to leave the Build Client step as before and remove build client from here.

That would have the benefit that we could re-run the e2e test without having to wait for the app to be built (which adds several minutes).

ciyer (changes requested):
I think this is really good, and it's great that you took the initiative to implement this so quickly after we had discussed it.

The only think I do not understand is the changes to when the CI pipeline tries to build the UI. I do not see why these changes were necessary, and I think the previous structure was better.

.github/workflows/test-and-ci.yml
lorenzo-cavazzi: Oh right, building is actually not necessary.
At some point, I _had_ to build the client to speed up starting the server. Then I found about `wait-on` and adding `wait-on: http://localhost:3000` is actually enough to prevent running tests too early and fail.

Let me move out the "client build" step into his own place again. I'll also refresh the documentation since it's stale now.

PR #2445, opened 2023-03-29
Fix dashboard links to sessions and update ssh message

lorenzo-cavazzi:
This includes a few minor fixes:
- (urgent) fix the URL to running sessions on the dashboard
- update ssh message
- prepare a bugfix release

/deploy #persist #cypress


ciyer (approved):
I tested and the dashboard session links do work correctly.

PR #2450, opened 2023-03-31
wip: get description from the KG

ciyer:
* Retrieve project description from KG
* Set project description using renku core


lorenzo-cavazzi:
superseded by #2631

PR #2453, opened 2023-03-31
Query for datasets using RTK-Query

### LINKED ISSUE #2415: Refactoring datasets components

ciyer:
Use RTK-Query to retrieve dataset information.

Implements parts of #2415

/deploy #persist #cypress renku=renku-ui-3.4.1-tests

### DIFF
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -19,7 +19,13 @@
     "plugin:react/recommended",
     "plugin:jest/recommended"
   ],
-  "plugins": ["react", "react-hooks", "jest", "spellcheck", "@typescript-eslint"],
+  "plugins": [
+    "react",
+    "react-hooks",
+    "jest",
+    "spellcheck",
+    "@typescript-eslint"
+  ],
   "settings": {
     "react": {
       "version": "detect"
@@ -224,6 +230,7 @@
           "Ravey",
           "rect",
           "redux",
+          "reduxjs",
           "remark2rehype",
           "renderer",
           "renderers",
--- a/client/src/dataset/Dataset.present.js
+++ b/client/src/dataset/Dataset.present.js
@@ -46,16 +46,21 @@ import { EntityDeleteButtonButton } from "../components/entities/Buttons";
 
 function DisplayFiles(props) {
   if (!props.files || !props.files?.hasPart) return null;
-  if (props.files?.fetching || props.loadingDatasets) return "LOADING FILES...";
+  if (props.isFilesFetching || props.loadingDatasets) return "LOADING FILES...";
 
 
-  if (props.files.fetchError !== null) {
-    const error = props.files.fetchError;
+  if (props.filesFetchError != null) {
+    const error = props.filesFetchError;
     let errorObject;
-    if (CoreError.isValid(error))
+    if (CoreError.isValid(error)) {
       errorObject = (<CoreErrorAlert error={error}/>);
-    else
-      errorObject = (<span><strong>Error fetching dataset files:</strong> {props.files.fetchError.message}</span>);
+    }
+    else {
+      errorObject = (<span>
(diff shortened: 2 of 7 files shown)


lorenzo-cavazzi (commented):
Nice solution! This works well and the code looks good :+1:  I assume `datasets.list` will be migrated later since it touches the project page and it's bit more complicated.

Do you mind trying this with `renku-graph=1257-new-datasets-query` to verify if it works well even with the new mixed IDs from KG? Or should that be included later?

ciyer:
> Nice solution! This works well and the code looks good 👍 I assume `datasets.list` will be migrated later since it touches the project page and it's bit more complicated.

Yes, exactly. The call to datasets.list is used elsewhere in the project, so I did not want to get to lost in those changes, but I will work on those improvements once this PR is merged.

lorenzo-cavazzi (approved):

PR #2460, opened 2023-04-06
Skip the workflow list query if a migration is required for core

### LINKED ISSUE #2443: Prevent firing workflow queries on outdated projects
## Description

On the workflows pages, we invoke queries on renku-core even when the project is outdated and there is no backend support for them (E.G. project version 9 and support only for version 10 and 11)
We should first check the available core versions, and prevent queries (using `skip` on the RTK Query) when there is no backend support.

## Reproduce

Open a workflow page on any outdated project.
This is an example with version... 1 :fearful:
https://renkulab.io/projects/laurent.gatto/wsbim1207/workflows

leafty:
When such a migration is required, the workflow list query will return an error, so we should skip it.

Fixes #2443.

/deploy #persist #cypress


client/src/workflows/Workflows.container.tsx
```diff
@@ -79,7 +80,7 @@ function WorkflowsList({ fullPath, reference, repositoryUrl, versionUrl }: Workf
   const workflowsDisplay = useWorkflowsSelector((state: RootStateOrAny) => state[workflowsSlice.name]);
 
   // Fetch workflow list
-  const skipList = (!versionUrlAvailable || !repositoryUrl) ? true : false;
+  const skipList = !versionUrlAvailable || !repositoryUrl || coreMigrationRequired;
```
lorenzo-cavazzi, line 83: The idea here is correct: we need to skip the query when there is no support from the backend.
However, it's possible that the project requires to be updated but it's not necessary for the workflows query. This is an example of a project still at version 9 where you can still query the workflows: https://dev.renku.ch/projects/workflows-pet-projects/workflows-flights-tutorial/workflows

You need to check that the current project version is in the list of the supported version from renku-core; we query that from `GET /ui-server/api/renku/version` and store it in the Redux store in "environment/coreVersions"
image
Sadly, the query fetching it is not in RTK Query yet -- feel free to let it untouched and get the information from the current Redux (nothing prevents you from converting  the query to RTK Query, but it would require updating the other places where this information is currently used).

A few tips:
* You can access the old "stateModel" global object in the Redux store using `props.model.get("environment.coreVersions")` in the `ProjectViewWorkflows` component. If you need an example from the existing code base, search for `model.baseModel.get("environment.coreVersions");` to find one.
* `versionUrl` is a string containing the version number. You can check the `checkRenkuCoreSupport` function to understand how to derive the project's metadata version number from `versionUrl`. That's enough to solve this issue, but feel free to create a dedicated function or pass the "pure" version number through the components if you prefer a cleaner solution.

Additionally, the current solution crashes when skipping the fetch -- try opening https://renku-ci-ui-2460.dev.renku.ch/projects/workflows-pet-projects/workflows-flights-tutorial
Depending on the final solution you implement, you might need to handle the skipped case also in the `WorkflowsTreeBrowserPresent`.

P.S. To be on the safe side, you can also add `unsupported` to the elements causing `skipList` to be `true`; The `skipList` variable will look like
```
const skipList = !versionUrlAvailable || !repositoryUrl || unsupported || <something_else_here>;
```

lorenzo-cavazzi (changes requested):
Nice solution! You will also need to handle the presentational component to prevent crashes on the UI.

However, the situation is slightly more complicated cause we support multiple project versions at the same time. See inline for a (verbose) explanation and a few tips.

client/src/workflows/Workflows.container.tsx
```diff
@@ -21,13 +21,10 @@ import { RootStateOrAny, useDispatch } from "react-redux";
 import { useParams } from "react-router-dom";
 
 import { WorkflowsTreeBrowser as WorkflowsTreeBrowserPresent } from "./Workflows.present";
-import { checkRenkuCoreSupport } from "../utils/helpers/HelperFunctions";
 import { useGetWorkflowDetailQuery, useGetWorkflowListQuery } from "../features/workflows/WorkflowsApi";
 import { workflowsSlice, useWorkflowsSelector } from "../features/workflows/WorkflowsSlice";
 
 
-const MIN_CORE_VERSION_WORKFLOWS = 9;
```
lorenzo-cavazzi, line 29: We may want to keep this for a bit longer; there was no support for workflows on renku-core before version 9. That means we can't show info for projects v8 in any case.

client/src/workflows/Workflows.container.tsx
```diff
@@ -44,6 +41,8 @@ interface WorkflowsListProps {
   reference: string;
   repositoryUrl: string;
   versionUrl: string;
+  coreMigrationRequired: boolean;
```
lorenzo-cavazzi, line 44: Not necessary anymore

lorenzo-cavazzi (changes requested):
A tiny change request to keep the minimum version for the workflows -- pre v9, we have no support from the backend (more details inline).

The e2e tests seem to fail because the presentational component tries to read unavailable values when it should probably stop earlier; I guess the `unsupported` variable in the `WorkflowsTreeBrowserPresent` component might be `undefined` -- this is based on the error I read on the console when running the tests manually (see screenshot); please double check if this is actually the case.

You might need to update the tests to invoke the function to simulate the API response for the renku-core versions (the endpoint should be something like `check_migration`), and be sure `unsupported` is always `true` or `false`.

image

client/src/workflows/Workflows.container.tsx
leafty: Done in [65c223c](https://github.com/SwissDataScienceCenter/renku-ui/pull/2460/commits/65c223cf1f55e173a9793f79fd1b0c12c27f7a5c).

client/src/workflows/Workflows.container.tsx
leafty: Added back in [77b27fe](https://github.com/SwissDataScienceCenter/renku-ui/pull/2460/commits/77b27fec358690ebb6b0fb18740f0c1977502fdd).

lorenzo-cavazzi (approved):
Great, the fix works well!
Let's wait for the acceptance tests -- those failures didn't make much sense to me since they are unrelated to the changes in this PR :thinking:

PR #2461, opened 2023-04-11
Use dropdown for session options when many choices

### LINKED ISSUE #2237: Group buttons in sessions not reachable
## Description

When starting a new session "with options", some of the options might not be visible/reachable depending on how many options are provided and the window size. The worse combination seems to be "lots of options" + "medium size window".

image

## Reproduce

This is easy to reproduce on limited since there are many options for both memory and storage.
To quickly reproduce the problem while developing, the easiest way is to add a few extra fake options to the response of the `server_options` API.

## Solution

We agreed on this solution https://github.com/SwissDataScienceCenter/renku-ui/issues/2237#issuecomment-1463891326

* Show separate buttons when there are few elements: that should be decided based on the number of elements + the characters contained in it.
    * try a few combinations to find a magic number: E.G: <num of chars> * <5 * number of elements> <= 20
    * _always_ show the button on 1 element
* Show a dropdown through a standard `<select>` element when there are mote elements

leafty:
The button group used for session options is replaced with a dropdown
menu when there are more than 5 choices. This is done to avoid choices
spilling over other options.

Fixes #2237.

/deploy #persist #cypress


leafty:
Current UI, without a lot of styling:
leafty dev renku ch_projects_workflows-pet-projects_workflows-flights-tutorial_sessions_new_
leafty dev renku ch_projects_workflows-pet-projects_workflows-flights-tutorial_sessions_new_ (1)

lorenzo-cavazzi (changes requested):
Nice solution using CSS modules!

Apart from a couple of linting errors to fix, it would be nice to use the same thicker border already in use for the selected radio button. Otherwise, it feels "unselected".

image

I noticed the size is also slightly different; perhaps it's related to the borders? Please double-check it again after adjusting borders just to be on the safe side :grin: 

image

lorenzo-cavazzi (dismissed):
Excellent solution, it works well!

lorenzo-cavazzi (approved):

PR #2465, opened 2023-04-12
Dashboard: fix issue with trying to process deleted project data

### LINKED ISSUE #2437: Dashboard suddenly listing no projects
## Description

The dashboard occasionally doesn't show any project, even when projects have been visited and the `GET /last-projects` API returns values.

## Reproduce

I suspect this happens when one of the projects returns a 404 error (see screenshot) i.e. when the user has manually deleted them.
This problem might become more common when we add a Delete button on the UI.
Otoh this should be really easy to fix.

image

leafty:
Fixes a bug where the API `/projects/<project>?statistics=false&doNotTrack=true` 404 response is processed as project data. We now skip these items.

/deploy #persist #cypress renku=renku-ui-3.4.2-updates
Fixes #2437.

### DIFF
--- a/client/src/features/projects/ProjectsApi.ts
+++ b/client/src/features/projects/ProjectsApi.ts
@@ -127,9 +127,10 @@ export const projectApi = createApi({
           try {
             const resultAllProjectData = await Promise.allSettled(projectRequests);
             const projectList = [];
-            for (const projectData of resultAllProjectData)
+            for (const projectData of resultAllProjectData) {
               if (projectData.status === "fulfilled" && !projectData.value.error)
                 projectList.push(formatProjectMetadata(projectData.value.data));
+            }
 
             return { data: projectList };
           }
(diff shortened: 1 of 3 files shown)


client/src/features/projects/ProjectsApi.ts
```diff
@@ -128,7 +128,8 @@ export const projectApi = createApi({
             const resultAllProjectData = await Promise.allSettled(projectRequests);
             const projectList = [];
             for (const projectData of resultAllProjectData)
-              if (projectData.status === "fulfilled") projectList.push(formatProjectMetadata(projectData.value.data));
+              if (projectData.status === "fulfilled" && !projectData.value.error)
+                projectList.push(formatProjectMetadata(projectData.value.data));
```
leafty, line 132: Note: it seems that the `{ error: e as FetchBaseQueryError }` return below is not processed; the error seems to be silently swallowed in `client/src/utils/customHooks/useGetRecentlyVisitedProjects.ts`.

lorenzo-cavazzi (approved):
Lgtm!

PR #2466, opened 2023-04-12
fix(client): prevent dropping pinned images

### LINKED ISSUE #2458: Pinning an image does not set the image property in the payload to notebook service
Saw this on renkulab.io.

When an image is pinned the payload does not contain the `image` property. So the pinned image is ignored and the image based on the commit is used.

image

lorenzo-cavazzi:
This fixes a nasty bug accidentally introduced here #2410 that prevents pinned images from being picked up correctly.

In the screenshot, I tested a project with a pinned image that was picked up correctly -- it's from a different commit than the current one.

image

You can use this project to test it https://dev.renku.ch/projects/lorenzo.cavazzi.tech/1408-pinned-image

/deploy #persist #cypress renku=renku-ui-3.4.2-updates
fix #2458


leafty (approved):
LGTM

Unfortunately a typical pitfall of js classes and `this` not being bound by default.

PR #2468, opened 2023-04-13
Add prettier to format UI code

### LINKED ISSUE #2427: Use prettier to keep code formatting coherent
## Description

We set up some rules through our linter to [impose constraints on code formatting](https://github.com/SwissDataScienceCenter/renku-ui/blob/master/client/.eslintrc.json#L27).
That approach is not the best since the style is not coherent everywhere.

We should consider dropping all the formatting rules and switching to a proper formatter like [Prettier](https://prettier.io/)

## Tasks

That requires a few steps:
* [ ] Evaluate alternatives (Prettier seems to be the most used; double check alternatives to be sure we don't miss something obvious)
* [ ] Install the packages, remove the stale `ESlint` rules, and replace (or drop) them.
* [ ] Adjust all the files. Typically, the `--fix` command doesn't handle everything.

leafty:
Add prettier to the `client` project and scripts to format only changed files.

To format changed files:
```bash
npm run format-changed
```

Also, we add a pre-commit hook to check staged files, but we don't format them and re-stage them for now.

Fixes #2427.

For review, see:
* https://github.com/SwissDataScienceCenter/renku-ui/pull/2468/files/65b04c1a25eb6097376c95bd8f07fc2a1001b64c
* eslint fixes -> https://github.com/SwissDataScienceCenter/renku-ui/pull/2468/commits/72ead52d33cf8642f9670ab132b2c1a2e2c43dd4

/deploy #persist #cypress


package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "renku-ui-root",
+  "private": true,
+  "devDependencies": {
+    "husky": "^8.0.0"
+  },
+  "scripts": {
+    "prepare": "husky install"
+  }
+}
```
leafty, line 10: Note: we can only install `husky` at the same location as the root of the git project.

leafty:
Note: example of CI failing because of formatting -> https://github.com/SwissDataScienceCenter/renku-ui/actions/runs/4795881027/jobs/8530980159

lorenzo-cavazzi (changes requested):
Very nice! Finally no formatting discrepancies anymore :raised_hands: 

A couple of details:
1. If I run `npm run format` there are still a couple of minimal changes, just 4 files. Perhaps we should re-run it?
2. Could you update the base README.md file, adding a reference to "Prettier", the additional `npm install` step in the "Developing" sections, and updating the "Unit Tests and Linting" sections as well? Bonus for adding a quick description of how to set up VScode to auto-format on save (this last step not 100% necessary tbh). Thank you :pray:

leafty:
> Very nice! Finally no formatting discrepancies anymore raised_hands
> 
> A couple of details:
> 
> 1. If I run `npm run format` there are still a couple of minimal changes, just 4 files. Perhaps we should re-run it?
> 2. Could you update the base README.md file, adding a reference to "Prettier", the additional `npm install` step in the "Developing" sections, and updating the "Unit Tests and Linting" sections as well? Bonus for adding a quick description of how to set up VScode to auto-format on save (this last step not 100% necessary tbh). Thank you pray

Updated readme in 65b04c1a25eb6097376c95bd8f07fc2a1001b64c.

lorenzo-cavazzi (dismissed):
:rocket:

lorenzo-cavazzi (approved):

PR #2472, opened 2023-04-14
Remove conspicuous "," in the add dataset screen

leafty:
See screenshot:
image

This is most likely a typo, an extra "," is seen above "Add Dataset".


ciyer (approved):
👍

PR #2473, opened 2023-04-14
Remove unused route in client/src/project/Project.present.js

leafty:
The removed route is never hit as the `datasetUrl` one is always hit beforehand.
Also, displaying the project navbar would not follow the convention used in other screens using a "Back to <project>" button.

/deploy #persist #cypress


ciyer (dismissed):
Good catch!

PR #2452, opened 2023-03-31
Make the sessions list fork button work

ciyer:
This should fix the sessions list fork button, but the styling needs to be fixed to only show the "Fork" option and not the number of forks.

image

# Testing

- [ ] Fork button should appear as in the image above when on the sessions page of a project that you do not have commit access to
- [ ] Fork button should work on the sessions page
- [ ] Fork button should appear as before in the Project > Overview > General page
- [ ] Fork button should work on the Project > Overview page

Fix #2451

/deploy #persist #cypress renku=renku-ui-3.4.2-updates

PR #2474, opened 2023-04-14
fix(client): fix `notebookWarning()` broken invocation after #2452

leafty:
In PR #2452, the line https://github.com/SwissDataScienceCenter/renku-ui/blob/39853035225ac330e34beb8ce11392e9116181a5/client/src/project/Project.present.js#L903 was not updated like https://github.com/SwissDataScienceCenter/renku-ui/blob/39853035225ac330e34beb8ce11392e9116181a5/client/src/project/Project.present.js#L858-L859.

This PR should also fix the `UnprivilegedSessionSpec` Selenium acceptance test.

/deploy #persist #cypress

### DIFF
--- a/client/src/project/Project.present.js
+++ b/client/src/project/Project.present.js
@@ -900,7 +900,8 @@ class ProjectStartNotebookServer extends Component {
       backUrl,
       defaultBackButton,
     } = this.props;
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi (approved):
Thanks for spotting and fixing this!

PR #2481, opened 2023-04-18
feat(client): show timezones

### LINKED ISSUE #2402: Show timezone when showing times

lorenzo-cavazzi:
_still a draft_

/deploy #persist #notest
fix #2402

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Changes
 
+## [3.4.2](https://github.com/SwissDataScienceCenter/renku-ui/compare/3.4.1...3.4.2) (2023-04-18)
+
+### Features
+
+* use dropdown for session options when the number of choices is too large for a button group ([#2461](https://github.com/SwissDataScienceCenter/renku-ui/issues/2461), [#2237](https://github.com/SwissDataScienceCenter/renku-ui/issues/2237))
+
+### Bug Fixes
+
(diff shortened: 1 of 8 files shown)


lorenzo-cavazzi:
Something looks wrong with how we handle dates.
Double-check if everything is fine, and consider using a time library like [MomentJS](https://momentjs.com/) to simplify handling all the corner case

lorenzo-cavazzi:
Will be replaced by #2527

ciyer:
Replaced by https://github.com/SwissDataScienceCenter/renku-ui/pull/2527

PR #2482, opened 2023-04-18
fix(client): prevent content layout shift when selecting a template

leafty:
Issue: when hovering templates in the new project form, the content gets shifted due to the added border.
See: https://dev.renku.ch/projects/new.

The new template cards do not shift around when hovered or selected.
See: https://leafty.dev.renku.ch/projects/new.

/deploy #persist #cypress


lorenzo-cavazzi (approved):
:+1:

Issue #2447, opened 2023-03-30
Fix search page multiple redirects


lorenzo-cavazzi:
## Description

We added most of the search filters to the URL as query parameters to preserve the search state and make the URL sharable ( #2351 ) using the `histroy` component to set filters.
Accidentally, now clicking on items adds multiple entries in the history. As a side effect, users need to click multiple times on the browser's "Back" button to go back to the previous page.

We should identify where this happens and fix it, so that clicking around adds a single navigation event and the browser's navigation buttons can be used again.

## Reproduce

Try to click on a project from the search page, and then go back with the browser's buttons. See the example GIF

![Peek 2023-03-30 17-21](https://user-images.githubusercontent.com/43481553/228884904-9b54baa0-4986-4f10-9e88-5f14fb7d22cd.gif)

PR #2486, opened 2023-04-20
refactor: use redux for kg search

### LINKED ISSUE #2446: Preserve page number on search page during navigation
## Description

We recently introduced the KG-based search page and, just after, we added most of the search filters to the URL as query parameters to preserve the search state and make the URL sharable #2351

There is an issue with the navigation history, and some parameters might get lost when using the "Back" button of the browser.
It's easy to reproduce for the page number, but we should double-check that it doesn't happen in other cases too.

## Reproduce

* Go to the search page.
* Scroll down and go to another page (E.G. page 4).
* Open a random project.
* Click on the browser's "Back" button.

The page number is lost at this point, and the user returns to page 1 (mind that the `page` query parameter is gone).

Peek 2023-03-30 17-06

leafty:
WIP to fix #2446 and #2447.

/deploy #persist #cypress


leafty:
Closed in favour of #2491.

PR #2491, opened 2023-04-24
refactor(client): simplify the logic of the kg search page

### LINKED ISSUE #2448: Cache content from the KG entities query for up to 1 minute
## Description

Loading data on the search page takes quite some time.
At the moment, if the user clicks on a project and then tries to go back to the search page immediately, the UI fires another query to the KG entities endpoint, taking quite some time to show the content again.

## Solution

We should cache the response for some (short) time; try 30 or 60 seconds. That should be very easy using RTK Query since there is a parameter for keeping the data valid for a certain amount of time after the components using it is unmounted (it's already used in the Workflows)

leafty:
This refactor simplifies the logic of the kg search page (`/search`), which should eliminate bugs related to browser history.

Fixes https://github.com/SwissDataScienceCenter/renku-ui/issues/2446, fixes https://github.com/SwissDataScienceCenter/renku-ui/issues/2447 and fixes #2448.

/deploy #persist #cypress


client/src/features/kgSearch/KgSearch.d.ts
```diff
@@ -48,14 +53,25 @@ interface KgSearchResult {
...
+  since: string;
+  sort: SortingOptions;
+  type: TypeEntitySelection;
+  typeDate: DateFilterTypes;
+  until: string;
+  visibility: VisibilitiesFilter;
+}
 
-export type { ListResponse, KgAuthor, KgSearchResult, KgSearchResultLink };
+export const TOTAL_RESULT_PER_PAGE = 20;
```
lorenzo-cavazzi, line 77: We are not using this. We can drop it entirely

client/src/features/kgSearch/KgSearchState.ts
```diff
@@ -110,137 +68,134 @@ type KgStateVal<T extends KgStateSimpleKey> = T extends KgStateAuthorKey
...
-}
+};
 
 /**
  * The date parameters need to be handled specially.
  * @param qp Query parameters
  * @returns since, typeDate, until
  */
-function queryParameterDateStateValue(qp: URLSearchParams): Pick<KgSearchFormState, "since" | "typeDate" | "until"> {
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
lorenzo-cavazzi, line 140: I feel we could remove the `eslint-disable` rule here, and fix the code later.
The `Object.keys(val!)` sections trigger the warnings. I feel we can wrap them around an `if (val !== undefined) {` conditional -- something like the following:

```
if (val !== undefined) {
        Object.keys(val).forEach((k) => {
          if (val[k as keyof typeof val] === true) stateMap.push([key, k]);
        });
      }
```

:warning: Please double-check the code, I haven't tried it thoroughly

lorenzo-cavazzi (changes requested):
The code looks great! I agree it's simpler and less error-prone than the other solution.

I added a couple of inline comments. They don't target changes from you, but since we are already touching the code, it's worth considering them.

Another tiny change: in the "features/kgSearch/KgSearchApi.ts" file, we might add a parameter in the `createApi` section to cache the content for a bit longer (I believe it's `keepUnusedDataFor` but please double-check this). If we set it to something like 30 seconds , browsing back and forth in a short sequence (or ticking/unticking the same filters) won't trigger new queries.

P.S. I just realized that would actually also fix https://github.com/SwissDataScienceCenter/renku-ui/issues/2448
Feel free to include it in the PR or we can make a separate one later

client/src/features/kgSearch/KgSearch.d.ts
leafty: Done in [f45dbd1](https://github.com/SwissDataScienceCenter/renku-ui/pull/2491/commits/f45dbd16eb27e76019b2d67f2220b4547d21e842).

leafty:
> The code looks great! I agree it's simpler and less error-prone than the other solution.
> 
> I added a couple of inline comments. They don't target changes from you, but since we are already touching the code, it's worth considering them.
> 
> Another tiny change: in the "features/kgSearch/KgSearchApi.ts" file, we might add a parameter in the `createApi` section to cache the content for a bit longer (I believe it's `keepUnusedDataFor` but please double-check this). If we set it to something like 30 seconds , browsing back and forth in a short sequence (or ticking/unticking the same filters) won't trigger new queries.
> 
> P.S. I just realized that would actually also fix #2448 Feel free to include it in the PR or we can make a separate one later

Done in [e21a96f](https://github.com/SwissDataScienceCenter/renku-ui/pull/2491/commits/e21a96f1a9f0b764fdaadf313e865bab5ac6245b).

lorenzo-cavazzi (approved):
Lgtm :rocket:

PR #2493, opened 2023-04-25
fix(client): correctly show loading state on a project's workflows tab

### LINKED ISSUE #2492: Regression: non-supported message briefly flashes when loading a project's workflows tab
## Description

This small regression was introduced by #2460: when loading `/projects/<namespace>/<project>/workflows`, the "Interacting with workflows in the UI requires updating your project to a newer version." message briefly flashes while the UI is waiting for the `/api/renku/cache.migrations_check` API endpoint to respond.

## Reproduce

E.g. visit: https://dev.renku.ch/projects/lorenzo.cavazzi.tech/flights-tutorial-2022-04-19/workflows

Reload the page to be sure to see the flashing message.

Peek 2023-04-24 18-06

leafty:
Fixes #2492.

/deploy #persist #cypress

### DIFF
--- a/client/src/project/Project.present.js
+++ b/client/src/project/Project.present.js
@@ -647,13 +647,17 @@ class ProjectViewOverview extends Component {
 function ProjectViewWorkflows(props) {
   const reference = props.metadata?.defaultBranch ? props.metadata?.defaultBranch : "";
 
+  useEffect(() => {
+    console.log({ migration: props.migration })
+  }, [props.migration])
+
   return (
     <WorkflowsList
       fullPath={props.projectPathWithNamespace}
       reference={reference}
       repositoryUrl={props.externalUrl}
       versionUrl={props.migration?.core?.versionUrl}
-      backendAvailable={!!props.migration?.core?.backendAvailable}
+      backendAvailable={props.migration?.core?.backendAvailable}
     />
   );
 }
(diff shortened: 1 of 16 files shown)


lorenzo-cavazzi (changes requested):
If you load the project page straight into the `/workflows` tab, you still briefly get the warning as in the following gif. I tried with this link:
https://renku-ci-ui-2493.dev.renku.ch/projects/lorenzo.cavazzi.tech/flights-tutorial-2022-04-19/workflows


Peek 2023-04-25 16-45

Suggestion: in the redux store, together with the `migration.core.backendAvailable` boolean, we store a `fetched` property initialized as `null` ([Reference](https://github.com/SwissDataScienceCenter/renku-ui/blob/master/client/src/model/RenkuModels.js#L323-L331)).
Perhaps checking if no `!fetched` might help? Not sure tho if that's the broken piece or something else needs to be checked.
image

client/src/project/Project.present.js
```diff
@@ -653,7 +653,7 @@ function ProjectViewWorkflows(props) {
       reference={reference}
       repositoryUrl={props.externalUrl}
       versionUrl={props.migration?.core?.versionUrl}
-      backendAvailable={!!props.migration?.core?.backendAvailable}
+      backendAvailable={props.migration?.core?.backendAvailable}
```
leafty, line 656: @lorenzo-cavazzi I simply forgot to include this change. The message should not flash now.

lorenzo-cavazzi (approved):

PR #2495, opened 2023-04-25
feat(client): add a clone button to the project overview

### LINKED ISSUE #2440: Add a Clone button
## Motivation

We show the repository URLs and other commands to clone on the settings page. This is not very intuitive and hard to find.

image

## Proposal

We should adopt a solution [similar to GitLab](https://gitlab.dev.renku.ch/lorenzo.cavazzi.tech/updated-flight-tutorial-2023), with a dropdown Clone button that opens a panel with the Clone commands ready to copy

image

The new button could go to the Overview page, where we currently have "Fork" and "Open in Gitlab". The Fork button should stay, and the "Open in GitLab" might either stay or be integrated with the new Clone button.
We don't need the date there anymore since it's already in the new project's header

image

leafty:
Add a "Clone" button to the project overview which open a card showing the `renku clone` command and repository URLs.

Other changes:
* Update various components to use the new `CommandCopy` component to display a line of code to copy/paste.

Fixes #2440.

/deploy #persist #cypress renku=leafty/update-tests-renku-ui-2495


leafty:
### Update

Clone card:
image

Clone card with git:
image

Removed clone commands from settings:
image

Settings when access is lower than `maintainer`:
image

leafty:
### Other components updated

Create a shareable link:
image
image

Connect with SSH (ignore "SSH not supported"):
image

Unsaved work modal:
image

Session cheat sheet:
image

Workflow details:
image

Create link for new project:
image

client/src/components/Clipboard.tsx
```diff
@@ -15,62 +15,80 @@
...
     >
-      {" "}
-      {copied ? (
-        <FontAwesomeIcon icon={faCheck} color="success" />
-      ) : (
-        <FontAwesomeIcon icon={faCopy} />
-      )}
+      <Wrap>
+        {copied ? (
+          <FontAwesomeIcon icon={faCheck} size="1x" style={{ minWidth: 16 }} />
```
lorenzo-cavazzi, line 86: Why manually specify the pixels? Should we rely only on the size instead?

client/src/project/clone/CloneSettings.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+          <p>
+            If you want to work with the repo using renku, you need to run the
+            following after the <code>git clone</code> completes:
+          </p>
+          <CommandCopy command={gitHooksInstall} />
+        </>
+      )}
+
+      <div /*className="text-end"*/>
```
lorenzo-cavazzi, line 112:
```suggestion
      <div>
```

lorenzo-cavazzi (changes requested):
This works well already! And thank you for going through the code and updating copy/paste everywhere to make it consistent :pray: 
I have a couple of comments:

* Clicking again on the "Clone" button once opened gives inconsistent results: when clicking on the arrow, nothing happens; when clicking on the text "Clone", it closes the dropdown. I would keep the same behavior (my personal preference would be to close, but all other dropdown buttons don't close on the second click, so let's privilege consistency)

* Do you mind adding a bottom margin after "Do not have renku?"
It has non and looks visually very close to the following element
image

lorenzo-cavazzi:
Also, it might be necessary to update the acceptance tests (Cypress for sure, perhaps also the others? :thinking: )

image

lorenzo-cavazzi:
As discussed offline, let's use an `InfoAlert` element for logged-in users without permission to inform them the settings are available only to maintainers.

For non-logged users, let's use something similar to the Settings Session section

image

leafty:
Updated settings:

Screenshot 2023-05-03 at 10 49 02

Screenshot 2023-05-03 at 10 43 03

leafty:
> This works well already! And thank you for going through the code and updating copy/paste everywhere to make it consistent 🙏 I have a couple of comments:
> 
> * Clicking again on the "Clone" button once opened gives inconsistent results: when clicking on the arrow, nothing happens; when clicking on the text "Clone", it closes the dropdown. I would keep the same behavior (my personal preference would be to close, but all other dropdown buttons don't close on the second click, so let's privilege consistency)
> * Do you mind adding a bottom margin after "Do not have renku?"
>   It has non and looks visually very close to the following element
>   image

Updated margins in fbf2ebad012f8972fab5326828004e40a779c2a4.

Also I do not know how to fix the behaviour of the button parts, it's using reactstrap components the same way as in their docs.

client/src/components/Clipboard.tsx
leafty: Fixed with 9ebb1b6c98bd72e73196081dd956f2fc4531c80b.

client/src/project/clone/CloneSettings.tsx
leafty: Thanks, fixed.

lorenzo-cavazzi (approved):
This works well :tada: 
Let's merge once both acceptance tests are fixed :rocket:

PR #2497, opened 2023-04-26
fix(client): update broken documentation link

lorenzo-cavazzi:
Fix broken link

/deploy #persist


leafty (approved):
LGTM

Users should be redirected to: https://renku.readthedocs.io/en/stable/renku-python/docs/reference/commands/migrate.html.

PR #2502, opened 2023-04-27
fix(client): do not crash the UI when showing all workflows

### LINKED ISSUE #2488: Toggling active/inactive workflows crashes the ui
I noticed this on dev in this project: https://dev.renku.ch/projects/tasko.olevski/nccr-demo-test-1

Attached is a short video.

The project is public so you can try things out as well.

https://user-images.githubusercontent.com/16360283/233357313-58757cd9-5381-49c8-b008-0818ec57951c.mov

leafty:
The workflows list passed from RTK query is read-only, but we call `.sort()` to order the workflow list. As a fix, we simply copy the list into a new array when we list all workflows ("show inactive workflows" toggle).

Fixes #2488.

/deploy #persist #cypress

### DIFF
--- a/client/src/workflows/Workflows.present.tsx
+++ b/client/src/workflows/Workflows.present.tsx
@@ -306,7 +306,7 @@ function orderWorkflows(
 ) {
   const filtered = !showInactive
     ? workflows.filter((w) => w.active)
-    : workflows;
+    : [...workflows];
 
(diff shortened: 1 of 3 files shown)


lorenzo-cavazzi (dismissed):
Simple and effective :ok_hand:

Issue #2484, opened 2023-04-19
Support Resource Pools from the UI


lorenzo-cavazzi:
## Motivation

We are introducing [Resource Access Control](https://github.com/SwissDataScienceCenter/renku-design-docs/blob/main/rfcs/011-resource-access-control/011-resource-access-control.md) in RenkuLab to assign resource pools to specific users.

This requires a few changes in the UI since the session will start within the context of a specific resource pool with an assigned set of resources.

### Starting sessions

We will need to query an API to get the list of resources (it should be [the `GET /resource_pools` API here](https://github.com/SwissDataScienceCenter/renku-data-services/blob/main/src/api.spec.yaml)) available to the user.
We will also need to query the project options to get a project's default pool and try to set it.
For quick-started sessions, we will use whatever is the project default or just the global default if nothing local overwrites it; for sessions through the "Start with options" page, we will error if the project-defined pool is not available, and show a warning. User can pick any other available option from the pool on that page.

ℹ️ This means the single resources (CPU, RAM, disk, GPU) won't be selectable as single entities but just as a pre-set. We won't need the `notebook_options` anymore since there will be just pools with sets of resources.

![image](https://user-images.githubusercontent.com/43481553/233350191-94e00920-7136-4351-810c-9bca81a58103.png)

### Sessions settings
On the project page, users can set session settings. The abovementioned ones will disappear and be swapped with the new pool sets.
Keep in mind that not everything is disappearing from there. The pinned images won't change, nor the default_url (even tho it will be just a pure string), and _probably_ also the "Automatically fetch LFS data" should not change.

jdupre81:
Basically, and for practical and consistency purposes, the design follows the same design principle of the commit dropdown message. 

![Image](https://user-images.githubusercontent.com/117274853/234025859-ab75b16b-7f0a-4f6e-821a-8fc081916b66.jpg)



![Image](https://user-images.githubusercontent.com/117274853/234025882-debcda90-c98e-40c9-9a43-dee24cd0e8aa.jpg)



![Image](https://user-images.githubusercontent.com/117274853/234025897-26971139-ddae-4da8-a901-2fac7efb70d4.jpg)

jdupre81:
as a **second option** I did a design with a short list and put a "Expand Button", but I think the design above looks much better. 

![Image](https://user-images.githubusercontent.com/117274853/234027129-62196a9b-a0a8-4998-b9bc-b58b2b9c6785.jpg)



[...]

PR #2506, opened 2023-05-04
Add support for compute resource pools in the session start options

leafty:
Add support for selecting the session class from the new resource pool API.

Notes:
* Update for storage disk size slider will be done in another PR
* Using the real new API endpoints will be done in another PR
* Styling of the drop-down menu will be finalized in another PR

See #2484.

image
image


client/src/features/project/projectCoreApi.ts
```diff
@@ -107,9 +121,38 @@ export const projectCoreApi = createApi({
...
+        const params = {
+          git_url: projectRepositoryUrl,
+          ...(branch ? { branch } : {}),
+        };
+        return {
+          url: versionedUrlEndpoint("config.show", versionUrl),
+          params,
+        };
+      },
+      transformResponse: (response: GetConfigRawResponse) => ({
```
lorenzo-cavazzi, line 136: I'm wondering if we should get all options here, so that we can identify and overwrite the outdated `interactive` options (separating into props at `.` makes sense). Still, this PR doesn't include the sessions settings page so it can be delayed

lorenzo-cavazzi (approved):
Very promising!
The solution you picked (and the code) look clean! Just a comment but it can be taken care of later.

client/src/features/project/projectCoreApi.ts
leafty: :+1: It will make sense to have a look with the settings page.

PR #2520, opened 2023-05-12
refactor(client): update RTK slice selector hooks

leafty:
The new `createSliceSelector()` helper function creates a React hook which simplifies writing state selectors for RTK slices.

Other updates:
* Moved `client/src/utils/helpers/EnhancedState.js` to TypeScript and export only one function, `createStore()`.

/deploy #persist #cypress renku=leafty/update-tests-renku-ui-2495

### DIFF
--- a/client/src/components/Logs.tsx
+++ b/client/src/components/Logs.tsx
@@ -17,7 +17,7 @@
  */
 
 import React, { useEffect } from "react";
-import { RootStateOrAny, useDispatch } from "react-redux";
+import { useDispatch } from "react-redux";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faSave, faSyncAlt } from "@fortawesome/free-solid-svg-icons";
 import {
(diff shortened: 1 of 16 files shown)


client/src/components/Logs.tsx
```diff
@@ -318,9 +318,7 @@ interface EnvironmentLogsProps {
   name: string;
 }
 const EnvironmentLogs = ({ name, annotations }: EnvironmentLogsProps) => {
-  const displayModal = useDisplaySelector(
-    (state: RootStateOrAny) => state[displaySlice.name].modals.sessionLogs
-  );
+  const displayModal = useDisplaySelector((state) => state.modals.sessionLogs);
```
leafty, line 321: Note: here, `state` has type `Display` which means `state.modals.sessionLogs` is fully type-checked and supports auto-complete from vscode.

lorenzo-cavazzi (approved):
Oh, properly typed objects from the store! :tada: :pray:

Issue #2483, opened 2023-04-19
Fix dates by using a wrapper library


lorenzo-cavazzi:
## Description

While working on #2481 , I noticed we might be introducing errors in dates. Try to initiate a date object and pass it to `Time.toIsoString`; you will get a wrong output.

The easiest solution to fix this would be to use a date and time library, like [Luxon](https://moment.github.io/luxon/#/).
We can keep out Time abstraction to be sure we don't accidentally use different formats in different places.

The steps to fix this are:

* [ ] Re-write the `Time.js` utility in TypeScript, using Luxon to convert dates from and into strings.
* [ ] Update the tests -- they are not considering the time zones now. Be sure to include that "detail".

## Reproduce

Try to initiate a date object and pass it to `Time.toIsoString`; you will get a wrong output.

PR #2527, opened 2023-05-16
refactor(client): update DateTime and Duration utilities to use luxon

leafty:
Fixes https://github.com/SwissDataScienceCenter/renku-ui/issues/2483.

New datetime and duration utility functions are introduced, using `luxon`.

Some datetime rendering now include the timezone, see below.

Before:
Screenshot 2023-06-15 at 14 04 39

After:
Screenshot 2023-06-15 at 14 03 39

/deploy #persist #cypress


lorenzo-cavazzi:
Is this intended to replace #2508 ?

leafty:
> Is this intended to replace #2508 ?

Yes, I am keeping it around to pick some code pieces.

lorenzo-cavazzi:
Just a generic comment while I review: there are a few downgrades in `client/package.json`.
I assume they are not intended and are just a side effect of a merge

leafty:
> Just a generic comment while I review: there are a few downgrades in `client/package.json`. I assume they are not intended and are just a side effect of a merge

Ah yes, thanks for noticing. I think this is because the branch was started a while back. I will fix the `package.json` file.

lorenzo-cavazzi (commented):
The code looks great and the solution is really good! We can merge this after adjusting the package file :rocket: 

I wonder if it makes sense to change how we visualize dates in some places to keep consistency across the application (if you agree with the changes, this can be done either within this PR or in a separate one).

The following could be turned into relative distances with a tooltip for the specific time:
- File header indicating last commit author and time (from "Lorenzo Cavazzi 2023-06-14 15:32:40 GMT+2" to "Lorenzo Cavazzi, 2 hours ago")
  image
- Project's commits list in the "Overvieew --> Commits" tab, adding the tooltip
  image
- Workflow detail pages, both the "Creation date" and the "Last execution" fields (tbh I don't have a strong preference for this case, but consistency might be enough to justify changing them too)
  image

lorenzo-cavazzi (dismissed):

lorenzo-cavazzi (approved):
Lgtm! :rocket: 
Thank you for taking care of this! :pray:  I neglected glitches on dates for too long

PR #2534, opened 2023-05-17
Update the session class selector styles

leafty:
Updated the session class selector styling:

leafty dev renku ch_projects_johann thiebaut_flights-tutorial-2022-04-19_sessions_new_useFakeResourcePools=1

leafty dev renku ch_projects_johann thiebaut_flights-tutorial-2022-04-19_sessions_new_useFakeResourcePools=1 (1)

leafty dev renku ch_projects_johann thiebaut_flights-tutorial-2022-04-19_sessions_new_useFakeResourcePools=1 (3)

Also, on error display this message:
leafty dev renku ch_projects_johann thiebaut_flights-tutorial-2022-04-19_sessions_new (1)

### DIFF
--- a/client/src/notebooks/components/SessionClassOption.module.scss
+++ b/client/src/notebooks/components/SessionClassOption.module.scss
@@ -31,47 +31,46 @@
 
 .option,
 .singleValue {
-  grid-template-columns:
-    minmax(min-content, 6fr)
-    1fr 4ch
-    1fr 7ch
-    1fr 8ch
-    1fr 4ch;
+  // Two lines on small viewports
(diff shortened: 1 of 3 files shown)


lorenzo-cavazzi (approved):
Lgtm! :rocket: 

I really like the 2 rows design you implemented for the low-width window. I feel we might expand it to the medium-sized window to prevent very verbose names from taking too much vertical space (see screenshot). Since that happens only with very specific window sizes, we can merge the PR as is, but let's keep that solution in mind if ppl complain about messed-up output with long names
image

PR #2541, opened 2023-05-24
fix(client): update Time.js to not throw an error on invalid dates

leafty:
It seems that `tests/cypress/e2e/workflows.spec.ts` often fails due to fixture data not providing datetimes when expected.

/deploy #persist #cypress renku=0000-ui-3.5.0


lorenzo-cavazzi (approved):
Lgtm.
Acceptance tests were failing yesterday evening, but it doesn't seem related to the changes here. Let's see if they pass now

PR #2543, opened 2023-05-24
fix(client): remove redundant template fetching function

### LINKED ISSUE #2532: Error when creating a new project
## Description 

When opening the "new project" page, we get an error (check the [sentry issue](https://sentry.dev.renku.ch/organizations/sentry/issues/1677/?referrer=github_integration) for details).

The error says:
```
UnhandledRejection: Non-Error promise rejection captured with value: No project templates are available in this RenkuLab deployment. Please notify a RenkuLab administrator and ask them to configure a project template repository.
```

In short, it's triggered by the `getTemplates` function in the "client/src/project/new/ProjectNew.state.js" file invoked without any `sources`. Apparently, after this error, the function is invoked again correctly.
Double-check why there is this initial step and prevent it.

## Reproduce

Click on the "new project" page and it will trigger the error.

leafty:
Also, migrate `Template.js` to TypeScript + pure functional component.

Fixes #2532.

/deploy renku-core=bugfix/2124-fetch-template-errors renku=2315ui-project-status #persist #cypress


ciyer (approved):
This works well. All the tests pass, and I tried it out manually as well.

PR #2549, opened 2023-05-25
feat: fix add dataset to project functionality

ciyer:
The feature to add a dataset to an existing project

image

image

Is broken. This PR fixes that problem.

/deploy renku-core=develop renku=2315ui-project-status extra-values=core.sentry.dsn=https://0cc02f3a02d3474d9f7e90915873ab4d@sentry.dev.renku.ch/2,core.sentry.environment=renku-dev,core.sentry.enabled=true #persist #notest #cypress


tests/cypress/support/renkulab-fixtures/datasets.ts
```diff
@@ -125,6 +125,7 @@ function Datasets<T extends FixturesConstructor>(Parent: T) {
     ) {
       const fixture = this.useMockedData ? { fixture: resultFile } : undefined;
       cy.intercept("/ui-server/api/renku/*/datasets.import", fixture).as(name);
+      cy.intercept("/ui-server/api/renku/datasets.import", fixture).as(name);
```
leafty, line 128: Should the previous `.intercetp()` be removed?

client/src/App.js
```diff
@@ -46,7 +46,7 @@ import { StyleGuide } from "./styleguide";
 import { Url } from "./utils/helpers/url";
 import { Unavailable } from "./Maintenance";
 import { Loader } from "./components/Loader";
-import { AddDataset } from "./dataset/addtoproject/DatasetAdd.container";
+import DatasetAddToProject from "./dataset/addtoproject/DatasetAddToProject";
```
leafty, line 49: Non-blocking note: because default imports/exports can be renamed at the import site, they may make components less searchable with simple text matching. This is why I would only use named exports.

client/src/dataset/addtoproject/DatasetAdd-types.ts
```diff
@@ -0,0 +1,34 @@
+import { IDataset } from "../../features/project/Project";
+
+interface AddDatasetDataset extends IDataset {
```
leafty, line 3: Exports are more readable when done at the definition site.

```suggestion
export interface AddDatasetDataset extends IDataset {
```

client/src/dataset/addtoproject/DatasetAdd-types.ts
```diff
@@ -0,0 +1,34 @@
...
+  fetchError: unknown;
+  project?: ExistingProject;
+}
+type AddDatasetHandlers = {
```
leafty, line 7:
```suggestion
export type AddDatasetHandlers = {
```

client/src/dataset/addtoproject/DatasetAdd-types.ts
```diff
@@ -0,0 +1,34 @@
...
+  setCurrentStatus: (status: AddDatasetStatus | null) => void;
+  submitCallback: (project: SubmitProject) => void;
+  validateProject: (project: SubmitProject, isDatasetValid: boolean) => void;
+};
+
+type AddDatasetStatus = { status: string; text?: string };
```
leafty, line 13:
```suggestion
export type AddDatasetStatus = { status: string; text?: string };
```

client/src/dataset/addtoproject/DatasetAdd-types.ts
```diff
@@ -0,0 +1,34 @@
...
+
+type SubmitProject = {
```
leafty, line 15:
```suggestion
export type SubmitProject = {
```

client/src/dataset/addtoproject/DatasetAdd-types.ts
```diff
@@ -0,0 +1,34 @@
...
+  name: string;
+  value: string;
+};
+
+interface ExistingProject extends SubmitProject {
```
leafty, line 20:
```suggestion
export interface ExistingProject extends SubmitProject {
```

client/src/dataset/addtoproject/DatasetAdd-types.ts
```diff
@@ -0,0 +1,34 @@
...
+  path_with_namespace: string;
+}
+
+export type {
+  AddDatasetDataset,
+  AddDatasetHandlers,
+  AddDatasetStatus,
+  ExistingProject,
+  SubmitProject,
+};
```
leafty, line 34:
```suggestion
```

client/src/dataset/addtoproject/DatasetAdd-types.ts
```diff
@@ -0,0 +1,34 @@
+import { IDataset } from "../../features/project/Project";
```
leafty, line 1: I think I would use the filename `DatasetAdd.types.ts` or `DatasetAddTypes.ts`.

leafty (approved):
LGTM 👍 

Some minor comments.

tests/cypress/support/renkulab-fixtures/datasets.ts
ciyer: Both are necessary, since the version that is used depends on the situation. If it is an older project, then the first of those is used, if it is a new project, then the second.

client/src/App.js
ciyer: This would be a good topic for the coding guidelines: we should decide as a team on conventions for imports/exports.

client/src/dataset/addtoproject/DatasetAdd-types.ts
ciyer: The comment above applies here as well.

PR #2550, opened 2023-05-26
feat(client): add a delete button on project settings

### LINKED ISSUE #2416: Add a Delete button on project pages
Now that we use the KG also for searching projects, having an out-of-sync state due to manually deleted projects is worse than ever.

KG implemented a `DELETE /project/<id>` API that deletes the project from GitLab and from our backend (Reference: https://github.com/SwissDataScienceCenter/renku-graph/tree/development/knowledge-graph#delete-knowledge-graphprojectsnamespacename).
We should add a button in the UI.

We might want to put it in a non-too-visible location (settings tab?) and adopt a similar style to GitHub: red button, opening a "warning modal" on click, asking to enter the project name as a confirmation.

We need some basic design for that.

leafty:
Fixes #2416.


client/src/features/project/components/ProjectSettingsGeneralDeleteProject.tsx
```diff
@@ -0,0 +1,224 @@
...
+  return (
+    <Card data-cy="project-settings-general-delete-project">
+      <CardBody>
+        <CardTitle className="fs-6 lh-base text-danger">
+          Delete project
+        </CardTitle>
+        {result.error && <ShowError error={result.error} />}
+        <p className="mb-0">
+          Deleting the project will remove its repository and all related
+          resources, including datasets(?), issues and merge requests.
```
ciyer, line 111:
```suggestion
          resources, including datasets, issues and merge requests.
```

client/src/features/project/components/ProjectSettingsGeneralDeleteProject.tsx
```diff
@@ -0,0 +1,224 @@
...
+        </p>
+        <p>Learn more</p>
```
ciyer, line 113: Not sure we need to explain this in greater detail. We can check with the product team to see if they think this is necessary.

ciyer (changes requested):
This looks good!

I think there are a couple of elements in the code that should be cleaned up/finalized.

I also think the notification that appears could use a little padding. The text is a bit too close to the border, and it does not look great.

image

leafty:
New notification style:

Screenshot 2023-06-02 at 12 00 40

client/src/features/project/components/ProjectSettingsGeneralDeleteProject.tsx
leafty: Removed in 6c76637.

ciyer (approved):
This is very convenient!

PR #2558, opened 2023-05-31
refactor(client): remove all mentions of `border-rk-light`

leafty:
The CSS class `border-rk-light` has no effect.


ciyer (approved):
Good cleanup!

PR #2563, opened 2023-06-01
feat(client): synchronize all loading spinners

leafty:
All spinners will now have syncrhonized CSS animation when rendered.


ciyer (approved):
Nice fix! It makes the UI look less nervous.

PR #2567, opened 2023-06-02
fix(tests): fix flaky session test

leafty:
Wait for the session health response after clicking the save button.

Example run: https://github.com/SwissDataScienceCenter/renku-ui/actions/runs/5155141273/jobs/9284452650


ciyer (approved):
This can't hurt.

PR #2568, opened 2023-06-02
chore(ci/cd): update actions/setup-node to v3

leafty:
The v1 action is outdated and generated deprecation warnings from GitHub actions.


ciyer (approved):
The actions that this would affect still pass, so looks good.

PR #2574, opened 2023-06-05
fix(client): enable start session button after load

leafty:
Do not let the user start a session until the options have been populated from the project settings or defaults.

### DIFF
--- a/client/src/notebooks/NotebookStart.present.js
+++ b/client/src/notebooks/NotebookStart.present.js
@@ -203,6 +203,11 @@ function StartNotebookServer(props) {
   const { branch, commit, namespace, project } = props.filters;
   const location = useLocation();
 
+  const sessionClass = useSelector(
+    (state) => state.startSessionOptions.sessionClass
+  );
+  const hasSessionClass = sessionClass != 0;
+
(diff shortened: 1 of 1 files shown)


client/src/notebooks/NotebookStart.present.js
```diff
@@ -203,6 +203,11 @@ function StartNotebookServer(props) {
   const { branch, commit, namespace, project } = props.filters;
   const location = useLocation();
 
+  const sessionClass = useSelector(
+    (state) => state.startSessionOptions.sessionClass
+  );
+  const hasSessionClass = sessionClass != 0;
```
ciyer, line 209:
```suggestion
  const hasSessionClass = sessionClass !== 0;
```

ciyer (changes requested):
Just a minor suggestion about equality testing.

client/src/notebooks/NotebookStart.present.js
leafty: Note that in TypeScript, the compiler would error on `<string> != <number>`, so `==` vs `===` is less of an issue in TypeScript code.

ciyer (approved):
LGTM!

PR #2579, opened 2023-06-06
fix(client): fix the workflow view crashing the UI in some cases

### LINKED ISSUE #2578: Last execution tooltip breaks the UI in some cases
See example: https://renkulab.io/projects/patrick.paitz/neurips-renku-seismic-example/workflows

```
DOMException: Failed to execute 'querySelectorAll' on 'Document': 'duration-/plans/a61431e0b6f495e1e8900b67f8c6617a' is not a valid selector.
```

leafty:
Fixes #2578.

The "last execution" tooltip in the "Workflows" view can sometimes crash the UI because the worflow id is not sanitized to create a valid CSS id.

This code change makes use of a `ref` instead of assigning a CSS id.

Also, the component has been refactored to make the rendering logic more readable.

### DIFF
--- a/client/src/components/entities/Executions.tsx
+++ b/client/src/components/entities/Executions.tsx
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-import React, { FunctionComponent, ReactNode, forwardRef, useRef } from "react";
+import React, { ReactNode, forwardRef, useRef } from "react";
 import cx from "classnames";
 import { UncontrolledTooltip } from "reactstrap";
 import Time from "../../utils/helpers/Time";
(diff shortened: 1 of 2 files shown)


client/src/components/entities/Executions.tsx
```diff
@@ -33,17 +33,22 @@ export interface EntityExecutionsProps {
   workflowId: string;
 }
 
-function EntityExecutions({
+export const EntityExecutions: FunctionComponent<EntityExecutionsProps> = ({
```
ciyer, line 36: A general comment on this: I think how we declare components would be a good topic for the style discussion. I am not aware of any clear advantages of one over the other (e.g., in terms of better errors from the type system). Given that, I think we should hold back on changing the style of existing code until we have that discussion.

ciyer (approved):
The refactoring is nice, and the code is easier to follow now. I was not able to check if the exact bug is fixed, since I do not know how exactly to recreate it, but I did look at many projects with workflows and the code changes here do not change how those appear, so in any case, this is an improvement.

ciyer (approved):
👍

ciyer (approved):
👍

PR #2582, opened 2023-06-06
Remove form generator from dataset import

ciyer:
/deploy #persist #cypress


client/src/project/datasets/import/DatasetImport.tsx
```diff
@@ -0,0 +1,439 @@
...
+  accessLevel: number;
+  client: DatasetImportClient;
+  history: { push: (arg: unknown) => void };
+  httpProjectUrl: string;
+  fetchDatasets: (force: boolean, versionUrl: string) => void;
+  projectPathWithNamespace: string;
+  location: { pathname: string };
+  toggleNewDataset: DatasetImportFormProps["toggleNewDataset"];
+};
+function DatasetImport(props: DatasetImportProps) {
```
leafty, line 389: Since this component is the default export, it should in principle be declared first, or very close to the top of the file.

tests/cypress/e2e/newDataset.spec.ts
```diff
@@ -171,3 +171,56 @@ describe("Project new dataset", () => {
...
+    cy.wait("@getProject");
+    cy.contains("Import").click();
+    cy.get_cy("input-uri")
+      .click()
+      .type("https://www.doi.org/10.7910/DVN/WTZS4K");
+    cy.get_cy("submit-button").click();
+    cy.wait("@importToProject");
+    cy.contains("Creating Dataset...").should("be.visible");
+    cy.wait("@importJobCompleted", { timeout: 20000 });
+    cy.contains("Datasets List").should("be.visible");
```
leafty, line 207: Would it be possible to check that the new dataset appears in the list? Something like `cy.contains("<Dataset title").should("be.visible")`?

ciyer: I think the `cy.wait("@datasetList")` will achieve this.

client/src/project/datasets/import/DatasetImport.tsx
```diff
@@ -0,0 +1,439 @@
...
+    return;
+  }
+
+  const job_id = response.data.result.job_id;
+
+  // Monitor job status
+  await new Promise<void>((resolve) => {
+    let cont = 0;
+    const monitorJob = setInterval(() => {
+      client.getJobStatus(job_id, versionUrl).then((job) => {
```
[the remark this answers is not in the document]
leafty: The counting logic should be:

```ts
  .finally(() => {
    pollCount++;
        if (pollCount >= 50) {
          handlers.setSubmitLoader({ value: false, text: "" });
          handlers.setServerErrors(ImportStateMessage.TOO_LONG);
          clearInterval(monitorJob);
          resolve();
          return;
        }
  })
```

and remove it from `.then()`.

This means that any error / network issue from the client will not make this code poll forever.

leafty (changes requested):
I think a `.catch()` or `.finally()` branch should still be added for `client.getJobStatus()`. See code suggestion.

ciyer:
> I think a `.catch()` or `.finally()` branch should still be added for `client.getJobStatus()`. See code suggestion.

You are right, thanks for pointing this out.

leafty (approved):

PR #2591, opened 2023-06-07
fix(ci/cd): fix the test action self-cancelling on master

### LINKED ISSUE #2590: Test workflow is self-cancelling
Example: https://github.com/SwissDataScienceCenter/renku-ui/actions/runs/5196564496/jobs/9370462019

The "cleanup-runs" step is frequently cancelling the latest workflow which should run to completion.

Solution: use the action concurrency option, e.g. https://github.com/SwissDataScienceCenter/renku/blob/master/.github/workflows/pull-request-test.yml.

leafty:
Fixes #2590.

### DIFF
--- a/.github/workflows/test-and-ci.yml
+++ b/.github/workflows/test-and-ci.yml
@@ -1,16 +1,13 @@
 name: Test and CI
 
-on: [push]
+on:
+  push:
+
+concurrency:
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi (approved):
I'm unfamiliar with the github actions `concurrency` command; after quickly reading the reference, the changes here makes sense and the other cleanup actions seems to be obsolete :+1:

PR #2596, opened 2023-06-08
fix(client): set autocomplete to off for the delete project confirmation box

leafty:
The autocomplete suggestions offered by the browser are not useful (other project names which were deleted).

/deploy renku-core=develop renku=2315ui-project-status #persist #cypress


ciyer (approved):
👍

PR #2597, opened 2023-06-08
More robust handling of markdown to prevent crashes

ciyer:
# Testing

Compare
* https://renku-ci-ui-2597.dev.renku.ch/projects/cramakri/markdown-testing
* https://dev.renku.ch/projects/cramakri/markdown-testing

/deploy #persist #cypress


client/src/components/markdown/RenkuMarkdownWithPathTranslation.tsx
```diff
@@ -28,30 +29,72 @@ const REF_TYPES = {
   FILE_PREV: "file", // with ! in the front
 };
 
-const getFileExtension = (file_path) => {
-  if (!file_path) return null;
+type IFileRef = {
```
leafty, line 32: Naming is not conveying what the type is.

```suggestion
type FileReference = {
```

client/src/components/markdown/RenkuMarkdownWithPathTranslation.tsx
```diff
@@ -28,30 +29,72 @@ const REF_TYPES = {
...
+  projectPathWithNamespace: string;
+  /** If true, render the output as a single line without line breaks */
+  singleLine?: boolean;
+  /** Any styles to apply */
+  style?: object;
 };
 
-const fileIsImage = (file_path) =>
-  IMAGE_EXTENSIONS.indexOf(getFileExtension(file_path)) >= 0;
+const getFileExtension = (file_path?: string) => {
```
leafty, line 65: `arg?` should be reserved for optional arguments.

```suggestion
const getFileExtension = (file_path: string | undefined) => {
```

client/src/components/markdown/RenkuMarkdownWithPathTranslation.tsx
```diff
@@ -28,30 +29,72 @@ const REF_TYPES = {
...
+  if (file_path == null) return undefined;
+
+  if (file_path.match(/\.(.*)/) === null) return undefined;
+  const components = file_path.split(".");
+  if (components == null || components.length < 1) return undefined;
```
leafty, line 70: This seems unnecessary, `split()` will always produce an array of length >= 1. (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split)

```suggestion
  const components = file_path.split(".");
  if (components.length < 1) return undefined;
```

client/src/components/markdown/RenkuMarkdownWithPathTranslation.tsx
```diff
@@ -28,30 +29,72 @@ const REF_TYPES = {
...
+  return components.pop()?.toLowerCase();
+};
+
+const fileIsImage = (file_path?: string) => {
```
leafty, line 74:
```suggestion
const fileIsImage = (file_path: string | undefined) => {
```

client/src/components/markdown/RenkuMarkdownWithPathTranslation.tsx
```diff
@@ -28,30 +29,72 @@ const REF_TYPES = {
...
+  if (file_path == null) return false;
+  const ext = getFileExtension(file_path);
+  if (ext == null) return false;
+  IMAGE_EXTENSIONS.indexOf(ext) >= 0;
```
leafty, line 78: Use `.includes()` from es2016:

```suggestion
  const ext = getFileExtension(file_path);
  if (!ext) return false;
  return IMAGE_EXTENSIONS.includes(ext);
```

client/src/components/markdown/RenkuMarkdownWithPathTranslation.tsx
```diff
@@ -28,30 +29,72 @@ const REF_TYPES = {
...
+  filePathArray: string[]
+) => {
+  const filesRefs: IFileRef[] = [];
   let blockCounter = 0;
 
   const previewFiles = markdownHTML.getElementsByTagName("img");
 
-  for (let file of previewFiles) {
-    let src = fixRelativePath(file.getAttribute("src"), filePathArray);
+  for (const file of previewFiles) {
```
leafty, line 90: Ideally, this would use `.map()`:

```ts
  const filesRefs = previewFiles
  .filter((file) => !file.src.match(patterns.urlFilesRef))
  .map((file, iBlock) => {
    return {
      type: ...,
      refPath: src,
      data: undefined,
      iBlock,
      error: false,
    };
  });
```

leafty (commented):
This quick fix seems to work, only a few comments.

client/src/components/markdown/RenkuMarkdownWithPathTranslation.tsx
```diff
@@ -28,36 +29,73 @@ const REF_TYPES = {
...
+};
 
-    if (!file.src.match(patterns.urlFilesRef)) {
-      filesRefs.push({
+const getFilesRefs = (
+  markdownHTML: HTMLDivElement,
+  filePathArray: string[]
+) => {
+  const previewFiles = markdownHTML.getElementsByTagName("img");
+  const filesRefs: FileReference[] = [...previewFiles]
```
leafty, line 87: `.filter()` and `.map()` will not mutate the original array.

```suggestion
  const filesRefs: FileReference[] = previewFiles
```

ciyer: The method [getElementsByTagName](https://developer.mozilla.org/en-US/docs/Web/API/Element/getElementsByTagName) returns an [HTMLCollection](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection) (see also https://dom.spec.whatwg.org/#interface-htmlcollection), not a standard JavaScript [Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array). 

Unlike Array, HTMLCollection does not contain `filter` as part of its interface, so it is necessary to convert that object into one that understands `filter`.

leafty (approved):
LGTM thanks for adding the suggestions

PR #2599, opened 2023-06-13
refactor(client): use the recently-viewed API from KG

### LINKED ISSUE #2521: Use KG "recently-viewed" API to load content in the dashboard
## Motivation

We currently load recently visited projects from the UI server. That's too volatile since the content is stored in Redis.

## Proposal

We should use the new `GET /entities/current-user/recently-viewed` API from KG to load the projects and the datasets.
The reference is on the swagger page (now it's available [here](https://renku-kg-dev.dev.renku.ch/swagger/?urls.primaryName=knowledge%20graph#/default/get_users__userId__projects), but once merged it should be [somewhere here](https://dev.renku.ch/swagger/?urls.primaryName=knowledge%20graph))

We can also drop the corresponding API from the UI server, together with the logic to store the projects' visits in Redis (look at the `lastProjectsMiddleware` middleware in the "server" code -- probably others bits should be dropped as well)

leafty:
Fixes #2521.

Use the recently-viewed API from the knowledge graph to feed data into the dashboard.


lorenzo-cavazzi:
[We decided to drop this change](https://www.notion.so/2024-01-28-eec2f6110aa24a0ca65aef22cf5566bd?pvs=4#457f0a2022e74256923ec63c46967607) since requiring projects to be indexed to show up on the dashboard might be an issue for some users.

PR #2604, opened 2023-06-16
feat(client): implement a global error boundary

### LINKED ISSUE #2498: Render an error screen instead of a blank one when the UI crashes
## Motivation

When the UI crashes, it renders a blank screen which can be confusing to users.

See for example: https://github.com/SwissDataScienceCenter/renku-ui/issues/2488.

## Proposal

Use Sentry's error boundary component to render a fallback component (e.g. "An error occurred") which would let the user either reload the page or return to the home page of the UI.

See: https://docs.sentry.io/platforms/javascript/guides/react/features/error-boundary/.

leafty:
Fixes #2498.

Add an global error boundary to the UI.

localhost_3000_style-guide

See example here: https://renku-ci-ui-2604.dev.renku.ch/style-guide
(Click the button)

/deploy #persist #cypress


client/src/styleguide/StyleGuide.js
```diff
@@ -36,13 +36,28 @@ import { TimeCaption } from "../components/TimeCaption";
...
       </p>
+      <div>
+        <button
+          onClick={() => {
+            setError(new Error("ouch!"));
+          }}
+        >
+          Do not click me
+        </button>
+      </div>
```
leafty, line 60: Changes on this file will be reverted before merging. This is used to have an easy way to show the error boundary.

lorenzo-cavazzi (commented):
Great addition! No more blank pages :rocket: 
I can approve after removing the fake error generator

lorenzo-cavazzi (dismissed):

lorenzo-cavazzi (approved):

PR #2606, opened 2023-06-19
fix: allow up to 100 namespaces and sort them

ciyer:
Some users have more than 20 namespaces. Increase the limit used when showing the namespace list in the project creation dropdown.

/deploy #persist #cypress


lorenzo-cavazzi (approved):
This works as advertised. Since it's intended as a quick fix, we can merge it as is.

Ideally, it would be nice to move the `Namespace` type to a definition file and parametrize both `per_page` and `owned_only` since we are touching that code already.

rokroskar:
why not use pagination properly? This is just going to break for someone else later.

ciyer:
> why not use pagination properly? This is just going to break for someone else later.

Because this allows us to get a quick fix for the immediate problem, but longer-term, then UX for this component really needs to be rethought.

Currently, we assume that users have a small number of namespaces they can access, and we show them a full list of their namespaces.

image

If the user has access to so many namespaces that pagination is necessary, this component will be unusable anyway. Instead, we need to do something like show a list of likely namespaces the user might want to pick (for example, recently accessed ones), but let them type input, search against that and show them a list of matches, which they can pick from.

PR #2607, opened 2023-06-19
fix: dataset delete is possible again

ciyer:
Deleting a dataset was broken. This should fix that feature again.

/deploy #persist #cypress

### DIFF
--- a/client/src/project/datasets/delete/DeleteDataset.present.js
+++ b/client/src/project/datasets/delete/DeleteDataset.present.js
@@ -52,12 +52,39 @@ function ModalContent({
     );
   }
 
+  if (submitLoader != null && submitLoader.value) {
+    return (
+      <div className="d-flex flex-row-reverse">
+        <FormText color="primary">
+          <Loader size={16} inline margin={2} />
+          {submitLoader.text}
+          <br />
+        </FormText>
+      </div>
+    );
+  }
+
   return (
     <Col>
       <p>
         Are you sure you want to delete dataset <strong>{dataset.name}</strong>?
       </p>
-      <>
+      <div className="d-flex flex-row-reverse">
+        <Button
+          type="submit"
+          onClick={deleteDataset}
+          disabled={submitLoader.value}
+          className="mt-1 btn-rk-pink"
+        >
+          Delete dataset
+        </Button>
+        <Button
+          disabled={submitLoader.value}
+          className="mt-1 me-2 btn-outline-rk-pink"
+          onClick={closeModal}
+        >
+          Cancel
+        </Button>
         {submitLoader != null && submitLoader.value && (
           <FormText color="primary">
             <Loader size={16} inline margin={2} />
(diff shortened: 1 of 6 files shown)


client/src/project/datasets/delete/DeleteDataset.present.js
```diff
@@ -37,55 +37,62 @@ import {
...
+        Are you sure you want to delete dataset <strong>{dataset.name}</strong>?
+      </p>
+      <Fragment>
+        {submitLoader !== undefined && submitLoader.value ? (
+          <FormText color="primary">
+            <Loader size={16} inline margin={2} />
+            {submitLoader.text}
+            <br />
+          </FormText>
+        ) : null}
```
leafty, line 67:
```suggestion
        {submitLoader != null && submitLoader.value && (
          <FormText color="primary">
            <Loader size={16} inline margin={2} />
            {submitLoader.text}
            <br />
          </FormText>
        )}
```

tests/cypress/e2e/projectDatasets.spec.ts
```diff
@@ -115,6 +115,31 @@ describe("Project dataset", () => {
...
+    cy.wait("@getProject");
+    cy.wait("@datasetList")
+      .its("response.body")
+      .then((data) => {
+        const datasets = data.result.datasets;
+        const totalDatasets = datasets?.length;
+        cy.get_cy("list-card").should("have.length", totalDatasets);
+        const dataset = datasets[0];
+        const datasetIdentifier = dataset.identifier.replace(/-/g, "");
+        fixtures.datasetById(datasetIdentifier, "getDatasetById0");
```
leafty, line 129: Would there be a name clash with `getDatasetById`?

tests/cypress/e2e/projectDatasets.spec.ts
```diff
@@ -115,6 +115,31 @@ describe("Project dataset", () => {
...
+        cy.wait("@getDatasetById0");
+        cy.get_cy("delete-dataset-button").should("exist").click();
+        fixtures.datasetsRemove(datasetIdentifier);
+        cy.get("button").contains("Delete dataset").should("exist").click();
+        // dataset should be deleted
+        cy.wait("@datasetsRemove");
+        // list should be refreshed
+        cy.wait("@datasetList");
+        cy.contains("Datasets List").should("be.visible");
+      });
```
leafty, line 140: You can maybe add:
```ts
cy.get_cy("list-card-title").contains(dataset.title).should("not.exist");
```

leafty (approved):
LGTM

Code looks good! A few non-blocking code suggestions below.

tests/cypress/e2e/projectDatasets.spec.ts
ciyer: I was just reusing the structure established by an existing test, but I think this should be fine.

tests/cypress/e2e/projectDatasets.spec.ts
ciyer: That will actually fail given the way the tests are currently written. The test re-uses the result of the request to list datasets, so for that to work, it would be necessary to return the list with the deleted dataset removed, but I do not think that will improve the test, since the point of the test is to make sure that the `Delete Dataset` button works as intended (it invokes the endpoint on the backend and redirects to the dataset list). The dataset list is already tested elsewhere.

client/src/project/datasets/delete/DeleteDataset.present.js
```diff
@@ -57,14 +57,14 @@ function ModalContent({
       <p>
         Are you sure you want to delete dataset <strong>{dataset.name}</strong>?
       </p>
-      <Fragment>
-        {submitLoader !== undefined && submitLoader.value ? (
+      <>
```
leafty, line 60: Actually, the fragment does not make sense here. Sorry for realizing only now.

leafty (approved):

PR #2611, opened 2023-06-20
fix(client): improve showing the error details for sessions

### LINKED ISSUE #2504: Confusing instructions on how to see the error for a failed session
We have discussed this previously and we improved a lot in terms of how we show the errors about failed sessions to users.

I think we can do better.

This is what happened to me today:
1. I launched a session that I knew was going to fail on purpose
2. I was looking for the error notification but I saw this text instead: 
image

4. After this I tried multiple times to click on the red text that says `Error` only to realize that the `error icon` refers to the red square icon.
image

If this happens to me I think it is likely it happens to other users as well.

I have two proposals:
1. Simplify the text to say `Error: click here to see it` and make the whole text or `here` clickable and show the error. We can also keep the behaviour of clicking the red square icon.
2. Show the whole error message instead of the `Error (Click the error icon for details.)` text. I think this is ideal and it would enable users to immediately see the problem with their session. We can make the error message collapsible if it is bigger than a certain number of characters. So we can always show the first lets say 300 characters and after that show a little > or similar icon to expand and show the whole message. But almost all message the notebook service shows for a failed sessions are one simple sentence. So they would fit in the card.

leafty:
Fixes #2504.

/deploy #persist #notest

### DIFF
--- a/client/src/notebooks/components/SessionListStatus.tsx
+++ b/client/src/notebooks/components/SessionListStatus.tsx
@@ -82,9 +82,10 @@ function SessionListRowStatusExtraDetails({
   if (!details.message) return null;
   if (status == "failed")
     return (
-      <span className="text-muted">
-        &nbsp;(Click the error icon for details.)
-      </span>
+      <>
+        {" "}
+        <span className="text-muted">(Click the error icon for details.)</span>
+      </>
     );
   return (
     <>


lorenzo-cavazzi (approved):
Lgtm :rocket:

PR #2615, opened 2023-06-21
feat(client): update session project settings for resource pools

leafty:
Remove the session class selector and add input for existing settings.

Screenshot 2023-06-21 at 15 39 34

### DIFF
--- a/client/src/features/project/Project.d.ts
+++ b/client/src/features/project/Project.d.ts
@@ -309,7 +309,7 @@ export interface ProjectConfigSection {
 
     legacyConfig?: {
       cpuRequest?: number;
-      memoryRequest?: string;
+      memoryRequest?: number;
       gpuRequest?: number;
     };
     unknownConfig?: {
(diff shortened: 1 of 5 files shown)


leafty:
Updated style:

image

lorenzo-cavazzi (changes requested):
This looks generally good!

I suggest removing `max` on the inputs.
Apart from that, I would update the warning message to clarify what's the problem when we can't pick the flavor. As a user, I would need help understanding exactly _where_ the problem is.

Instead of `The session cannot be configured exactly as requested for this project.` we can explicitly mention the requirements, as in `The project requires 4 CPUs and there is no Session class that matches it`

image

leafty:
> This looks generally good!
> 
> I suggest removing `max` on the inputs. Apart from that, I would update the warning message to clarify what's the problem when we can't pick the flavor. As a user, I would need help understanding exactly _where_ the problem is.
> 
> Instead of `The session cannot be configured exactly as requested for this project.` we can explicitly mention the requirements, as in `The project requires 4 CPUs and there is no Session class that matches it`

I am not updating the "Start with options" page in this PR because it needs to be updated to support the CRC service telling the user which session classes meet their compute resource requirements.

lorenzo-cavazzi (approved):

PR #2619, opened 2023-06-23
chore: expand the coding guidelines to cover more topics

### LINKED ISSUE #2575: Expand the coding guidelines to cover more topics
The coding guidelines should be expanded over time to cover more topics and help us achieve more consistent and more readable code.

The suggestion would be to split the guidelines into their own file, e.g. `Coding_Guidelines.md`, and expand them to cover more topics.

Topics to cover (list is not exhaustive):
- [x] do not construct CSS class names
- [x] standardize functional component code style
- [x] avoid `condition ? true : false`
- [x] avoid nested `if/else`
- [x] exports at the definition site
- [x] named/default exports
- [ ] ordering code, following the step-down rule

leafty:
Fixes #2575.

Move the coding guidelines to a new file and expand them with an agreed set of rules.


lorenzo-cavazzi:
Should we consider merging this and opening follow-up PRs to expand it later?

lorenzo-cavazzi (approved):
Lgtm :rocket: 
We can keep expanding the guidelines in follow-up PRs

PR #2620, opened 2023-06-26
feat(client): update session start options and flow for resource pools 

leafty:
* Update the session start options to show which session classes match the compute requirements.
* Update the quick start flow to auto-select a suitable session class or fall back to the option screen if non is suitable.

Start options:
renku-ci-ui-2620 dev renku ch_projects_johann thiebaut1_another-playground-project_sessions_new
renku-ci-ui-2620 dev renku ch_projects_johann thiebaut1_another-playground-project_sessions_new (1)
renku-ci-ui-2620 dev renku ch_projects_johann thiebaut1_another-playground-project_sessions_new (2)

Quick start not finding a suitable session class:
renku-ci-ui-2620 dev renku ch_projects_johann thiebaut1_another-playground-project_sessions_new (4)

/deploy extra-values=crc.image.tag=sha-d9cef4d renku=crac-service-helm-chart renku-notebooks=feat-work-with-crac #persist #cypress

### DIFF
--- a/client/src/features/dataServices/dataServices.d.ts
+++ b/client/src/features/dataServices/dataServices.d.ts
@@ -40,6 +40,8 @@ export interface ResourceClass {
   default_storage: number;
 
   default: boolean;
+
+  matches?: boolean;
 }
 
 export interface Resources {
--- a/client/src/features/dataServices/dataServices.d.ts
+++ b/client/src/features/dataServices/dataServices.d.ts
@@ -50,3 +50,10 @@ export interface Resources {
   gpu: number;
   storage: number;
 }
+
+export interface ResourcePoolsQueryParams {
+  cpuRequest?: number;
+  gpuRequest?: number;
+  memoryRequest?: number;
+  storageRequest?: number;
+}
--- a/client/src/features/dataServices/dataServices.d.ts
+++ b/client/src/features/dataServices/dataServices.d.ts
@@ -21,6 +21,9 @@ export interface ResourcePool {
   name: string;
   classes: ResourceClass[];
   quota: Resources;
+  // TODO(@leafty): remove "?"
+  default?: boolean;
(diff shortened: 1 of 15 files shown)


lorenzo-cavazzi (changes requested):
The code looks good!

Thanks for updating this! I know it's a pain to read through functions like the hectic `triggerAutoStart`.
Hopefully, rewriting and simplifying some code (together with the changes for autosaves) will save us headaches when going through Notebooks code again.

I've found a couple of UX details to adjust:

**Project settings**
We should hide or disable the button to reset the value when no value is set. Otherwise, we get this error when clicking it.
image

**Start with options**
I would move the error message `This session class does not match the compute requirement` just after the dropdown. This has 2 advantages:
1.  The dropdown element doesn't move while selecting an invalid element (see gif)
    Peek 2023-06-29 17-24
2. The message isn't above the bigger warning when there is no valid class available
    image

leafty:
> **Project settings**
> We should hide or disable the button to reset the value when no value is set. Otherwise, we get this error when clicking it.

Will do, but this seems to be an actual bug from `renku-core`.

> **Start with options**
> I would move the error message `This session class does not match the compute requirement` just after the dropdown.

👍

lorenzo-cavazzi:
> > **Project settings**
> > We should hide or disable the button to reset the value when no value is set. Otherwise, we get this error when clicking it.
> 
> Will do, but this seems to be an actual bug from `renku-core`.

Fair point. Should we also open an issue in renku-python?

lorenzo-cavazzi (approved):
Lgtm :rocket: 
Let's merge it to the main feature branch

leafty:
> Fair point. Should we also open an issue in renku-python?

https://github.com/SwissDataScienceCenter/renku-python/issues/3544

PR #2621, opened 2023-06-28
build(server): set up prometheus metrics

lorenzo-cavazzi:
Set up Prometheus

/deploy #persist

### DIFF
--- a/server/src/config.ts
+++ b/server/src/config.ts
@@ -99,10 +99,16 @@ const WEBSOCKET = {
   timeoutActivationStatus: 120, // in minutes
 };
 
+const PROMETHEUS = {
+  enabled: [true, "true"].includes(process.env.PROMETHEUS_ENABLED),
+  path: "/metrics",
+};
+
 const config = {
   auth: AUTHENTICATION,
   data: DATA,
   deployment: DEPLOYMENT,
+  prometheus: PROMETHEUS,
   redis: REDIS,
   routes: ROUTES,
   sentry: SENTRY,
--- a/server/src/config.ts
+++ b/server/src/config.ts
@@ -45,12 +45,16 @@ const DEPLOYMENT = {
 };
 
 const SENTRY = {
-  enabled: [true, "true"].includes(process.env.SENTRY_ENABLED),
+  enabled: ["true", "1"].includes(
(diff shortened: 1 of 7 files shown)


lorenzo-cavazzi:
- Move the setup to a separate function specific for setting up monitoring.
- Activate `includePath` and set it up https://www.npmjs.com/package/express-prom-bundle
- We need to setup the annotations on the pod. E.G: this is the current one, ofc to adapt
```
metadata:
  annotations:
    prometheus.io/path: /ui-server/metrics
    prometheus.io/port: '8080'
    prometheus.io/scrape: 'true'
```

server/src/config.ts
```diff
@@ -99,10 +99,16 @@ const WEBSOCKET = {
   timeoutActivationStatus: 120, // in minutes
 };
 
+const PROMETHEUS = {
+  enabled: [true, "true"].includes(process.env.PROMETHEUS_ENABLED),
```
ciyer, line 103: I believe process.env values will always be strings, so no need to check for the boolean, but maybe lowercase and/or allow 1 or "yes". 🤷

server/src/utils/prometheus/prometheus.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+import express from "express";
+import promBundle from "express-prom-bundle";
+
+import config from "../../config";
+import logger from "../../logger";
+
+export function initializePrometheus(app: express.Application): void {
+  if (config.prometheus.enabled === false) {
```
ciyer, line 26: `config.prometheus` is a boolean, so I think this is simpler. 
```suggestion
  if (!config.prometheus.enabled) {
```

ciyer (changes requested):
It will be nice to be able to access Prometheus metrics! I have some minor code suggestions.

server/src/config.ts
lorenzo-cavazzi: Fair point, I guess "1" is also pretty common. What about using this?

```
enabled: ["true", "1"].includes(
    (process.env.PROMETHEUS_ENABLED ?? "").toLowerCase()
),
```

ciyer (approved):
LGTM! 🎉

PR #2627, opened 2023-06-30
fix(client): fix `versionedUrlEndpoint()`

leafty:
This can cause issue for projects using an unsupported metadata version.

/deploy #persist #cypress

### DIFF
--- a/client/src/features/project/projectCoreApi.ts
+++ b/client/src/features/project/projectCoreApi.ts
@@ -34,8 +34,12 @@ function versionedUrlEndpoint(
   endpoint: string,
   versionUrl: string | null | undefined
 ) {
-  const urlPath = versionUrl ? `${versionUrl}/${endpoint}` : `/${endpoint}`;
-  return `/renku${urlPath}`;
+  const endpoint_ = endpoint.startsWith("/") ? endpoint.slice(1) : endpoint;
+  const versionUrl_ = versionUrl?.startsWith("/")
+    ? versionUrl.slice(1)
+    : versionUrl;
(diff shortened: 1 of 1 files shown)


client/src/features/project/projectCoreApi.ts
```diff
@@ -30,8 +30,11 @@ import type {
 } from "./Project.d";
 import { MigrationStartScopes } from "./projectEnums";
 
-function versionedUrlEndpoint(endpoint: string, versionUrl?: string) {
-  const urlPath = versionUrl ? `${versionUrl}/${endpoint}` : endpoint;
+function versionedUrlEndpoint(
+  endpoint: string,
+  versionUrl: string | null | undefined
+) {
+  const urlPath = versionUrl ? `${versionUrl}/${endpoint}` : `/${endpoint}`;
```
lorenzo-cavazzi, line 37: Should we make this a bit more robust by always checking whether `urlPath` starts with `/`, adding it when that's not the case?

lorenzo-cavazzi (changes requested):

client/src/features/project/projectCoreApi.ts
ciyer: I think this is a good idea. I believe there are code paths that add a `/`, so checking for that and adding a slash if necessary seems like a good strategy.

leafty: Done in fc16765.

lorenzo-cavazzi (approved):
Lgtm! Thanks for fixing this

PR #2628, opened 2023-06-30
Remove form generator completely

### LINKED ISSUE #1930: default to not expanding folders in dataset view
If a folder has many files inside (or many other folders), it can make it difficult to navigate e.g. the dataset edit or view pages. We should default to _not_ expanding folders automatically in those views.

### LINKED ISSUE #2356: Validate dataset keyword input
## Description

The input into the dataset `keywords` field is not validated: unregistered input is just ignored. So a user can type some keywords

image

and then click create/modify, and the input in the  keywords field is just silently dropped (no error, but keywords are not changed either).

### LINKED ISSUE #2622: Remove info about datasets in KG on the project datasets tab
## Motivation

We show whether every dataset is in KG on the datasets cards in the project datasets tab.
We don't need that bit anymore since there is already a warning on the same page when KG is not activated.

## Proposal

Let's remove it

image

ciyer:
* Remove form generator from dataset edit path
* Remove form generator code

Fix #2622
Fix #2356
Fix #1930 

/deploy #persist #cypress renku=renku-ui-3.10-tests extra-values=core.horizontalPodAutoscaling.minReplicas=1,core.replicaCount=1


ciyer:
Can use https://renku-ci-ui-2628.dev.renku.ch/projects/sean/another-playground-project/datasets to test the changes to when folders are opened or left closed in dataset view.

ciyer:
Should use renku branch renku-ui-3.10-tests, but seem to be having problems with deploy

lorenzo-cavazzi (changes requested):
Wow, quite some changes in this PR :rocket: 

The code looks great already! I found a few little issues -- some not necessarily related to changes you introduced but just in the code you moved around or touched:

* The cards in the datasets project tab look a bit broken
image

* The same goes for some of the Cards inside the dataset details page
image

* (Not new to this PR, but worth fixing) the "Deleting datasets" message we show in the "Delete Dataset" modal floats to the right
image

* This is not UI related, but I couldn't fully test a few operations due to the renku-core problems. E.G. importing a dataset into a project never ended even though it was imported. 
Do you mind adding these extra values to the deploy string? It should also help with the tests `extra-values=core.horizontalPodAutoscaling.minReplicas=1,core.replicaCount=1`
image

lorenzo-cavazzi (commented):
This looks ready!

One detail: does it actually fix #1930 ? I still see the folders expanded on dataset view and edit pages
image

ciyer:
Thanks for the review. All good points! I have addressed all your requests, I think. The dataset delete now looks like this: 
image

ciyer:
> One detail: does it actually fix #1930 ? I still see the folders expanded on dataset view and edit pages image


Yes, there is now a more intelligent logic to determine if the folder should be displayed open or closed. The top level is shown opened if there are less than 5 elements there.  See https://renku-ci-ui-2628.dev.renku.ch/projects/sean/another-playground-project/datasets/using_open_citation_data/

image

lorenzo-cavazzi:
> Yes, there is now a more intelligent logic to determine if the folder should be displayed open or closed. The top level is shown opened if there are less than 5 elements there. 

I see. It's a bit different than what the original issue suggested and I fear it might not work well in some scenarios (e.g. more deeply nested folder structures with fewer elements per folder) but let's give this a go

lorenzo-cavazzi (approved):
:rocket:

PR #2629, opened 2023-07-03
feat: surface backend error message on dataset list page

ciyer:
The date list page now displays the backend error if one occurs when trying to determine project support status.

/deploy #persist #cypress


lorenzo-cavazzi (approved):
Lgtm!

PR #2630, opened 2023-07-03
chore: explicitly mention telepresence is running

lorenzo-cavazzi:
Since we set up a dashboard message in the telepresence script, let's explicitly mention we are running telepresence.

### DIFF
--- a/client/run-telepresence.sh
+++ b/client/run-telepresence.sh
@@ -41,8 +41,6 @@ else
   WELCOME_PAGE=`echo "${WELCOME_MESSAGE}" | base64`
   echo "Warning! your OS has not been tested yet"
 fi
-DASHBOARD_MESSAGE_TEXT=$(echo "# Welcome to Renku! 🐸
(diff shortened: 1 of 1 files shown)


client/run-telepresence.sh
```diff
@@ -124,6 +122,9 @@ else
   fi
 fi
 
+DASHBOARD_MESSAGE_TEXT=$(echo "# Welcome to Renku! 🐸
+You are running **telepresence** on **${DEV_NAMESPACE}** 🔗" | node -e "let content = ''; process.stdin.setEncoding('utf-8').on('data', (chunk) => content += chunk).on('end', () => {console.log(JSON.stringify(content))})")
+
```
leafty, line 127: Should I move this to a node.js script? This is done to ensure the text content is encoded for JSON.

leafty (approved):
LGTM

client/run-telepresence.sh
lorenzo-cavazzi: I think we can leave it there since it's just for the internally used telepresence script.

PR #2631, opened 2023-07-03
feat: use KG to fetch project data

lorenzo-cavazzi:
Use KG to fetch project details.

It uses renku-core to update the project settings and KG to fetch the data. Since there is a delay between the 2,  I added some logic to check the status again (mind that it might miss the update when this change takes too long to be picked up by KG) and re-fetch the project's metadata when the indexing finishes.

The description on the Dashboard is still outdated but that will be fixed by #2599 since we will use data from KG instead of GitLab for the project.

Mind that this PR includes minimal changes to the new Visibility section that was moved from the older `ProjectSettings.present.js` to the newer `ProjectSettings.tsx` file.

#fix 2438


client/src/components/formlabels/FormLabels.tsx
```diff
@@ -61,11 +64,12 @@ const InputLabel = ({ text, isRequired = false }: InputLabelProps) => {
...
 
-const LoadingLabel = ({ text }: LabelProps) => {
+const LoadingLabel = ({ className, text }: LabelProps) => {
   return (
-    <FormText className="loading-label">
-      <span>{text}</span>
-      <Loader inline size={16} />
+    <FormText className={cx(className, "loading-label")}>
+      <span>
+        {text} <Loader inline size={16} />
```
leafty, line 71: We should avoid inserting extra `[space]` characters:

```suggestion
        {text}
        <Loader className="ms-1" inline size={16} />
```

client/src/components/buttons/Button.tsx
```diff
@@ -209,35 +209,33 @@ type InlineSubmitButtonProps = {
...
 }: InlineSubmitButtonProps) {
   if (isDone) return <SuccessLabel text={doneText} />;
-  if (isSubmitting) {
+  if (isSubmitting)
     return (
-      <div className="inlineSubmit">
-        {" "}
-        <Loader className="mx-1" inline size={16} />
-      </div>
+      <LoadingLabel className="feedback mx-1" text={submittingText || " "} />
```
leafty, line 228: This should render fine, without adding a space character:

```suggestion
      <LoadingLabel className="feedback mx-1" text={submittingText} />
```

client/src/components/entities/Description.tsx
```diff
@@ -16,105 +16,122 @@
...
+
+function EntityDescriptionContainer({
+  children,
+  className,
+  isHeightFixed,
+  numberLines,
+}: EntityDescriptionContainerProps) {
+  const descriptionStyles: CSSProperties = {
+    display: "-webkit-box",
+    height: isHeightFixed ? `${25 * numberLines}px` : undefined,
```
leafty, line 118: Note: this will break if the font size is not the one expected. Maybe a `lh` multiplier would work better?

https://developer.mozilla.org/en-US/docs/Web/CSS/length

client/src/components/entities/Description.tsx
```diff
@@ -16,105 +16,122 @@
...
+  numberLines: number;
+}
+
+function EntityDescriptionContainer({
+  children,
+  className,
+  isHeightFixed,
+  numberLines,
+}: EntityDescriptionContainerProps) {
+  const descriptionStyles: CSSProperties = {
```
leafty, line 116: Nitpick: name it the same way as the prop.

```suggestion
  const style: CSSProperties = {
```

client/src/components/entities/Description.tsx
```diff
@@ -16,105 +16,122 @@
...
+    textOverflow: "ellipsis",
+    WebkitBoxOrient: "vertical",
+    WebkitLineClamp: isHeightFixed ? numberLines : undefined,
+  };
+
   return (
     <div
-      className={`card-text ${className}`}
+      className={cx("card-text", className)}
       style={{ ...descriptionStyles, margin: "12px 0 0 0" }}
```
leafty, line 131: Margin is added again here. We can just have:

```suggestion
      style={style}
```

client/src/components/entities/Description.tsx
```diff
@@ -16,105 +16,122 @@
...
-import { Link } from "react-router-dom";
-import { RootStateOrAny, useSelector } from "react-redux";
-import { RenkuMarkdown } from "../markdown/RenkuMarkdown";
-
 /**
  *  renku-ui
  *
  *  Entity Description.tsx
  *  Entity Description component
  */
```
leafty, line 24: Nitpick: I would remove this, the component is named `EntityDescription` which is plenty clear.

client/src/components/entities/Description.tsx
```diff
@@ -16,105 +16,122 @@
...
-    overflow: "hidden",
-    textOverflow: "ellipsis",
-    display: "-webkit-box", // eslint-disable-line
-    lineClamp: isHeightFixed ? numberLines : undefined,
-    WebkitLineClamp: isHeightFixed ? numberLines : undefined, // eslint-disable-line
-    WebkitBoxOrient: "vertical", // eslint-disable-line
-    minHeight: isHeightFixed ? `${25 * numberLines}px` : undefined,
-    height: isHeightFixed ? `${25 * numberLines}px` : undefined,
-  };
+  let content: React.ReactNode = null;
```
leafty, line 55: Avoid `let` and use ternaries (coding guidelines).

client/src/components/formlabels/FormLabels.css
```diff
@@ -3,6 +3,11 @@
   font-style: italic;
 }
 
+.loading-label.feedback {
+  line-height: 16px;
+  padding-top: 6px;
+}
```
leafty, line 9: We should avoid adding global CSS rules if possible. I think this can be achieved by:
* Adding bootstrap class names to `InlineSubmitButton`
* Or by adding a CSS module for `InlineSubmitButton`

client/src/components/formlabels/FormLabels.tsx
```diff
@@ -23,17 +23,20 @@
  *  FormLabels components.
  */
 import * as React from "react";
```
leafty, line 25: We can modernize imports, maybe:
```suggestion
import React from "react";
```

client/src/components/visibility/Visibility.tsx
```diff
@@ -87,7 +87,7 @@ export interface VisibilitiesInputProps {
   namespaceVisibility: Visibilities;
 
   /** Default value */
-  value: Visibilities | null;
+  value?: Visibilities;
```
leafty, line 90: I think the original typing is more semantically correct:

```suggestion
  value: Visibilities | null;
```

client/src/features/project/components/ProjectEntityHeader.tsx
```diff
@@ -20,48 +20,53 @@ import React from "react";
...
+import { useProjectMetadataQuery } from "../../projects/projectsKgApi";
 
 type ProjectEntityHeaderProps = EntityHeaderProps & {
   branch: string;
   projectId: number;
 };
 
 export function ProjectEntityHeader(props: ProjectEntityHeaderProps) {
-  const { fullPath } = props;
+  const { branch, devAccess, fullPath, gitUrl, projectId } = props;
```
leafty, line 33: Nitpick: deconstruct props

```tsx
export function ProjectEntityHeader({
  branch,
  devAccess,
  [...]
}: ProjectEntityHeaderProps) {
  [...]
}
```

client/src/features/project/components/ProjectEntityHeader.tsx
```diff
@@ -20,48 +20,53 @@ import React from "react";
...
-    }
-  );
+  const projectIndexingStatus = useGetProjectIndexingStatusQuery(projectId, {
+    skip: !fullPath || !projectId,
+  });
 
-  const { data: projectInKgData } = useProjectMetadataQuery(
-    { projectPath: fullPath },
-    { skip: !fullPath || !projectIndexingStatus.data?.activated }
+  const projectMetadata = useProjectMetadataQuery(
```
leafty, line 39: Nitpick: variable name for readability

```suggestion
  const projectMetadataQuery = useProjectMetadataQuery(
```

client/src/features/project/components/ProjectEntityHeader.tsx
```diff
@@ -20,48 +20,53 @@ import React from "react";
...
+    { projectPath: fullPath, projectId },
+    { skip: !fullPath || !projectId || !projectIndexingStatus.data?.activated }
   );
 
+  // overwrite description when available from KG
+  const descriptionKg: EntityHeaderProps["description"] = {
+    isLoading: projectMetadata.isLoading || projectIndexingStatus.isLoading,
+    value: projectMetadata.data?.description || "",
```
leafty, line 47:
```suggestion
    value: projectMetadata.data?.description ?? "",
```

client/src/features/project/components/ProjectSettingsDescription.tsx
```diff
@@ -0,0 +1,164 @@
...
+        description,
+        gitUrl,
+        projectId,
+      }),
+    [updateDescriptionMutation, gitUrl, projectId]
+  );
+
+  const submit = () => {
+    updateDescription(description);
+  };
```
leafty, line 70: Create only one callback function here:

```suggestion
  const onSubmit = useCallback(() => {
    updateDescriptionMutation({
      description,
      gitUrl,
      projectId,
    });
  }, [description, gitUrl, projectId, updateDescriptionMutation]);
```

client/src/features/project/components/ProjectSettingsUtils.tsx
```diff
@@ -0,0 +1,37 @@
...
+import cx from "classnames";
+
+import { faExclamationTriangle } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+interface SettingRequiresKgProps {
+  className?: string;
+}
+export function SettingRequiresKg({ className }: SettingRequiresKgProps) {
+  const localClass = className ? cx(className) : "d-block mt-2";
```
leafty, line 29: No need for `cx` here:

```suggestion
  const localClass = className ?? "d-block mt-2";
```

client/src/features/project/components/ProjectSettingsUtils.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  return (
+    <small className={localClass}>
+      <FontAwesomeIcon icon={faExclamationTriangle} /> This requires Knowledge
+      Graph integration
```
leafty, line 34: Avoids extra `[space]`:

```suggestion
      <FontAwesomeIcon className="me-5" icon={faExclamationTriangle} />
      This requires Knowledge Graph integration
```

client/src/features/project/components/migrations/ProjectStatusIcon.tsx
```diff
@@ -70,8 +69,8 @@ export function ProjectStatusIcon({
   const maintainerText =
     isMaintainer && !matchRoute ? " Click to see details." : "";
 
-  if (isLoading) return null;
-  if (!kgActivated)
+  // if (isLoading) return null;
```
leafty, line 72:
```suggestion
```

client/src/features/project/projectCoreApi.ts
```diff
@@ -109,7 +113,7 @@ function urlWithQueryParams(url: string, queryParams: any) {
 export const projectCoreApi = createApi({
   reducerPath: "projectCore",
   baseQuery: fetchBaseQuery({ baseUrl: "/ui-server/api" }),
-  tagTypes: ["project", "project-status", "ProjectConfig"],
+  tagTypes: ["project-status", "project", "ProjectConfig"],
```
leafty, line 116: I think we should use the same convention as [RTK docs](https://redux-toolkit.js.org/rtk-query/usage/automated-refetching) for tags, which is `CamelCase` like types, interfaces, classes, etc.

```suggestion
  tagTypes: ["ProjectStatus", "Project", "ProjectConfig"],
```

client/src/features/project/projectCoreApi.ts
```diff
@@ -282,6 +286,41 @@ export const projectCoreApi = createApi({
...
+        return {
+          body: { description: data.description, git_url: data.gitUrl },
+          method: "POST",
+          url: `/renku/project.edit`,
+          validateStatus: (response, body) => {
+            return response.status < 400 && !body.error?.code;
+          },
+        };
+      },
+      // ? Invalidating immediatley won't work since KG needs a moment to pick up the change.
```
leafty, line 303: What does `// ?` mean? (Question mark at start of comment)

client/src/features/project/projectCoreApi.ts
```diff
@@ -282,6 +286,41 @@ export const projectCoreApi = createApi({
...
+      onCacheEntryAdded: (params, { dispatch, cacheDataLoaded }) => {
```
leafty, line 304: Non-blocking: This may not be the most reliable way to wait for status.
Maybe polling until the status is correct would work? e.g. https://github.com/SwissDataScienceCenter/renku-ui/blob/b5a035b1e7d6ae52bc67c330fc28cfdd72c4e7a7/client/src/features/session/useWaitForSessionStatus.hook.ts

client/src/features/project/projectCoreApi.ts
```diff
@@ -282,6 +286,41 @@ export const projectCoreApi = createApi({
...
+                { type: "project-indexing", id: params.projectId },
+              ])
+            );
+            // ? invalidate also metadata in the unlikely case the change is quicker than the timeout.
+            dispatch(
+              projectsKgApi.util.invalidateTags([
+                { type: "project-kg-metadata", id: params.projectId },
+              ])
+            );
+          }, 1000 * 5);
```
leafty, line 319: Define a constant for this.

client/src/project/new/components/Visibility.tsx
```diff
@@ -82,7 +92,7 @@ interface EditVisibilityModalConfirmationProps {
   isLoading: boolean;
   isSuccess: boolean;
   message: ReactNode;
-  visibility: Visibilities;
+  visibility?: Visibilities;
```
leafty, line 95: This feels suspicious. The prop should always be defined I think.

```suggestion
  visibility: Visibilities;
```

client/src/project/new/components/Visibility.tsx
```diff
@@ -179,15 +190,22 @@ const EditVisibility = ({
     data: forkProjectData,
     isFetching: isFetchingForkProject,
     isLoading: isLoadingForkProject,
-  } = useGetProjectByIdQuery(forkedProjectId, { skip: !forkedProjectId });
+  } = useGetProjectByIdQuery(forkedProjectId as number, {
```
leafty, line 193:
```suggestion
  } = useGetProjectByIdQuery(forkedProjectId ?? 0, {
```

client/src/project/new/components/Visibility.tsx
```diff
@@ -256,49 +274,51 @@ const EditVisibility = ({
...
     );
 
-  if (!isLoadingProject && !isFetchingProject && !projectData)
-    return (
-      <WarnAlert dismissible={false}>
-        Knowledge Graph integration must be enabled to change visibility from
-        the RenkuLab web interface.
-      </WarnAlert>
-    );
+  const isLoadingSomething =
```
leafty, line 277: Nitpick: feels more serious to simply call it `isLoading`

```suggestion
  const isLoading =
```

client/src/project/new/components/Visibility.tsx
```diff
@@ -256,49 +274,51 @@ const EditVisibility = ({
...
+    isLoadingNamespace || isLoadingProject || isLoadingIndexingStatus;
 
-  if (isLoadingProject && isFetchingProject && !projectData) return <Loader />;
+  const content = isLoadingSomething ? (
+    <>
+      <label>Visibility</label>
```
leafty, line 282: DOM semantics (#2653):

```suggestion
      <div className="form-label">Visibility</div>
```

client/src/components/formlabels/FormLabels.tsx
```diff
@@ -23,17 +23,20 @@
...
+import cx from "classnames";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import {
   faExclamationTriangle,
   faCheck,
 } from "@fortawesome/free-solid-svg-icons";
 
 import "./FormLabels.css";
+
+import { FormText, FormFeedback, Label } from "../../utils/ts-wrappers";
```
leafty, line 35: Import these from `reactstrap`

leafty (changes requested):
Good to see the project description editing going to RTK

The loading state looks broken:
image

client/src/project/new/components/Visibility.tsx
lorenzo-cavazzi: Actually `isLoading` is already used for something else in that component :sweat_smile: 
I will rename that to `isLoadingMutation` and change this to `isLoadingAnything`. `isLoading` would be too confusing since there are 4 isLoading variables

client/src/components/entities/Description.tsx
lorenzo-cavazzi: `lh` cannot be used for `height`, it's marked as invalid in Firefox. I tried a couple of other solutions but they didn't work either.

If it's ok for you I would leave this change to another issue since it doesn't impact the settings page but rather the listings (e.g. Datasets tab, search results, ...)

client/src/features/project/components/ProjectEntityHeader.tsx
lorenzo-cavazzi: There are quite props we don't use here but still need to pass down to `EntityHeader` as `{...props}`. Not sure restructuring here would be beneficial

client/src/features/project/components/ProjectSettingsDescription.tsx
lorenzo-cavazzi: Could you briefly explain why this is a better solution? Just to be more concise or is there any impact on performances?

client/src/features/project/projectCoreApi.ts
lorenzo-cavazzi: I'm not sure about this. We actually use kebab case in all the other tags so far, with the one exception `ProjectConfig` that is pascal case. I personally find pascal case the most confusing choice here since we use that only for types and react components, and I wouldn't consider switching to yet another convention (camel case)

client/src/components/entities/Description.tsx
leafty: Yes, this is not a blocking comment.

I think it could be done like this: https://tailwindcss.com/docs/line-clamp

client/src/features/project/projectCoreApi.ts
lorenzo-cavazzi: We use it in many places in the UI to explain non-trivial choices in the code.

I guess the main added benefit is that many plugins color it, immediately distinguishing it from the other comments. I believe we use also `// !` for critical and `// TODO` for things to remember when further touching some code)

client/src/features/project/components/ProjectSettingsDescription.tsx
leafty: It's simpler, and it's less code to read. Performance-wise, it's also better since doing `onClick={onSubmit}` will be a stable reference.

client/src/features/project/projectCoreApi.ts
leafty: The idea of that convention is that the tag type should be the return type of the query (e.g. `builer.query<ProjectStatus, GetProjectStatusParams>({ ... })`). In that way, it makes sense to adopt the same identifier as the type being returned.

client/src/features/project/projectCoreApi.ts
lorenzo-cavazzi: That's a fair point. Your solution sounds a bit better, but we can still miss the updates in several cases. I guess we really need a message queue system notifying the UI about changes in the project status/indexing to really address this

client/src/features/project/projectCoreApi.ts
leafty: Also, I am already using that convention in all my RTK endpoints since I followed the docs :wink:

lorenzo-cavazzi: I see the reason, but this doesn't seem easy in all cases. E.G: in the KG API we use `KgMetadataResponse`, that doesn't sound like a good tag name.

I'm ok with using pascal case, but since this PR doesn't touch the tag directly I would rather make a PR that changes all the params to a new convention, and add that the coding guides

lorenzo-cavazzi:
Thank you @leafty for the review! I either addressed the points or responded inline.
I also improved the UX for anonymous users for both the description and the visibility to prevent unnecessary errors

client/src/project/new/components/Visibility.tsx
leafty: I would avoid "Something" and "Anything" because they do not help explain what is happening in the code. Something like `isLoadingQueries` or `isLoadingData` gives a hint about what we are loading.

client/src/features/project/projectCoreApi.ts
leafty: OK :+1: I have not found any syntax highlighter which does this in vscode though.

client/src/features/project/components/ProjectSettingsDescription.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  const submitButton = readOnly ? null : (
+    <InlineSubmitButton
+      className="updateProjectSettings"
+      doneText="Updated"
+      id="update-desc"
+      isDone={updateDescriptionStatus.isSuccess}
+      isReadOnly={readOnly || pristine}
+      isSubmitting={updateDescriptionStatus.isLoading}
+      onSubmit={() => onSubmit()}
```
leafty, line 117: Technicality: `() => onSubmit()` creates a new function at each render, while `onSubmit` may stay the same.

```suggestion
      onSubmit={onSubmit}
```

andre-code (commented):
I'm glad we use RTK to update the description and get the project data.
The delay doesn't convince me much in terms of UX because the user doesn't know what's going on, why the value doesn't change, could navigate and see the previous value again.
We also lost the "updating description" message in the project header when it was changing (2nd image attached).

 Is it possible to update the value in the store when the change is successful? if something fails which would be in minor percentage of cases 🤞 then it will revert to the fetched value after 5 sec. By the way the wait to see the updated value was more than 5 seconds 😔.

update desc


update desc before

client/src/features/project/projectCoreApi.ts
```diff
@@ -50,6 +54,10 @@ interface GetConfigRawResponse {
   error?: unknown;
 }
 
+// Expected seconds of delay before the KG is aware of the latest changes.
+// TODO: this won't be useful anymore once we get status updates though WebSockets.
```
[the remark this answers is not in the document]
lorenzo-cavazzi: Fair point!
In this case, we don't have an issue yet cause the feature isn't defined yet

leafty (approved):
LGTM

Issue #2515, opened 2023-05-11
Project visibility button


bethcg:
## Motivation
Some users want to change the visibility (private, public, internal) of their projects. Currently, this can be only done through the Gitlab settings. It is not intuitive how to do it, if you are not an experienced user. 
![Screenshot 2023-05-11 at 11 07 03](https://github.com/SwissDataScienceCenter/renku-ui/assets/127967525/cd77a390-69bb-4746-8f39-0dc0668a7d38)
View of Renku Project/Settings page

![screenshot_2023-05-11_at_11 06 17_720](https://github.com/SwissDataScienceCenter/renku-ui/assets/127967525/4cd9cccc-7eb4-499d-81b2-ae9090221836)
View of Gitlab Project/Settings page

## Proposal

There are mainly two options:
- Create a button in the Project page / Settings that allows the user to change the visibility of the project (this may have some corner cases that need to be tackled).
- Create a button that redirects the user to the Gitlab settings project page within the Renku Project page/ Settings and give indications to the user on how to do the change (and include it in the documentation page).

jdupre81:
**Forked project** - change visibility

https://github.com/SwissDataScienceCenter/renku-ui/assets/117274853/c4b382bd-9297-4f30-b277-774774f800ab

**Own project** - change visibility

https://github.com/SwissDataScienceCenter/renku-ui/assets/117274853/4a3c3a9b-b6a9-4926-b699-730b30149196

jdupre81:
as discussed last week, here is the design for the use case:
**Change visibility of my own project + Message on a Confirmation box**
[Figma prototype ](https://www.figma.com/proto/8yrdRbT2CuKYeUlFqNKaqe/Project-visibility-button%232515?page-id=307%3A3412&type=design&node-id=307-3684&viewport=144%2C-54%2C0.16&t=kEkvzkuznxCobBpk-1&scaling=min-zoom&starting-point-node-id=307%3A3417&mode=design)

https://github.com/SwissDataScienceCenter/renku-ui/assets/117274853/980a464c-b16e-4a16-8560-e8a47607f2e3

lorenzo-cavazzi:
A few details:
- For the visibilities, re-use the same text/description we have in the "New project" page (mind they are swapped in the prototype above)
![image](https://github.com/SwissDataScienceCenter/renku-ui/assets/43481553/bce72fa5-43b2-4190-b538-257c6921a49f)
[...]

PR #2648, opened 2023-07-11
feat: change visibility on an existing project (#2515)

andre-code:
PR to include change the visibility on an existing project.

edit visibility

/deploy renku-graph=development renku=auto-update/renku-ui-3.10.0 #persists


client/src/utils/helpers/TextUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+// eslint-disable-next-line
+function extractTextFromObject(obj: Record<string, any>): string[] {
+  // eslint-disable-line
+  const textValues: string[] = [];
+
+  for (const key in obj) {
```
leafty, line 23: nitpick: I would favour using `.map()` or `.flatpMap()` here:

```ts
export function extractTextFromObject(obj: Record<string, unknown>): string[] {
  const textValues = Object.values(obj).flatMap((value) => {
    if (typeof value === 'string') {
      return [value as string]
    }
    if (Array.isArray(value)) {
      const mapped = (value as unknown[]).flatMap((element) => {
        if (typeof element === 'string') {
          return [element as string]
        }
        if (typeof element === 'object') {
          return extractTextFromObject(element as Record<string, unknown>)
        }
        return []
      })
      return mapped;
    }
    if (typeof value === 'object') {
      return extractTextFromObject(value as Record<string, unknown>)
    }
    return []
  })
  return textValues;
}
```

client/src/styles/components/_renku_general.scss
```diff
@@ -123,3 +123,9 @@ svg:focus-visible {
     height: auto;
   }
 }
+.rk-form-section {
+  padding: 18px;
+  background-color: white;
+  border-radius: 8px;
+  margin-bottom: 20px;
+}
```
leafty, line 131: Would it be possible to add this inside a CSS or SCSS module instead? This will avoid CSS pollution (for example grouped inputs or broken because of bootstrap misconfiguration at the moment).

leafty (changes requested):
Overall, the logic is good.

Sorry for all the code quality comments, but these are important for readability and avoiding bugs.

client/src/styles/components/_renku_general.scss
leafty: Actually, using bootstrap classes would be the most favoured option.

client/src/utils/helpers/TextUtils.test.ts
```diff
@@ -0,0 +1,39 @@
...
+      "Array text 2",
+      "Array text 3",
+      "Array text 4",
+      "Array text 5",
+      "Array text 6",
+    ];
+
+    const textValues = extractTextFromObject(obj);
+    expect(textValues).toEqual(expectedTextValues);
+  });
```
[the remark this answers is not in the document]
andre-code: Thanks for the feedback.
As I understand the definition of expected result (expected) is defined within the Arrange section.

client/src/utils/helpers/TextUtils.ts
andre-code: Good idea. That would be cleaner and remove some of the complexity. I have redone this, you can find it in the last commit.

client/src/styles/components/_renku_general.scss
andre-code: I am not sure if I understood this suggestion correctly, could you please clarify which module you are referring to, one per component? In the meantime I have added the class in the css module we have created for the forms `_renku-form.scss`.

andre-code:
> Overall, the logic is good.
> 
> Sorry for all the code quality comments, but these are important for readability and avoiding bugs.

Thanks @leafty  for your time in reviewing my PR.
I have carefully considered all your suggestions and implemented them in the latest commit, only a couple of cases are missing as they are not quite clear. I have left some comments.

client/src/utils/helpers/TextUtils.test.ts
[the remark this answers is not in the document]
leafty: This is a nitpick, you can ignore it if you think it makes more sense in the arrange section.

It would be nice to have:
```
<arrange>
new line
<act>
new line
<assert>
```
if possible.

client/src/styles/components/_renku_form.scss
```diff
@@ -199,3 +199,10 @@
   display: flex;
   align-items: center;
 }
+
+.form-rk-section {
+  padding: 18px;
+  background-color: white;
+  border-radius: 8px;
+  margin-bottom: 20px;
+}
```
leafty, line 208: Adding global CSS classes should be avoided when possible as they leak style globally (see next comment for how to achieve the same style).

```suggestion
```

client/src/project/settings/ProjectSettings.present.js
```diff
@@ -89,26 +90,42 @@ function ProjectSettingsGeneral(props) {
...
-      </Row>
-      <Row>
-        <Col xs={12}>
-          <ProjectAvatarEdit
-            externalUrl={props.externalUrl}
-            avatarUrl={props.metadata.avatarUrl}
-            onAvatarChange={props.onAvatarChange}
-            settingsReadOnly={props.settingsReadOnly}
-          />
+          <div className="form-rk-section">
```
leafty, line 102: To avoid defining a new global CSS class, I have found that we can do:

```tsx
  return (
    <div className="form-rk-green mb-4">
      <Row className="gy-4">
        <SettingsCard>
          <ProjectTags
            tagList={props.metadata.tagList}
            onProjectTagsChange={props.onProjectTagsChange}
            settingsReadOnly={props.settingsReadOnly}
          />
        </SettingsCard>
        <SettingsCard>
          <ProjectDescription {...props} />
        </SettingsCard>
        <SettingsCard>
          <EditVisibility
            projectId={props.metadata.id}
            namespace={namespace}
            forkedProjectId={props.forkedFromProject?.id}
            visibility={props.metadata.visibility}
          />
        </SettingsCard>
        <SettingsCard>
          <ProjectAvatarEdit
            externalUrl={props.externalUrl}
            avatarUrl={props.metadata.avatarUrl}
            onAvatarChange={props.onAvatarChange}
            settingsReadOnly={props.settingsReadOnly}
            includeRequiredLabel={false}
          />
        </SettingsCard>
      </Row>
    </div>
  );
}

function SettingsCard({ children }) {
  return (
    <Col xs={12}>
      <Card>
        <CardBody className="py-2">
          <Row>
            <Col>{children}</Col>
          </Row>
        </CardBody>
      </Card>
    </Col>
  );
}
```

Note that this makes all the cards have the same spacing around them (which was not the case before).

client/src/components/visibility/Visibility.tsx
```diff
@@ -132,107 +173,152 @@ const VisibilityInput = ({
...
-      <InputLabel text="Visibility" isRequired={isRequired} />
+      <legend>
+        {" "}
+        Visibility{" "}
+        {includeRequiredLabel ? (
+          <RequiredLabel isRequired={isRequired} />
+        ) : (
+          ""
+        )}{" "}
+      </legend>
```
leafty, line 233: The legend renders with a large font size, you can style it with:

```suggestion
      <legend className="form-label fs-6">
        Visibility{" "}
        {includeRequiredLabel && <RequiredLabel isRequired={isRequired} />}
      </legend>
```

client/src/components/visibility/Visibility.tsx
```diff
@@ -132,107 +173,151 @@ const VisibilityInput = ({
...
-    },
-  ];
-
-  const options = items.map((item) => {
-    const disabledByNamespace = visibilities.disabled.includes(item.value);
-    const isDisabled = disabled || disabledByNamespace;
 
+  const disableByNamespaceOptions = {
+    public: "",
+    private: `Public and Internal options are not available due to ${
```
lorenzo-cavazzi, line 179:
```suggestion
    private: `Public and Internal not available due to ${
```
slightly shorter

client/src/components/visibility/Visibility.tsx
```diff
@@ -132,107 +173,151 @@ const VisibilityInput = ({
...
+      isForked ? "forked project or " : ""
+    }group namespace restrictions. `,
+    internal: `Public is not available due to ${
```
lorenzo-cavazzi, line 182:
```suggestion
    internal: `Public not available due to ${
```
same pattern as the previous message

client/src/project/new/components/Visibility.tsx
```diff
@@ -58,4 +73,209 @@ const Visibility = ({ handlers, meta, input }: VisibilityProps) => {
...
+          title="documentation"
+        />
+        .
+      </>
+    );
+
+  return (
+    <Modal isOpen={isOpen} toggle={() => toggleModal()} size="lg">
+      <ModalHeader toggle={() => toggleModal()}>
+        Edit visibility to{" "}
```
lorenzo-cavazzi, line 138:
```suggestion
        Change visibility to{" "}
```
Mind that this requires adjusting the e2e test -- I'll add that as a suggestion too

tests/cypress/e2e/project.spec.ts
```diff
@@ -178,6 +182,39 @@ describe("display a project", () => {
     );
     cy.wait("@updateProject");
     cy.get_cy("entity-description").should("contain.text", "description abcde");
+    // Change visibility
+    cy.wait("@getProjectsById");
+    // should keep visibility status after cancel update
+    cy.get_cy("visibility-private").click();
+    cy.get(".modal-title").should("contain.text", "Edit visibility to Private");
```
lorenzo-cavazzi, line 189:
```suggestion
    cy.get(".modal-title").should("contain.text", "Change visibility to Private");
```

client/src/project/new/components/Visibility.tsx
```diff
@@ -58,4 +73,209 @@ const Visibility = ({ handlers, meta, input }: VisibilityProps) => {
...
+
+  const content =
+    isError || isSuccess ? (
+      message
+    ) : isLoading ? (
+      <LoadingLabel text="Updating visibility" />
+    ) : (
+      <>
+        Please note that users will not need your explicit permission to see
+        this project. If you need more information about visibility, please
```
lorenzo-cavazzi, line 124: This message is correct for public and internal projects but misleading for private projects; it's actually the opposite.
In that case, I would change it to something like "Users will not find this project anymore unless you explicitly add them to the Members list", with a link to the members page in GitLab

image

client/src/project/new/components/Visibility.tsx
```diff
@@ -58,4 +73,209 @@ const Visibility = ({ handlers, meta, input }: VisibilityProps) => {
...
+  isOpen,
+  toggleModal,
+  onConfirm,
+  isError,
+  isLoading,
+  isSuccess,
+  message,
+  visibility,
+}: EditVisibilityModalConfirmationProps) {
+  const buttons = !isError && !isLoading && !isSuccess && (
```
lorenzo-cavazzi, line 97: Should we show a "Close" button in all the other cases?
Feel free to disagree and keep the current solution here :slightly_smiling_face:

lorenzo-cavazzi (changes requested):
Very useful addition!
I added just a couple of inline comments about working or tiny changes.

There is just one additional thing I would change. We should ignore clicks to the current visibility. E.G. if the project is Private already and the user tries to click on Private, we should not open the modal to confirm changing it. Mind that this happens only when clicking on the label, while the radio button already ignores the click correctly (see video)

Peek 2023-07-20 19-04

andre-code:
> Very useful addition! I added just a couple of inline comments about working or tiny changes.
> 
> There is just one additional thing I would change. We should ignore clicks to the current visibility. E.G. if the project is Private already and the user tries to click on Private, we should not open the modal to confirm changing it. Mind that this happens only when clicking on the label, while the radio button already ignores the click correctly (see video)
> 
> Peek 2023-07-20 19-04

Thanks @lorenzo-cavazzi for your suggestions, I have implemented them and also those discussed in the product ui meeting.
In addition I have included here the endpoint kg to update the visibility, if the project is not yet activated we show an Alert.

client/src/project/new/components/Visibility.tsx
```diff
@@ -58,4 +75,231 @@ const Visibility = ({ handlers, meta, input }: VisibilityProps) => {
...
+    ) : (
+      ""
+    );
+
+  if (!isLoadingProject && !isFetchingProject && !projectData)
+    return (
+      <WarnAlert>
+        Knowledge Graph integration must be enabled to change visibility from
+        the RenkuLab web interface.
+      </WarnAlert>
```
lorenzo-cavazzi, line 265: Should we make this non-dismissible? Otherwise, we are left with a white space.

Peek 2023-07-24 14-39

lorenzo-cavazzi (changes requested):
This looks good! It does include #2675 already, right?

I added a couple of minor comments inline. We can then merge it

client/src/components/formlabels/FormLabels.tsx
```diff
@@ -69,6 +75,7 @@ const HelperLabel = ({ text }: LabelProps) => {
 };
 
 const InputHintLabel = ({ text }: LabelProps) => {
+  // TODO: remove this component
```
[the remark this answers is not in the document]
leafty: See https://github.com/SwissDataScienceCenter/renku-ui/pull/2648#discussion_r1267722707. This adds one level of indirection for no benefit. Just use `FormText`.

client/src/components/formlabels/FormLabels.tsx
```diff
@@ -68,8 +68,12 @@ const HelperLabel = ({ text }: LabelProps) => {
...
 };
 
-const InputHintLabel = ({ text }: LabelProps) => {
-  return <FormText className="input-hint">{text}</FormText>;
+const InputHintLabel = ({ text, children }: LabelProps) => {
+  return (
+    <FormText className="input-hint">
+      {text} {children}
+    </FormText>
+  );
```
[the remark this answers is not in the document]
leafty: Done in 84a42dde5fcb849d5ce683668a7632c2fc20c111

client/src/components/visibility/Visibility.tsx
```diff
@@ -132,107 +173,149 @@ const VisibilityInput = ({
...
+            {item.title}
+          </label>
+          <FontAwesomeIcon
+            icon={item.icon}
+            className={isDisabled ? "icon-disabled" : ""}
+          />
+        </div>
+      </div>
+      {isDisabled ? (
+        <ThrottledTooltip
```
leafty, line 308: Use `UncontrolledTooltip` from `reactstrap`.

lorenzo-cavazzi (approved):
Lgtm! :rocket:

lorenzo-cavazzi (approved):

PR #2649, opened 2023-07-11
fix(client): prevent endless loader on migration errors

lorenzo-cavazzi:
When the check_migrations API errors, we show an endless spinning wheel and the error doesn't surface. This PR fixes it.

Peek 2023-07-11 15-01

/deploy #persist #cypress


lorenzo-cavazzi:
Superseded by #2650

PR #2650, opened 2023-07-11
fix(client): prevent endless loader on migration errors

lorenzo-cavazzi:
When the check_migrations API errors, we show an endless spinning wheel and the error doesn't surface. This PR fixes it.

This is an alternative implementation to #2649 . Do not merge both!

Peek 2023-07-11 15-01

/deploy #persist #cypress

### DIFF
--- a/client/src/features/project/components/migrations/ProjectCoreMigrations.tsx
+++ b/client/src/features/project/components/migrations/ProjectCoreMigrations.tsx
@@ -263,7 +263,7 @@ interface ProjectMigrationStatusDetailsProps {
   buttonDisable: boolean;
   data: MigrationStatus | undefined;
   isMaintainer: boolean;
-  isSupported: boolean;
+  isSupported: boolean | undefined;
   showDetails: boolean;
   updateProject: (scope: MigrationStartScopes) => void;
 }
--- a/client/src/features/project/dataset/ProjectDatasetsView.tsx
+++ b/client/src/features/project/dataset/ProjectDatasetsView.tsx
@@ -197,6 +197,7 @@ function ProjectDatasetsView(props: any) {
   const {
     backendAvailable,
     computed: coreSupportComputed,
+    backendErrorMessage,
     versionUrl,
   } = coreSupport;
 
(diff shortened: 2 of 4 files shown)


client/src/features/project/components/migrations/ProjectCoreMigrations.tsx
```diff
@@ -263,7 +263,7 @@ interface ProjectMigrationStatusDetailsProps {
   buttonDisable: boolean;
   data: MigrationStatus | undefined;
   isMaintainer: boolean;
-  isSupported: boolean;
+  isSupported: boolean | undefined;
```
leafty, line 266: I think `isSupported` should be `false` in case of an error, not `undefined`.

client/src/features/project/dataset/ProjectDatasetsView.tsx
```diff
@@ -232,6 +233,14 @@ function ProjectDatasetsView(props: any) {
     versionUrl,
   ]);
 
+  if (coreSupportComputed && backendErrorMessage)
+    return (
+      <ErrorAlert>
+        <b>There was an error verifying support for this project.</b>
+        <p>{backendErrorMessage}</p>
+      </ErrorAlert>
```
leafty, line 241: Note: do we offer to reload the page here?

client/src/features/project/useProjectCoreSupport.ts
```diff
@@ -23,10 +23,16 @@ import { useGetMigrationStatusQuery } from "./projectCoreApi";
 export type CoreSupport =
   | {
       backendAvailable: undefined;
-      backendErrorMessage: string | undefined;
+      backendErrorMessage: undefined;
       computed: false;
       versionUrl: undefined;
     }
+  | {
+      backendAvailable: undefined;
```
leafty, line 31: I think this should be `false` when there is an error.

leafty (commented):
LGTM! 🚀

leafty (approved):

client/src/features/project/utils/migrations.ts
```diff
@@ -86,7 +86,7 @@ export function getCompareUrl(
  */
 export function getMigrationLevel(
   migrationStatus: MigrationStatus | undefined,
-  backendAvailable: boolean
+  backendAvailable: boolean | undefined
```
leafty, line 89: Do we need this type change?

lorenzo-cavazzi: Ops sorry I missed this comment. It's actually not really needed since we always invoke `getMigrationLevel`  by passing a `boolean` . I can make another PR to change this bit

lorenzo-cavazzi: https://github.com/SwissDataScienceCenter/renku-ui/pull/2651

PR #2656, opened 2023-07-13
refactor session buttons

leafty:
* `SessionButton` shows a button to start or connect to a session with additional options accessible from a dropdown
* `SimpleSessionButton` shows the same but without the dropdown
* Update the sessions RTK Query API
* Add comments identifying where the new `hibernating` session state needs to be handled

/deploy #persist #cypress extra-values=core.horizontalPodAutoscaling.minReplicas=1,core.replicaCount=1


leafty:
> Nice work! The code looks good already.
> 
> I've found a couple of issues with the start button/dropdown:
> 
> * The height of the loading button is different; can we make it the same? Otherwise, the content moves down when changing from "loading" to "start"
> 
> * When the session runs, the button should say "Connect" instead of "Start". It occasionally stays outdated for a while.
>  
> * The other way around is also true: I got the "Connect" button after stopping the session, but clicking on it brought me to a "Session not available" page
> 
> * Occasionally, the Logs modal doesn't show the commit details
>  

Fixed the update issue and apply `btn-sm` when loading.

lorenzo-cavazzi (commented):
It works well now :+1: 
Can you give a look at the failing local test? It seems that the `/servers` API is invoked too frequently in a test scenario

leafty:
> It works well now +1 Can you give a look at the failing local test? It seems that the `/servers` API is invoked too frequently in a test scenario

@lorenzo-cavazzi I'm not looking at the tests now, this is going to be merged into a feature branch, not `master`. Tests will be updated to reflect the new persistent sessions later.

PR #2658, opened 2023-07-13
update the loader spinner

leafty:
Refactor the loader spinner:
* Simplify props: now it only has `className`, `inline` and `size`
* The spinner will be of the same colour as the current text colour
* Alignment and positioning is simpler and has less quirks
* When `inline=true`, a spinner of the same size as the font size will fit naturally with text

Example:

Screenshot 2023-07-26 at 09 09 20

/deploy #persist #cypress renku=auto-update/renku-ui-3.10.0


lorenzo-cavazzi (approved):
Lgtm! :rocket:

PR #2660, opened 2023-07-13
fix: properly render rstudio components

olevski:
This is just a test to see if we can improve the infrequent weird rending that occurs in rstudio.

I suspect it has to do with how we render/style the iframe but I could be wrong.

This is just a test.

EDIT: After the test the messed up layout never happened in this CI deployment but it happened every time on dev. So the problem is the display property on the iframe.

/deploy #persist #cypress extra-values=core.horizontalPodAutoscaling.minReplicas=1,core.replicaCount=1


olevski:
@lorenzo-cavazzi I tested on this CI deployment and on dev with the same project: https://dev.renku.ch/projects/tasko.olevski/test-r-project-4

On this PR's deployment I did never got the messed up rstudio layout out of 10 session launches.

On dev, on the same project I got the messed up rstudio layout every single time (out of 5 launches). I did not do more because I dont think there is a need.

So the problem is that the rstudio renders when the iframe display property is set to none. And there is a race condition between that and the loading of the rstudio assets. But it happens very often it seems.

olevski:
The messed up layout of rstudio looks like this:
image

The proper look is this:
image

olevski:
@lorenzo-cavazzi also please note that with this change there is a brief period when a blank screen shows up when the session loads. But we can probably find a way to avoid this... I just wanted to test that the problem is with the display property.

lorenzo-cavazzi:
Thanks for looking into this.
To prevent flashing blank pages, we could create the iframe component earlier and keep it invisible until all it's ready. I played with `visibility: hidden` instead of `display: none` to achieve this result. This should also speed up the visualization time when freshly launched.

I pushed the commit already, please give it a try

olevski:
@lorenzo-cavazzi that is really nice! And we dont have the blank page flashing anymore. I launched 6-7 sessions over and over and the layout was always good!

lorenzo-cavazzi (approved):

PR #2661, opened 2023-07-14
fix(client): fix card background with bootstrap 5.3.0

leafty:
The bootstrap update changes the default card color from `white` to be `$body-bg`. This change sets the default card background to be white.

/deploy #persist #notest


lorenzo-cavazzi (changes requested):
Background still transparent
image

leafty:
It's not?!? Maybe you loaded the URL too soon?

lorenzo-cavazzi (approved):
> Maybe you loaded the URL too soon?

That's probably the reason :shrug: 
It looks effective; thank you for the fix!

PR #2675, opened 2023-07-19
feat: change visibility using kg API (#2515)

andre-code:
PR to use kgAPI to modify visibility

/deploy renku=renku-ui-3.10-tests renku-graph=development #persists #notest


andre-code:
included in PR #2648

PR #2676, opened 2023-07-19
ci: update Cypress acceptance tests action

lorenzo-cavazzi:
Updating the Cypress acceptance tests action to support changes in https://github.com/SwissDataScienceCenter/renku-actions/pull/54 and https://github.com/SwissDataScienceCenter/renku/pull/3150

/deploy #persist #notest #cypress renku=renku-ui-3.10-tests

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -16,15 +16,11 @@ on:
       - .github/workflows/**
       - chartpress.yaml
 
-jobs:
-  cleanup-previous-runs:
-    runs-on: ubuntu-22.04
-    if: github.event.action != 'closed'
-    steps:
-      - uses: rokroskar/workflow-run-cleanup-action@v0.3.3
-        env:
-          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+concurrency:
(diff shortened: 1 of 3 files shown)


.github/workflows/acceptance-tests.yml
```diff
@@ -108,19 +104,29 @@ jobs:
...
+      (needs.check-deploy.outputs.test-enabled == 'true' || needs.check-deploy.outputs.test-cypress-enabled == 'true')
     needs: [check-deploy, deploy-pr]
-    if: github.event.action != 'closed' && needs.check-deploy.outputs.pr-contains-string == 'true' && needs.check-deploy.outputs.test-cypress-enabled == 'true'
     runs-on: ubuntu-22.04
+
+    strategy:
+      fail-fast: false
+      max-parallel: 1
+      matrix:
+        tests: [publicProject, updateProjects, useSession]
```
leafty, line 119: Should we run the R session test here?

leafty (approved):

.github/workflows/acceptance-tests.yml
lorenzo-cavazzi: I was wondering if it's necessary since it basically replicates what we test in `useSession` but picks a Rstudio template.
I would rather implement a test specifically to create a project and start a session in each template we provide.

PR #2686, opened 2023-07-21
feat!(client): support persistent sessions in the UI

### LINKED ISSUE #2526: 🔶 Support session persistence
## Motivation

We are reworking how autosaves work on the backend. All the info is available in [this RFC document](https://github.com/SwissDataScienceCenter/renku-design-docs/pull/28) made on [this ShapeUp pitch](https://www.notion.so/Make-autosaves-not-be-so-bad-120264b524ce45d485d407c14a597fb6).

In short, sessions will change as follows:
- Only 1 session can run per user per project.
- Running sessions will split into 2 different states, either "running" or "sleeping" (double check the exact words). A sleeping session can resume quicker and might need to be re-created if it should run from a specific branch/commit different from the current one.
- We won't have autosaves from GitLab anymore. The sleeping session will take its place. Any change to running or sleeping sessions requires informing the user (E.G: we need to start from a different branch/commit, throwing away the previous content)

## Proposal

This requires quite a few changes.
Ideally, we can use this to rewrite stale session components. This includes:
- [ ] Dropping the project sessions tab. That won't be useful anymore since we allow just 1 session per project/user.
    - This requires updating the session buttons (they are used in the dashboard and the project header).
    - We can drop the `Notebooks` component from the notebooks containers, together with presentational components used only by `Notebooks`.
    - Double-check all the places where we redirect to "/projects/[slug]/sessions" and redirect to just "/projects/[slug]" instead.
    - :exclamation: anonymous users still have a global sessions page using the `Notebooks` component in the Notebooks.container.js file; it should be accessible by the `/sessions` page/route. We can swap that page with a new one showing the running sessions using the component from the Dashboard (or a simplified version).
- [ ] Simplify the logic to start a new project (`StartNotebookServer` component). 
    - That's _really_ a beast. We need to update the logic to support the new simplified autosave flow. We can simplify quite a bit all the rest, but beware that changing too much will result in a huge PR!
    - Remember we need to support the existing flows: Quickstart and Start with options. The quickstart might provide specific `branch`, `commit` and `env variables` from the URL and the settings coming from the project ".ini" file (read through `GET /renku/config.read`); when failing on picking any values, we fall back to the normal Start with options showing a warning.
    - The flow to read Images, pipelines, jobs, ... is very convoluted. We have [a graph here](https://github.com/SwissDataScienceCenter/renku-ui/blob/master/client/src/notebooks/README.md) of how it works. Be sure you familiarize with it, and perhaps create a set ob projects in specific conditions to test.
- [ ] Drop the polling mechanism we use in some cases. That's used by other old components in the Notebooks.container.js file (`CheckNotebookStatus`, `ShowSession`).
    - We need to update those components, but converting all of them at once might be too time-consuming since we already have a lot of changes here. Consider adding just a wrapper to inject the `sessions` data from the RTK query and create follow-up issues if required.

leafty:
Fixes #2526.

The components to support interactive sessions (e.g. Jupyter notebooks, R studio) have been refactored and updated to support persistent sessions.

/deploy #persist extra-values=notebooks.userSessionPersistentVolumes.enabled=true,notebooks.amalthea.image.repository=leafty/renku-amalthea,notebooks.amalthea.image.tag=pitch-session-hibernation,notebooks.amalthea.image.pullPolicy=Always renku=leafty/update-tests-persistent-sessions renku-notebooks=1.19.0


ciyer (changes requested):
This looks good and works well. I have one suggestion for a change of name, and one proposal for improving UX.

client/src/features/session/components/SessionButton.tsx
```diff
@@ -0,0 +1,512 @@
...
+
+  return <SessionActions className={className} session={runningSession} />;
+}
+
+interface SessionActionsProps {
+  className?: string;
+  session: Session;
+}
+
+function SessionActions({ className, session }: SessionActionsProps) {
```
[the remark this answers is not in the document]
leafty: 🤔 yes it seems that logs are simply not available, so I will remove the option.

`Create Session Link` should always be available (as this is about creating `Launch in Renku` badges).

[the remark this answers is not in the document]
leafty: Done in [005eee2](https://github.com/SwissDataScienceCenter/renku-ui/pull/2686/commits/005eee268e679548344fc69582296bef924b67ef)

lorenzo-cavazzi:
Something else I noticed while playing around. I believe we should not give the `pause` option when the session is in an error state. Pieces might be missing, and this state usually happens at the start time. Pausing can easily fail.
Should we only offer to delete the session in this case?
image

leafty:
> Something else I noticed while playing around. I believe we should not give the `pause` option when the session is in an error state. Pieces might be missing, and this state usually happens at the start time. Pausing can easily fail. Should we only offer to delete the session in this case? image

We should definitely offer a chance for users to cycle their session and not lose their work

lorenzo-cavazzi:
It's unlikely that "Pause" will work when sessions are in that state.
If you wish to preserve this action, we should find a way to handle the error we will probably get.
Currently, we keep showing the spinning wheel even when the API errors. That's not a good UX
image

leafty:
> 

🤔 I will fix that

leafty:
> It's unlikely that "Pause" will work when sessions are in that state. If you wish to preserve this action, we should find a way to handle the error we will probably get. Currently, we keep showing the spinning wheel even when the API errors. That's not a good UX

Done in [9f31c08](https://github.com/SwissDataScienceCenter/renku-ui/pull/2686/commits/9f31c0845c1f9c3a64c64ff3892605fecffeba19).

lorenzo-cavazzi:
Thanks!

I tested the system quite a bit. It works pretty well! I can't wait to see this deployed on RenkuLab :star_struck: 
I got confusing feedback only with failures on the migrations endpoint -- sadly, that seems to happen quite often lately...
When that happens, I get the following:

* On the quickstart page, it hangs in Step 1 on "Loading project settings". We should probably show a red cross on that step and add a notification
image
* On the "Start session" page the feedback is messed up cause we keep re-invoking the `GET /renku/config.show` API. We should show an error alert and not invoke the API if we don't have the project version
Peek 2023-08-31 16-48

rokroskar:
Pausing a session always brings up an error - is it supposed to work at the moment?

leafty:
> Pausing a session always brings up an error - is it supposed to work at the moment?

Should be working now.

leafty:
> Thanks!
> 
> I tested the system quite a bit. It works pretty well! I can't wait to see this deployed on RenkuLab 🤩 I got confusing feedback only with failures on the migrations endpoint -- sadly, that seems to happen quite often lately... When that happens, I get the following:
> 
> * On the quickstart page, it hangs in Step 1 on "Loading project settings". We should probably show a red cross on that step and add a notification
>   image
> * On the "Start session" page the feedback is messed up because we keep re-invoking the `GET /renku/config.show` API. We should show an error alert and not invoke the API if we don't have the project version
>   Peek 2023-08-31 16-48

I just added [728fa15](https://github.com/SwissDataScienceCenter/renku-ui/pull/2686/commits/728fa1525b6d6ecf77c8b6ee5115248c364a3feb), I hope this will cover enough cases on `core` not supporting the project or failing.

client/src/features/session/components/options/AutostartSessionOptions.tsx
```diff
@@ -129,14 +129,17 @@ function useAutostartSessionOptions(): void {
...
+  const {
+    data: projectConfig,
+    error: errorProjectConfig,
+    isFetching: projectConfigIsFetching,
+  } = usePatchedProjectConfig({
+    commit,
+    gitLabProjectId: gitLabProjectId ?? 0,
+    projectRepositoryUrl,
+    versionUrl,
+    skip: !coreSupportComputed || !commit,
```
lorenzo-cavazzi, line 141: I believe this is not enough to properly handle an error from the `migrations_check` endpoint. When I get an error, the `coreSupport` object contains the following:
```
{
  "backendAvailable": false,
  "backendErrorMessage": "Our servers generated an unexpected error while processing data.",
  "computed": true
}
```

`computed` is true, so it fetches the config anyway. You should check both `computed` and `backendAvailable`. Otherwise, it fetches the config with a URL not containing the project version (it works the same only when projects are on the latest version)

Peek 2023-09-01 10-10

lorenzo-cavazzi (changes requested):

client/src/features/session/components/options/AutostartSessionOptions.tsx
leafty: Would [4e40d77](https://github.com/SwissDataScienceCenter/renku-ui/pull/2686/commits/4e40d772b3ac838572b14ee0a1b4fa98bb55b7d0) be OK? The UI will error if the project is not supported by core.

lorenzo-cavazzi: Yes, that's much better!

image

Ideally, we could show the error message from renku-core since it's more on point (it's already in `backendErrorMessage`). In the sample above, there was an error on the server side and it might work again later. Saying the project is not supported is a bit confusing for a user who could wonder what should be done to make the project supported. Still, we can keep the current solution if it's a bigger change.

leafty: I think I will try, but I would like to do it in another PR.

client/src/project/Project.present.js
lorenzo-cavazzi, line 1: I think the code that triggers infinite queries for the servers is the `constructor` in the `ProjectViewHeader`. It invokes `fetchNotebooks` from `NotebooksCoordinator`.
It can _probably_ be removed entirely without any change since now we query sessions differently, but it's worth checking whether we need to add `useGetSessionsQuery` to compensate.
The same goes for the Dashboard component in "features/dashboard/Dashboard.tsx".

Could you give it a look? Otherwise, I can do it sometimes tomorrow

ciyer (approved):
Thanks for addressing my suggestions.

lorenzo-cavazzi (dismissed):
I played around a little bit and all worked well. I can't wait so see this merged! :star_struck:  :rocket: 
Thank you for the big effort on this!

lorenzo-cavazzi:
A tiny detail I noticed: when reaching the "Start with options" page, the UI briefly flashes the error message for the project settings. It's easy to reproduce if you refresh the page from the browser

Peek 2023-09-18 10-57

leafty:
> A tiny detail I noticed: when reaching the "Start with options" page, the UI briefly flashes the error message for the project settings. It's easy to reproduce if you refresh the page from the browser

Should be fixed in [c149bf5](https://github.com/SwissDataScienceCenter/renku-ui/pull/2686/commits/c149bf56fe03672be89cd12d0da700244f111379)

lorenzo-cavazzi (approved):

lorenzo-cavazzi:
P.S. Before merging, we might need to update the Cypress acceptance tests here https://github.com/SwissDataScienceCenter/renku/pull/3240

PR #2687, opened 2023-07-21
support the "hibernated" session state

leafty:
renku-ci-ui-2626 dev renku ch_projects_johann thiebaut1_another-playground-project_sessions


client/src/notebooks/components/StartSessionProgressBar.tsx
```diff
@@ -33,7 +34,7 @@ export interface SessionStatusData {
   details: StatusDetail[];
   message: string;
   readyNumContainers: number;
-  state: string;
+  state: SessionStatusState;
```
lorenzo-cavazzi, line 37: We could have an enum of string instead, and here the type could be
```suggestion
  state: keyof typeof SessionStatusState;
```

client/src/notebooks/Notebooks.present.js
```diff
@@ -121,7 +120,7 @@ function SessionJupyter(props) {
 
   let content = null;
   const status = notebook.data?.status?.state;
-  if (status === SessionStatus.running) {
+  if (status === "running") {
```
lorenzo-cavazzi, line 123: Using an enum here seems more intuitive than a string. Especially in JS where we cannot assign a type to `status`.
Why this change?

lorenzo-cavazzi (changes requested):
Nice work!

I'm unsure about dropping the enum for the status. What's the reason for that?

lorenzo-cavazzi (approved):
We can keep the string literals if we drop the JS files entirely before merging the feature branch. I would stay with the enums for the time being but we can move forward with the currently implemented solution

lorenzo-cavazzi (approved):

PR #2695, opened 2023-07-25
refactor the session view page

leafty:
* Refactor the components used to show a session
* For logged in users, "Stop session" now sets the session in hibernation

### DIFF
--- a/client/src/features/session/components/PullSessionModal.tsx
+++ b/client/src/features/session/components/PullSessionModal.tsx
@@ -1,0 +1,283 @@
+/*!
+ * Copyright 2022 - Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React, { useCallback, useState } from "react";
+import cx from "classnames";
+import { Button, Col, Modal, ModalBody, ModalHeader, Row } from "reactstrap";
+import { Loader } from "../../../components/Loader";
+import {
+  CenteredLoader,
+  InformationalBody,
+  commitsPhrasing,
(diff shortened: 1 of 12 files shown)


client/src/features/session/components/PullSessionModal.tsx
```diff
@@ -0,0 +1,283 @@
...
+  });
+
+  if (isLoading) {
+    return <CenteredLoader />;
+  }
+
+  if (
+    error != null ||
+    data == null ||
+    data.status !== "running" ||
```
lorenzo-cavazzi, line 85: This seems to be a plain `string`. Should we allow to pick only the possible statuses?

lorenzo-cavazzi (changes requested):
Very nice work!
I added a couple of comments inline, but overall good and almost ready to merge already!

One problem I identified: when stopping/deleting a running session, we go back to the Session tab in the project, but the state is outdated and shows a running session for a while. In the video it's an example from an anonymous user but that happens also when hibernating.
Would it make sense to modify the status for the running session? Or save somewhere that we just clicked the button to stop/hibernate and override for a while the value we get from the sessions as far as it's running?

image

lorenzo-cavazzi:
P.S: Not sure if this belongs to the current PR since it's not on the Show Session component, but be sure to align the Stop behavior of the dropdown button to what we offer on the Session page. Here I can close a running session but it should hibernate instead.

Peek 2023-07-25 15-39

client/src/features/session/components/PullSessionModal.tsx
leafty: This is from `client/src/features/session/sidecarApi.ts`. I can try to have a look later, but for now, we don't type the status response from this endpoint.

leafty:
> Very nice work! I added a couple of comments inline, but overall good and almost ready to merge already!
> 
> One problem I identified: when stopping/deleting a running session, we go back to the Session tab in the project, but the state is outdated and shows a running session for a while. In the video it's an example from an anonymous user but that happens also when hibernating. Would it make sense to modify the status for the running session? Or save somewhere that we just clicked the button to stop/hibernate and override for a while the value we get from the sessions as far as it's running?

Added waiting on the UI side with the help of the `useWaitForSessionStatus` hook.

lorenzo-cavazzi (approved):
Nice solution! It feels so much smoother now :100: 

One tiny detail: clicking on Resume should start the session and connect to it, as for the Start action. Currently, it resumes the session but doesn't connect. The video is from the Sessions tab but is the same on the Dashboard. Feel free to address this on another PR, I remember it was actually working as expected the last time I tried, perhaps something was accidentally changed :shrug: 

Peek 2023-07-27 00-46



P.S: I'll drop this here as something to remember before merging the feature branch but not strictly for this PR: double check we don't duplicate the logic for fetching servers on specific pages like during the Start sessions sequence. The WebSocket channel also triggers the `GET /sessions` API. The screenshot might not be immediately obvious, but we fetch both the full list servers and the specific server while starting a new session. We can probably spare one of the set of queries.

image

PR #2698, opened 2023-07-25
feat: Render bigger notebooks and fetch content from files stored in lfs (#2616)

### LINKED ISSUE #2616: Render bigger notebooks and fetch content from files stored in lfs
## Motivation

Jupyter notebooks can easily have sizes >10Mb ad end up in git lfs.
The UI allows customizing the max size for previews but doesn't try to fetch content from git lfs.

## Proposal

We should check whether recent versions of GitLab simplify fetching content stored with lfs, and relax the preview limit -- at least the hard limit, so users can try to preview big-ish files at risk of crashing the UI.

Here is a sample file that renkulab.io can't review now and would benefit from fetching content from lfs:
https://renkulab.io/projects/lorenzo.cavazzi.tech/laue3dndt-experiment-analysis/files/blob/notebooks/plot_segmentation_fega10_s1.ipynb

andre-code:
PR to render big files and lfs files in the file view

/deploy #persists
fix #2616

### DIFF
--- a/client/src/api-client/project.js
+++ b/client/src/api-client/project.js
@@ -16,6 +16,7 @@
  * limitations under the License.
  */
 
+import { getFileExtension } from "../components/markdown/RenkuMarkdownWithPathTranslation";
 import { projectKgApi } from "../features/project/projectKg.api";
 import { sleep } from "../utils/helpers/HelperFunctions";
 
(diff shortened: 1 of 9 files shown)


andre-code:
Postpone it until the product view redesign is done.

PR #2701, opened 2023-07-27
Leafty/2526 persistent sessions 5

leafty:


lorenzo-cavazzi (changes requested):
I love this type of PRs :star_struck: 
image


About the styling, the sessions tab content seems to be slightly too big, it should be as large as the headers as in the other tabs.

image

Another detail about the sessions listed there. Having the large full-name slug feels weird, and clicking it redirects to the Overview section. Should we hide it when it's on that tab? We still want to preserve it on the sessions page for anonymous users

Peek 2023-07-27 16-17

leafty:
> About the styling, the sessions tab content seems to be slightly too big, it should be as large as the headers as in the other tabs.
> 

Fixed in 8cdcb27.

> Another detail about the sessions listed there. Having the large full-name slug feels weird, and clicking it redirects to the Overview section. Should we hide it when it's on that tab? We still want to preserve it on the sessions page for anonymous users

I will update this as another PR for the persistent sessions feature.

lorenzo-cavazzi (approved):

PR #2703, opened 2023-07-28
Show version of Renku and components in UI

### LINKED ISSUE #2664: Reference to the currently deployed release in the UI
## Motivation

As a user, I would like to understand which is the current version of RenkuLab that I am using, and refer it to the releases in GitHub.

## Proposal

The footer of RenkuLab could include the version of Renku, matching the latest deployed release:
image

ciyer:
Fix #2664

# Screenshot

image

/deploy #persist #cypress


client/src/help/HelpRelease.tsx
```diff
@@ -0,0 +1,224 @@
...
+      />
+    </>
+  );
+}
+
+function CoreRelease() {
+  const { data, isFetching } = useGetCoreVersionsQuery();
+  return (
+    <ComponentVersion
+      componentUrl="https://github.com/SwissDataScienceCenter/renku-python"
```
lorenzo-cavazzi, line 126: Let's keep the renku repositories URL in a single place. Check out /utils/constants/Repositories.ts

```suggestion
      componentUrl={RenkuRepositories.Python}
```

client/src/help/HelpRelease.tsx
```diff
@@ -0,0 +1,224 @@
...
+      taggedVersion={data?.coreVersions[0]}
+    />
+  );
+}
+
+function NotebookRelease() {
+  const { data, isFetching } = useGetNotebooksVersionsQuery();
+  return (
+    <ComponentVersion
+      componentUrl="https://github.com/SwissDataScienceCenter/renku-notebooks"
```
lorenzo-cavazzi, line 137: This should also come from `/utils/constants/Repositories.ts`; renku-notebooks still needs to be added there

client/src/help/HelpRelease.tsx
```diff
@@ -0,0 +1,224 @@
...
+      isFetching={isFetching}
+      taggedVersion={data?.version}
+    />
+  );
+}
+
+function RenkuRelease({ chartVersion }: { chartVersion: string }) {
+  const { taggedVersion, devHash } = parseChartVersion(chartVersion);
+  const releaseUrl = componentDocsUrl(
+    "https://github.com/SwissDataScienceCenter/renku",
```
lorenzo-cavazzi, line 147: See above

client/src/help/HelpRelease.tsx
```diff
@@ -0,0 +1,224 @@
...
+      <ExternalLink role="text" title={taggedVersion} url={releaseUrl} />
+      {devHash != null && <span>&nbsp;(dev version {devHash})</span>}
+    </>
+  );
+}
+
+function UiRelease({ uiVersion }: { uiVersion: string }) {
+  const { taggedVersion, devHash } = parseChartVersion(uiVersion);
+  const releaseUrl = componentDocsUrl(
+    "https://github.com/SwissDataScienceCenter/renku-ui",
```
lorenzo-cavazzi, line 162:
```suggestion
    RenkuRepositories.UI,
```

client/src/help/HelpRelease.tsx
```diff
@@ -0,0 +1,224 @@
...
+  isFetching,
+  taggedVersion,
+}: ComponentVersionsProps) {
+  if (isFetching) {
+    return <Loader inline size={16} />;
+  }
+  const releaseUrl =
+    taggedVersion != null
+      ? `${componentUrl}/releases/tag/${taggedVersion}`
+      : `${componentUrl}/releases/`;
```
lorenzo-cavazzi, line 110: I wonder if we should still link to the release tag for dev versions since the link is right on the version. It doesn't matter too much when it's the latest version, but perhaps it's useful for older versions

image

lorenzo-cavazzi (changes requested):
Great to see the version number(s) in the UI :tada: 

The solution is neat and works well already! I added a couple of inline comments, just minor details to address before merging.

client/src/help/HelpRelease.tsx
ciyer: For released versions, it is safe to assume that the version is a tag, but this is not correct for dev versions. There is no tag yet for 0.32.1 -- this will be the version _once it is tagged_.

client/src/help/HelpRelease.tsx
```diff
@@ -0,0 +1,224 @@
...
+        </Col>
+      </Row>
+    </>
+  );
+}
+
+export default HelpRelease;
+
+// For testing
+export { parseChartVersion };
```
leafty, line 224: Exports at definition site

ciyer:
Thanks for the reviews! I think I have addressed the key comments and the PR is ready for re-review.

lorenzo-cavazzi (approved):
Lgtm! :rocket:

PR #2707, opened 2023-08-02
feat: Add project slug tooltip to each link in the linked projects list (#2617)

### LINKED ISSUE #2617: Show projects using datasets with fully qualified names
## Description

When viewing a dataset, there is a list shown of projects using the dataset.

image

The problem is, if multiple projects use the dataset, then it is not possible to distinguish between them. 

As a quick fix, we can show the path with namespace. We should also think about better solutions for showing this information with more structural information.

andre-code:
PR to add a tooltip in the list of linked projects in the dataset view.
The friendly name was already displayed, so we included the slug as a tooltip to differentiate the project name, it is necessary when the dataset is used in several projects that have the same name.

tooltip in project list

Fix #2617 


client/src/components/entities/LinkedEntitiesByItemType.tsx
```diff
@@ -17,6 +17,8 @@
  */
 
 import * as React from "react";
+import { ReactNode, useRef } from "react";
```
lorenzo-cavazzi, line 20: Let's keep a single import from `"react"`

```suggestion
import React,  { ReactNode, useRef } from "react";
```

lorenzo-cavazzi (changes requested):
This works well!
Just a tiny detail in the `import` section.

lorenzo-cavazzi (approved):

PR #2711, opened 2023-08-03
Allow navigating back after login

### LINKED ISSUE #2682: The login flow seems to push one too many pages in the browser history
On the login page, clicking the back button usually goes back to `/login` which redirects to the login form again. This means the user has to quickly press "back" twice to go back one page.

Investigate why there is an extra page in the browser history.

ciyer:
Fix #2682

/deploy #persist #cypress


client/src/authentication/Authentication.container.js
```diff
@@ -130,12 +130,12 @@ class Login extends Component {
...
-    const redirectUrl = LoginHelper.createLoginUrl(url);
+    const redirectParam = encodeURIComponent(LoginHelper.createLoginUrl(url));
+    const uiServerUrl = this.props.params.UISERVER_URL;
+    const authUrl = `${uiServerUrl}/auth/login?redirect_url=${redirectParam}`;
 
     // set new location
-    window.location = `${
-      this.props.params.UISERVER_URL
-    }/auth/login?redirect_url=${encodeURIComponent(redirectUrl)}`;
+    window.location.replace(authUrl);
```
leafty, line 138: Doing this in a render is probably not OK!

An alternative is to render a `<Redirect push to={url} />` for the same effect. (from `react-router`)

leafty (changes requested):

client/src/authentication/Authentication.container.js
ciyer: You are right that having a side-effect like this in the render method is not conforming to the expectations of the framework. 

`react-router` is for routing within a single page application, and the `Redirect` component does not do what we need. In this case, it is necessary to make an HTTP request to get the log in page, since it is served by another application (Keycloak), but `Redirect` just modifies the browser history without triggering a request. 

I have implemented a different solution that plays better with React and achieves the desired result.

client/src/authentication/Authentication.container.js
```diff
@@ -124,20 +124,20 @@ const LoginHelper = {
 };
 
 // always pass "previous" with the current `location.pathname`
-class Login extends Component {
-  render() {
+function Login(props) {
```
leafty, line 127: Would it be OK to pull this out into a new component, e.g. `src/authentication/LoginRedirect.tsx`?

client/src/authentication/Authentication.container.js
leafty: I see. Using `useEffect()` is correct here, as location.replace()` is a side-effect.

leafty (dismissed):

client/src/authentication/Authentication.container.js
ciyer: So, ```

ciyer: So this?
```
function LoginRedirect(props) {
  React.useEffect(() => {
    // build redirect url
    let url = props.params.BASE_URL;
    if (props.location.state && props.location.state.previous)
      url += props.location.state.previous;
    const redirectParam = encodeURIComponent(LoginHelper.createLoginUrl(url));
    const uiServerUrl = props.params.UISERVER_URL;
    const authUrl = `${uiServerUrl}/auth/login?redirect_url=${redirectParam}`;

    // set new location
    window.location.replace(authUrl);
  });
  return <div className="bg-primary h-100"></div>;
}

// always pass "previous" with the current `location.pathname`
function Login(props) {
  return <LoginRedirect {...props} />;
}
```

leafty: Why the wrapping? You would import the new one in `App.js`

leafty (approved):
LGTM 👏  Thanks for fixing this 🎉

client/src/authentication/LoginRedirect.tsx
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from "react";
+import type { Params } from "../App.types";
```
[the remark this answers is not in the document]
ciyer: You are right, in this case, the `import type` is not necessary. 

In general, there are situations where `import type` is necessary even when the target is a type or interface because the compiler cannot tell if it is just a type or actually a value ([announcing-typescript-3-8-beta](https://devblogs.microsoft.com/typescript/announcing-typescript-3-8-beta/#type-only-imports-exports)).

My preference is to always be explicit and use `[import/export] type` when dealing with types, but I recognize that this is a stylistic choice.

PR #2715, opened 2023-08-03
fix(client): restore maintenance page

lorenzo-cavazzi:
Back in the days, there was a mechanism to disable most UI functionalities. It was conceived for urgent maintenances and was meant to prevent working on the UI while backend services were intermittently available (E.G. updates to GitLab).

This hasn't been used in a while because better mechanisms were implemented (we generally rely on Statuspage now) but it can still be useful in some situations and the code it's already there, it just requires minimal updates.

:information_source:  Mind that I'm not running the acceptance tests since the test deployment has the maintenance turned on, therefore no page will be available.

image

/deploy #persists #notest extra-values=ui.client.maintenance=1


olevski (approved):

client/src/features/maintenance/Maintenance.tsx
```diff
@@ -16,47 +16,41 @@
...
-import { MaintenanceNavBar, FooterNavbar } from "./landing";
-import { StatuspageDisplay, isStatusConfigured } from "./statuspage";
+import { FooterNavbar } from "../../landing";
+import { StatuspageDisplay, isStatusConfigured } from "../../statuspage";
 
 interface MaintenanceProps {
   info: string;
 }
 function Maintenance(props: MaintenanceProps) {
   const { info } = props;
```
leafty, line 33:
```suggestion
function Maintenance({ info }: MaintenanceProps) {
```

client/src/features/maintenance/Maintenance.tsx
```diff
@@ -16,47 +16,41 @@
...
-          </section>
-        </main>
+      <div className="min-vh-100 d-flex flex-column">
+        <div className="flex-grow-1">
+          <main role="main" className="container-fluid">
+            <section className="jumbotron-header rounded px-3 px-sm-4 py-3 py-sm-5 text-center mb-3">
+              <h1 className="text-center text-primary">
+                <FontAwesomeIcon icon={faWrench} /> {headerText}{" "}
+                <FontAwesomeIcon icon={faWrench} />
+              </h1>
```
leafty, line 48: No extra `[space]`:

```suggestion
            <h1 className="text-center text-primary">
              <FontAwesomeIcon className="me-3" icon={faWrench} />
              Maintenance ongoing
              <FontAwesomeIcon className="ms-3" icon={faWrench} />
            </h1>
```

client/src/features/maintenance/Maintenance.tsx
```diff
@@ -16,47 +16,41 @@
...
+              <br />
+              <p className="text-center">{body}</p>
```
leafty, line 50: No extra `[newline]`:

```suggestion
              <p className="text-center mt-4">{body}</p>
```

client/src/features/maintenance/Maintenance.tsx
ciyer: The spaces are necessary to achieve the correct rendering. The proposal here yields 
image
 (compare to the screenshot in the issue description).

ciyer (dismissed):
Good idea to revive this! It is obviously not something that is needed everyday, but, **when** needed, it is very helpful to have it available.

client/src/features/maintenance/Maintenance.tsx
leafty: ?!??!? It works as intended for me.

lorenzo-cavazzi: Both solutions work well for me.
Having the margins instead of whitespace with `{" "}` looks a bit cleaner to me, but I'm not sure we should enforce this on old components. I changed it nevertheless.

leafty: My rule of thumb is to select the text and paste it into a text editor. The end result should be "clean" if possible, i.e. respect punctuation, no trailing spaces, etc.

PR #2716, opened 2023-08-03
chore: prevent scaling the client server to zero

olevski:
/deploy #persist


lorenzo-cavazzi:
This is a bit shaky. The UI pod goes down after a few minutes, and a new one reappears. Ironically the UI service says there were 0 restarts, it seems like it's actually scaling down to 0 and then back to 1.

Perhaps there is something wrong with the syntax.

image

image

olevski:
...that is why you saw the ui pod go up and down

PR #2717, opened 2023-08-03
feat: pass gitlab urls to core APIs instead of project ids fetched from the backend cache

### LINKED ISSUE #2713: Phase out cached project IDs
## Motivation

Renku-python is removing support for passing project IDs locally cached in the renku-core service.
We still use that in a few places (probably only the datasets APIs -- double check that!) and we should remove it asap.

## Proposal

If this impacts only the datasets APIs, we might want to convert that to use RTK query

lorenzo-cavazzi:
Phase out the code to invoke deprecated renku-core APIs.
Those were primarily datasets APIs requiring to pass the renku-core cached project id reference.

Mind that the PR includes a few minor fixes on the datasets page:
* Avatar elements take the required vertical space.
* Avatar not accidentally removed when editing a dataset.
* No grey-background cells for tables on the dataset page.
* Stable datasets URL (not everywhere).

/deploy #persist renku-core=develop renku=renku-ui/3.11-acceptance-tests
fix #2713


lorenzo-cavazzi:
The biggest changes are with:
* Creating datasets.
* Editing datasets.
* Deleting datasets.

The logic to import datasets is mostly unchanged.

client/src/utils/types/coreService.types.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export interface CoreErrorResponse {
+  error: CoreErrorContent;
+}
+
+export interface CoreResponse {
+  error?: CoreErrorContent;
+  // ? we should avoid `any`, but this is meant to be re-defined when extending the interface
+  result?: any; // eslint-disable-line @typescript-eslint/no-explicit-any
+}
```
ciyer, line 45: I think defining the structure, but leaving the type of `result` to subtypes of `CoreResponse` could be better handled this way:

```suggestion
export interface CoreResponse<T> {
  error?: CoreErrorContent;
  result?: T;
}
```

client/src/features/project/projectCoreApi.ts
```diff
@@ -298,7 +287,7 @@ export const projectCoreApi = createApi({
         return {
           body: { description: data.description, git_url: data.gitUrl },
           method: "POST",
-          url: `/renku/project.edit`,
+          url: `/project.edit`,
```
ciyer, line 290: Is this not a versioned endpoint?

client/src/dataset/Dataset.present.js
```diff
@@ -356,10 +369,13 @@ function EditDatasetButton({
       className="float-right mb-1"
       id="editDatasetTooltip"
       data-cy="edit-dataset-button"
-      to={{
-        pathname: "modify",
+      to={(location) => ({
+        ...location,
+        pathname: location.pathname.endsWith("/")
+          ? location.pathname + "modify"
+          : location.pathname + "/modify",
```
ciyer, line 376: Could this be shared with the similar code in `ErrorAfterCreation`?

client/src/features/datasets/datasets.types.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+interface DatasetOperationResponse {
+  name: string;
+  remote_branch: string;
+}
+
+export interface DeleteDatasetResponse extends CoreResponse {
+  result?: DatasetOperationResponse;
+}
```
ciyer, line 39: Assuming the changes suggested in `coreService.types`:

```suggestion
export type DeleteDatasetResponse = CoreResponse<DatasetOperationResponse>
```

client/src/features/datasets/datasets.types.ts
```diff
@@ -0,0 +1,85 @@
...
+  file_url?: string;
+  job_id?: string;
+}
+export interface AddFilesOperationResponse extends DatasetOperationResponse {
+  files: DatasetFile[];
+}
+
+export interface AddFilesResponse extends CoreResponse {
+  result?: AddFilesOperationResponse;
+}
```
ciyer, line 79: Assuming the changes suggested in coreService.types:

```suggestion
export type AddFilesResponse = CoreResponse<AddFilesOperationResponse>
```

client/src/features/datasets/datasets.types.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export interface PostDataset {
+  name: string;
+  remoteBranch: string;
+}
+
+export interface PostDatasetResponse extends CoreResponse {
+  result?: DatasetOperationResponse;
+}
```
ciyer, line 60: Assuming the changes suggested in coreService.types:

```suggestion
export type PostDatasetResponse = CoreResponse<DatasetOperationResponse>;
```

client/src/features/project/dataset/DatasetModify.tsx
```diff
@@ -369,23 +317,15 @@ function formatServerWarning(
...
 function formatServerError(error: ServerError) {
   if (error.source === "jobs") return null;
   if (error.source === "edit")
     return (
       <div>
-        <p>
-          Metadata changes were successfully applied, but some files could not
-          be updated.
-        </p>
+        <p>Some files could not be updated.</p>
```
ciyer, line 328: Is the old message no longer accurate? If it is, I think we should communicate to the user that some of their changes were applied, but the files in the dataset were not changed.

ciyer (changes requested):
This is a much needed and long overdue refactoring! For the most part, it looks good to me. I have a few small suggestions for improvements to the code, but that is all.

client/src/features/project/projectCoreApi.ts
lorenzo-cavazzi: Good catch, this should be versioned. I'll move that to another PR since it touches a different functionality

ciyer (approved):
Fantastic!

PR #2729, opened 2023-08-10
minor: remove Knowledge Graph from user message

ciyer:
Parallel to the PR that removed the phrase "Knowledge Graph" from user messages, another one introduced it -- this PR should fix that.

/deploy #persist


ciyer:
# Screenshots

Wording updated in many places, here are a few examples

image

image

image

image

image

client/src/features/dashboard/components/InactiveKgProjects.tsx
```diff
@@ -46,7 +46,7 @@ export function ProjectsInactiveKGWarning() {
   return (
     <WarnAlert>
       <div data-cy="inactive-kg-project-alert">
-        You have {totalProjects} projects that are not in the Knowledge Graph.{" "}
+        You have {totalProjects} projects with deactivated metadata processing.{" "}
```
lorenzo-cavazzi, line 49: nitpick: I'm unsure about using the word `deactivated` here, it sounds like it can be turned on and off.

Would it make sense to re-phrase it to say that processing was never activated? Something like

> Metadata processing isn't activated yet for {totalProjects} projects.

I would slightly change the message at line 50 too to avoid saying `Activate your projects`; perhaps `Activate metadata processing [...]` or just `Activate it [...]`.

lorenzo-cavazzi (commented):
Thanks for these changes!
I suggest not using the word `deactivated` since metadata processing cannot be turned on/off; I suggested a few alternatives (see inline) but feel free to re-word better -- or disagree with this point

lorenzo-cavazzi (approved):
Lgtm! :rocket: 

I realized from this PR that we have quite some duplicates for the indexing components that we might easily remove now that we have RTK query handing the status.
Ofc this is out of scope for this PR.

PR #2730, opened 2023-08-10
minor: support dev versions of core and notebooks

ciyer:
If dev versions of core and/or notebooks were deployed, they were incorrectly displayed as tagged versions. This PR fixes that problem.

image

### DIFF
--- a/client/src/help/HelpRelease.tsx
+++ b/client/src/help/HelpRelease.tsx
@@ -81,34 +81,6 @@ function parseChartVersion(version: string | undefined) {
   };
 }
 
-type ComponentVersionProps = {
-  componentUrl: string /* URL of the component's repository, without a trailing slash */;
-  isFetching: boolean;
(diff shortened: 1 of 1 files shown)


client/src/help/HelpRelease.tsx
```diff
@@ -163,6 +146,7 @@ function RenkuRelease({ chartVersion }: { chartVersion: string }) {
     <ComponentAndDevVersion
       componentUrl={RenkuRepositories.Renku}
       devHash={devHash}
+      isFetching={false}
```
leafty, line 149: Nitpick: if you type the prop as `isFetching?: boolean;` then you can omit it here.

leafty: An alternative is to handle this in the same component as where you use the RTK query hook, and remove this prop. This way of handling loading/fetching is usually more intuitive to read.

leafty (approved):

PR #2732, opened 2023-08-11
Log redis failures during operations 

### LINKED ISSUE #2610: Failed writes to Redis are potentially ignored
The Redis adapter in the ui-server returns true or false [here](https://github.com/SwissDataScienceCenter/renku-ui/blob/master/server/src/storage/RedisStorage.ts#L97).

But in several places the result from the `save` method is not checked. And the code assumes that the write to Redis was successful. Or perhaps Redis raises an error if the write was not successful. If Redis raises an error on failed writes then it would be good to check the code and ensure that ignoring the return value from `storage.save` is correct or some checks should be added. Or if the true/false value being returned from the save method indicates a successful write to Redis then we should be always checking the output and raising an appropriate error.

Here is where this happens:
- https://github.com/SwissDataScienceCenter/renku-ui/blob/master/server/src/authentication/index.ts#L182
- https://github.com/SwissDataScienceCenter/renku-ui/blob/master/server/src/authentication/index.ts#L189
- https://github.com/SwissDataScienceCenter/renku-ui/blob/master/server/src/authentication/index.ts#L365 (indirectly but the return value is still ignored)
- https://github.com/SwissDataScienceCenter/renku-ui/blob/master/server/src/utils/middlewares/lastProjectsMiddleware.ts#L52

There are a few more cases.

ciyer:
Check return values from redis and log errors when appropriate.

Fix #2610

### DIFF
--- a/server/src/utils/middlewares/lastProjectsMiddleware.ts
+++ b/server/src/utils/middlewares/lastProjectsMiddleware.ts
@@ -16,10 +16,13 @@
  * limitations under the License.
  */
 
-import { Storage, TypeData } from "../../storage";
+import * as Sentry from "@sentry/node";
 import express from "express";
-import config from "../../config";
+
 import { getUserIdFromToken } from "../../authentication";
+import config from "../../config";
+import logger from "../../logger";
+import { Storage, TypeData } from "../../storage";
 
 function projectNameIsId(projectName: string): boolean {
   return projectName.match(/^[0-9]*$/) !== null;
@@ -49,15 +52,22 @@ const lastProjectsMiddleware =
 
         const userId = getUserIdFromToken(token);
         // Save as ordered collection
-        storage.save(
-          `${config.data.projectsStoragePrefix}${userId}`,
-          projectName,
-          {
+        storage
+          .save(`${config.data.projectsStoragePrefix}${userId}`, projectName, {
             type: TypeData.Collections,
             limit: config.data.projectsDefaultLength,
             score: Date.now(),
-          }
-        );
+          })
+          .then((value) => {
+            if (!value) {
+              const errorMessage = `Error saving project ${projectName} for user ${userId}`;
+              logger.error(errorMessage);
+              Sentry.captureMessage(errorMessage);
+            }
+          })
+          .catch((err) => {
+            Sentry.captureException(err);
+          });
       });
     }
     next();
--- a/server/src/authentication/index.ts
+++ b/server/src/authentication/index.ts
@@ -16,6 +16,7 @@
  * limitations under the License.
  */
 
+import * as Sentry from "@sentry/node";
 import express from "express";
 import { Issuer, generators, Client, TokenSet } from "openid-client";
 
@@ -35,6 +36,11 @@ import jwt from "jsonwebtoken";
 const verifierSuffix = "-verifier";
 const parametersSuffix = "-parameters";
 const maxAttempts = config.auth.retryConnectionAttempts;
+
+type GetStorageValueReturn = {
+  storageKey: string;
+  value: string | null;
+};
 
 class Authenticator {
   authServerUrl: string;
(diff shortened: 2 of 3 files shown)


server/src/utils/middlewares/lastProjectsMiddleware.ts
```diff
@@ -49,15 +52,22 @@ const lastProjectsMiddleware =
...
+          .then((value) => {
+            if (!value) {
+              const errorMessage = `Error saving project ${projectName} for user ${userId}`;
+              logger.error(errorMessage);
+              Sentry.captureMessage(errorMessage);
+            }
+          })
+          .catch((err) => {
+            Sentry.captureException(err);
+          });
```
lorenzo-cavazzi, line 70: The code here makes sense to me, but when I try to simulate a storage failure just before navigating to a project from the UI (I manually delete all the Redis instances from Rancher) I get a `500 internal error` response; from the code, I assume it should log the error to Sentry and jump to the next step.

image

Perhaps something with `storage.save` is off?

lorenzo-cavazzi (changes requested):
The code makes sense to me, but I spotted a strange case with the "lastVisistedProjects" middleware (more info inline).

After a few more trials, I wonder if the problem is in the `RedisStorage` class. I got the following error when simulating a failure; we are not catching exceptions from `ioredis`, perhaps we should wrap `this.redis.<someAction>` around a `try/catch`

```
[start:dev-debug] [ioredis] Unhandled error event: Error: All sentinels are unreachable. Retrying from scratch after 10ms. Last error: Connection is closed.
[start:dev-debug]     at SentinelConnector.<anonymous> (/home/lorenzo/development/sdsc/renku-ui/server/node_modules/ioredis/built/connectors/SentinelConnector/index.js:73:31)
[start:dev-debug]     at Generator.next (<anonymous>)
[start:dev-debug]     at /home/lorenzo/development/sdsc/renku-ui/server/node_modules/ioredis/built/connectors/SentinelConnector/index.js:8:71
[start:dev-debug]     at new Promise (<anonymous>)
[start:dev-debug]     at __awaiter (/home/lorenzo/development/sdsc/renku-ui/server/node_modules/ioredis/built/connectors/SentinelConnector/index.js:4:12)
[start:dev-debug]     at connectToNext (/home/lorenzo/development/sdsc/renku-ui/server/node_modules/ioredis/built/connectors/SentinelConnector/index.js:59:37)
[start:dev-debug]     at SentinelConnector.<anonymous> (/home/lorenzo/development/sdsc/renku-ui/server/node_modules/ioredis/built/connectors/SentinelConnector/index.js:125:24)
[start:dev-debug]     at Generator.throw (<anonymous>)
[start:dev-debug]     at rejected (/home/lorenzo/development/sdsc/renku-ui/server/node_modules/ioredis/built/connectors/SentinelConnector/index.js:6:65)
```

leafty:
> The code makes sense to me, but I spotted a strange case with the "lastVisistedProjects" middleware (more info inline).
> 
> After a few more trials, I wonder if the problem is in the `RedisStorage` class. I got the following error when simulating a failure; we are not catching exceptions from `ioredis`, perhaps we should wrap `this.redis.<someAction>` around a `try/catch`
> 
> ```
> [start:dev-debug] [ioredis] Unhandled error event: Error: All sentinels are unreachable. Retrying from scratch after 10ms. Last error: Connection is closed.
> [start:dev-debug]     at SentinelConnector.<anonymous> (/home/lorenzo/development/sdsc/renku-ui/server/node_modules/ioredis/built/connectors/SentinelConnector/index.js:73:31)
> [start:dev-debug]     at Generator.next (<anonymous>)
> [start:dev-debug]     at /home/lorenzo/development/sdsc/renku-ui/server/node_modules/ioredis/built/connectors/SentinelConnector/index.js:8:71
> [start:dev-debug]     at new Promise (<anonymous>)
> [start:dev-debug]     at __awaiter (/home/lorenzo/development/sdsc/renku-ui/server/node_modules/ioredis/built/connectors/SentinelConnector/index.js:4:12)
> [start:dev-debug]     at connectToNext (/home/lorenzo/development/sdsc/renku-ui/server/node_modules/ioredis/built/connectors/SentinelConnector/index.js:59:37)
> [start:dev-debug]     at SentinelConnector.<anonymous> (/home/lorenzo/development/sdsc/renku-ui/server/node_modules/ioredis/built/connectors/SentinelConnector/index.js:125:24)
> [start:dev-debug]     at Generator.throw (<anonymous>)
> [start:dev-debug]     at rejected (/home/lorenzo/development/sdsc/renku-ui/server/node_modules/ioredis/built/connectors/SentinelConnector/index.js:6:65)
> ```

Just a non-blocking remark here: wrappers like the RedisStorage class are usually prone to create bugs and issues over time. I would rather use the redis client library directly and handle error states at the call site.

lorenzo-cavazzi:
> wrappers like the RedisStorage class are usually prone to create bugs and issues over time.

I agree with this.
The main driver to have a wrapper was the idea to move away from Redis soon (delegate auth to Gateway, and use permanent storage for everything else). How confusing are responses from `ioresdis` was another point for that solution. Ofc that played out badly and in insight I would not go for this solution again.

At this point we will probably repurpose the UI server soon, I would say we can keep the current approach for the time being.

lorenzo-cavazzi (approved):
Lgtm! :rocket:

PR #2737, opened 2023-08-15
Persistent sessions: session button, session cards, new session message

leafty:
Details:
* The session card will now tell if a paused session has unsaved work
* The session button now fully supports persistent session
* Pause session and delete session are both offered in the drop-down
* Start session page checks whether a session exists

Note that the share link functionality will be updated and fixed in the next PR

Session card:
image

Session button drop-down:
image

Delete warning:
image

Pause modal:
Screenshot 2023-08-15 at 13 08 04

Start session page:
renku-ci-ui-2737 dev renku ch_projects_johann thiebaut1_another-playground-project_sessions_new

/deploy #persist extra-values=notebooks.userSessionPersistentVolumes.enabled=true,notebooks.amalthea.image.repository=renku/amalthea,notebooks.amalthea.image.tag=db70f80ef1983d49a1b167faa57d02d9d8b9554f renku-notebooks=pitch/session-persistence


client/src/features/session/components/ShowSession.tsx
```diff
@@ -396,9 +397,13 @@ function StopSessionBtn({ toggleStopSession }: StopSessionBtnProps) {
 
   const ref = useRef<HTMLButtonElement>(null);
 
-  const buttonId = logged ? "stop-session-button" : "delete-session-button";
-  const icon = logged ? faStop : faTrash;
-  const tooltip = logged ? "Stop session" : "Delete session";
+  const buttonId = logged ? "pause-session-button" : "delete-session-button";
+  const icon = logged ? (
+    <SessionPausedIcon className="text-rk-dark" size={16} />
```
lorenzo-cavazzi, line 402: Since `FontAwesomeIcon` adds a few classes, this looks misaligned on the dropdown menu. The easiest thing would be to add the missing classes here.
For the horizontal alignment we can add `fa-fw me-2`; for the vertical one `svg-inline--fa`

lorenzo-cavazzi (approved):
This works very well :rocket: 

I have a comment, but I feel we don't need to spend time on it.

* The header and session tab buttons are temporarily out of sync. I guess overriding the local state temporarily is easier than doing that in redux; if a solution to have them synced would be too tedious, we can live with this -- especially since it lasts just very few seconds
image

client/src/features/session/components/ShowSession.tsx
leafty: Done in [84739b3](https://github.com/SwissDataScienceCenter/renku-ui/pull/2737/commits/84739b36924ad31f29ecff882983179352127c05).

leafty:
> This works very well 🚀
> 
> I have a comment, but I feel we don't need to spend time on it.
> 
> * The header and session tab buttons are temporarily out of sync. I guess overriding the local state temporarily is easier than doing that in redux; if a solution to have them synced would be too tedious, we can live with this -- especially since it lasts just very few seconds

Yes, I think overriding the state from RTK query globally is not that easy. I would have to add a specific slice for it or a fake API call in RTK query.

Issue #2212, opened 2022-12-13
Better take advantage of storybook


ciyer:
Follow up to #2163.

- [ ] Upgrade from storybook 6.4.19 to latest storybook ~(>=6.5.14)~ 7+ (_updated on 27.07.23_)
    - ℹ️ last time we tried, there were a few issues. After upgrading the libraries, try to remove the `node_modules` folder, `npm install` again, and run telepresence to connect to a deployment
- [ ] Decide on and implement the best testing strategy (see below)
- [ ] Move the style-guide to storybook

We should also figure out how we want to take advantage of storybook for testing. Here are two options.

1. Jest-based: using storybook/testing-library integration https://storybook.js.org/blog/interaction-testing-with-storybook/ 
2. Cypress-based: start storybook (similar to how we start the UI) and click around the components

PR #2748, opened 2023-08-17
test(storybook): add test runner and interactive stories (#2212)

andre-code:
This pull request introduces **Storybook  Interactive stories**  that are useful when we want to demonstrate and verify the behavior of the components in a realistic and dynamic manner. 

And also include **Storybook test runner**, which transforms our stories into executable tests, covering both Chromium and Firefox browsers.

- For those without a play function: it verifies whether the story renders without any errors.
- For those with a play function(interactive stories): it also checks for errors in the play function and that all assertions passed.

## How to test
To run tests `npm run test-storybook`

For interactive stories:

interaction stories

Note: Pending Visual tests to cover in other PR

/deploy #persist


client/package.json
```diff
@@ -88,7 +88,8 @@
     "format-changed": "pretty-quick --pattern \"client/**/*.*\"",
     "format-commit": "pretty-quick --staged --pattern \"client/**/*.*\"",
     "storybook": "storybook dev -p 6006 -s public",
-    "build-storybook": "storybook build -s public"
+    "build-storybook": "storybook build -s public",
+    "test-storybook": "test-storybook --browsers firefox chromium"
```
lorenzo-cavazzi, line 92: Should this also start storybook? 
If I run just `test-storybook`, I get

> [test-storybook] It seems that your Storybook instance is not running at: http://127.0.0.1:6006. Are you sure it's running?

client/src/components/form-field/CreatorsInput.tsx
```diff
@@ -278,6 +280,7 @@ export function FormGeneratorCreatorsInput({
         <Col>
           <Button
             data-cy="addCreatorButton"
+            data-testId="creator-add"
```
lorenzo-cavazzi, line 283: Could we re-use the `data-cy` here since it's about testing? Or perhaps we should have a generic `data-test`, but I would not distinguish between Cypress or Storybook, it seems redundant

client/src/components/form-field/CreatorsInput.tsx
```diff
@@ -129,6 +130,7 @@ function CreatorForm(props: CreatorFormProps) {
             disabled={props.disabled}
             className="mb-3 p-0 border-0 text-danger bg-transparent"
             onClick={props.deleteCreator}
+            data-testId="creator-delete"
```
lorenzo-cavazzi, line 133: same as above

lorenzo-cavazzi (changes requested):
This is very interesting!

client/src/components/form-field/CreatorsInput.tsx
andre-code: Storybook test uses Jest, which has limited query capabilities, in this scenario, we can address the need to utilize an attribute other than data-testId by implementing a  [Custom Queries in DOM Testing Library](https://testing-library.com/docs/dom-testing-library/api-custom-queries/).

In my view, maintaining consistency with the choice of a single attribute for testing purposes is important. While data-cy is closely associated with Cypress, it would be advantageous to modify the tests across different repositories (both Cypress in the UI and Renku) to employ the data-test attribute. However, this might warrant a separate issue.

At the moment, we have the option to either work with the default attribute used by Jest or to create a custom query. I am open to both solutions.

client/package.json
andre-code: yes,  test runner requires either a locally running Storybook instance or a published Storybook to run the tests.

lorenzo-cavazzi: Ok. Can we make the action start storybook and run the tests without requiring 2 separate steps?

client/src/components/form-field/CreatorsInput.tsx
lorenzo-cavazzi: Thanks for pointing me to that documentation.

I thought about this and I feel implementing a selector to reuse `data-cy` would be nice. That way, we will have only one custom attribute for all tests. Also, that might incentives us to use it even more and it will automatically be available to other set of tests (E.G: the Cypress acceptance tests)

client/package.json
andre-code: This might be a bit tricky to implement straightforwardly. Even if I utilize tools like `concurrently` to wait for server accessibility, there's still the possibility that the server might be reachable but not fully prepared for testing.

client/src/components/form-field/CreatorsInput.tsx
andre-code: I have found a way to do it without using the custom queries 🙂

README.md
```diff
@@ -171,6 +171,15 @@ The full documentation for Storybook can be found [here](https://storybook.js.or
 - Wherever relevant, please include stories that demonstrate responsive behavior across various devices.
 - You can use `addon-redux` for state management on components that require to get data from the Redis store.
 
+4. Testing Stories:
```
lorenzo-cavazzi, line 174: Let's adapt this to the current format from the `2298-fix-broken-components` branch: 

```suggestion
#### Test stories
```

The content might need to be slightly adjusted if we for for this solution https://github.com/SwissDataScienceCenter/renku-ui/pull/2748/files#r1309106040

lorenzo-cavazzi:
This looks good! And we are on the right way!
Give a look at the comment here https://github.com/SwissDataScienceCenter/renku-ui/pull/2748#discussion_r1309106040 (sorry I messed up when using the review buttons so this is just a comment now :facepalm: )

lorenzo-cavazzi (commented):
This looks good now :rocket: 

I added a few commits to streamline some details (readme formatting, new npm commands, gitignore file). I pushed them directly since it's just formatting/chores, but ofc feel free to change anything you don't like from my last changes.

Storybook tests work well now :raised_hands: 
The last missing piece is running them in a GitHub action. Since we don't need any existing deployment, we can run them in the `test-and-ci.yml` action. I suggest adding a `run-storybook` job similar to `run-e2e`.
The setup should be simple since we just need to checkout, setup node, and run the new all-in-one action to build and test.
P.S. if you need a reference, there is a section in the Storybook documentation https://storybook.js.org/tutorials/ui-testing-handbook/react/en/automate/
We need only a subset of the `interaction-and-accessibility` job; it's probably simpler to copy from our other existing jobs. Text me if you need any help

andre-code:
@lorenzo-cavazzi Thanks for your review! I appreciate your suggestion regarding the tests. I've incorporated it into the PR #2773 , which also includes the deploy. This way, we can easily run the tests on the deployed Storybook.

README.md
```diff
@@ -171,6 +171,12 @@ The full documentation for Storybook can be found [here](https://storybook.js.or
 - Wherever relevant, please include stories that demonstrate responsive behavior across various devices.
 - You can use `addon-redux` for state management on components that require to get data from the Redis store.
 
+#### Testing Stories:
+
+Storybook's test runner transforms stories into executable tests, supporting Chromium and Firefox browsers. It ensures error-free rendering for non-play stories. For interactive stories with play functions, it validates both play function errors and successful assertions.
+
+    $ npm run start-storybook-and-test
```
lorenzo-cavazzi, line 178: We changed the command here

```suggestion
    $ npm run storybook-compile-and-test
```

lorenzo-cavazzi (changes requested):
Great, thank you for adding the tests in the GitHub action!
Just a tiny change on the Readme file and this can be merged :rocket:

lorenzo-cavazzi (approved):
:rocket:

lorenzo-cavazzi (approved):
Let's merge this :star_struck:

PR #2751, opened 2023-08-18
chore(client): update eslint rules about jsx runtime

### LINKED ISSUE #2576: Update eslint rule to not require React import
The current React version and webpack build chain do not require our source files to include the global React import.

```ts
// This import is not needed
import React from "react";
// This is also not needed
```

Note that the current situation is creating a conflict between vscode `Organize imports` commands and `eslint` rules. The former removes the import (as it is not needed) while `eslint` requires it.

leafty:
Fixes #2576.

Importing `React` is not necessary as we are using `react >= 18`. Also, vscode often detects it as an unused dependency which clashes with the previous linting rules.

After this change is merged, we can use `Organize imports` in vscode to clean-up and keep imports organized.

/deploy renku=renku-ui/3.11-acceptance-tests #persist


lorenzo-cavazzi (approved):
Lgtm :rocket: 
P.S. I noticed there are a few conflicts; I'll re-approve after resolving them.

lorenzo-cavazzi (approved):

PR #2757, opened 2023-08-22
fix(client): silence session warning about mem_request

leafty:


lorenzo-cavazzi (approved):
This fix works well! Let's merge it :rocket:

PR #2766, opened 2023-08-24
fix(client): use the migrations_check endpoint on the latest core version

lorenzo-cavazzi:
This fixes a bug accidentally introduced in #2717 
/deploy #persist


ciyer (approved):
👍

PR #2772, opened 2023-08-29
chore(client): fix uses of bootstrap icons

leafty:
Details:
* remove icon imports from `ts-wrappers.js`
* adjust icon alignment in entity header card
* add "bi" (bootstrap icon) class name for inline use with text

Screenshot 2023-09-05 at 15 45 34

/deploy #persist


ciyer (approved):
Looks good, nice cleanup!

PR #2776, opened 2023-08-31
fix: correctly determine if a deployment is a dev deployment

ciyer:
There is a mistake in the logic to determine if a deployment is a dev deployment.

**Before**

image

**After**

image

### DIFF
--- a/client/src/landing/NavBar.js
+++ b/client/src/landing/NavBar.js
@@ -545,9 +545,9 @@ function FooterNavbar({ location, params }) {
   const displayVersion =
     taggedVersion == null
       ? "unknown"
-      : isDevVersion == null
-      ? taggedVersion
-      : `${taggedVersion} (dev)`;
+      : isDevVersion
+      ? `${taggedVersion} (dev)`
+      : taggedVersion;
 
   const footer = (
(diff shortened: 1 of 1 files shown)


client/src/landing/NavBar.js
```diff
@@ -545,9 +545,9 @@ function FooterNavbar({ location, params }) {
   const displayVersion =
     taggedVersion == null
       ? "unknown"
-      : isDevVersion == null
-      ? taggedVersion
-      : `${taggedVersion} (dev)`;
+      : isDevVersion
```
leafty, line 548: This should be more robust (checked in a `.ts` file):

```js
  const chartVersion = params && params["RENKU_CHART_VERSION"];
  const parsedChartVersion = chartVersion ? parseChartVersion(chartVersion) : null;
  const taggedVersion = parsedChartVersion?.taggedVersion;
  const isDevVersion = parsedChartVersion?.devHash != null;
  const displayVersion =
    taggedVersion == null
      ? "unknown"
      : isDevVersion
      ? `${taggedVersion} (dev)`
      : taggedVersion;
```

lorenzo-cavazzi (approved):

PR #2786, opened 2023-09-06
fix: unzip multiple dataset files (#2754)

### LINKED ISSUE #2754: Unzipped dataset files aren't handled correctly.
## Description

The UI allows users to unzip when loading dataset files. That is currently not working properly because the code in the "client/src/components/form-field/DropzoneFileUploader.js" file doesn't handle multiple files correctly.
The code in `onCompletedUpload` filters only the first file `let resultFile = resultFiles ? resultFiles[0] : null;` while all the files should be handled in a for loop.

## Reproduce

Try to upload a zipped file that contains at least 2 files. That will result in an error cause the `onCompletedUpload` function will return a file with multiple `file_id` and later the `POST /renku/datasets.add` API will fail.

:warning:  Mind that some zip structures work because specific files are ignored. E.G: a zip with just one file and any number of subfolders should work because subfolders are ignored. The final dataset won't contain the whole zip anyway.

andre-code:
Pull request for resolving the issue with uploading dataset files in the "Add/Modify Dataset" form. 

1. Issue with displaying subfolders in unzipped dataset files.

* Before
expand files Error
* After
expand files OK

2. Error when sending several dataset files
* Before
send files error

* After
send files ok

Fix #2754 

/deploy #persist

### DIFF
--- a/client/src/components/form-field/DropzoneFileUploader.js
+++ b/client/src/components/form-field/DropzoneFileUploader.js
@@ -773,7 +773,7 @@ function FileStatusComp({ file, uploadCompressedFile, uploadThresholdSoft }) {
           icon={faExclamationTriangle}
         />
         <span className="mb-1">&nbsp;Unzip on upload?</span>
-        <span className="me-1">
+        <span className="me-1 d-flex  gap-2">
           <span
             data-cy="upload-compressed-yes"
             className="text-primary text-button"
@@ -786,7 +786,7 @@ function FileStatusComp({ file, uploadCompressedFile, uploadThresholdSoft }) {
             }
           >
             Yes
-          </span>{" "}
+          </span>
           or
           <span
             data-cy="upload-compressed-no"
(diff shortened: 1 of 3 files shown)


client/src/components/form-field/DropzoneFileUploader.js
```diff
@@ -773,7 +773,7 @@ function FileStatusComp({ file, uploadCompressedFile, uploadThresholdSoft }) {
           icon={faExclamationTriangle}
         />
         <span className="mb-1">&nbsp;Unzip on upload?</span>
-        <span className="me-1">
+        <span className="me-1 d-flex  gap-2">
```
lorenzo-cavazzi, line 776:
```suggestion
        <span className="me-1 d-flex gap-2">
```

lorenzo-cavazzi (approved):
Great fix, it works well now! :rocket: 
Tiny fix inline (just an extra space to remove).

lorenzo-cavazzi (approved):
:rocket:

PR #2789, opened 2023-09-07
fix(client): handle embedded template variables in project creation links

lorenzo-cavazzi:
There was a rather annoying bug causing an infinite loop when using project creation links embedding custom variables.
Actually, some of the `useEffect` hooks in the ProjectNew container were triggered indefinitely (it would have been easier to spot if we didn't explicitly ignore the `react-hooks/exhaustive-deps` warnings...) but that was not causing issues in most other cases.

This PR also introduces better feedback messages when using the project creation links (see Cypress tests for examples).

Cypress tests were added to be sure we don't accidentally introduce further regressions on the feedback messages or the optional template variables.

/deploy #persist

### DIFF
--- a/client/src/project/new/ProjectNew.container.js
+++ b/client/src/project/new/ProjectNew.container.js
@@ -23,7 +23,14 @@
  *  Container components for new project
  */
 
-import { useEffect, useState, useRef, useContext, Component } from "react";
+import {
+  useCallback,
+  useEffect,
+  useState,
+  useRef,
+  useContext,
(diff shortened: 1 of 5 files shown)


leafty (approved):
LGTM for the code changes.

We can merge and release this. Updates to e2e tests will be reviewed later.

ciyer (approved):
LGTM! (And yes, we do need fix code that ignores react-hook lint rules)

PR #2790, opened 2023-09-07
build: release 3.12.1

lorenzo-cavazzi:
This is a bugfix release including the fix from 3.11.1
Using the renku-core versioning is also included, but that's probably not enough to mark this as a full release: the change is nothing visible to users, more of a refactoring on how we invoke renku-core APIs.

/deploy #persist


ciyer (approved):
👍

PR #2791, opened 2023-09-11
fix(client): avatar url

### LINKED ISSUE #2765: Support avatars with external URLs
## Description

Users can set avatars to external images from the CLI. In that case, the `datasets.list` API returns a response containing something like

```
{
    "images": [{
        "content_url": "https://zenodo.org/static/img/logos/zenodo-gradient-round.svg",
        "position": 1
    }]
}
```

Currently, we take it for granted the `content_url` is a relative path to the GitLab repository. As a consequence, we try to fetch from

> https://gitlab.dev.renku.ch/lorenzo.cavazzi.tech/test-dataset-avatars/-/raw/master/https://zenodo.org/static/img/logos/zenodo-gradient-round.svg

We should check if the `content_url` is formatted as a URL and in that case fetch it from the source.

## Reproduce

Try with https://dev.renku.ch/projects/lorenzo.cavazzi.tech/test-dataset-avatars/datasets

andre-code:
PR to resolve the dataset avatar URL issue. GitHub URL is included only if the content_url is a GitHub URL.
### Testing
Every dataset in this project should display an image. https://renku-ci-ui-2791.dev.renku.ch/projects/lorenzo.cavazzi.tech/test-dataset-avatars/datasets

Fix #2765 

/deploy #persist


client/src/utils/helpers/DatasetFunctions.ts
leafty, line 1: [CODING_GUIDELINES.md](https://github.com/SwissDataScienceCenter/renku-ui/blob/master/CODING_GUIDELINES.md#r005-file-naming-conventions)

This file should probably be renamed to `dataset.utils.ts` or similar.

client/src/utils/helpers/Dataset.utils.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+/**
+ * Add the URL for the marquee image to the dataset if is an image git url. Modifies the dataset object.
+ * @param {string} gitUrl
+ * @param {Dataset} dataset
+ */
+export function addMarqueeImageToDataset(gitUrl: string, dataset: IDataset) {
+  const urlRoot = cleanGitUrl(gitUrl) + "/-/raw/master";
+  let mediaUrl = dataset?.images?.[0]?.content_url ?? null;
+  if (mediaUrl !== null && mediaUrl.startsWith(".renku/dataset_images/"))
```
leafty, line 36: I find this logic non-intuitive here. Why do we not consider any relative path inside the repository? Non-relative locations would start with a URL scheme (`<scheme>://`). Also, we should probably validate external URLs and only allow those starting with `https://`.

leafty (changes requested):

client/src/utils/helpers/Dataset.utils.ts
leafty: My suggestion is to use the [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) class:

```ts
const mediaUrl = new URL("https://example.com/image.jpg", "https://gitlab.renkulab.io/namespace/name/-/raw/main/");
mediaUrl.toString() // 'https://example.com/image.jpg'
```

```ts
const mediaUrl = new URL("path/to/image.jpg", "https://gitlab.renkulab.io/namespace/name/-/raw/main/");
mediaUrl.toString() // 'https://gitlab.renkulab.io/namespace/name/-/raw/main/path/to/image.jpg'
```

```ts
const mediaUrl = new URL("s3://example/image.jpg", "https://gitlab.renkulab.io/namespace/name/-/raw/main/");
mediaUrl.toString() // 's3://example/image.jpg'
```

The URL constructor will handle relative vs absolute locations for you without needing to re-implement it.

client/src/utils/helpers/Dataset.utils.ts
```diff
@@ -0,0 +1,51 @@
...
+    mediaUrl = `${urlRoot}/${mediaUrl}`;
+
+  dataset.mediaContent = mediaUrl;
+  return dataset;
```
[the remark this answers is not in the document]
andre-code: I understand the suggestion to avoid mutating arguments in functions. In this specific context, we intend to mutate the object when fetching datasets in order to add the mediaContent property. While we could refactor this code chain to return a new object, it's not within the scope of this PR. We may consider making such a change when transitioning to the use of Redux Toolkit

leafty (approved):

PR #2796, opened 2023-09-14
fix(client): remove extra title on hover for session buttons

leafty:
Removes the extra title tooltip on hover and adds hidden text for accessibility.

Removes this:
image


lorenzo-cavazzi (approved):
Lgtm :heavy_check_mark:

PR #2800, opened 2023-09-15
fix adding files by URL to datasets

ciyer:
Adding files to a dataset by URL was broken, it should be fixed now

image

image

/deploy #persist


lorenzo-cavazzi (approved):
The fix works well! :rocket: 

The URL validation function is a bit too strict, but that's not related
image

PR #2802, opened 2023-09-18
chore: test new renku-actions

lorenzo-cavazzi:
Testing changes to renku-actions

/deploy


lorenzo-cavazzi:
This is not necessary anymore. We can merge https://github.com/SwissDataScienceCenter/renku-ui/pull/2804 instead

Issue #2778, opened 2023-08-31
Switch to using the new KG project update API


lorenzo-cavazzi:
## Motivation

There is a new API from KG to update visibility, description, keywords, images/avatars
> PATCH `/projects/{project_full_slug}`

Changes will be instantaneous on KG and core and we won't need to invoke the core APIs anymore.

## Proposal

We should use that for updating:
- [ ] Description (also remove the logic in the RTK mutation to invalidate the KG tag after a few seconds to re-fetch from KG; just invalidate the tag immediately)
- [ ] Visibility
- [ ] Keywords/tags
- [ ] Images (this might be more time consuming -- defer the change to another issue if that's the case)

Bonus: handle protected branches, or create a follow-up issue to do that. KG throws a 409 error when the main branch is protected, and will return a `data` variable with the new branch (mind it's not in their first implementation).
In the UI, we could reuse the same logic we use in datasets to deal with that case.

PR #2806, opened 2023-09-20
feat: use kg to update avatar (#2778)

andre-code:
PR to use kg api to update project's avatar.

Contents of this PR:
- Modification of the updateProject mutation in the projectKgApi to override the project-kg-metadata tag when a project is modified. Avatar should be updated in the Project Header immediately after the change.
- Introduction of a new component for online submit Image input.
- Integration of a storybook for the online submit image input component: [Storybook Link](https://renku-ci-ui-2806.dev.renku.ch/storybook/?path=/docs/components-forms-inlinesubmitimageinput--docs).

update project avatar

/deploy renku-graph=development  #persist


client/src/features/project/components/ProjectSettings.tsx
```diff
@@ -65,35 +83,52 @@ export function ProjectSettingsGeneral(props: ProjectSettingsGeneralProps) {
...
-        userLogged={props.user.logged}
-      />
+      {props.settingsReadOnly ? (
+        <InfoAlert dismissible={false} timeout={0}>
+          <p className="mb-0">
+            Project settings can be changed only by maintainers.
+          </p>
+          {loginElement}
+        </InfoAlert>
+      ) : (
```
lorenzo-cavazzi, line 93: I'm not sure if this is the best approach. If we do this, users without permission will not see the settings at all. I suggest disabling them instead as it already happens for Visibility and Description.
P.S. I realize I missed this when reviewing the other PR
image

client/src/project/shared/ProjectAvatar.container.js
```diff
@@ -157,36 +74,44 @@ function ProjectAvatarEdit({
...
+      onChange={(value) => {
+        changeValue(value);
+      }}
+      onSubmit={(f) => submitAvatar(f)}
+      readOnly={settingsReadOnly}
+      submitButtonId="update-avatar"
+      initialValue={initial}
+    />
   );
 }
```
lorenzo-cavazzi, line 117: I might be missing something here. It seems to me that `ProjectAvatarEdit` is imported only in the index file, re-exported from there, but no other component is actually importing or using it. Is this correct?

lorenzo-cavazzi (dismissed):
Good to see this change!

I added a few inline comments, and I have a couple more points here:

- Error message: should we format it better? Showing a JSON string isn't the best solution. In this case, it should be "avatar is too big ...". Also, there should be some spacing before the error message instead of touching the button.
image

- File size: I thought we were downsizing the avatar. I tried to zoom it in so much that it should be very few pixels, but I still get the 200kb size error. Why is that? Also, the message "max 3 MB" is misleading in this case.

client/src/features/project/components/ProjectSettings.tsx
andre-code: yes, you are right. In order to maintain transparency and clarity in the user interface the combination of read-only inputs and a permission-related message would be the best approach.

that how looks now: 
Screenshot 2023-10-04 at 15 23 59

client/src/project/shared/ProjectAvatar.container.js
andre-code: yes, I removed that code

andre-code:
Thank you for your review, @lorenzo-cavazzi. I have updated the input helper text to display a more accurate message regarding the limits of the image to upload compared to the image after cropping. It's possible that the calculation of the file size after crop and resize could be greater than the expected final size, and this message reflects that accurately.

I have also improved the parsing of error messages when there is an issue with changing the image. With these two changes, it now looks like this:

Screenshot 2023-10-04 at 15 23 21

client/src/components/form-field/FormGeneratorImageInput.js
```diff
@@ -301,7 +304,8 @@ function ImageContentInput({
           Select an image file (max size {formatBytes(maxSize)}).
           <br />
           Images will be cropped to {CARD_IMAGE_DIMENSIONS.width}px &times;{" "}
-          {CARD_IMAGE_DIMENSIONS.height}px.
+          {CARD_IMAGE_DIMENSIONS.height}px to achieve the desirable size{" "}
```
ciyer, line 307:
```suggestion
          {CARD_IMAGE_DIMENSIONS.height}px to achieve the target size{" "}
```

client/src/features/project/components/ProjectSettingAvatar.tsx
```diff
@@ -0,0 +1,195 @@
...
+    })
+      .unwrap()
+      .then(() => setSucceeded(true))
+      .catch(() => setSucceeded(false));
+  }, [avatar, projectFullPath, updateProject, projectId]);
+
+  const setAvatarAndReset = (newAvatar: ImageValue) => {
+    setAvatar(newAvatar);
+    reset();
+  };
```
ciyer, line 102: I think useCallback should be used here...
```suggestion
  const setAvatarAndReset = useCallback((newAvatar: ImageValue) => {
    setAvatar(newAvatar);
    reset();
  }, [setAvatar, reset]);
```

client/src/features/project/components/ProjectSettingAvatar.tsx
```diff
@@ -0,0 +1,195 @@
...
+          options: [
+            {
+              [Prop.NAME]: CURRENT_AVATAR_NAME,
+              [Prop.URL]: getEntityImageUrl(images),
+            },
+          ],
+          selected: 0,
+        } as ImageValue)
+      : INITIAL_IMAGE_VALUE;
+  };
```
ciyer, line 116: Could this function be lifted to outside the function? It does not use anything from the lexical context, and I think it would be cleaner/easier if it were declared at the module level.

client/src/features/project/components/ProjectSettingAvatar.tsx
```diff
@@ -0,0 +1,195 @@
...
+  }, [projectMetadata.data?.images]);
+
+  const onCancel = () => {
+    setAvatar(
+      getCurrentImageAvatar(
+        projectMetadata.data?.images as unknown as ImagesLinks[]
+      )
+    );
+    reset();
+  };
```
ciyer, line 133:
```suggestion
  const onCancel = useCallback(() => {
    setAvatar(
      getCurrentImageAvatar(
        projectMetadata.data?.images as unknown as ImagesLinks[]
      )
    );
    reset();
  }, [reset, setAvatar]);
```

client/src/features/project/components/ProjectSettingAvatar.tsx
```diff
@@ -0,0 +1,195 @@
...
+            onCancel={onCancel}
+            onChange={(value) => {
+              setAvatarAndReset(value as unknown as ImageValue);
+            }}
+            onSubmit={() => onSubmit()}
+            readOnly={readOnly}
+            submitButtonId="update-avatar"
+            value={avatar ?? initialValue}
+          />
+          {errorAlert}
```
ciyer, line 190: I think adding some spacing to the margin would be good here. The alert is a little to close to the buttons here.

image

ciyer (changes requested):
Overall, this looks very good! 🎉 I'm still doing some manual testing, but I have a couple of small suggestions about the code.

client/src/features/project/components/ProjectSettingAvatar.tsx
```diff
@@ -0,0 +1,198 @@
...
+  isMaintainer,
+  projectFullPath,
+  projectId,
+  branch,
+  gitUrl,
+}: ProjectSettingsAvatarProps) {
+  const [avatar, setAvatar] = useState<ImageValue>();
+  const [succeeded, setSucceeded] = React.useState<boolean | undefined>(
+    undefined
+  );
```
leafty, line 75:
```suggestion
  const [succeeded, setSucceeded] = useState<boolean | undefined>(
    undefined
  );
```

leafty (approved):

leafty (approved):

PR #2810, opened 2023-09-22
tests(client): update cypress action

leafty:
Updates the cypress action and cypress itself. This seems to fix the `ECONNRESET` issue.


ciyer (approved):
This is good -- we want to be on the latest version of Cypress anyway, so this is a nice way to fix this bug.

PR #2820, opened 2023-09-27
fix(tests): try to fix the dataset test

lorenzo-cavazzi:
/deploy renku=ui2686-update-tests renku-notebooks=1.19.0 extra-values=notebooks.userSessionPersistentVolumes.enabled=true


lorenzo-cavazzi:
We just merged https://github.com/SwissDataScienceCenter/renku-ui/commit/8b7780d67105b60926f2d57104668b66b2684ea1 to fix this case

PR #2827, opened 2023-10-02
fix(client): handle chunk load errors

### LINKED ISSUE #2826: React Suspense components can crash the app if there was an update
If a client has loaded a UI which has been replaced by another version (upgrade), then trying to load extra js chunks will fail:

```
bundle.js:399176 Uncaught ChunkLoadError: Loading chunk vendors-node_modules_highlight_js_es_index_js failed.
```

Maybe we should provide an error boundary for this case and make the user reload the page, or even reload the page automatically.

leafty:
Fixes #2826.

The UI will reload the page when `ChunkLoadError` happens. The handler will also avoid reloading loops.

### DIFF
--- a/client/src/error-boundary/ErrorBoundary.tsx
+++ b/client/src/error-boundary/ErrorBoundary.tsx
@@ -16,9 +16,10 @@
  * limitations under the License.
  */
 
-import { ReactNode } from "react";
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi (approved):
Lgtm

PR #2831, opened 2023-10-04
fix(client): add core version in url when upload dataset files

### LINKED ISSUE #2821: Be sure we always provide the project version to the cache.files_upload API
## Description

Uploading files to datasets occasionally uses a non-version API (see screenshot)

image

Don't forget to change the e2e tests accordingly. Currently, we use the function `uploadDatasetFile` in the tests and it matches both the versioned and the non-versioned API :scream:

## Reproduce

Using the e2e tests is enough to reproduce this case.
We should double-check every place where we use the `POST /cache.files_upload` API and be sure we pass the project version.

andre-code:
PR to add core version in url when upload dataset files.

add files datase

Fix #2821 
/deploy #persist


ciyer (approved):
Very clean fix! Looks great. 🎉

PR #2832, opened 2023-10-04
chore(client): perform more bundle optimization

leafty:
Details:
* Lazy load the Lineage component
* Update the lodash imports
* Removed unused function which loads a yaml parser
* Split off `jszip` and `file-saver` (UI tested by downloading session logs)
* Set 365 days cache expiry for assets which are hashed by webpack

/deploy


client/nginx.vh.default.conf
```diff
@@ -39,6 +39,16 @@ server {
...
+    location ^~ /static/css/ {
+      expires 365d;
+      add_header Cache-Control "public";
+    }
+
+    location ^~ /static/media/ {
+      expires 365d;
+      add_header Cache-Control "public";
+    }
+
```
ciyer, line 51: I'm a little concerned about this, since it makes assumptions about how the module bundler works, which could bite us if we change to different bundler.

ciyer (commented):
Most of the changes here are great -- removing unused code, importing more selectively from lodash, etc -- these we should merge ASAP.

The only one that I am a little unsure of is the change to the nginx config. I'm concerned that this could trip us up if we move away from webpack and switch to a different bundler.

client/nginx.vh.default.conf
leafty: I think it's expected that the production nginx configuration is coupled to how the bundler emits files.

ciyer (approved):
Nice refactorings that should improve cold start time.

PR #2834, opened 2023-10-04
fix(client): improve string validation when trying to upload a dataset file by URL

### LINKED ISSUE #2801: URL validation when uploading files to datasets is too strict
## Description

Some URLs are marked as not valid even if they are valid -- E.G. anything with `:` in it, like
https://commons.wikimedia.org/wiki/File:Location_dot_black.svg

## Reproduce

Try to upload from the link above.

andre-code:
PR to fix bug when upload a dataset file by url.

Changes:
- Modify function to validate an url

Fix #2801 
/deploy extra-values=core.sentry.dsn=https://0cc02f3a02d3474d9f7e90915873ab4d@sentry.dev.renku.ch/2,core.sentry.environment=renku-dev,core.sentry.enabled=true


lorenzo-cavazzi:
@andre-code Is this PR waiting for a renku-core fix?

lorenzo-cavazzi:
I think the renku-core requirement is not a problem anymore (renku-core 2.8.0 is out)

lorenzo-cavazzi (changes requested):
This works fine now :+1: 

We can use the new function instead of the older `isUrl`; that's actually used only in a single other place where the new function should be more reliable. I suggest replacing the import with `isValidURL` and removing the older function `isUrl` (changing the code is also fine; I find the new name you gave to the new function more intuitive)

lorenzo-cavazzi (approved):
:+1:

PR #2847, opened 2023-10-12
fix(client): make the settings alert consistent

leafty:


leafty:
Closed: made the intended changes in #2990.

PR #2848, opened 2023-10-13
fix(client): update the Renku logo and Renku browser icons

leafty:
Updates the Renku logo and Renku icons in the UI

/deploy


lorenzo-cavazzi (commented):
The changes look good :+1: 
Should we merge this or is there a blocker?

leafty:
> The changes look good 👍 Should we merge this or is there a blocker?

The previous logos use non-standard colors, but they offer better contrast. I think some discussion is needed to find out where these come from and how we proceed.

lorenzo-cavazzi (approved):
Lgtm! Thanks for wrapping this up :rocket:

PR #2853, opened 2023-10-17
Show status banner as loud during major incidents

### LINKED ISSUE #2767: The status banner should be loud during major outages
(See `StatusPage.present.js`)

The logic to render the status banner on every page or only on the home page should be updated to take into account the severity of the current outage:

* if `status.indicator` is `major` or `critical` then the banner should be "loud"
* if any `incident.impact` is `major` or `critical` then the banner should be "loud"

ciyer:
If there is a major or critical incident reported on the statuspage, show information about it in a banner.

image

Fix #2767

/deploy

### DIFF
--- a/client/src/statuspage/Statuspage.present.js
+++ b/client/src/statuspage/Statuspage.present.js
@@ -111,7 +111,7 @@ function SiteStatusDetails(props) {
       </div>
     );
   }
-  return <WarnAlert>{status.description}</WarnAlert>;
+  return <WarnAlert dismissible={false}>{status.description}</WarnAlert>;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


ciyer:
# Testing

You can look at cypress tests in the `maintenance.spec.ts` spec, or you can run in telepresence and provide a statuspage id using the `STATUSPAGE_ID` environment variable to see how changes to the status affect the display of the banner.

client/src/statuspage/Statuspage.present.js
```diff
@@ -120,14 +120,20 @@ function SiteStatusDetails(props) {
...
+
+  const alertStyle = loud ? { fontSize: "larger" } : {};
   return (
-    <WarnAlert className="container-xxl renku-container">
-      RenkuLab is unstable: {status.description}. See{" "}
-      <b>
-        <Link to={siteStatusUrl}>RenkuLab Status</Link>
-      </b>{" "}
-      for more details.
+    <WarnAlert className="container-xxl renku-container" dismissible={!loud}>
```
andre-code, line 129: Would be nice to use here classnames (cx)

client/src/statuspage/Statuspage.present.js
```diff
@@ -120,14 +120,20 @@ function SiteStatusDetails(props) {
...
+      <div style={alertStyle}>
```
andre-code, line 130: Instead of applying inline styles, consider using the className `fs-5.`

andre-code (changes requested):
The banner works as expected. I've only got a couple of minor suggestions left.

PR #2854, opened 2023-10-17
feat(client): adapt slug and name of dataset (#2723)

### LINKED ISSUE #2723: Adapt slug and names on KG datasets APIs
The KG team is harmonizing responses from their APIs, introducing `slug` and `name` everywhere, replacing other properties.

As soon as https://github.com/SwissDataScienceCenter/renku-graph/pull/1741 is ready, we should adapt the UI

andre-code:
PR to adjust the dataset slug and name in alignment with the updated Renku service and KG API, which now return slug and name instead of name and title.

Please note that the requests to Renku service for datasets now require the use of the `slug` rather than the `name`.

In this update, what was previously referred to as `title` has been changed to `name`, and `name` has now become `slug`.

- Pending
- [x] Update renku cypress test
- [x] Test Renku service for compatibility, given the issue with the version, as Renku Core has not yet released an updated version. 

Fix #2723 
/deploy renku=ds-slug-title-name-change renku-graph=1511-ds-slug-name-only renku-core=develop extra-values=global.renku.cli_version=2.8.0rc3  #persist


ciyer (approved):
I looked around and made changes to datasets, and everything looks good after this change. 🚀

PR #2855, opened 2023-10-17
chore: always use read-only for cloud storage

olevski:
The 3rd party library we use for mounting cloud storage breaks down when in read/write mode.

We will swap this for a better library. But we want to release this with read-only support and not delay the feature further. Because swapping this library cannot be done that quickly and it needs some testing to make sure it works well.

You can see this deployed at https://github.com/SwissDataScienceCenter/renku/pull/3214

### DIFF
--- a/client/src/features/project/components/AddCloudStorageButton.tsx
+++ b/client/src/features/project/components/AddCloudStorageButton.tsx
@@ -321,7 +321,7 @@ function AdvancedAddCloudStorage({
               </FormText>
             </div>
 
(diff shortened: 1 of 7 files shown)


ciyer (changes requested):
In principle, this looks fine as a quick-fix. I think it would be better to use a constant like ` ALLOW_READ_WRITE_MOUNTS` and then have:

```
{
  !ALLOW_READ_WRITE_MOUNTS ? 
    null :
   [element]
}
```

That way, the code will be easier to revive later, especially if there are changes or refactorings done before we are able to give users this option again.

PR #2856, opened 2023-10-19
feat: show a confirmation text when saving a session

### LINKED ISSUE #2792: The save button on sessions doesn't provide adequate feedback
## Motivation

When clicking on the "Save" icon in a running session, we get no feedback that the content was successfully created. Instead, we immediately show again the message "nothing to be saved". This seems to be confusing for a few users.

## Proposal

Let's fix also a couple of tiny details.

- Show clear feedback that the content was saved, and keep it until the user closes the modal.
- The double "Saving session" is unnecessary. We can show it only on the button (where we should fix the spacing anyway)
  image
- Done in #2796 ~We show a double popup on the Save button (and others). Let's remove it~
  image

andre-code:
PR to introduce a confirmation message upon a successful session save. Additionally, it addresses minor issues, such as removing duplicate "saving session" messages and adjusting the spacing between text and the icon in the button.

Fix #2792 

/deploy #persist


ciyer (dismissed):
Very nice! 🎉

client/src/features/session/components/SaveSessionModal.tsx
```diff
@@ -295,8 +308,8 @@ function SaveSessionBody({
     undefined
   );
   const saveText = saving ? (
-    <span>
-      <Loader className="me-1" inline size={16} />
+    <span className="d-flex gap-2">
```
leafty, line 311: We should avoid doing this, the alignment is not correct here.

See here:
image


When `inline` is passed to `<Loader>`, it is meant to be inserted in the normal flow of a text. Separation can be achieved by using margin classes, e.g. `me-1` or `ms-1`.

See correct alignment here:
image

```tsx
<span>
  <Loader className="me-2" inline size={16} />
  Saving Session
</span>
```

client/src/features/session/components/SaveSessionModal.tsx
```diff
@@ -53,6 +52,7 @@ import {
   useRenkuSaveMutation,
 } from "../sidecarApi";
 import styles from "./SessionModals.module.scss";
+import { SuccessAlert } from "../../../components/Alert";
```
leafty, line 55: fix import order (vscode "Organize imports" command)

leafty (changes requested):

leafty (approved):
LGTM 🌟

Issue #2283, opened 2023-01-09
Prevent flashing the RenkuLab down page


lorenzo-cavazzi:
## Description

A few users recently mentioned they saw a RenkuLab down page flashing for a moment before seeing the correct content -- I could see something similar last Friday while working on Cypress e2e tests.
Apparently, it's now impossible to reproduce this problem.

After checking into possible reasons, I found that we don't fully handle unexpected responses from the Statuspage API endpoint.
If the `GET /user` endpoint is slow to respond and the Statuspage API return quickly a failure, there is actually the possibility of flashing that page.

To cover this case, we should:
- Implement additional tests on the landing page to verify that wrong responses from the Statuspage endpoint are handled correctly (it probably requires updating those components).
  :information_source: there is already an unused `getStatusPage` fixture from #2281 -- use that and add fake failed responses
- Cover the case of the user endpoint responding slowly

PR #2858, opened 2023-10-20
tests: add tests for slow and error responses from user and statuspage

ciyer:
Work on #2283

### DIFF
--- a/tests/cypress/e2e/home.spec.ts
+++ b/tests/cypress/e2e/home.spec.ts
@@ -46,32 +46,25 @@ describe("404 page", () => {
   });
 });
 
-describe("display the maintenance page", () => {
+describe("display the home page even when APIs return strange responses", () => {
   beforeEach(() => {
-    new Fixtures(cy).config().versions();
+    new Fixtures(cy)
+      .config()
+      .versions()
+      .statuspageDown()
+      .userNone(undefined, 1000);
+    cy.visit("/");
   });
 
-  it("displays an error when trying to get status page information", () => {
-    // ! we plan to change this behavior and ignore statuspage info when unavailable #2283
-    new Fixtures(cy).renkuDown();
-    cy.visit("/");
+  it("displays the home page intro text", () => {
+    cy.wait("@getUser");
     cy.get("h1").should("have.length", 1);
-    cy.get("h1").contains("RenkuLab Down").should("be.visible");
-    cy.get(".alert-content")
-      .contains("Could not retrieve status information")
-      .should("be.visible");
-  });
-
-  it("displays status page information", () => {
-    // if the call to get the user fails (e.g., no .userNone() fixture)
-    // then show the status page
-    new Fixtures(cy).getStatuspageInfo();
-    cy.visit("/");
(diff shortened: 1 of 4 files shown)


ciyer:
I added the tests suggested in [#2283](https://github.com/SwissDataScienceCenter/renku-ui/issues/2283), but I still do not see the behavior described in that issue. Though I believe all those that have reported it, from looking at the code, it isn't clear how this situation could happen in the way described

I think a more likely scenario that could cause this would be an error response from the `user` endpoint, followed by a UI refresh, followed by a valid response from `user`.

tests/cypress/e2e/home.spec.ts
```diff
@@ -37,6 +37,28 @@ describe("display the home page", () => {
   });
 });
 
+describe("display the home page even when APIs return strange responses", () => {
+  beforeEach(() => {
+    new Fixtures(cy)
+      .config()
+      .versions()
+      .statuspageDown()
+      .userNone("getUser", 5000);
```
lorenzo-cavazzi, line 46: This can probably be much lower than 5000, I'd say 1000

lorenzo-cavazzi (commented):
I suggest having a single place where we verify the maintenance messages work as expected.
Perhaps we should move the maintenance-related tests (that includes the `"display the maintenance page"` section, to be updated or merged with the new tests) to the maintenace.spec.ts file and let the home specs check only home page-specific components.

ciyer:
> I suggest having a single place where we verify the maintenance messages work as expected. Perhaps we should move the maintenance-related tests (that includes the `"display the maintenance page"` section, to be updated or merged with the new tests) to the maintenace.spec.ts file and let the home specs check only home page-specific components.

I moved the `display maintenance page` test to the `mainenance.spec.ts` file, but I kept the new test in the home page, since it tests that the home page, and not the mainenance page, is shown. I reduced the delay to 1s, as suggested.

tests/cypress/e2e/maintenance.spec.ts
```diff
@@ -45,6 +45,35 @@ describe("display the maintenance page", () => {
   });
 });
 
+describe("display the maintenance page when there is no user response", () => {
+  beforeEach(() => {
+    new Fixtures(cy).config().versions();
+  });
+
+  it("displays an error when trying to get status page information", () => {
+    // ! we plan to change this behavior and ignore statuspage info when unavailable #2283
```
lorenzo-cavazzi, line 54:
```suggestion
```
We can remove this line

lorenzo-cavazzi (changes requested):
Lgtm!

Before merging, let's remove the comment about the #2283 issue since it's closed by this PR.

lorenzo-cavazzi:
The test `display the home page even when APIs return strange responses` seems to fail consistently -- I tried it locally as well

ciyer:
> The test `display the home page even when APIs return strange responses` seems to fail consistently -- I tried it locally as well

I see that now -- it was a problem introduced as a side effect of merging changes from master. Everything should work now.

lorenzo-cavazzi (approved):
Lgtm!

PR #2861, opened 2023-10-23
fix(server): properly close all WebSocket connections upon shutdown

leafty:
This PR fixes an issue when the server receives a `TERM` signal, which is most apparent when running the server through `telepresence`. After the fix, the server properly shuts down all web socket connections and will not hang forever.


server/src/index.ts
```diff
@@ -96,23 +96,60 @@ const server = app.listen(port, () => {
...
+      process.exit(1);
+    } else {
+      logger.info("Shutting down storage");
+      storage.shutdown();
+      logger.info("Shutdown completed.");
+      setImmediate(() => {
+        process.exit(0);
+      });
+    }
+  });
```
ciyer, line 129: Is this a better fix than

```
createTerminus(server, {
  signals: ["SIGTERM", "SIGINT"],
  onShutdown: async () => {
    storage.shutdown();
    logger.info("Shut down completed.");
  },
  onSignal: async () => {
    logger.info("Interrupted, shutting down...");
  },
});
```

?

leafty: Most likely yes, because here the web socket connections are closed gracefully. Also, we do not need to add yet another dependency.

ciyer (approved):
This works and fixes the problem.

PR #2862, opened 2023-10-24
build: release 3.14.0

lorenzo-cavazzi:
Release 3.14.0

/deploy


ciyer (approved):
👍

leafty (approved):

leafty:
Note: `swiper` is adding a little bit of weight into the main bundle, but we should go ahead.

PR #2863, opened 2023-10-24
refactor(tests): reorganise commands used in Cypress tests 

leafty:
The custom Cypress commands have been re-organised into `tests/cypress/support/commands`, and they have proper types (de-referencing in VSCode leads to the function definition).

Note: a follow-up PR will refactor fixtures.


ciyer (approved):
Nice refactoring -- brings the code more inline with standard JS/TS style

PR #2866, opened 2023-10-27
build: release 3.15.0

leafty:
/deploy renku=leafty/feat-admin-panel renku-gateway=master renku-data-services=main extra-values=global.gitlab.url=https://gitlab.dev.renku.ch,dataService.image.pullPolicy=Always,notebooks.cloudstorage.s3.enabled=true,amalthea.extraChildResources[0].group=com.ie.ibm.hpsys,amalthea.extraChildResources[0].name=datasets,notebooks.cloudstorage.s3.installDatashim=false

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,13 @@
 ### Features
 
 - Show a confirmation text when saving a session ([#2856](https://github.com/SwissDataScienceCenter/renku-ui/pull/2856)).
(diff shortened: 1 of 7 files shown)


lorenzo-cavazzi (changes requested):
Looks good!
Could you also add the PR links next to the bullet points, similar to the previous releases? The PR seems more immediate to identify what was implemented because, unfortunately, what's implemented doesn't always match the issue description (especially true for issues with multiple bullet points or closed by multiple separate PRs)

PR #2871, opened 2023-10-30
fix: correctly handle statuspage down

ciyer:
Handle a bad response from the statuspage API


leafty (approved):

client/src/statuspage/StatuspageAPI.js
leafty, line 1: non-blocking: it would probably not be much effort converting this to RTK query.

leafty (dismissed):

client/src/statuspage/StatuspageAPI.js
ciyer: Good suggestion. Certainly be worth doing!

leafty (approved):

lorenzo-cavazzi (approved):

PR #2872, opened 2023-10-31
tests: fix some flakiness with e2e tests

leafty:


ciyer (approved):
This fixes the problems that I was seeing locally!

client/src/project/Project.present.js
```diff
@@ -800,55 +798,55 @@ class ProjectViewFiles extends Component {
...
+//             </Route>
+//             <Route exact path={props.settingsCloudStorageUrl}>
+//               <ProjectSettingsCloudStorage />
+//             </Route>
+//           </Switch>
+//         </Col>
+//       </Row>
+//     </Col>
+//   );
+// }
```
lorenzo-cavazzi, line 849:
```suggestion
```
Should we remove this before merging?

lorenzo-cavazzi (changes requested):
I dislike adding a re-try strategy for local tests, but it seems effective and gives us time to improve the UI wherever needed.

image

client/src/project/Project.present.js
leafty: Done in [fc2b93c](https://github.com/SwissDataScienceCenter/renku-ui/pull/2872/commits/fc2b93c4611d14f78b7f2e2a1163b05b4f447cfd).

lorenzo-cavazzi (approved):
Thank you for the joint effort on this :pray:

PR #2874, opened 2023-10-31
fix(client): prevent session issues for uppercase namespaces

lorenzo-cavazzi:
Projects with uppercase namespaces were messing up sessions, resulting in:
* Wrong feedback on the sessions dropdown menu in both the header and the dashboard (in the screenshot, all projects are paused but the last one shows "Start" instead of "Resume").
* Missing sessions in the project's session tab.

This PR fixes it (in the screenshot, left is on "dev" and right is on the current PR).

image

/deploy


lorenzo-cavazzi:
P.S: making a namespace uppercase requires some effort; the easiest way is to go to GitLab, create a new group, and manually update the group namespace to make it uppercase.
Here is a test project: https://gitlab.dev.renku.ch/uppercase-group/UPPER/uppercase-namespace

client/src/features/session/sessions.utils.ts
```diff
@@ -38,7 +38,9 @@ export function getRunningSession({
       namespace: annotations.namespace,
       path: annotations.projectName,
     });
-    return thisAutostartUrl === autostartUrl;
+    return typeof thisAutostartUrl !== "string"
+      ? thisAutostartUrl === autostartUrl
+      : thisAutostartUrl.toLowerCase() === autostartUrl.toLowerCase();
```
leafty, line 43: This is always a string:

```suggestion
    return thisAutostartUrl.toLowerCase() === autostartUrl.toLowerCase();
```

In [client/src/utils/helpers/url/Url.js - L260](https://github.com/SwissDataScienceCenter/renku-ui/blob/66c40019326a8f66aa687d69cbb1f83617c8d939/client/src/utils/helpers/url/Url.js#L260), update the types:

```js
/**
 * Create a Url based on the target page. Depending on the specific page, it may require context data.
 *
 * @param {object | string} target - the page you are targeting, as contained in the `pages`  member
 *   (e.g. pages.landing, pages.project.stats, ...).
 * @param {object} [data] - the context data you need to provide, if any
 *   (e.g. for project, you need to provide a `namespace` and a `path`).
 * @param {boolean} [full] - switch between full or relative path. The default is `false`.
 * @returns {string}
 */
function get(target, data, full = false) {
```

leafty (changes requested):
Some suggested changes:

In [client/src/components/list/ListBarSessions.tsx - L255](https://github.com/SwissDataScienceCenter/renku-ui/blob/66c40019326a8f66aa687d69cbb1f83617c8d939/client/src/components/list/ListBarSessions.tsx#L255), you can also add the session name:

```tsx
<SessionButton fullPath={fullPath} gitUrl={gitUrl} runningSessionName={notebook.name} />
```

leafty (approved):
LGTM :rocket:

PR #2875, opened 2023-11-01
[Shape up] Renku Native Projects

### LINKED ISSUE #2870: Renku native projects creation
## Description

Page/flow for creating renku-native projects [renku-projects pitch](https://www.notion.so/Exploration-Define-Renku-Projects-and-Users-6f15f7870af04bc9b66362a7936b34cf).

ciyer:
Work on the Renku Native Projects pitch

Fix #2870 

/deploy renku=renku-data-services-users-db renku-data-services=main extra-values=initDb.image.repository=registry.dev.renku.ch/tasko.olevski/renku-images/init-db,initDb.image.tag=0.44.1-1.ebb0e80,keycloakx.initRealm.image.repository=registry.dev.renku.ch/tasko.olevski/renku-images/init-realm,keycloakx.initRealm.image.tag=0.44.1-1.ebb0e80.git.2200.h26267d0e


ciyer:
# Testing

The feature is not exposed in the UI, but you can navigate to it directly:

https://renku-ci-ui-2875.dev.renku.ch/projectsV2/

Here, you can create and modify V2 Projects. If you know the email address of a user in the instance, you can add them to the project.

client/src/features/projectsV2/edit/projectMemberFields.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+  return (
+    <>
+      <ModalBody>
+        <Form
+          className="form-rk-green"
+          noValidate
+          onSubmit={handleSubmit(onSubmit)}
+        >
+          <div className="d-flex flex-row align-items-baseline mb-3">
```
lorenzo-cavazzi, line 208: Nitpick comment: I see you occasionally use `cx` when there are more than 2 classes but there are some exceptions; would it make sense to keep it consistent?

client/src/features/projectsV2/show/ProjectV2EditForm.tsx
lorenzo-cavazzi, line 1: Naming is a bit confusing. There are a `show` and an `edit` folder. Should this file go to the `edit` folder? Or perhaps would it be easier to have a single "show" folder?

client/src/features/projectsV2/api/projectV2.enhanced-api.ts
```diff
@@ -0,0 +1,94 @@
...
+          ? parseInt(headerPerPage)
+          : queryArg.perPage ?? 20;
+        const total = headerTotal
+          ? parseInt(headerTotal)
+          : projects
+          ? projects.length
+          : 0;
+        const totalPages = headerTotalPages
+          ? parseInt(headerTotalPages)
+          : total / perPage;
```
lorenzo-cavazzi, line 40: It would be nice to handle the pagination headers through a shared function instead of having the logic in the `transformResponse` section.
I'm wondering if we should add it the existing `pagination.utils.ts` in the `utils/helpers` section

lorenzo-cavazzi (commented):
I went through the code and added a few comments.
I must admit the output from `@rtk-query/codegen-openapi` is messier than I expected, especially with the multitude of unnecessary types created, added directly the main `.api` file. I assume it's just a matter of writing cleaner OpenAPI specs without defining custom types that correspond to simple types like `string`

client/src/features/projectsV2/api/projectV2.enhanced-api.ts
leafty: These already exist:
* `client/src/utils/types/pagination.types.ts`
* `client/src/utils/helpers/pagination.utils.ts`

leafty (changes requested):
[Partial review]
Good work, the code is readable and easy to follow.

* All barrel files should be removed (`index.ts`). See https://vitejs.dev/guide/performance.html#avoid-barrel-files
* As a general comment, help text should not disappear because of validation.
* <FormText> should ideally be pointed to with `aria-describedby` -> https://getbootstrap.com/docs/5.3/forms/form-control/#form-text
  See `client/src/features/admin/AddManyUsersToResourcePoolButton.tsx` as an example

Note: styling comments have been omitted.

client/src/features/projectsV2/api/projectV2.enhanced-api.ts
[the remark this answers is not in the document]
ciyer: > Is this file manually created? Is the purpose to manually extend the APIs automatically created since they don't support some interaction?

Yes, exactly. https://redux-toolkit.js.org/rtk-query/api/created-api/code-splitting#injectendpoints

client/src/features/projectsV2/edit/index.tsx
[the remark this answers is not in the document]
ciyer: I have to confess, I'm not a big fan of this advice. 

When structuring code, we should focus on understandability for humans, and only make concessions for machines (compilers, bundlers, etc.) when there is a compelling reason to do so, and I'm not convinced there is in this case. 

1. Using directory modules makes it possible break code down into multiple files and still clearly express the public interface that is intended to be exposed.
2. In production, Vite does tree-shaking, so there is no penalty to using the barrel files
3. There is potentially a penalty in dev, but many users would like Vite to do tree shaking in dev as well (see the issue referenced in the link), Webpack does this, and someone (see that issue again) has implemented a plugin that does this; In any case, it would be good to know how large the penalty is, since we might be doing an optimization that has almost no actual benefit for us.

client/src/features/projectsV2/edit/projectMemberFields.tsx
ciyer: My preference is to use cx when the classes are dynamically computed and to set them inline, regardless of how many there are, when they are static, but if we want to have a different standard, that is ok with me.

client/src/features/projectsV2/show/ProjectV2EditForm.tsx
ciyer: I renamed the folder `edit` to `fields`, which better communicates the purpose, I think. 7b35b99bcc69da3782b7e4e32bca4e023a82ec9a

client/src/features/projectsV2/edit/projectMemberFields.tsx
lorenzo-cavazzi: That works fine but it doesn't seem to be consistent. Here `cx` isn't used, but it is for the next input component despite the content being static `className={cx("form-control", "ms-3")}`. Further down I see again `className="btn-outline-rk-green"` and then next line `className={cx("bi", "me-1")}`.

Also, in most of the recent code, we settled on using `cx` whenever there is more than one single class. That's not enforced in the coding rules, though, and it's mainly for readability. No need to use that approach in this PR since it would require more changes

client/src/features/projectsV2/edit/ProjectRepositoryFormField.tsx
```diff
@@ -0,0 +1,103 @@
...
+            required: true,
+            pattern: /^(http|https):\/\/[^ "]+$/,
+          }}
+        />
+        <Button className="ms-1" onClick={onDelete}>
+          X
+        </Button>
+      </div>
+      {errors.repositories && errors.repositories[index] ? null : index >
+        0 ? null : (
```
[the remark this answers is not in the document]
ciyer: Fixed in b9d525b0ccd046b832968ef0491a398f32b57037

client/src/features/projectsV2/edit/index.tsx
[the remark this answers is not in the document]
lorenzo-cavazzi: Without testing, I assume the penalty is minimal considering we just re-export small components, so that's probably not a good argument from my side.
Still, I find having these index files confusing unless we consistently use them in all folders; currently, it's unclear whether the components are only exported directly from the file or re-exported through the index, so I always find myself browsing the folder to find it out.

We only discussed using `export default` and limiting the number of components per file to make it immediately intuitive what's available ([reference](https://github.com/SwissDataScienceCenter/renku-ui/blob/main/CODING_GUIDELINES.md#r004-include-a-default-export-when-appropriate)). We didn't set any rule for the index files; in this case, the component `ProjectRepositoryFormField` is exported by default from the `ProjectRepositoryFormField.tsx`, making it easy to find, while `AddProjectMemberModal` is exported by a file with a different name. If I need to work on the projects a few months from now, it won't be immediately clear what is available: I can't rely on the index file exports since it's frequently incomplete, and the file names won't help me.

Here I would at least rename the file "projectMemberFields" and use `export default` -- or split that into multiple files to have `AddProjectMemberModal.tsx`. For the use of the index file, we can leave it as is if you find it useful but I would like to discuss this all together and decide on a consistent solution.

client/src/features/projectsV2/fields/ProjectRepositoryFormField.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { Controller } from "react-hook-form";
+import type { FieldValues } from "react-hook-form";
+
+import { Button, FormText, Input, Label } from "reactstrap";
+
+import type { Repository } from "../projectV2.types";
+import type { GenericProjectFormFieldProps } from "./formField.types";
+
+interface ProjectRepositoryFormFieldProps<T extends FieldValues>
+  extends GenericProjectFormFieldProps<T> {
```
leafty, line 30: This interface does not need to be generic:

```suggestion
interface ProjectRepositoryFormFieldProps extends GenericProjectFormFieldProps<ProjectV2Repositories> {
```

client/src/features/projectsV2/list/ProjectV2List.tsx
```diff
@@ -0,0 +1,121 @@
...
+          <ProjectV2ListProject key={project.id} project={project} />
+        ))}
+      </div>
+      <Pagination
+        currentPage={data.page}
+        perPage={perPage}
+        totalItems={data.total}
+        onPageChange={(page: number) => {
+          setPage(page);
+        }}
```
leafty, line 92:
```suggestion
        onPageChange={setPage}
```

client/src/features/projectsV2/show/ProjectV2Show.tsx
```diff
@@ -0,0 +1,219 @@
...
+      </div>
+    </>
+  );
+}
+
+export default function ProjectV2Show() {
+  const location = useLocation();
+  const pathname = location.pathname;
+  const components = pathname.split("/");
+  const projectId = components[components.length - 1];
```
leafty, line 169: I think all of this can be replaced by just one line with `useParams()` from `react-router`.

leafty (commented):
some small comments

leafty (commented):
On my side, I think this can be merged after the comments have been addressed. This part of the UI will have code churn as we work on it, so I don't think the nitpicks are worth fighting for here.

Good work 👏 💯

client/src/features/projectsV2/edit/index.tsx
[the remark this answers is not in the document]
ciyer: Fixed in 66b03a608e5179cf113dd438096f094aece9c503

client/src/App.jsx
```diff
@@ -271,6 +276,30 @@ function CentralContentContainer(props) {
...
+            )}
+          />
+          <Route
+            path={Url.get(Url.pages.projectsV2.list)}
+            render={(p) => (
+              <ContainerWrap>
+                <ProjectV2List key="listProjectV2" {...p} />
+              </ContainerWrap>
+            )}
+          />
```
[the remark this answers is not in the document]
ciyer: Fixed in 66b03a608e5179cf113dd438096f094aece9c503

client/src/features/projectsV2/edit/projectMemberFields.tsx
ciyer: Fixed in d2035b509828c8d128273e5a48c44022295c1bab

client/src/features/projectsV2/api/projectV2.enhanced-api.ts
ciyer: The referenced files do not work for this case. They expect pagination information to be structured differently (from GitLab, I think) than how our services provide the information. I have made the changes in f8c13ef47f62625adbccf23da02acab6e4849972

client/src/features/projectsV2/fields/ProjectRepositoryFormField.tsx
ciyer: Fixed in af82cda0cfb8688eb6000b8934c8115ac37016fa

client/src/features/projectsV2/edit/projectMemberFields.tsx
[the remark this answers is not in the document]
ciyer: Fixed in f9b29504ada5935608a9b6345534a00930f4604f

client/src/features/projectsV2/show/ProjectV2Show.tsx
ciyer: I did not initially implement it like this because `useParams` requires coordination with the lexical context in which the route is defined (`App.jsx`), but maybe that isn't such a big deal. I made this change in 2f94a8041d536444bdfff74f7c78f9aaf296855d.

lorenzo-cavazzi (approved):
Thank you for addressing all the reviews!
This is a great first step to the new Renku 1.0 features :rocket:

leafty (approved):

PR #2883, opened 2023-11-07
Fix a few visual glitches

lorenzo-cavazzi:
There are quite a few visual glitches that piled up with time, most of them coming from updates to bootstrap defaults.

This PR addresses some of them. Here is a list to simplify the review:

* Remove badges customization
    01_badge
* Update the list items defaults -- a few pages were slightly adjusted, like the commits tab:
    02_commits
* Set transparent background on popovers:
    03_popovers
* Set transparent background on tables:
    04_transparent_background
* Adjust other custom elements:
    image

/deploy


client/src/App.css
leafty, line 1: Good seeing a lot of global rules being removed! 👍

client/src/components/commits/Commits.js
```diff
@@ -153,19 +153,19 @@ function SingleCommit(props) {
...
               <Button
-                color="rk-background rounded-0"
+                color="rk-background"
                 className="border"
                 id={idCopyButton}
               >
-                <Clipboard clipboardText={props.commit.id} />
+                <Clipboard clipboardText={props.commit.id}>
+                  <code>{props.commit.short_id}</code>
+                </Clipboard>
```
leafty, line 163: It doesn't make sense to have a `<Clipboard>` inside a `<Button>`.
Maybe try to use `<span>` instead of `<Button>`, and add `btn` to the `<Clipboard>` classname.

```tsx
<span id={idCopyButton}>
  <Clipboard classname={cx("btn", "border")} clipboardText={props.commit.id}>
```

client/src/components/commits/Commits.js
```diff
@@ -153,19 +153,19 @@ function SingleCommit(props) {
...
               </Button>
               <UncontrolledTooltip placement="top" target={idCopyButton}>
                 Copy commit SHA
               </UncontrolledTooltip>
+              <Button className="d-none"></Button>
```
leafty, line 168: Why is there a hidden button?

client/src/dataset/Dataset.present.js
```diff
@@ -244,21 +242,16 @@ function DisplayInfoTable(props) {
...
-              style={{ width: "120px" }}
-            >
+            <td className="text-dark fw-bold" style={{ width: "120px" }}>
               Source
             </td>
-            <td className="bg-transparent">{source}</td>
+            <td>{source}</td>
           </tr>
         ) : null}
         {dataset.published?.creator?.length >= 3 ? (
```
leafty, line 251: non-blocking: the logic here is weird and creates an empty card above the description.

client/src/styles/bootstrap/_custom_bootstrap_variables.scss
```diff
@@ -79,20 +81,28 @@ $theme-colors: (
 
 $theme-colors: map-merge($theme-colors, $rk-colors);
 
+$enable-rounded: false;
+$input-bg: #ffffff;
```
leafty, line 85: maybe use:
```suggestion
$input-bg: $rk-white;
```

client/src/styles/bootstrap_ext/_badge.scss
```diff
@@ -1,11 +1,5 @@
 .badge {
-  font-weight: lighter;
   border-radius: 5px;
-  font-size: 13px;
-  letter-spacing: -0.13px;
-  &.bg-warning {
-    color: var(--bs-dark);
-  }
```
leafty, line 8: For contrast, this rule should stay.

client/src/components/commits/Commits.css
leafty, line 1: Can you check the borders? They do not seem consistent:

image

The date headers should probably have borders too.

leafty (changes requested):

client/src/components/commits/Commits.js
lorenzo-cavazzi: I kept the buttons to avoid spending time updating the styling.
I agree this solution isn't optimal, but this tab might disappear soon so I wouldn't spend more time on it

client/src/components/commits/Commits.js
lorenzo-cavazzi: To use the buttons group styling. Same reason as above: the tab might disappear soon so let's just fix the visual appearance

client/src/dataset/Dataset.present.js
lorenzo-cavazzi: I don't get empty cards, could you share an example? Or perhaps I'm missing it :thinking: 

Still, the logic is indeed weird :smile: If it's about improving what we show on the dataset page, let's keep it separate from the CSS/classes changes in this PR

lorenzo-cavazzi:
Thanks for the review! I added the style/visual changes, but I didn't update the components used for the commits since that tab is likely to disappear soon (or be re-written anyway). I agree the current solution is... non-optimal :grimacing:

client/src/dataset/Dataset.present.js
lorenzo-cavazzi: Oh, interesting! I got contradicting outputs and, after digging a bit into this, I found out that KG occasionally return responses for non-indexed projects (i.e. when the same database has been used in another project that happens to be indexed) :exploding_head: 

Anyway, I managed to reproduce the issue
image

I couldn't understand how authors/source work, but I changed it slightly. Actually, that was probably a bad decision and I should have made a separate PR (or issue) to prevent "patchwork PRs" changing random stuffs together but... it was just quicker and I couldn't resist the temptation :shrug: 
image

client/src/components/commits/Commits.js
lorenzo-cavazzi: Vertical alignment seems a bit off with that solution. For comparison, on the left side is the code from the PR, while on the right it's your suggestion. Mind the vertical alignment
image.

I removed the unnecessary hidden button (see https://github.com/SwissDataScienceCenter/renku-ui/pull/2883/commits/5c9b3af87d55bbab7c7a14679df13f24b63ca425 ) since that was admittedly bad, and replaced it by playing around with the `rounded-*` classes.
I suggest to keep the button with clipboard to avoid spending more time on this -- that was already the previous solution after all.

leafty (approved):
LGTM 🚀 thanks for the changes 🙏

PR #2887, opened 2023-11-10
fix(client): fix cards shown on the search page

leafty:
Before the fix, the cards shown on the search page do not have proper clickable areas:

Screenshot from 2023-11-10 10-23-15
Screenshot from 2023-11-10 10-23-27

Also, when the entity has an avatar image, then the whole header is not clickable.

This PR fixes this issue by making the whole card clickable.


lorenzo-cavazzi (approved):
Nice fix! Thank you!

PR #2889, opened 2023-11-10
refactor(client): update clipboard component

leafty:
Remove the `react-clipboard.js` library.

/deploy


client/src/components/clipboard/Clipboard.tsx
```diff
@@ -67,19 +60,34 @@ export const Clipboard = ({
...
-      data-clipboard-text={clipboardText}
-      onSuccess={onSuccess}
-      className={className}
-      style={{ textDecoration: "none" }}
+    <a
+      className={cx(className, "text-decoration-none")}
+      onClick={onCopyToClipboard}
     >
       <Wrap>
         <FontAwesomeIcon
```
lorenzo-cavazzi, line 93: Not strictly necessary as part of this PR, but it would be nice to swap this for the Bootstrap icons

lorenzo-cavazzi (approved):
Lgtm! :rocket:

client/src/components/clipboard/Clipboard.tsx
leafty: Yes, I would do that in another PR.

PR #2890, opened 2023-11-13
feat: add hero section landing page anonymous users

### LINKED ISSUE #2885: Modify Hero landing page
## Proposal

X-Small _320px
Small 576px
Hero  Large - 992px
Hero  - Extra extra large - 1400px

andre-code:
PR to add new hero section for landing page anonymous users.

Fix #2885 

/deploy #persist


client/src/landing/HeroLanding/HeroLanding.tsx
```diff
@@ -0,0 +1,72 @@
...
+                "flex-md-row",
+                "flex-sm-row",
+                "flex-column",
+              ])}
+            >
+              <Link
+                className={cx([
+                  "btn",
+                  "btn-rk-green",
+                  "heroBtn",
```
lorenzo-cavazzi, line 44: Is this a leftover? I believe it doesn't do anything; the "real" `heroBtn` class comes from `styles.heroBtn`.

client/src/landing/HeroLanding/HeroLanding.tsx
```diff
@@ -0,0 +1,72 @@
...
+                id="link-try-it-out"
+                to="#rk-anon-home-get-started"
+              >
+                Try it out
+              </Link>
+              <Link
+                className={cx([
+                  "btn",
+                  "btn-outline-secondary",
+                  "heroBtn",
```
lorenzo-cavazzi, line 57: Same here

client/src/landing/HeroLanding/HeroLanding.tsx
```diff
@@ -0,0 +1,72 @@
...
+              </span>
+            </h3>
+            <p>Data, Code, and Compute all under one roof.</p>
+            <div
+              className={cx([
+                "d-flex",
+                "gap-3",
+                "pt-5",
+                "pt-md-3",
+                "flex-md-row",
```
lorenzo-cavazzi, line 35: This is not necessary since `flex-sm-row` already implies also "flex-md-row" and all the bigger sizes

client/src/landing/HeroLanding/HeroLanding.tsx
```diff
@@ -0,0 +1,72 @@
+import cx from "classnames";
+import { Link } from "react-router-dom";
+import { HomeHeader } from "../AnonymousHome";
+import heroGraphic from "../Graphics/hero_graph.svg";
+import { AnonymousHomeConfig } from "../anonymousHome.types";
+import styles from "./HeroLanding.module.scss";
```
lorenzo-cavazzi, line 6: Big thumb-up for importing classes this way :slightly_smiling_face:

client/src/landing/Assets/Visual_head_dots.svg
lorenzo-cavazzi, line 1: There are a few things to adjust with this image:
- We should still upload the solid image and play with the CSS property `opacity` if we need semi-transparent images. I know there is no `background-opacity` and that's tedious when using images as backgrounds, but there are many easy pure CSS "workarounds" (E.G: [this](https://coder-coder.com/background-image-opacity/))
- That's actually a rather heavy image for the size; looking closer, the SVG seems wrong. If you open it with a text editor, you will notice it includes an image `image id="image0_301_201" width="4096" height="2560"` while the parent SVG element is `svg width="1400" height="594"`. For comparison, a correctly generated SVG looks like "Visual_Head.svg" (or the other new SVG in this PR) and should include a bunch of `g` and `path` elements instead of an `image`. Could you re-export that as native SVG?
- (non-blocking): this image and "Visual_Head.svg" seem to overlap quite a bit. A better approach would be uploading one image and taking the desired slice. Ofc that's not always the best solution, but I feel it would work well here because the 2 single images would be bigger, and they are shown on the same page (i.e. we never load just 1). This is non-blocking, but something to keep in mind in these cases

client/src/landing/Graphics/hero_graph.svg
lorenzo-cavazzi, line 1: Not sure about the differentiation between "Assets" and "Graphics" here. This isn't blocking for this PR, but since you are restructuring the landing page, I think we can move all the images together, perhaps moving all the files from the `Assets` folder to the `Graphics` folder -- they are all SVG images

client/src/landing/AnonymousHome.tsx
```diff
@@ -96,7 +96,7 @@ type SearchInputFormFields = {
   phrase: string;
 };
 
-function SearchInput() {
+export function SearchInput() {
```
lorenzo-cavazzi, line 99: This isn't used anywhere else. Should we remove it instead of exporting it, or do you plan to use it later?

client/src/styles/bootstrap_ext/_button.scss
```diff
@@ -110,11 +110,15 @@ $borderRadius: 1000px;
   }
 }
 
+.btn-secondary {
+  -#{$prefix}btn-active-bg: #{$rk-green-highlight};
+}
```
lorenzo-cavazzi, line 115: At line 132 there is already something just for `.btn-secondary`. Should we merge the code? It will be something like:

```
.btn-secondary {
  -#{$prefix}btn-active-bg: #{$rk-green-highlight};
  &:focus { ... }
}
```

lorenzo-cavazzi (changes requested):
I really like this new section!

The code looks good overall! There are a few things to fix; I added comments inline. Mainly just minor classes/styles improvements, apart from a problem with one of the SVG files.

Here is another comment, still in the category "minor classes/styles improvements" but it doesn't fit inline.
The Renku logo on top looks misaligned on the left, particularly on smaller screens. The PR doesn't directly introduce this, but it's more evident now that the text in the hero section is left-justified. Adding a left margin on the logo is enough to get it aligned with the rest of the application -- here is a GIF to better explain what I mean
Peek 2023-11-15 11-57

client/src/landing/AnonymousHome.tsx
andre-code: Yes, I keep it because there is a high probability of using it in another section.

lorenzo-cavazzi: Ok :+1: 
Could you add a comment, something like "Currently not used; planned for #shapeup-issues"? We can later remove the comment once the function is used. Should we not use it, the comment will remind us why we kept this.

andre-code:
@lorenzo-cavazzi Thanks for your review 🙏🏼
I added your suggestions in the last commit.

lorenzo-cavazzi (approved):
Looks great! :rocket:

PR #2892, opened 2023-11-13
fix(server): save lower case project path for last-visited

leafty:
The full path for a project is case-insensitive (e.g. foo/bar is equivalent to FOO/bar), so we should normalize project paths before saving.

/deploy


lorenzo-cavazzi (approved):
Nice fix! This should prevent unwanted duplicates on the Dashboard :raised_hands:

Issue #2865, opened 2023-10-27
Add pin/unpin buttons to the UI


leafty:
Part of https://github.com/SwissDataScienceCenter/renku/issues/3325.

Add pin/unpin buttons:
* on project cards?

PR #2893, opened 2023-11-14
chore: Add APIs for user preferences

leafty:
Part of #2865.

Contains only the changes to RTK Query to have a small PR for review.

Note: merging into a feature branch.


ciyer (approved):
LGTM!

PR #2896, opened 2023-11-14
feat: add get started & Who is Behind sections landing page anonymous users (#2891)

### LINKED ISSUE #2891: Add Get Started Landing Page Section
### Proposal

Image
Image
Image
Image

Figma files: https://www.figma.com/file/EzfEbjRNebKfpqjcxNA7gv/Design-Visual-Copy-(low-fidelity)?type=design&node-id=309-777&mode=design

andre-code:
PR to add new getStarted and Who is Behind sections for landing page anonymous users.

### Testing
To include or modify a project for testing the start session on the landing page, configure the HOMEPAGE_PROJECT_PATH in the client/run-telepresence.sh script.

No included here: 
- Storybook for Entity card skeleton loading https://github.com/SwissDataScienceCenter/renku-ui/issues/2909

Fix https://github.com/SwissDataScienceCenter/renku-ui/issues/2891
Fix  https://github.com/SwissDataScienceCenter/renku-ui/issues/2897

/deploy renku=2891-get-started-section extra-values=global.renku.cli_version=2.7.0 #persist


client/src/components/list/EntityCardSkeleton.tsx
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import skeletonCardImage from "../../styles/assets/SkeletonCard.svg";
+import "./EntityCardSkeleton.scss";
```
lorenzo-cavazzi, line 20: Would you mind importing the style as you did for the other CSS files?

lorenzo-cavazzi (changes requested):
The changes are amazing! I really like how the landing page is being transformed, and it looks more professional now :100: 

Also, big :+1: for the code quality; there is hardly anything to say on the code.

Here is a list of more generic considerations; feel free to ignore them, especially if they have already been discussed/decided with the product team:
- I see an error message when no highlighted project is set by the RenkuLab admin. It's a bit confusing as a user because it looks like I should do something. I suggest not showing error messages targeting the administrators. More in general, I would just hide the card when no value is provided -- it catches the attention with the (amazing :star_struck: ) animation, but there is no possible interaction for the user.
Peek 2023-11-21 14-59
- When I click on the Start button, and then I click back, I go "back" to the project and not the landing page. This is minor, but I assume users might click there and think "oh no, what did I do? Bring me back" -- but the back button brings the user somewhere else. P.S: not sure how easy this is to change, perhaps using another query param could be the quick solution? The session link would be `/sessions/new?autostart=1&fromLanding=1` and the back button could use that to redirect back to landing instead of the project.
Peek 2023-11-21 15-10
- This is very generic: in the section "Renku is open source and free to use!" I would mention "You can deploy your own instance in your Kubernetes cluster.". I guess it would be a bigger change; I'll mention that at the next meeting to the product team :grin:

client/src/components/list/EntityCardSkeleton.tsx
leafty: @andre-code see [client/src/components/Loader.module.scss](https://github.com/SwissDataScienceCenter/renku-ui/blob/aa826b45f31df04cf638d0bb226ae7f6c663b482/client/src/components/Loader.module.scss) for an example of using `@keyframes` with SCSS modules.

client/src/components/list/ListCard.tsx
```diff
@@ -74,7 +75,11 @@ function ListCard({
               </div>
             )}
           </Link>
-          <EntityButton type={itemType} slug={path as string} />
+          <EntityButton
+            type={itemType}
+            slug={path as string}
+            animated={animated}
+          />
```
leafty, line 82: This doesn't seem to be doing what is intended when the button renders above the fold. It animates immediately which is a bit odd.

andre-code:
@lorenzo-cavazzi, thank you for your review. I have incorporated the suggestions for the backButton, the link to the howToGuide, and the import. As for the changes in the texts, I propose discussing them with the product team and incorporating them into another PR. The same approach can be taken for the suggestions regarding animations.

Regarding the message in the skeleton card when there is an error loading the project, I added it to quickly identify errors in deployment, such as missing the project path or an invalid path. The purpose is to quickly recognize the mistake and address it. I believe that acceptance tests may not capture this error, as it could occur during the deployment of Renku.

lorenzo-cavazzi (approved):
> Regarding the message in the skeleton card when there is an error loading the project, I added it to quickly identify errors in deployment, such as missing the project path or an invalid path. The purpose is to quickly recognize the mistake and address it.

If the project isn't reachable anymore (E.G. accidentally deleted or otherwise not available) users will see the error.
In general, the UI should only show errors targeting users, not administrators. Those should be logged though Sentry instead.

Not blocking for merging the PR since we are short on time, but it would be nice to have this changed later.

lorenzo-cavazzi (approved):

PR #2901, opened 2023-11-15
feat: enable RW option on cloud storage

### LINKED ISSUE #2899: Add a Write option on cloud storage
## Motivation

Backend will soon support writing to external storage. We should allow users to enable/disable it on each external storage.

## Proposal

This is simple, just add an `<option>` for turning "Write" on/off. Default should be off

lorenzo-cavazzi:
Enable Read-Write mode by default, leaving the possibility to limit the access to "read-only"

image

image

/deploy
fix #2899

### DIFF
--- a/client/src/features/project/components/AddCloudStorageButton.tsx
+++ b/client/src/features/project/components/AddCloudStorageButton.tsx
@@ -362,7 +362,7 @@ function AdvancedAddCloudStorage({
                           className={cx("form-check-label", "ms-2")}
                           for="addCloudStorageReadWrite"
                         >
-                          Read/Write
+                          Read-Write
                         </Label>
                       </div>
                     </>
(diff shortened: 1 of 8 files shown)


client/src/features/project/components/AddCloudStorageButton.tsx
```diff
@@ -675,55 +653,34 @@ function SimpleAddCloudStorage({
...
-                        >
-                          Read/Write
-                        </Label>
-                      </div>
-                    </>
-                  )}
-                />
-              </div>
-            )}
+            <div className="mb-3">
```
leafty, line 656: Because this is the last item:

```suggestion
            <div>
```

leafty (approved):
I think the message could be "Check this box to mount the storage in read-only mode. Use this setting to prevent accidental data modifications."

lorenzo-cavazzi:
> I think the message could be "Check this box to mount the storage in read-only mode. Use this setting to prevent accidental data modifications."

That text reads better than the previous one :+1:

PR #2903, opened 2023-11-17
fix(client): handle errors from the getSessions query in the Dashboard

### LINKED ISSUE #2873: Handle errors from the servers endpoint in the Dashboard
## Description

The dashboard, and specifically the projects section, keeps refreshing at short intervals when the `GET /servers` endpoint from renku-notebooks fails to respond.
We should handle the error better and prevent this behavior. Best would be to show an error message so that users know sessions aren't currently shown.

## Reproduce

Use the e2e tests, pick one for rendering the dashboard and remove the servers endpoint response (or fail it)

https://github.com/SwissDataScienceCenter/renku-ui/assets/43481553/008500fd-6eba-4734-842b-723c6702cd36

leafty:
Fixes #2873.

Introduces a new context and component to handle RTK Query errors in the case when nested components use an errored query.

/deploy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changes
 
+## [3.15.1](https://github.com/SwissDataScienceCenter/renku-ui/compare/3.15.0...3.15.1) (2023-11-14)
+
+### Bug Fixes
+
+- Address visual glitches on many pages ([#2883](https://github.com/SwissDataScienceCenter/renku-ui/issues/2883)).
+- Fix links on the search page cards ([#2887](https://github.com/SwissDataScienceCenter/renku-ui/issues/2887)).
+- Prevent showing wrong options on the Session dropdown menu when the project namespace includes uppercase letters ([#2874](https://github.com/SwissDataScienceCenter/renku-ui/issues/2874)).
+- Correctly handle Statuspage down ([#2871](https://github.com/SwissDataScienceCenter/renku-ui/issues/2871), [#2871](https://github.com/SwissDataScienceCenter/renku-ui/issues/2871)).
+- Ignore the case when storing last visited projects to prevent duplicates on the Dashboard ([#2892](https://github.com/SwissDataScienceCenter/renku-ui/issues/2892)).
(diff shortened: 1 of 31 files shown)


lorenzo-cavazzi (commented):
Nice solution and the code looks good :+1: 

Should we show a banner on the Dashboard mentioning there is a problem with sessions when we get errors? Otherwise, users do not know that's the case and all looks fine
image

leafty:
> Should we show a banner on the Dashboard mentioning there is a problem with sessions when we get errors? Otherwise, users do not know that's the case and all looks fine

I add 3befaf0, but I am not sure if users will understand the error message.

lorenzo-cavazzi:
Oh, I meant something more generic, like a Warning saying "Sessions are currently unavailable"

leafty:
> Oh, I meant something more generic, like a Warning saying "Sessions are currently unavailable"

Done in 65d40d3.

lorenzo-cavazzi (changes requested):
Very nice solution!

I noticed the function `extractRkErrorMessage` isn't working as expected. I'm unsure whether the Notebooks error is different from the Core error, or if something else is wrong in the logic there.
It would be great if you had some spare time to look at it. Otherwise, feel free to remove the specific error message and we can make a follow-up issue to fix the "RkErrorMessage" component(s) -- it would be great to have one that "just works", or to clearly distinguish between "renku-core" and "renku-notebooks" errors.

image

leafty:
> Very nice solution!
> 
> I noticed the function `extractRkErrorMessage` isn't working as expected. I'm unsure whether the Notebooks error is different from the Core error, or if something else is wrong in the logic there. It would be great if you had some spare time to look at it. Otherwise, feel free to remove the specific error message and we can make a follow-up issue to fix the "RkErrorMessage" component(s) -- it would be great to have one that "just works", or to clearly distinguish between "renku-core" and "renku-notebooks" errors.

I think it's not as simple because if I remove the notebooks service I get a proper error:
image

lorenzo-cavazzi (approved):

PR #2907, opened 2023-11-20
fix(client): adjust dropdown menus  with anchors nested in buttons

lorenzo-cavazzi:
The `DropdownItem` component should render either a button or a link based on its properties.
Since we use `Link` lo link local resources, we ended up nesting the 2 components instead of using `<DropdownItem href=...>`. That renders an anchor inside a button, with some undesired side effects, like making the areas next to the borders clickable but with no effect (in the gif, the fixed version is on the right).

Peek 2023-11-20 16-01

/deploy

### DIFF
--- a/client/src/components/ssh/ssh.tsx
+++ b/client/src/components/ssh/ssh.tsx
@@ -65,7 +65,7 @@ function SshDropdown({ fullPath, gitUrl }: SshDropdownProps) {
   };
 
   return (
-    <DropdownItem onClick={() => handleClick()}>
+    <DropdownItem onClick={handleClick}>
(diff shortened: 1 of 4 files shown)


client/src/components/ssh/ssh.tsx
```diff
@@ -66,10 +66,7 @@ function SshDropdown({ fullPath, gitUrl }: SshDropdownProps) {
 
   return (
     <DropdownItem onClick={() => handleClick()}>
```
leafty, line 68: nitpick:

```suggestion
    <DropdownItem onClick={handleClick}>
```

leafty (approved):

leafty (approved):

PR #2916, opened 2023-11-22
feat(client): add tolerations and node affinities to resource classes

leafty:
Renku admins can now add tolerations and node affinities to resource classes

renku-ci-ui-2916 dev renku ch__renku_login=1

/deploy


lorenzo-cavazzi (approved):
The code looks good!
I'm approving but I'm not 100% sure about how this is meant to work. E.G. why can't I add tolerations/affinities in a new class but only when editing an existing one? Is that intended?
It would be good if a YAT person (or some user) could give it a quick look before merging

image

P.S. as a side note, I noticed that loading the Admin page flashes a 404 page. Not a big deal, and not related to this PR either :grin: 

Peek 2023-12-12 07-49

leafty:
> E.G. why can't I add tolerations/affinities in a new class but only when editing an existing one? Is that intended?

Fixed in [6307d39](https://github.com/SwissDataScienceCenter/renku-ui/pull/2916/commits/6307d3987b310c577440e0d7be3cc327be60e5ec).

> P.S. as a side note, I noticed that loading the Admin page flashes a 404 page. Not a big deal, and not related to this PR either 😁

Maybe can be fixed later.

client/src/features/admin/AddResourceClassButton.tsx
```diff
@@ -279,6 +303,131 @@ function AddResourceClassModal({
               )}
             />
           </div>
+
+          <div className="mb-3">
```
lorenzo-cavazzi, line 307: This is probably missing a margin top

image

Btw why not make the component reuseable and share it in the new/edit sections?

leafty: > Btw why not make the component reuseable and share it in the new/edit sections?

I could do this later, but this is not trivial as the form actions are not the same. For this reason I prefer to write readable code first and figure out if it is factorizable later.

lorenzo-cavazzi (approved):
Lgtm :rocket:

PR #2919, opened 2023-11-23
refactor(client): more robust App params

leafty:
Refactor app params to be typed and validated before they are set in `AppContext`.

Uses existing validation for object values which means some parameters (e.g. `HOMEPAGE`) are not type-safe.

/deploy extra-values=global.renku.cli_version=2.7.0 renku=build/3325-pinned-projects


client/src/utils/context/appParams.utils.ts
```diff
@@ -0,0 +1,261 @@
...
+  AppParams,
+  AppParamsBooleans,
+  AppParamsStrings,
+  PreviewThresholdParams,
+  PrivacyBannerLayoutParams,
+  TemplatesParams,
+  UploadThresholdParams,
+} from "./appParams.types";
+
+export function validateAppParams(params: unknown): AppParams {
```
ciyer, line 33: `validateAppParams` makes me think that the function will just check if the AppParams are valid, but in this case it returns an AppParams object that has been cleaned.

```suggestion
export function validatedAppParams(params: unknown): AppParams {
```

client/src/index.js
```diff
@@ -20,6 +20,7 @@ import { UserCoordinator } from "./user";
 import { Sentry } from "./utils/helpers/sentry";
 import { Url, createCoreApiVersionedUrlConfig } from "./utils/helpers/url";
 import { AppErrorBoundary } from "./error-boundary/ErrorBoundary";
+import { validateAppParams } from "./utils/context/appParams.utils";
```
ciyer, line 23: (See comment in appParams.utils)
```suggestion
import { validatedAppParams } from "./utils/context/appParams.utils";
```

ciyer (approved):
This is a very nice refactoring. I have one proposal about renaming a function to give it a clearer name, otherwise once the tests run and pass, I think this can be merged.

ciyer (approved):
🎉 !

PR #2920, opened 2023-11-23
fix(client): use password fields for credentials

leafty:
Screenshot 2023-11-23 at 16 42 37

/deploy renku=auto-update/renku-core-2.8.0 extra-values=global.renku.cli_version=2.8.0rc3


lorenzo-cavazzi (approved):
Nice solution! :rocket:

leafty:
Merging decision: failing test is unrelated and unsolvable (`CLI test failing to install renku==2.8.0` which is not released yet).

PR #2924, opened 2023-11-27
refactor(client): use Vite instead of create-react-app

### LINKED ISSUE #2879: Migrate from Create React App to Vite
## Motivation

Create React App isn't maintained anymore. In the long term, we are considering server-side rendering but migrating to solutions like Next.js requires some time, while swapping CRA with [Vite](https://vitejs.dev/) should be quicker.

## Proposal

Give it a try, but time limit the effort to 2 days. If it turns out to be too time-consuming, we might want to reconsider this intermediate step.

leafty:
Fixes #2879.

Note: reviewing and merging can be delayed to after the next tagged release.


.github/workflows/test-and-ci.yml
```diff
@@ -76,6 +76,8 @@ jobs:
           node-version: "18.17.1"
       - name: Install dependencies
         run: npm clean-install
+      - name: Install Playwright
+        run: npx playwright install --with-deps
```
ciyer, line 80: **question:** why is it necessary to install playwright? I think replacing cypress is something that might be worth considering, but I would keep that to a separate issue/PR if possible.

client/src/components/list/ListCard.stories.tsx
```diff
@@ -1,5 +1,5 @@
 import { Meta, StoryObj } from "@storybook/react";
-import { ARG_REDUX_PATH, PARAM_REDUX_MERGE_STATE } from "addon-redux";
+// import { ARG_REDUX_PATH, PARAM_REDUX_MERGE_STATE } from "addon-redux";
```
ciyer, line 2: **question:** Is this work in progress? Otherwise, I think we can just remove the unused code, no need to comment it out.

ciyer (commented):
**praise:** This is very good work here! A lot of tedious changes to enable this transition, but they are straightforward to follow. 

Code-wise, there were two small things that I did not understand. I still need to run the branch locally, but the changes all made sense to me.

.github/workflows/test-and-ci.yml
leafty: If I remember correctly, this is due to the updates made to Storybook (storybook testing library). With the updated setup, the command `npm run storybook-compile-and-test` fails because there is no headless browser.

This is the error log (https://github.com/SwissDataScienceCenter/renku-ui/actions/runs/7020244553/job/19099502002):
```
[TEST] Error: Executable doesn't exist at /home/runner/.cache/ms-playwright/chromium-1091/chrome-linux/chrome
[TEST] ╔═════════════════════════════════════════════════════════════════════════╗
[TEST] ║ Looks like Playwright Test or Playwright was just installed or updated. ║
[TEST] ║ Please run the following command to download new browsers:              ║
[TEST] ║                                                                         ║
[TEST] ║     npx playwright install                                              ║
[TEST] ║                                                                         ║
[TEST] ║ <3 Playwright Team                                                      ║
[TEST] ╚═════════════════════════════════════════════════════════════════════════╝ Failed to launch browser.
```

client/src/components/list/ListCard.stories.tsx
leafty: I could not make `addon-redux` work as part of the migration to Vite. Should we remove the commented code and list it as a change in the PR description?

.github/workflows/test-and-ci.yml
leafty: I am not sure I understand how this was working beforehand. The `@storybook/test-runner` package documents that it is relying on playwright to run tests. :thinking:

client/src/components/list/ListCard.stories.tsx
ciyer: I see -- I think it's ok to leave the code commented out and just add a note that reviving the addon-redux is a TODO.

.github/workflows/test-and-ci.yml
ciyer: Thanks for the clarification.

client/src/components/list/ListCard.stories.tsx
leafty: Done in [b71fa07](https://github.com/SwissDataScienceCenter/renku-ui/pull/2924/commits/b71fa07b3e124d9d95f95d5106b1608bbefdaad0).

ciyer (approved):
This works very well. I tested running telepresence with the new config, and that worked, so I think this is ready to merge.

PR #2927, opened 2023-11-29
feat: auto-caroussel on the landing page

leafty:
https://github.com/SwissDataScienceCenter/renku-ui/assets/951086/a005fb01-e79c-49dd-8344-70719ef48f76


andre-code (commented):
Great, thanks for adding this slide. I think 5 seconds would be enough instead of 10 seconds, and it would also be amazing to include a countdown progress line. These are ideas that we can add later. Also, preload the images so the slide effect looks smooth.

andre-code (approved):
Gorgeous 😍
Thanks for adding the animation; the code looks good to me.

PR #2928, opened 2023-11-29
feat: add /version endpoints

### LINKED ISSUE #1652: Add /version endpoint
The UI and the UI server should expose a public `GET /version` endpoint.

The response format should be the following one:
```
{
    "name": string,
    "versions": [
        {
            "version": string,
            "data": object
        }
    ]
}
```

* "name" will be either `renku-ui` or `renku-ui-server`
* "version" should be a SemVer valid value, possibly the tag or some combination of last tag + commit short-sha if it's a dev version
* "data" is optional, and it can be any arbitrary valid JSON object. I would either leave it out initially, or add the commit short-sha of the last commit `data: { commit: <short-sha> }`

leafty:
Fixes #1652.

Version endpoint: https://renku-ci-ui-2928.dev.renku.ch/ui-server/api/version.

Note: for the UI client, we can hit `/config.json` and get the information that way.


andre-code (approved):
LGTM 🚀

PR #2938, opened 2023-12-08
chore(client): let users delete a session from within it

### LINKED ISSUE #2825: Check if there is unsaved work before deleting a session
Right now when I go to delete a running session. I get this message

image

But this happens even when I have no unsaved or unpushed changes in the repo. The ui can call the sidecar in the session when the delete session button is pressed and display a proper message. In other words we can tell for certain whether there is or isnt unsaved data. And only display the unsaved data message when appropriate.

### LINKED ISSUE #2894: Add a Delete session button on the session page
## Motivation

Many sessions are meant to be temporary and can be deleted right away without pausing first.
We already allow Pause and Delete from the Sessions dropdown button, but users can only pause from the full-screen sessions page.

image

## Proposal

We could modify the Pause Session modal to allow stopping the session right away.
The text should clarify that pausing is safer and encourage it while deleting is permanent. Since users easily skip texts, we could try to:

- Move the "`x` weeks inactivity" info in the main text and re-use the alert to mention that deleting is permanent (perhaps a warning instead of info, but feel free to try different combinations and pick the best)
- Add a `Delete session` danger button and shorten the text on the "Back to Session" button -- "Cancel" should do it.

Ofc adapt the rest or the text where needed (E.G. the title)

### Additional information

There are concerns that this could confuse the users since they don't always read texts and might have no idea about the difference between pausing and stopping.
After pushing the PR, we can discuss and decide whether we like the solution or not. Closing a small PR instead of merging it isn't necessarily bad :slightly_smiling_face:

leafty:
Fixes #2894, fixes #2825.

* Adds a "Delete session" button on the session page
* The delete session modals have been updated to fetch the current `git` status.

renku-ci-ui-2938 dev renku ch_projects_flora thiebaut_a-playground-project_sessions_show_flora-2eth-a-2dplayground-2dproject-a4570512
renku-ci-ui-2938 dev renku ch_projects_flora thiebaut_a-playground-project_sessions_show_flora-2eth-a-2dplayground-2dproject-a4570512 (1)
renku-ci-ui-2938 dev renku ch_projects_flora thiebaut_a-playground-project_sessions_show_flora-2eth-a-2dplayground-2dproject-a4570512 (2)
renku-ci-ui-2938 dev renku ch_projects_flora thiebaut_a-playground-project_sessions_show_flora-2eth-a-2dplayground-2dproject-a4570512 (3)

/deploy


lorenzo-cavazzi (approved):
Very nice solution!

Just a detail: I can still click on the Delete session button while loading the session status on the Delete session modal. Should that be disabled until the session status is fetched?

leafty:
> Just a detail: I can still click on the Delete session button while loading the session status on the Delete session modal. Should that be disabled until the session status is fetched?

I don't think so, it's going to frustrate users who know what they are doing.

PR #2939, opened 2023-12-11
chore(client): update the session start warning

### LINKED ISSUE #2906: Change the unavailable image warning once the image is available
## Motivation

When trying to quick start a session just after a commit has been created (mind that many interactions requiring invoking renku-core APIs end up creating a commit) it's rather common to get the yellow warning because the new base image hasn't been built yet. Still, it will probably be ready within a few seconds, so the user can just try to click again.

image (1)

This turns out to be confusing for some users who don't know exactly what was the problem.

## Proposal

As soon as the image is available, we could either:

- Change the alert to Info and mention there was an error cause the image wasn't ready, but now it's available and trying to start the session again should work. Or...
- Add another Info alert below the Warning to explain the image is no available and trying to start the session again should work.

I generally don't like swapping alerts, but adding multiple alerts might be overkill and users might ignore the latest addition.

leafty:
Fixes #2906.

Update the session start warning when the image has finished building.

Screenshot from 2023-12-11 13-14-05
Screenshot from 2023-12-11 13-16-04
Screenshot from 2023-12-11 13-18-58

/deploy

### DIFF
--- a/client/src/features/project/GitLab.types.ts
+++ b/client/src/features/project/GitLab.types.ts
@@ -48,6 +48,7 @@ export interface GitLabPipelineJob {
   status:
     | "success"
     | "running"
+    | "created"
     | "pending"
     | "stopping"
     | "failed"
--- a/client/src/features/project/projectGitLab.api.ts
+++ b/client/src/features/project/projectGitLab.api.ts
@@ -82,6 +82,7 @@ const projectGitLabApi = createApi({
 
           const jobs = result.data as GitLabPipelineJob[];
           const found = jobs.find(({ name }) => name === jobName);
+          console.log({ jobs, found });
           if (found) {
             return { data: found };
           }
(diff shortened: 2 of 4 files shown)


lorenzo-cavazzi (changes requested):
I like the solution but something might be off in the logic. I made a new project and started a session right away, I got a yellow warning but it turned into a red one for failed images when the job _succeded_.
Refreshing the page shows the image as available

image

leafty:
> I like the solution but something might be off in the logic. I made a new project and started a session right away, I got a yellow warning but it turned into a red one for failed images when the job _succeded_. Refreshing the page shows the image as available

Solved in [4e013cf](https://github.com/SwissDataScienceCenter/renku-ui/pull/2939/commits/4e013cf8508eb270df797e6ffff96416487e10b2).

lorenzo-cavazzi (approved):
Now that I see it, the additional start button in the warning looks a little weird... but that fixed the issue as intended :+1: 

@rokroskar what do you think about this solution?

PR #2940, opened 2023-12-11
fix(client): fix numeric id redirect for projects

### LINKED ISSUE #1490: Support numeric namespaces
## Description

Users with a numeric namespace, intended as numbers **only**, can't access their projects (nor can anyone else).
The problem is that the URL has the form `<renkulab_deployment>/projects/12345/<project_name>` and the numeric part is interpreted as a numeric project id.

## Reproduce

Try to access this project created in the user namespace `12345`. The user registered through an email address like `12345@email.com`.
https://dev.renku.ch/projects/12345/test

leafty:
Fixes #1490.

/deploy

### DIFF
--- a/client/src/project/Project-Route.test.js
+++ b/client/src/project/Project-Route.test.js
@@ -138,37 +138,14 @@ describe("id route extraction", () => {
     expect(pathComponents.baseUrl).toEqual("/projects/1");
     expect(pathComponents.namespace).toEqual(null);
   });
-  it("handles project-id sub routes", () => {
-    const pathComponents = splitProjectSubRoute("/projects/1/overview");
-    expect(pathComponents.projectPathWithNamespace).toEqual(null);
-    expect(pathComponents.projectId).toEqual("1");
-    expect(pathComponents.baseUrl).toEqual("/projects/1");
-    expect(pathComponents.namespace).toEqual(null);
-  });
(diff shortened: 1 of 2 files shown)


lorenzo-cavazzi (changes requested):
The project works well when created, but the application crashes when picking a numeric namespace on the new project page. Mind that it happens only if that's the first namespace to be selected, and not when something else was already manually selected first (see gif)

Peek 2023-12-12 08-14

leafty:
> The project works well when created, but the application crashes when picking a numeric namespace on the new project page. Mind that it happens only if that's the first namespace to be selected, and not when something else was already manually selected first (see gif)

The crash happens because the group endpoint, e.g. `/ui-server/api/groups/7698` responds with 404. When you select another group first, then the failure still happens but the hooks we use propagate old data (from the previously selected group).

lorenzo-cavazzi:
> The issue is caused by GitLab, so I think I would like to mark the issue as non-resolvable and not support numeric project paths. Users can modify the path in GitLab and things should work fine after that.

That sounds like a reasonable solution. We might want to enforce this by filtering out numeric-only namespaces and showing a warning mentioning we don't support numbers-only namespaces (I assume it's rare, but perhaps some groups?).
Considering we are moving towards a different backend for the projects, we don't necessarily need to do anything for now.

PR #2950, opened 2023-12-20
refactor(client): update branch and commit options

### LINKED ISSUE #2725: Start session options: load more commits with RTK query (pagination)
This doesn't need to change in this PR, but users find this cog icon non-intuitive and I regret coming up with that solution.
Perhaps we can set here `100` as default, and in a follow-up issue we could improve the UX: one solution would be showing the first few commits (25 would be ok) and adding a "load more" row at the bottom that actually incrementally loads more commits (100 by 100?)

_Originally posted by @lorenzo-cavazzi in https://github.com/SwissDataScienceCenter/renku-ui/pull/2719#discussion_r1287317205_

leafty:
Fixes #2725.

/deploy


client/src/features/session/components/options/SessionBranchOption.tsx
```diff
@@ -57,35 +72,87 @@ export default function SessionBranchOption() {
...
     { skip: !gitLabProjectId }
   );
+  const {
+    data: branchesFirstPage,
+    isError: branchesFirstPageIsError,
+    isFetching: branchesFirstPageIsFetching,
+    requestId: branchesFirstPageRequestId,
+  } = useGetRepositoryBranchesQuery(
+    {
+      projectId: `${gitLabProjectId ?? 0}`,
```
lorenzo-cavazzi, line 93: `?? 0` isn't necessary considering `skip: !gitLabProjectId`

client/src/features/session/components/options/SessionBranchOption.tsx
```diff
@@ -57,35 +72,87 @@ export default function SessionBranchOption() {
...
+    fetchedPages: 0,
+    hasMore: true,
+    currentRequestId: "",
+  });
+
+  const [fetchBranchesPage, branchesPageResult] =
+    projectGitLabApi.useLazyGetRepositoryBranchesQuery();
+  const onFetchMore = useCallback(() => {
+    const request = fetchBranchesPage({
+      projectId: `${gitLabProjectId ?? 0}`,
```
lorenzo-cavazzi, line 124: Should the query be skipped if `gitLabProjectId` isn't set instead of using `0`?

client/src/features/session/components/options/SessionBranchOption.tsx
```diff
@@ -57,35 +72,87 @@ export default function SessionBranchOption() {
...
+    fetchedPages,
+    gitLabProjectId,
+  ]);
 
   const currentBranch = useAppSelector(
     ({ startSessionOptions }) => startSessionOptions.branch
   );
 
+  // Handle forced refresh
+  const [refetch] = useRefetchBranchesMutation();
```
lorenzo-cavazzi, line 144: Oh, interesting... is this to invalidate the tags?

client/src/features/session/components/options/SessionCommitOption.tsx
```diff
@@ -68,25 +72,84 @@ export default function SessionCommitOption() {
...
+    hasMore: true,
+    currentRequestId: "",
+  });
+
+  const [fetchCommitsPage, commitsPageResult] =
+    projectGitLabApi.useLazyGetRepositoryCommitsQuery();
+  const onFetchMore = useCallback(() => {
+    const request = fetchCommitsPage({
+      branch: currentBranch,
+      projectId: `${gitLabProjectId ?? 0}`,
```
lorenzo-cavazzi, line 90: Same here

lorenzo-cavazzi (changes requested):
Nice solution!
Just a couple of inline comments.

P.S. One generic comment about this library: it seems to re-draw all the dropdown elements even when just hovering one. Not great, but I guess there isn't much we can do
Peek 2024-01-15 16-53

client/src/features/session/components/options/SessionBranchOption.tsx
leafty: It is needed for typing.

client/src/features/session/components/options/SessionBranchOption.tsx
leafty: Same as above, this is done for typing, the `onFetchMore()` callback is never used when the gitlab project id is not known.

client/src/features/session/components/options/SessionBranchOption.tsx
leafty: Yes, the mutation only invalidates tags.

client/src/features/session/components/options/SessionBranchOption.tsx
lorenzo-cavazzi: I don't get typing errors when removing `?? 0` since it's converting to string.  `${null}` should work fine

lorenzo-cavazzi (dismissed):
Mind that, until the PRs from the other renku repositories are merged (probably tomorrow), the `useSessions` tests won't pass correctly unless you set up a csi driver

client/src/features/session/components/options/SessionBranchOption.tsx
leafty: I removed the `?? 0`.

client/src/features/session/components/options/SessionBranchOption.tsx
leafty: Fixed it.

lorenzo-cavazzi (approved):
Lgtm :rocket:

lorenzo-cavazzi (changes requested):
Oh actually, I noticed a strange behavior while playing around with the branches (apart from the flashing error on the resource pools :grin: unrelated to this PR).
I select a bunch of branches and it always works well, until I choose the same branch already selected on the list and then it breaks: it keeps loading and never shows some other fields again.

Peek 2024-01-16 17-37

leafty:
> Oh actually, I noticed a strange behavior while playing around with the branches (apart from the flashing error on the resource pools 😁 unrelated to this PR). I select a bunch of branches and it always works well, until I choose the same branch already selected on the list and then it breaks: it keeps loading and never shows some other fields again.

Fixed in 063f786.

lorenzo-cavazzi (approved):
Lgtm :rocket:

PR #2959, opened 2024-01-10
Adapt test setup for vite-based server

ciyer:
The change to vite requires changes to how the tests are started.


leafty (approved):
LGTM 🎉 thanks

PR #2962, opened 2024-01-10
[ShapeUp] improve UX when adding/editing storage

lorenzo-cavazzi:
This PR fixes a few issues when adding/editing storage:
- Prevents a bunch of default values to be set automatically, especially when many advanced options exist.
- Correct use of placeholders for default values and examples.
- Remove values correctly when editing an existing storage.
- Handles non-stringy values correctly when listing the storage options in the storage list.
- Use `<datalist>` to pick values from the examples list.
    - Mind that examples from the backend are not 100% reliable; we can't use a `<select>` here, and I would rather keep a simple solution than mess up with autocomplete examples and "use-react-form." Support for the datalist tag is pretty good on Firefox, while the UX on Chrome is still not that good, especially for longer lists.

/deploy renku=rclone-sessions renku-notebooks=rclone-csi amalthea=main renku-data-services=main extra-values=notebooks.cloudstorage.enabled=true,notebooks.cloudstorage.storageClass=csi-rclone-test,notebooks.cloudstorage.installRCloneCSI=false

### DIFF
--- a/client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
+++ b/client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
@@ -17,7 +17,14 @@
  */
 
 import cx from "classnames";
-import { RefObject, useCallback, useMemo, useRef, useState } from "react";
+import {
+  RefObject,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
 import {
   ExclamationTriangleFill,
   EyeFill,
@@ -35,9 +42,6 @@ import {
   Label,
   ListGroup,
   ListGroupItem,
-  PopoverBody,
-  PopoverHeader,
-  UncontrolledPopover,
   UncontrolledTooltip,
 } from "reactstrap";
 
@@ -49,7 +53,7 @@ import {
   AddCloudStorageState,
   CloudStorageDetails,
   CloudStorageSchema,
-  CloudStorageSchemaOptions,
+  CloudStorageSchemaOptionExample,
 } from "./projectCloudStorage.types";
 import {
   convertFromAdvancedConfig,
@@ -203,8 +207,8 @@ function AddStorageAdvancedToggle({
     </>
   );
 }
-// *** Add storage: helpers *** //
 
+// *** Add storage: helpers *** //
 interface AddStorageStepProps {
   schema: CloudStorageSchema[];
   setStorage: (newDetails: Partial<CloudStorageDetails>) => void;
@@ -537,53 +541,6 @@ function AddStorageType({
 }
 
 // *** Add storage: page 2 of 3, with storage options *** //
-
-interface AddStorageOptionsExamplesProps {
-  examples: CloudStorageSchemaOptions["examples"];
-  name: string;
-  provider?: string;
-}
-function AddStorageOptionsExamples({
-  examples,
-  name,
-  provider,
-}: AddStorageOptionsExamplesProps) {
-  // TODO: We should use examples as enum, or items in an autocomplete. But sometimes we get just "None" 🥴
-  if (!examples?.length) return null;
-  const filterExamples = provider
-    ? examples.filter((e) => !e.provider || e.provider === provider)
-    : examples;
-  if (!filterExamples?.length) return null;
-
-  const popoverId = `popover-${name}`;
-  const exampleItems = filterExamples.map((e) => {
-    return (
-      <li className="mb-1" key={e.value}>
-        <code>{e.value}</code>
-        <br />
-        <small>{e.help}</small>
-      </li>
-    );
-  });
-
-  return (
-    <>
-      <QuestionCircle id={popoverId} className={cx("bi", "ms-1")} />
-      <UncontrolledPopover
-        target={popoverId}
-        placement="right"
-        style={{ maxHeight: "98vh", overflow: "auto" }} // eslint-disable-line spellcheck/spell-checker
-        trigger="hover focus"
-      >
-        <PopoverHeader>Examples</PopoverHeader>
-        <PopoverBody>
-          <ul>{exampleItems}</ul>
-        </PopoverBody>
-      </UncontrolledPopover>
-    </>
-  );
-}
-
 interface SecretOptionWarningProps {
   isSecret?: boolean;
   name: string;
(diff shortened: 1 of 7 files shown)


client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
```diff
@@ -608,6 +565,81 @@ function SecretOptionWarning({
...
+  collapsedLines = 2,
+  linesTolerance = 1,
+  text,
+}: TruncatedTextProps) {
+  const [isTruncated, setIsTruncated] = useState(true);
+  const contentRef = useRef<HTMLDivElement>(null);
+
+  const toggleTruncate = () => {
+    setIsTruncated(!isTruncated);
+  };
```
leafty, line 599: Prefer `useCallback()` version:

```suggestion
  const toggleTruncate = useCallback(() => {
    setIsTruncated((isTruncated) => !isTruncated);
  }, []);
```

client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
```diff
@@ -608,6 +565,81 @@ function SecretOptionWarning({
...
+  }, [collapsedLines, linesTolerance, text]);
+
+  const textStyle = {
+    maxHeight: isTruncated ? `${collapsedLines * 1.5}em` : "none",
+  };
+
+  return (
+    <div>
+      <div
+        className={cx("overflow-hidden", "local-style")}
```
leafty, line 621: What's `"local-style"`?

client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
```diff
@@ -648,6 +680,7 @@ function AddStorageOptions({
     setStorage({ options: validOptions });
   };
 
+  // ? We treat checkbox and password separately, while the rest is an input
   const optionItems =
     options &&
     options.map((o) => {
```
leafty, line 686: This is difficult to read, consider moving this to components (e.g. `<CheckboxOptionItem>`, `<PasswordOptionItem>`, `<DropdownOptionItem>`, etc.).

client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
```diff
@@ -683,7 +704,7 @@ function AddStorageOptions({
       return (
         <div className="mb-3" key={o.name}>
           <label htmlFor={o.name}>
-            {o.friendlyName ?? o.name} {examples} {warning}
+            {o.friendlyName ?? o.name} {warning}
           </label>
 
           {inputType === "checkbox" ? (
```
leafty, line 710: Checkboxes should be placed to the left of their label, not to the right.

client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
```diff
@@ -751,35 +772,64 @@ function AddStorageOptions({
...
+                      )}
+                    </>
+                  );
+                }}
+              />
+            </>
           )}
-          <div className={cx("form-text", "text-muted")}>{o.help}</div>
+          <div className={cx("form-text", "text-muted")}>
+            {/* {o.help} */}
```
leafty, line 830:
```suggestion
```

client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
```diff
@@ -816,6 +866,10 @@ function AddStorageOptions({
 
   const sourcePath = (
     <div className="mb-3">
+      <Label className="form-label" for="add-storage-name">
```
leafty, line 869: Wrong `for=` here.

client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
```diff
@@ -608,6 +565,81 @@ function SecretOptionWarning({
     </>
   );
 }
+interface OptionDataListProps {
+  examples: CloudStorageSchemaOptionExample[];
+  listId: string;
+}
+function OptionDataList({ examples, listId }: OptionDataListProps) {
```
leafty, line 572: Note: the same functionality can be achieved with `react-autosuggest`.

leafty (changes requested):

client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
lorenzo-cavazzi: Ouch, that's a leftover from a previous test

client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
lorenzo-cavazzi: That's a good point; I split that long chunk of code into multiple components

lorenzo-cavazzi:
@leafty thank you for the review. I addressed your comments in ea48f28c400cfc9dee8614c01df04e39ca941646

There are a couple of additional commits that add minimal changes:
- Hide a couple of storage services that we don't want to support in the UI, at least initially.
- Restore the correct tag when invoking renku actions

client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
```diff
@@ -343,6 +347,246 @@ function AddStorageAdvanced({ storage, setStorage }: AddStorageStepProps) {
...
+  control,
+  defaultValue,
+  inputType,
+  onFieldValueChange,
+  option,
+}: InputOptionItemProps) {
+  const additionalProps: Record<string, string> = {};
+  if (inputType === "dropdown") {
+    additionalProps.list = `${option.name}__list`;
+  }
```
leafty, line 537:
```suggestion
  const additionalProps: Record<string, string> = {
    ...(inputType === "dropdown" ? { list: `${option.name}__list` } : {}),
  };
```

leafty (approved):
LGTM :100: just some small things

lorenzo-cavazzi:
@leafty fixes from the last review are in this commit https://github.com/SwissDataScienceCenter/renku-ui/pull/2962/commits/0a995c9717ac8a0de739a85d72fdac53d7e7bfbb

client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
```diff
@@ -343,6 +347,246 @@ function AddStorageAdvanced({ storage, setStorage }: AddStorageStepProps) {
...
+            Show all
+          </Button>
+        </>
+      )}
+    </div>
+  );
+}
+
+interface CheckboxOptionItemProps {
+  control: Control<FieldValues, any>; // eslint-disable-line @typescript-eslint/no-explicit-any
```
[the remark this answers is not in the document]
leafty: I think we can force this to be `void` if we call `useForm<Values, void>({<defaultValues>})` when using `react-hook-form`. The `context: TContext` field is used only if we use a third-party library for field validation.

`unknown` can work too I think. We don't need to address the use of `any` in this PR anyway.

leafty (approved):

PR #2965, opened 2024-01-11
chore(client): fix eslint config with vite

leafty:
* Add a plugin to run eslint with the dev server to print linting warnings during development
* Fix eslint rules to have `.jsx` files linted (they were previously skipped)

### DIFF
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -27,6 +27,7 @@
     }
   },
   "overrides": [
+    { "files": ["**/*.,s", "**/*.jsx"] },
     {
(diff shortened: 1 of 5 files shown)


client/.eslintrc.json
```diff
@@ -27,6 +27,7 @@
     }
   },
   "overrides": [
+    { "files": ["**/*.,s", "**/*.jsx"] },
```
ciyer, line 30: **question** why is there a glob for `"**/*.,s"`?

leafty: it's a typo

ciyer (approved):
Good catch!

PR #2970, opened 2024-01-15
Remove "author" search filter and add "role" search filter

### LINKED ISSUE #2523: Update the search page "owned by me" option to list exactly projects owned by the user
## Description

Currently the search is performed using the user's full name and may inaccurately match projects not owned by that user.

The KG will add a search param to match exactly based on the user's credentials. The issue on their side is the following: https://github.com/SwissDataScienceCenter/renku-graph/issues/1486

## Reproduce

1. Create alternate accounts with the same full name
2. Create a project from one account
3. Log out
4. Log in with an alternate account
5. Search projects with "Owned by me" option

EDIT: the KG issue has been solved; use the `development` branch, or any KG version >= 2.46.0
~Depends on: https://github.com/SwissDataScienceCenter/renku-graph/issues/1486~

leafty:
Fixes #2523.

Removed:
Screenshot 2024-01-17 at 14 35 28

Added:
Screenshot 2024-01-17 at 14 35 15

The new controls are finer ("owner", "maintainer", "reader") and should correspond to the state in GitLab.

/deploy


lorenzo-cavazzi (approved):
Lgtm! :rocket: 

As a side note, I'm not sure it's intuitive that removing all the ticks from the User Role section will show the public projects, but it might be just me. No need to change anything now; we can adjust it later if needed

PR #2972, opened 2024-01-16
Always specify the branch when dealing with datasets

lorenzo-cavazzi:
There is a breaking change from the renku-core APIs that requires the clients to either _always_ or _never_ specify the branch name. The reason is that the branch is used for caching, and not sending one results in falling back to a default name instead of picking the default branch name.

Since we know the default branch and we are already using it somewhere, this PR updates the code to always pass the default branch name.

/deploy renku=remove-v9 renku-core=develop


leafty (approved):
LGTM

PR #2977, opened 2024-01-17
fix(client): always specify branch for the renku-core API

### LINKED ISSUE #2974: Set branches on all the renku-core APIs
Renku-core recently changed how the cache works, making it "per branch".
As a side effect, invoking APIs without providing a branch potentially leads to unpredictable outcomes because it doesn't fall back to the default branch as one would expect, but instead clones the default branch into a sort of "no-branch" folder that gets eventually synchronized to the remote.

That turns out to be problematic for datasets: some operations were done in the "no-branch" folder, while others were in the default branch folder. The UI was giving confusing feedback (some successful changes seemed to disappear), and after a while, the backend was synchronized again with even more unpredictable outcomes due to conflicts.

To prevent similar side effects in the future, we should go through the code and _always_ set a branch when invoking the APIs. It shouldn't be too difficult since the UI only supports the default branch.
The sections to update include the settings page, the workflows tab, and potentially others.
This PR https://github.com/SwissDataScienceCenter/renku-ui/pull/2972 can be used as an example of what should be changed.

leafty:
FIxes #2974.

Changes:
- Update query and mutation param types to require `branch` when is appropriate in `core` APIs.
- Update corresponding query and mutation hooks.
- Fix `Project.d.ts` typings file into `project.types.ts`.
- Removed some unused legacy API client functions which are not used.
- Removed unused List.jsx component.
- [Oops!] Fixes bug introduced in #2965 which crashed the Stats page.

/deploy


client/src/features/project/project.types.ts
```diff
@@ -243,13 +233,13 @@ export interface ProjectActivateIndexingResponse {
 
 export interface MigrationStartParams
   extends Pick<CoreVersionUrl, "apiVersion"> {
-  branch?: string;
+  branch: string;
```
ciyer, line 236: **question:** I wonder if it wouldn't be better to type this as
```
branch: string | null
```
There are several places in the code where the query is skipped if there is no branch, and you see things like:
```
{ ..., branch: branch ?? "" },
```
at the call site.

I think
```
{ ..., branch: branch ?? null },
```

Would be clearer in intention.

ciyer (commented):
**Praise** this looks very good, and I see that you removed some previously dead code as part of it. I have a question about how one field is typed.

client/src/features/project/project.types.ts
leafty: No, the type `branch: string` communicates that a proper branch is **required** to perform the query. Using `string | null` would communicate another intent. `branch: branch ?? ""` is used due to the limitation that the typescript compiler will expect `branch` to be of type `string` and doesn't understand that we skip the query when a proper `branch` is not provided.

We can see if using a skip token makes code more readable in another PR -> https://redux-toolkit.js.org/rtk-query/usage-with-typescript#skipping-queries-with-typescript-using-skiptoken.

ciyer (approved):
👍

PR #2979, opened 2024-01-18
refactor(client): rename files from *.d.ts to *.types.ts

ciyer:
Files named *.d.ts have a particular semantics which is not appropriate for most code that is just type decls. 

/deploy


leafty (approved):
LGTM 🙏 Thanks

Note: for consistency with the guidelines we could also rename some of the files to e.g. `workflows.types.ts`.

ciyer:
> Note: for consistency with the guidelines we could also rename some of the files to e.g. `workflows.types.ts`.

I made this change, but I had to rewrite commit history because otherwise things would have been a bit messy on macOS.

leafty (approved):

PR #2984, opened 2024-01-19
refactor(client): transform pages into lazy components

leafty:
This refactor moves all pages imported at the root of the UI (from `App.jsx`) into lazy components. This change reduces the root bundle size and slightly improves the Lighthouse metrics. The main benefit of this change however is to improve the DevX around using `telepresence`: when reloading the UI after a code change, a much smaller number of source files will be reloaded by `vite` which makes development of the UI much smoother.

* Added `<LazyXYZPage />` components for each page defined in `App.jsx`
* Removed corresponding barrel files (see https://vitejs.dev/guide/performance.html#avoid-barrel-files)
* Moved around some components to fix build warnings

/deploy


lorenzo-cavazzi (commented):
Early comment while still going through the review: the change looks great, but we might want to tune the fallback components in a few cases.
When loading the page for the first time, it shows a Loader and the footer almost on top; it would be best to either show just the loader, or forcing the footer to the bottom
Peek 2024-01-23 11-51

lorenzo-cavazzi:
Code looks otherwise good :rocket:

lorenzo-cavazzi (approved):
Lgtm! :rocket:

PR #2985, opened 2024-01-19
feat: improve feedback when starting sessions on outdated projects

### LINKED ISSUE #2983: Show a better error message when session can't start on outdated projects
Currently, we show a generic error message when sessions cannot start cause the project is too old (see below).

We should show a more specific message:
- Customize it when the project is just too old, and add a link to the settings section where users can update the project.
- Add the specific error message for any other error.

For context, this was a somewhat rare occurrence until we recently phased out v9 on renkulab.io which has been supported for a long time, and now suddenly many users can't start a session and don't understand why from this error message.

image

ciyer:
If the project is outdated, provide a better explanation and a link to the sessions page.

**Project can be updated with one click**

image

**Project requires more work to update**

image

Fix #2983 

/deploy


lorenzo-cavazzi (approved):
Lgtm! :rocket:

PR #2989, opened 2024-01-22
chore: update telepresence scripts

leafty:
Update the `telepresence` to be inline with the current version of [Telepresence OSS](https://www.getambassador.io/docs/telepresence-oss/latest/install).

The current version prefers that the `connect` command is used before the `intercept` command. The scripts now check the `telepresence status` and re-connect to the correct k8s namespace if needed.

### DIFF
--- a/client/run-telepresence.sh
+++ b/client/run-telepresence.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright 2018 - Swiss Data Science Center (SDSC)
+# Copyright 2024 - Swiss Data Science Center (SDSC)
 # A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
 # Eidgenössische Technische Hochschule Zürich (ETHZ).
 #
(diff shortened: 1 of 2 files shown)


client/run-telepresence.sh
```diff
@@ -189,9 +189,16 @@ EOF
 
 ./scripts/generate_sitemap.sh "${BASE_URL}" "./public/sitemap.xml"
 
+CURRENT_TELEPRESENCE_NAMESPACE=$( telepresence status | grep "Namespace" | cut -d ":" -f2 | tr -d " " )
```
lorenzo-cavazzi, line 192: In my setup, `namespace` is lowercase; that's always set to the same value, which is != from `$DEV_NAMESPACE`.

client/run-telepresence.sh
```diff
@@ -189,9 +189,16 @@ EOF
...
+if [[ $DEV_NAMESPACE != $CURRENT_TELEPRESENCE_NAMESPACE ]]
```
lorenzo-cavazzi, line 193: What's the idea behind this? Prevent starting a new instance whenever one is already running?
In that case, since setups varies, I suggest checking the `Intercepts`
```
  Intercepts        : 0 total
```
```
  Intercepts        : 1 total
    renku-ci-ui-2994-ui-renku-ci-ui-2994: lorenzo@lorenzo-SDSC-9310

```

lorenzo-cavazzi (commented):
My setup seems to be slightly different, and checking for `Namespace` from `telepresence status` might not work as expected -- see inline

leafty:
> My setup seems to be slightly different, and checking for `Namespace` from `telepresence status` might not work as expected -- see inline

What is your version of telepresence? I used the latest Telepresence OSS install -> https://www.getambassador.io/docs/telepresence-oss/latest/install

client/run-telepresence.sh
leafty: No, this is to comply with how the current version of telepresence works. We should `connect` to the cluster with the correct namespace before doing an intercept. If we are already connected to the correct namespace, then there is nothing to do.

ciyer:
> > My setup seems to be slightly different, and checking for `Namespace` from `telepresence status` might not work as expected -- see inline
> 
> What is your version of telepresence? I used the latest Telepresence OSS install -> https://www.getambassador.io/docs/telepresence-oss/latest/install

I also had to update telepresence before things started working for me.

lorenzo-cavazzi (approved):
I was on 2.12; it works with the latest version of telepresence (2.17) :+1:

leafty:
Thanks @lorenzo-cavazzi and @ciyer. I made this PR to remove the warning logs I saw using the current telepresence version. :+1:

PR #2994, opened 2024-01-22
Add a Test connection button when adding/editing storage

### LINKED ISSUE #2993: Add a test connection button when adding/editing a storage
We now support a broader range of external storage services. Configuring one isn't always straightforward, and we currently don't give feedback on wrong configurations until launching a new session. Even worse, some errors only appear when starting sessions, or in the running session when trying to access a mounted folder.

To improve the UX, we want to add a "Test connection" button on the modal to add/edit storage, so that users can immediately verify the configuration.

lorenzo-cavazzi:
/deploy

fix #2993


ciyer:
Replaced by #3194

PR #2997, opened 2024-01-23
chore(client): update the project session list when empty

### LINKED ISSUE #2918: sessions page is very empty with no sessions
when there are no sessions, the sessions page is pretty empty and it might not be immediately  apparent for a new user that the button all the way in the top right is how you start a session.

Maybe we could add a sessions start button here as well or make it clearer how to start a session if there are none
image
in any case, feel free to ignore, it's really just a trivial suggestion

leafty:
FIxes #2918.

Screenshot 2024-01-24 at 09 16 25

Screenshot 2024-01-24 at 09 16 35

/deploy


lorenzo-cavazzi (approved):
Changes look good :+1: 

I'm not sure if we have a written convention for the button colors, but I kind of expected that button to be blue or secondary dark blue (like the Login button in the double alert screenshot) since it's on the info alert.

PR #2999, opened 2024-01-24
feat(client): update the template cards

### LINKED ISSUE #2387: Improve the display of templates and add an SSH enabled badge
## Motivation

We get the info on SSH support for templates when fetching the list of templates.

## Proposal

We should use that to show an SSH-enabled icon/text next to the template supporting it.

We will use the style shown here https://github.com/SwissDataScienceCenter/renku-ui/issues/2387#issuecomment-1462099847 -- consider the left-aligned text.

* Adjust the current style to match the one in the preview (mind removing the highlighted border that changes the card's size and adding proper padding)
* Add the flipped side of the card on mouse over with the description and the SSH icon _only on deployments supporting ssh_ (that info is already fetched with RTK query from the `GET /notebooks/verion` API). ℹ️ mind that very long template descriptions should not change the card size; show a vertical scrollbar instead

leafty:
Fixes #2387.

Update the template cards on the new project page.

Screenshot 2024-01-25 at 15 13 56

/deploy


client/src/components/templateSelector/TemplateSelector.module.scss
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  .templateCardDescription,
+  .templateCardSsh {
+    display: none;
+    background-color: $rk-green-highlight;
+  }
+
+  &:hover,
+  input:focus + & {
```
lorenzo-cavazzi, line 29:
```suggestion
  &:hover {
```
I guess that's to enable keyboard usability. As a consequence, it makes the interaction through the mouse a little confusing because the card I click keeps the same hover style instead of the "selected" style until I click somewhere else. So when I hover another card, it's not clear which one is actually hovered.
Not sure about an easy fix; maybe clicking should lose the focus?

Peek 2024-01-26 09-54

client/src/components/templateSelector/TemplateSelector.tsx
```diff
@@ -167,94 +169,205 @@ function TemplateGalleryRow({
...
-      : defaultTemplateIcon;
-    const id = "id" + simpleHash(repository.name) + simpleHash(t.id);
-    const selectedClass = selectedTemplate === t.id ? "selected" : "";
-    const invalidTemplate = isInvalid ? "template--invalid" : "";
-    const statusTemplate = isDisabled
-      ? "template--disabled cursor-not-allowed"
-      : "template--active cursor-pointer";
+  return (
+    <div>
+      <Row>
```
lorenzo-cavazzi, line 194: We might want to add space on top in case we have multiple sources (in the screenshot, there is no space between `Source: Telepresence` and the content above). This might require removing bottom spacing from `Template` to prevent excessive space in the first line

image

lorenzo-cavazzi (changes requested):
The code looks good- okay, I admit I had to read it twice to understand why there are so many `CardBody` :joy: 

I feel the hover/selected styling collides a little too much making the UX sub-optimal when using the mouse. Not sure how easy it would be to improve it -- I suggest a possible solution inline

client/src/components/templateSelector/TemplateSelector.module.scss
lorenzo-cavazzi: Sorry, I realize now the "Suggested change" is confusing; the actual suggestion is something totally different

leafty: Done in a6d1794.

client/src/components/templateSelector/TemplateSelector.tsx
leafty: Done in a6d1794.

lorenzo-cavazzi (approved):
Lgtm :rocket:

PR #3001, opened 2024-01-25
fix(client): hide button to add storage when the feature is turned off

lorenzo-cavazzi:
We should not show the button to add a new storage service in deployments that do not support it

image

/deploy renku=000-cypress-storage-off


leafty (approved):
LGTM

Issue #2605, opened 2023-06-19
Refactor UI routing


leafty:
Currently, routing in the UI is done in source files found in various places which makes searching and finding UI components based on the URL difficult to do, e.g. `/projects/:projectId/sessions/new` goes through `index.js` > `App.js` (2 components with `<Route>`) > `project/Project.present.js` (2 components with `<Route>`).

### Suggestion

Refactor routing to have it defined from `client/src/routes` so that finding which components are rendered based on the URL is made easier.

PR #3009, opened 2024-01-31
refactor(client): update UI routing

leafty:
See #2605.

First part of refactoring UI routing:
* Refactor routing to remove the use of the `render` in favor of a child component
* Refactored some components to remove one level of prop drilling and replace the propagation of `location` and `history` to use the hooks from `react-router`

/deploy renku=release-0.47.1


lorenzo-cavazzi (approved):
Lgtm! Thank for going through this :pray:

PR #3010, opened 2024-01-31
fix(client): fix landing page parallax background

leafty:
Fixes an issue with the parallax background on the landing page. The size of the background image was dependent on the browser's viewport which would create cases when the background image is repeated many times at a small size.

Before:
localhost_3000_ (5)
localhost_3000_ (6)

After:
localhost_3000_ (7)
localhost_3000_ (8)


ciyer (dismissed):
This graphic may need to be re-thought, since, after the fix, the dots look disproportionately big on large viewports.

image

But I think your fix is an improvement over the status quo. 👍

leafty:
> This graphic may need to be re-thought, since, after the fix, the dots look disproportionately big on large viewports.
> 
> But I think your fix is an improvement over the status quo. 👍

I'm using a fixed size now which seems to stay consistent across all viewport sizes.

ciyer:
> I'm using a fixed size now which seems to stay consistent across all viewport sizes.

Good idea, looks good on all viewport sizes now!

PR #3012, opened 2024-01-31
chore(ci/cd): add a timeout for the unit tests

leafty:
The current client unit tests (`npm run test`) usually run in ~11 seconds, but in some cases the test is stuck and takes more than 10 minutes (e.g. https://github.com/SwissDataScienceCenter/renku-ui/actions/runs/7727771816/attempts/1?pr=3011).

This adds a timeout to the corresponding step so that it fails fast and can be re-tried if needed.


lorenzo-cavazzi (approved):
Lgtm :+1:

PR #3014, opened 2024-02-02
fix(clent): restore logged in/out notifications

leafty:
Restore logged in and out notifications in the UI.

Screenshot 2024-02-01 at 10 35 29

/deploy


lorenzo-cavazzi (approved):
Thanks for restoring this

PR #3017, opened 2024-02-06
fix(client): handle jsonrpc improper redirects

### LINKED ISSUE #2966: Trying to delete a session sometimes crashes the UI
## Description

When clicking the dropdown menu on the session button and selecting "Delete session", the UI may crash. This seems to happen when the `/jsonrpc` session endpoint returns HTTP 500.

## Reproduce

Not sure when this happens, as it is dependent on `/jsonrpc` failing.

leafty:
Fixes #2966.

How to reproduce the original issue:
1. Start a session
2. Once the session is starting immediately go back
   Screenshot 2024-02-06 at 14 42 59
3. On the sessions view, try to delete the session

/deploy


ciyer (approved):
I was not able to provoke the original bug that this fixes, so I cannot verify that the problem existed before and has been fixed by these changes, but the code changes make sense.

PR #3019, opened 2024-02-07
fix(client): temporarily fix search bar styles

leafty:
Component styles which leaked global classes where used in other places; now that pages are loaded lazily, this breaks styling for components which used to rely on "hidden" global styles.

Before:
Screenshot 2024-02-07 at 10 24 46
Screenshot 2024-02-07 at 10 25 02

After:
Screenshot 2024-02-07 at 10 27 38
Screenshot 2024-02-07 at 10 29 20


ciyer (approved):
Good catch!

PR #3025, opened 2024-02-08
build: release 3.20.1

leafty:
Release 3.20.1

/deploy renku=release-0.48.x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changes
 
-## [3.20.1](https://github.com/SwissDataScienceCenter/renku-ui/compare/3.20.0...3.20.1) (2024-02-8)
+## [3.20.1](https://github.com/SwissDataScienceCenter/renku-ui/compare/3.20.0...3.20.1) (2024-02-08)
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,11 @@
 
 ### Bug Fixes
 
-- update UI routing https://github.com/SwissDataScienceCenter/renku-ui/pull/3009 by @leafty
-- fix landing page parallax background https://github.com/SwissDataScienceCenter/renku-ui/pull/3010 by @leafty
-- temporarily fix search bar styles https://github.com/SwissDataScienceCenter/renku-ui/pull/3019 by @leafty
-- update the font part of the style guide https://github.com/SwissDataScienceCenter/renku-ui/pull/3018 by @leafty
-- handle jsonrpc improper redirects https://github.com/SwissDataScienceCenter/renku-ui/pull/3017 by @leafty
+- update UI routing ([#3009](https://github.com/SwissDataScienceCenter/renku-ui/pull/3009), [#2605](https://github.com/SwissDataScienceCenter/renku-ui/issues/2605)) by @leafty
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,23 +1,31 @@
 # Changes
 
+## [3.20.1](https://github.com/SwissDataScienceCenter/renku-ui/compare/3.20.0...3.20.1) (2024-02-08)
+
+### Bug Fixes
+
+- update UI routing ([#3009](https://github.com/SwissDataScienceCenter/renku-ui/pull/3009), [#2605](https://github.com/SwissDataScienceCenter/renku-ui/issues/2605)) by @leafty
```
ciyer, line 7:
```suggestion
- refactor UI routing ([#3009](https://github.com/SwissDataScienceCenter/renku-ui/pull/3009), [#2605](https://github.com/SwissDataScienceCenter/renku-ui/issues/2605)) by @leafty
```

ciyer (changes requested):
Thanks for compiling this release. I have a suggestion about the wording of one of the entries to make it clearer that no behavior was changed in the UI as a result.

**note:** The markdown spec states that asterisks, pluses, and hyphens can be used interchangeably to generate unordered lists (https://daringfireball.net/projects/markdown/syntax#list); I think your editor has been configured to only use hyphens and convert all other ways an unordered list has been specified to use hyphens. As a result, the diffs for this commit are confusing, since it looks like the changelog from earlier releases has been modified. I think it would be preferable to either only add the actually new entries to the commit, or, split the PR into two commits, one with the conversion from `*` to `-` and another with the new content.

leafty:
> Thanks for compiling this release. I have a suggestion about the wording of one of the entries to make it clearer that no behavior was changed in the UI as a result.
> 
> **note:** The markdown spec states that asterisks, pluses, and hyphens can be used interchangeably to generate unordered lists (https://daringfireball.net/projects/markdown/syntax#list); I think your editor has been configured to only use hyphens and convert all other ways an unordered list has been specified to use hyphens. As a result, the diffs for this commit are confusing, since it looks like the changelog from earlier releases has been modified. I think it would be preferable to either only add the actually new entries to the commit, or, split the PR into two commits, one with the conversion from `*` to `-` and another with the new content.

I have "format on save" enabled from vscode, so I have `prettier` prettify all the things.

CHANGELOG.md
```diff
@@ -1,23 +1,31 @@
 # Changes
 
+## [3.20.1](https://github.com/SwissDataScienceCenter/renku-ui/compare/3.20.0...3.20.1) (2024-02-08)
+
+### Bug Fixes
+
+- refactor UI routing ([#3009](https://github.com/SwissDataScienceCenter/renku-ui/pull/3009), [#2605](https://github.com/SwissDataScienceCenter/renku-ui/issues/2605))
+- fix landing page parallax background ([#3010](https://github.com/SwissDataScienceCenter/renku-ui/pull/3010))
+- temporarily fix search bar styles ([#3019](https://github.com/SwissDataScienceCenter/renku-ui/pull/3019))
+- update the font part of the style guide ([#3018](https://github.com/SwissDataScienceCenter/renku-ui/pull/3018))
```
lorenzo-cavazzi, line 10: Not sure if we should even mention this

lorenzo-cavazzi (approved):
:+1:

ciyer (approved):

PR #3026, opened 2024-02-08
refactor: change the url structure for renku-native projects

ciyer:
Change the URL structure for renku-native projects from `projectsV2` to `native/projects` to allow introducing other native (v1) entities under the `native` path.

/deploy

### DIFF
--- a/client/src/App.jsx
+++ b/client/src/App.jsx
@@ -199,17 +199,17 @@ function CentralContentContainer(props) {
               <LazyNotificationsPage />
             </ContainerWrap>
           </Route>{" "}
-          <Route path={Url.get(Url.pages.projectsV2.new)}>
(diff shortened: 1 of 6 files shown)


leafty (commented):
Has the word "native" been approved through Product? It can have negative connotation, related to racism and xenophobia so I would rather we check if we instead should use "next" or "v2".

ciyer:
> Has the word "native" been approved through Product? It can have negative connotation, related to racism and xenophobia so I would rather we check if we instead should use "next" or "v2".

I can imagine that a bigger discussion is needed around naming, and I just want to structure things a bit more flexibly so I can add the group/namespace support in a place that keeps it near the new projects. I propose using `v2` for now and we can rename later. Is this ok with you?

leafty (approved):
LGTM :raised_hands:

PR #3030, opened 2024-02-13
feat(client): add an additional top bar for Renku 1.0

leafty:
Adds an additional top bar when navigating paths under `/v2`.

localhost_3000_v2

/deploy renku=release-0.48.x


lorenzo-cavazzi (dismissed):
Works great, thank you!

lorenzo-cavazzi (approved):

PR #3032, opened 2024-02-13
[ShapeUp] Search and Discovery 1.0

lorenzo-cavazzi:
Feature branch for Search and Discovery 1.0
Pitch: https://www.notion.so/Renku-1-0-Project-and-User-Search-Service-e3df82dab0734f91bb8dc128f5e3aa54
Build page: https://www.notion.so/Renku-1-0-Search-Discovery-Services-Build-Team-cc4782794e084adca8fe9c36571d7b99

/deploy renku=build/search-and-discovery-1.0 renku-data-services=main renku-gateway=master


andre-code:
Items for review:
- Searching by Query Type, Visibility, and Creation Date for authenticated users
search


- Searching by Query Type and Creation Date for non-authenticated users
- Sorting by Score, Creation Date, and Name
- Filtering by user ID
search by user


On hold to review:
- Search by Role, awaiting implementation in the search service

client/run-telepresence.sh
leafty, line 1: Minor comment: we should undo the changes in the telepresence scripts. See https://github.com/SwissDataScienceCenter/renku-ui/pull/3047.

client/src/features/rootV2/NavbarV2.tsx
```diff
@@ -30,10 +30,15 @@ export default function NavbarV2() {
           <span className="me-1">Renku 1.0</span>
           <WipBadge />
         </div>
-        <Nav className="navbar-nav">
+        <Nav className={cx("navbar-nav", "flex-row", "gap-4")}>
+          <NavItem>
+            <RenkuNavLinkV2 end to="search" title="Search">
+              Search v2
```
leafty, line 36:
```suggestion
              Search
```

client/src/features/rootV2/NavbarV2.tsx
```diff
@@ -30,10 +30,15 @@ export default function NavbarV2() {
...
+            </RenkuNavLinkV2>
+          </NavItem>
           <NavItem>
             <RenkuNavLinkV2 end to="projects" title="Projects">
-              Projects
+              Projects v2
```
leafty, line 41:
```suggestion
              Projects
```

client/src/features/searchV2/components/SearchV2Results.tsx
```diff
@@ -0,0 +1,193 @@
...
+          }}
+          key={entity.id}
+          project={entity}
+        />
+      );
+    } else if (entity.type === "User") {
+      return <SearchV2ResultUser key={entity.id} user={entity} />;
+    }
+    // Unknown entity type, in case backend introduces new types before the UI catches up
+    const fakeId = simpleHash(Math.random().toString());
```
ciyer, line 110: I'm not sure this is a good idea. I think it would be better to use the index in the collection as the key (since this should not collide with the entity ids, which have a different structure).

ciyer (changes requested):
Overall looks very good! 🚀 I have two minor change requests which should be easy to implement.

client/src/utils/helpers/url/Url.js
```diff
@@ -492,6 +492,12 @@ const Url = {
         "/v2/projects/id",
       ]),
     },
+    v2Users: {
```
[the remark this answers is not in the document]
ciyer: As discussed in the stand up, let us set this change aside, since projects is also done this way and decide what to do in a later PR.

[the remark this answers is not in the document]
leafty: Anyway, I think we should remove the `Url.js` file at some point. See https://github.com/SwissDataScienceCenter/renku-ui/issues/3052.

client/src/features/searchV2/components/SearchV2Results.tsx
andre-code: sure, I have modified the component key to use the index instead

ciyer (approved):
LGTM! 🚀

ciyer (approved):
Tested the new changes, and everything works for me! 🚀

PR #3033, opened 2024-02-13
[ShapeUp] Search and Discovery - API and frontend

lorenzo-cavazzi:
This first part includes a basic fronted for the new search API that supports manually overriding filters by typing things like `type:project` in the search bar.

https://renku-ci-ui-3033.dev.renku.ch/v2/search

image

It also includes support for the backend but the integration is still not ready.

For the review, you can add a fake response in the `getSearchResults` endpoint, transforming the response to something like the following:

```
[
   {
      "type":"Project",
      "id":"0caf4c73-f50d-4514-a041-c9eddb025a35",
      "name":"first-project",
      "slug":"first-Xs9GL",
      "repositories":[
         "https://gh/first-project"
      ],
      "visibility":"public",
      "description":"first project desc",
      "createdBy":{
         "id":"ee53f3b7-f808-4a81-84f4-8290e874b9a7"
      },
      "creationDate":"2024-02-16T10:14:28.426Z",
      "members":[
         {
            "id":"ee53f3b7-f808-4a81-84f4-8290e874b9a7"
         }
      ]
   }
]
```

/deploy #notest renku=build/new-sessions renku-data-services=main extra-values=initDb.image.repository=registry.dev.renku.ch/tasko.olevski/renku-images/init-db,initDb.image.tag=0.44.1-1.ebb0e80,keycloakx.initRealm.image.repository=registry.dev.renku.ch/tasko.olevski/renku-images/init-realm,keycloakx.initRealm.image.tag=0.44.1-1.ebb0e80.git.2200.h26267d0e


client/src/features/searchV2/components/SearchV2Filters.tsx
```diff
@@ -0,0 +1,148 @@
...
+          <div
+            className={cx("form-rk-green", "d-flex", "align-items-center")}
+            key={id}
+          >
+            <input
+              checked={checked}
+              className="form-check-input"
+              data-cy={`user-role-${key}`}
+              id={id}
+              onChange={() => toggleOption(key)}
```
andre-code, line 134: It would be clearer if the user could see results when the filters change. Currently, the user needs to click the Search button to refresh the results according to the filters.

andre-code, line 134: It would be clearer if the user could see results when the filters change. Currently, the user needs to click the Search button to refresh the results according to the filters.

client/src/features/searchV2/searchV2.utils.ts
```diff
@@ -0,0 +1,93 @@
...
+    friendlyName: "Title: reverse",
+    sortingString: "title-desc",
+  },
+};
+
+export const SORT_KEY = "sort";
+
+export const FILTER_ASSIGNMENT_CHAR = ":";
+
+export const buildSearchQuery = (searchState: SearchV2State): string => {
```
andre-code, line 64: Could you please add a unit test for this function?

andre-code (commented):
I very much like the simplicity of the form, and the code was very easy to read. I have the following doubts and suggestions:

- It is distracting to see the full string query in the last query list, since the format is unfriendly and also doesn't sync with the form status when selected. I believe it needs to be rethought to be easy to understand. At the moment, I would exclude the full format from the list and only include the text the user typed.

- Question: If the user is searching for a user, how would the Role filter affect the search?

- Question: If the user is accessing as an anonymous user, would it make sense to leave the private filter?

- Could you please include ui e2e testing to facilitate testing and also include examples of how it looks for a user-type result?

client/src/features/searchV2/components/SearchV2Filters.tsx
lorenzo-cavazzi: Sure! Tests added in https://github.com/SwissDataScienceCenter/renku-ui/pull/3033/commits/43a7f1e372ffcbddad77bf2e5b36bd66382cbc39

lorenzo-cavazzi:
Thank you for the review @andre-code .

> It is distracting to see the full string query in the last query list [...] At the moment, I would exclude the full format from the list and only include the text the user typed.

I agree, fixed in https://github.com/SwissDataScienceCenter/renku-ui/pull/3033/commits/38c92750d4925a066e315e3f5e90bd7569904b5d

> Question: If the user is searching for a user, how would the Role filter affect the search?

Not all filters can be applied to all the entities (actually, probably no filter will work across all entities).
You won't get any results if you filter by user type _and_ by a specific role.
That scenario is currently difficult to test and requires the backend to be up and working (adding support for users will be the last step).
Still, I added basic logic for supporting users in https://github.com/SwissDataScienceCenter/renku-ui/pull/3033/commits/f219157ce0bcf8b40799df2d682473dd32bec0a6 and https://github.com/SwissDataScienceCenter/renku-ui/pull/3033/commits/dc64192b788cf7a43fe156462b188735dc82d297 so that the UI behaves correctly when getting users.

> Question: If the user is accessing as an anonymous user, would it make sense to leave the private filter?

Oh yes, good point! Fixed in https://github.com/SwissDataScienceCenter/renku-ui/pull/3033/commits/1c5e4a57578bec22f2c02c85a9100cfc34eea915

> Could you please include ui e2e testing to facilitate testing and also include examples of how it looks for a user-type result?

Fair point -- I planned to add e2e in a separate PR but it's hard to test without backend _nor_ tests  :facepalm: 
Added in https://github.com/SwissDataScienceCenter/renku-ui/pull/3033/commits/3ba1ee5b6aec9775f9ed483f089b97673b1eddfa

client/src/features/searchV2/searchV2.utils.ts
lorenzo-cavazzi: Sure! Tests added in https://github.com/SwissDataScienceCenter/renku-ui/pull/3033/commits/43a7f1e372ffcbddad77bf2e5b36bd66382cbc39

client/src/features/searchV2/components/SearchV2Filters.tsx
lorenzo-cavazzi: I was actually going down that way, but then I changed my mind and revered the "search on click" behavior.
Perhaps it's more intuitive to trigger the search as you suggest; if you don't mind, I'll quickly sync with the build team (to be sure we are all on the same page) and then I'll make that change in another PR

client/src/features/searchV2/components/SearchV2Results.tsx
```diff
@@ -66,34 +67,64 @@ function SearchV2ResultsContent() {
...
+
+interface SearchV2ResultsCardProps {
+  cardId: string;
+  children: React.ReactNode;
+  url?: string;
+}
+function SearchV2ResultsCard({
+  cardId,
+  children,
+  url = "#",
```
andre-code, line 119: This component should always expect a valid url, I suggest to remove the default value and include it in the mocked data.

client/src/features/searchV2/components/SearchV2Results.tsx
```diff
@@ -66,34 +67,64 @@ function SearchV2ResultsContent() {
...
+          <p className={cx("form-text", "mb-0")}>
+            {entity.slug} - {entity.visibility}
+          </p>
+          <p className={cx("form-text", "text-rk-green")}>
+            user-{entity.createdBy.id}
+          </p>
+          <p>{entity.description}</p>
+          <p className="form-text mb-0">
+            <TimeCaption datetime={entity.creationDate} prefix="Created" />
+          </p>
```
andre-code, line 84: Could you please add this in a separate component

client/src/features/searchV2/components/SearchV2Results.tsx
```diff
@@ -66,34 +67,64 @@ function SearchV2ResultsContent() {
...
+        </SearchV2ResultsCard>
+      );
+    } else if (entity.type === "User") {
+      const url = Url.get(Url.pages.v2Users.show, { id: entity.id });
+      return (
+        <SearchV2ResultsCard key={entity.id} url={url} cardId={entity.id}>
+          <h4 className="mb-0">{entity.id}</h4>
+          <p className="form-text mb-0">
+            <TimeCaption datetime={entity.creationDate} prefix="Created" />
+          </p>
```
andre-code, line 94: same here

client/src/features/searchV2/components/SearchV2Results.tsx
```diff
@@ -66,34 +67,64 @@ function SearchV2ResultsContent() {
...
-              <p className="form-text mb-0">
-                <TimeCaption datetime={entity.creationDate} prefix="Created" />
-              </p>
-            </CardBody>
-          </Card>
-        </Link>
-      </Col>
+      <SearchV2ResultsCard key={fakeId} cardId={fakeId}>
+        <h4 className="mb-0">Unknown entity</h4>
+        <p className="form-text mb-0">This entity type is not supported yet.</p>
```
andre-code, line 103: And here also add this in a separate component. It is nice to have a fallback component

andre-code (commented):
Thank you, @lorenzo-cavazzi , for including the e2e tests. I have only a few suggestions:

lorenzo-cavazzi:
@andre-code changes done in the last commit.

P.S. If you wish to give it a try with a working deployment, you can run this on https://ci-renku-3478.dev.renku.ch/v2/search using telepresence

andre-code (approved):
🚀

PR #3047, opened 2024-02-20
chore: accept a namespace when running the telepresence script

lorenzo-cavazzi:
Tiny little change to accept passing a `NAMESPACE` variable to the run-telepresence scripts.
This is helpful when messing around with non-UI PRs, for which we already have the convenient `PR` variable.


lorenzo-cavazzi:
@ciyer now I understand your comment, there is an issue in the script -- I accidentally messed up the logic in the final cleanup and the script was ignoring `NAMESPACE` when passed from the console :facepalm: .
This should fix it https://github.com/SwissDataScienceCenter/renku-ui/pull/3102

Issue #3049, opened 2024-02-22
Update the UI to enable starting a session


leafty:
Part of https://github.com/SwissDataScienceCenter/renku/issues/3475.

Note: Example payload is as follows.
```json
{
  "branch":"main",
  "commit_sha":"1cd64c01db40b0eb2c9c90fef8b493fa97f799e0",
  "default_url":"/lab",
  "image": "...",
  "environment_variables":{},
  "lfs_auto_fetch":false,
  "namespace":"flora.thiebaut",
  "project":"a-playground-project",
  "resource_class_id":4,
  "storage":8
}
```

leafty:
When there is one repository in the project, we can try to launch a session like so:

1. Identify if the repository is hosted on GitLab and if yes try to load its metadata.
2. For `branch`, pick the default branch of the repository
3. For `commit`, pick the head of the repository
4. For `default_url`, pick `/lab` (until we can get this from the launcher?)
5. For `image`, resolve the image from the launcher
6. For `environment_variables`, use `{}` (empty object)
7. Set `lfs_auto_fetch` to `false`
8. For `namespace` and `project`, use the repository namespace and project from GitLab
9. Pick the default resource class and storage

PR #3053, opened 2024-02-28
feat(client): initial support for Renku 2.0 sessions

leafty:
Fixes https://github.com/SwissDataScienceCenter/renku-ui/issues/3049.

Add initial support for Renku 2.0 sessions.

* Can start a session from a session launcher
* Can open a running session
* Can perform session actions (pause, delete, modify)
* Renku 1.0 and Renku 2.0 sessions are not co-mingled: Renku 1.0 sessions continue to appear in the dashboard, etc. while Renku 2.0 sessions will appear only in `/v2` pages.

Limitations:
* Sessions can work with projects with zero repositories.
* Sessions for projects with at least one repository will only work if the first repository is in GitLab.

Screenshot 2024-03-07 at 13 29 57

/deploy #notest renku=build/new-sessions renku-data-services=leafty/134-default-url renku-notebooks=leafty/have-sessions-that-work


client/src/features/sessionsV2/SessionsV2.tsx
```diff
@@ -81,10 +95,42 @@ function SessionLaunchersListDisplay() {
...
+                annotations["projectId"] === projectId &&
+                launchers.find(({ id }) => id === annotations["launcherId"]) ==
+                  null
+              );
+            })
+            .reduce(
+              (prev, [name, session]) => ({ ...prev, [name]: session }),
+              {} as Sessions
+            )
+        : null,
```
ciyer, line 130: **suggestion:** I think this would read clearer if the order of the branches were swapped, and if the variable were always initialized to an object, which would simplify the logic below.

Also, there the filter/reduce logic is quite similar to code below. This could be extracted to shared function:

```
function filteredSessionMap(sessions, predicate) {
  Object.entries(sessions)
     .filter(([, session]) => {
     const annotations = NotebooksHelper.cleanAnnotations(
        session.annotations
      ) as NotebookAnnotations;
              return (
                annotations["renkuVersion"] === "2.0" &&
                annotations["projectId"] === projectId &&
                predicate(annotations)
              );
            })
            .reduce(
              (prev, [name, session]) => ({ ...prev, [name]: session }),
              {} as Sessions
}
```

(Not completely sure the syntax is correct below, but I hope the idea is clear.)
```suggestion
      launchers == null || sessions == null
        ? {}
        : filteredSessionMap(sessions, (annotations) => 
                launchers.find(({ id }) => id === annotations["launcherId"]) ==
                  null
              );
            })
```

client/src/features/sessionsV2/SessionsV2.tsx
```diff
@@ -98,7 +144,12 @@ function SessionLaunchersListDisplay() {
     return <RtkErrorAlert error={error} />;
   }
 
-  if (!launchers || launchers.length == 0) {
+  if (
+    !launchers ||
+    (launchers.length == 0 &&
+      orphanSessions != null &&
```
ciyer, line 150: **suggestion:** If `orphanSessions` is initialized to an object, the null check can be omitted.

client/src/features/sessionsV2/SessionsV2.tsx
```diff
@@ -112,6 +163,10 @@ function SessionLaunchersListDisplay() {
             projectId={projectId ?? ""}
           />
         ))}
+        {orphanSessions &&
```
ciyer, line 166: See comment above about initializing the variable to an empty object instead of null.

client/src/features/sessionsV2/SessionsV2.tsx
```diff
@@ -139,6 +195,34 @@ function SessionLauncherDisplay({
...
+                annotations["renkuVersion"] === "2.0" &&
+                annotations["projectId"] === projectId &&
+                annotations["launcherId"] === launcher.id
+              );
+            })
+            .reduce(
+              (prev, [name, session]) => ({ ...prev, [name]: session }),
+              {} as Sessions
+            )
+        : null,
```
ciyer, line 217: **suggestion:** Following the comments above.

```suggestion
      sessions == null
        ? {} 
        : filteredSessionMap(sessions, (annotations) =>
                annotations["launcherId"] === launcher.id
              );
```

client/src/features/sessionsV2/SessionsV2.tsx
```diff
@@ -139,6 +195,34 @@ function SessionLauncherDisplay({
...
+    [launcher.id, projectId, sessions]
+  );
+  const filteredSessionsLength = useMemo(
+    () =>
+      filteredSessions != null ? Object.keys(filteredSessions).length : null,
```
ciyer, line 222: This can be simplified if `filteredSessions` is initialized to an empty object instead of `null`.

client/src/features/sessionsV2/SessionsV2.tsx
```diff
@@ -174,29 +258,106 @@ function SessionLauncherDisplay({
...
           </CardText>
-          <div className="mt-auto">
-            <StartSessionButton
-              launcherId={launcher.id}
-              projectId={projectId}
-            />
-          </div>
+          {filteredSessions &&
+          filteredSessionsLength != null &&
+          filteredSessionsLength > 0 ? (
```
ciyer, line 284: With the above changes, this can be simplified to `filteredSessionsLength > 0`

client/src/features/sessionsV2/ShowSessionPage.tsx
```diff
@@ -0,0 +1,346 @@
...
+
+  const { height } = useWindowSize();
+  const iframeHeight = height ? height - 42 : 800;
+
+  useEffect(() => {
+    // Wait 4 seconds before setting `isTheSessionReady` for session view
+    if (thisSession?.status.state === "running") {
+      const timeout = window.setTimeout(() => {
+        setIsTheSessionReady(true);
+      }, 1_000);
```
ciyer, line 118: **question:** This value and the comment do not seem to be in sync.

ciyer (changes requested):
**praise:** This is very impressive! It is especially cool to be able to start a session on a project that has 0 or more than 1 repo!

I have a few suggestions for code simplification. Also one **nitpick(non-blocking:** The status about the session in card seems to be shown twice in some situations:

image

Is this because of sharing code between buttons and badges for the old and new sessions? If so, then I think it is better to live with it for the time being and fix it later, but if it is an easy fix, then maybe look at it now.

client/src/features/sessionsV2/SessionsV2.tsx
leafty: Done in b991ed84be186609593531b96fd9444369d9701b.

client/src/features/sessionsV2/SessionsV2.tsx
leafty: Done in b991ed84be186609593531b96fd9444369d9701b.

leafty:
> I have a few suggestions for code simplification. Also one **nitpick(non-blocking:** The status about the session in card seems to be shown twice in some situations:
>
> Is this because of sharing code between buttons and badges for the old and new sessions? If so, then I think it is better to live with it for the time being and fix it later, but if it is an easy fix, then maybe look at it now.

This is not a duplicate status: the first component is the session status while the second is the session actions button. When the session is being touched, the actions are unavailable until the session status is resolved.

How should de-duplicating work? Removing any of these two components temporarily would make the layout shift and be annoying to users.

PR #3056, opened 2024-03-01
feat: support pagination and refresh query automatically

lorenzo-cavazzi:
This PR adds the following:

- Handles pagination (feel free to raise the default `perPage` value, currently at `10`, before merging).
- Automatically refresh the search result when changes on the UI require it (e.g. sorting, filtering, page number, ...)
- Search by Creation Date

/deploy renku=build/search-and-discovery-1.0 #notest renku-data-services=main renku-gateway=master

### DIFF
--- a/client/src/features/searchV2/components/SearchV2Results.tsx
+++ b/client/src/features/searchV2/components/SearchV2Results.tsx
@@ -157,10 +157,11 @@ function SearchV2ResultUser({ user }: SearchV2ResultUserProps) {
   const url = Url.get(Url.pages.v2Users.show, { id: user.id });
   return (
     <SearchV2ResultsCard key={user.id} url={url} cardId={user.id}>
-      <h4 className="mb-0">{user.id}</h4>
-      <p className="form-text mb-0">
-        <TimeCaption datetime={user.creationDate} prefix="Created" />
-      </p>
+      <p className="form-text mb-0">{user.id}</p>
+      <h4 className="mb-0">
+        {user.firstName} {user.lastName}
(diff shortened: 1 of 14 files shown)


ciyer (commented):
Code changes look good, but I cannot test, since the deployment seems to have some problems.

ciyer (approved):
Works well, code looks good, and it is very fast! 🚀 

I noticed a small problem in testing, which is that if you do a search and then wait a while (whatever the cache clearance interval is), then the results disappear. I don't think this is a blocker right now, but it is something that should eventually be fixed.

client/src/features/searchV2/components/SearchV2Results.tsx
```diff
@@ -133,10 +157,11 @@ function SearchV2ResultUser({ user }: SearchV2ResultUserProps) {
   const url = Url.get(Url.pages.v2Users.show, { id: user.id });
   return (
     <SearchV2ResultsCard key={user.id} url={url} cardId={user.id}>
-      <h4 className="mb-0">{user.id}</h4>
-      <p className="form-text mb-0">
-        <TimeCaption datetime={user.creationDate} prefix="Created" />
-      </p>
+      <p className="form-text mb-0">{user.id}</p>
```
ciyer, line 160: This looks great! One suggestion: this should not be a link until we have a page for displaying users. It currently leads to a page that gives a 404.

ciyer (changes requested):

ciyer (approved):
Looks great! 🎉

PR #3066, opened 2024-03-11
chore(tests): update tsconfig.json

leafty:
Update `tsconfig.json` in `tests/` to allow the use of `tsc` (the TypeScript compiler) without emitting `.js` files.

/deploy


ciyer (approved):
Tested locally and it works there; the tests also pass in the server environment.

PR #3067, opened 2024-03-11
feat(client): add support for entity tags on Renku 2.0 projects

leafty:
Adds support for entity tags on Renku 2.0 projects.

* Send `If-Match` header with `ETag` value if such a value exists.

With this update, the UI supports the Renku 2.0 projects API both with ETags and without ETags.

/deploy renku=release-0.50.x


ciyer (approved):
This is a good approach to identifying concurrent updates so that the UI can inform the user and allow them to take an appropriate action.

PR #3069, opened 2024-03-12
feat(client): support Renku 2.0 sessions with multiple repositories

leafty:
Update the Renku 2.0 sessions UI:
* Check if each repository in a project is accessible, i.e. it is found in GitLab and can be accessed by the user
* Request mounting each supported repository when starting sessions

/deploy #notest renku=build/new-sessions renku-data-services=build/new-sessions-demo renku-notebooks=pitch/compute-sessions-final


ciyer (approved):
Looks good and works well!

PR #3073, opened 2024-03-15
feat(client): add "email us" to the session class option

leafty:
Add "Need more compute resources? Email us!" below the session class option in the new session page.

image

This line does not show by default and can be configured through `helm` chart values.

See https://github.com/SwissDataScienceCenter/renku/pull/3545 for the `helm` chart changes.

Clicking on the "Email us!" link should open an email client like so:
Screenshot 2024-03-19 at 14 45 09

Also, update the "Contact us" button on the landing page to use the same email template.

/deploy renku=leafty/add-email-compute-resources extra-values=ui.client.sessionClassEmailUs.enabled=true,ui.client.sessionClassEmailUs.email.to=hello@renku.io,global.gitlab.url=https://gitlab.dev.renku.ch,global.gitlab.registry.host=registry.dev.renku.ch


lorenzo-cavazzi (approved):
Lgtm :rocket:

PR #3074, opened 2024-03-15
chore(ci/cd): add an action to cleanup old deployments

### LINKED ISSUE #3068: Can we clean-up deployments?
There are more than 500 deployments that GitHub sees as active at the moment:

image

Can we handle cleaning this up?

leafty:
Fixes #3068.

Add a manual action which cleans up old deployment environments from GitHub.

After cleanup, the PR will show that the git branch was previously deployed, but is not active:
Screenshot 2024-03-15 at 15 37 49

Removing old environments will de-clutter the "Deployments" section from the GitHub page:
Screenshot 2024-03-15 at 15 39 34

### DIFF
--- a/.github/workflows/cleanup-old-deployments.yml
+++ b/.github/workflows/cleanup-old-deployments.yml
@@ -1,0 +1,17 @@
+name: Cleanup old deployments
+
+on:
+  workflow_dispatch:
+
+jobs:
+  cleanup-old-deployments:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/github-script@v7
+        with:
+          script: |
+            const script = require('./.github/workflows/cleanup-old-deployments.js')
+            script({ core, github })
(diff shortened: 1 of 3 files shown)


ciyer (approved):
👍 Looks good, and we can remove some accumulated junk with this.

.github/workflows/cleanup-old-deployments.yml
```diff
@@ -0,0 +1,17 @@
+name: Cleanup old deployments
+
+on:
+  workflow_dispatch:
+
+jobs:
+  cleanup-old-deployments:
```
ciyer, line 7: One small request -- could we make clear in the name that it only cleans up in GH and not in k8s?

```suggestion
  cleanup-old-deployments-from-gh
```

ciyer (changes requested):
One small request -- could we make clear in the name of the action that it does not change k8s?

.github/workflows/cleanup-old-deployments.yml
leafty: Done in [90fcbcb](https://github.com/SwissDataScienceCenter/renku-ui/pull/3074/commits/90fcbcb7867b54e1eed38e731eb5999475c7b169).

ciyer (approved):
🙏

PR #3080, opened 2024-03-20
fix(client): call /user endpoint when logged in

leafty:
/deploy #notest


ciyer (approved):
Code looks good, and I tested that it works as promised: the `api/data/user` endpoint is hit when the user is logged in, and not when the user is not.

PR #3093, opened 2024-04-04
build: release 3.22.0

ciyer:
Release support for v2 namespaces.

/deploy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changes
 
+## [3.22.0](https://github.com/SwissDataScienceCenter/renku-ui/compare/3.21.0...3.22.0) (2024-04-04)
+
+### Features
+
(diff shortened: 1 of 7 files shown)


lorenzo-cavazzi (approved):
Lgtm :rocket: 

I've noticed [the release draft](https://github.com/SwissDataScienceCenter/renku-ui/releases/tag/untagged-88529fceb281de18ce5c) has an additional entry compared to the changelogs here. It would be nice to keep them aligned; since it's about the tests, we can just drop that in the release draft.
If you prefer to keep it, let's add the bugfix here too and drop the `chore(tests):` prefix.

ciyer:
> I've noticed [the release draft](https://github.com/SwissDataScienceCenter/renku-ui/releases/tag/untagged-88529fceb281de18ce5c) has an additional entry compared to the changelogs here. It would be nice to keep them aligned; since it's about the tests, we can just drop that in the release draft.

Good point -- I have removed the reference to the commit that is only internally relevant.

PR #3094, opened 2024-04-05
feat: test styles for Renku 2.0

andre-code:
Draft styles for Renku 2.0 and current Renku

/deploy


client/src/features/rootV2/StylesV2.tsx
```diff
@@ -0,0 +1,4 @@
+import "../../styles/indexV2.scss";
+
+const StylesV2 = () => <></>;
```
lorenzo-cavazzi, line 3: Is this file just to import files? This shouldn't be necessary, you can import the style directly on the component where you need it

client/src/features/rootV2/LazyRootV2.tsx
```diff
@@ -21,10 +21,12 @@ import { Suspense, lazy } from "react";
...
 const RootV2 = lazy(() => import("./RootV2"));
+const StylesV2 = lazy(() => import("./StylesV2"));
 
 export default function LazyRootV2() {
   return (
     <Suspense fallback={<PageLoader />}>
+      {StylesV2 && <StylesV2 />}
       <RootV2 />
     </Suspense>
   );
```
lorenzo-cavazzi, line 32: What about just importing? Something like this.
```suggestion
export default function LazyRootV2() {
  useEffect(() => {
    import("../../styles/indexV2.scss");
  }, []);
  return (
    <Suspense fallback={<PageLoader />}>
      <RootV2 />
    </Suspense>
  );
```

client/src/utils/customHooks/ReloadOnV2Navigation.hook.ts
```diff
@@ -0,0 +1,23 @@
...
+  const location = useLocation();
+  const [prevPathname, setPrevPathname] = useState("");
+
+  useEffect(() => {
+    if (
+      prevPathname.startsWith("/v2") &&
+      !location.pathname.startsWith("/v2")
+    ) {
+      // Reload the page if the previous URL starts with /v2 and the current URL does not
+      window.location.reload();
```
lorenzo-cavazzi, line 14: It would be great to re-assign the SCSS variables instead; not sure what's the best method but it's a matter of re-executing the code in "styles/bootstrap/_custom_bootstrap_variables.scss". Re-importing index.scss won't work just because it's cached, but there might be a workaround.

P.S. the function would need to work slightly different cause the variables should be re-defined before rendering the components. You can see what I mean by swapping this with something like `import("../../styles/indexV1.scss");` where `indexV1.scss` is a copy of `index.scss` and trying to move around. The first page will _usually_ render with the previous colors, and all the others will have the correct colors again

lorenzo-cavazzi (changes requested):
See inline

client/src/styles/bootstrap/_custom_bootstrap_variables2.0.scss
```diff
@@ -0,0 +1,109 @@
...
+  "info": $info,
+  "warning": $warning,
+  "danger": $danger,
+  "light": $light,
+  "dark": $dark,
+);
+
+$theme-colors: map-merge($theme-colors, $rk-colors);
+
+$enable-rounded: false;
```
leafty, line 82: Note: I think we should set this to `true`. Having this set to `false` while we actually use rounded corners breaks many components (e.g. input groups where we have to manually set which corners are rounded).

andre-code: Agree, I changed this in the PR https://github.com/SwissDataScienceCenter/renku-ui/pull/3104

client/src/features/rootV2/LazyRootV2.tsx
andre-code: Thank you @lorenzo-cavazzi . The strategy for managing styles has changed, so this code is no longer needed.

leafty (approved):

PR #3095, opened 2024-04-08
feat: new renku navigation

### LINKED ISSUE #3097: V2 Navbar and dashboard
Work on the "Nav Bar" part of the Build B / Renku V2 Nav pitch https://www.notion.so/Renku-2-0-Navigation-consolidation-df113c5fb395483694fd818ac5bffa03

ciyer:
Changes to V2 Navigation
* Added content to `/v2/

Fix #3097 


client/src/features/dashboardV2/HelpV2.tsx
```diff
@@ -0,0 +1,227 @@
...
+        </NavItem>
+      )}
+    </Nav>
+  );
+}
+
+function HelpInfo() {
+  return (
+    <Row className="mb-3">
+      <Col xs={6}>
```
lorenzo-cavazzi, line 98: Is `xs={6}` a leftover from a copy/paste? It makes this section smaller than the others on the help page.

client/src/features/dashboardV2/HelpV2.tsx
```diff
@@ -0,0 +1,227 @@
...
+      {privacyPolicyConfigured && (
+        <NavItem>
+          <RenkuNavLinkV2 to="privacy">Privacy Policy</RenkuNavLinkV2>
+        </NavItem>
+      )}
+    </Nav>
+  );
+}
+
+function HelpInfo() {
```
lorenzo-cavazzi, line 95: Actually, this section seems redundant since it's the very same content as the Dashboard.
Personally, I would entirely get rid of it. If you prefer to keep it, I suggest making a component that can be re-used in the Dashboard as well

client/src/features/dashboardV2/DashboardV2Sessions.tsx
```diff
@@ -0,0 +1,186 @@
...
+    return <div>No running sessions.</div>;
+  }
+
+  return (
+    <div
+      data-cy="dashboard-session-list"
+      className="d-flex flex-column gap-3 mb-sm-2 mb-md-4"
+    >
+      {Object.entries(v2Sessions).map(([key, session]) => (
+        <DashboardSession key={key} session={session} />
```
lorenzo-cavazzi, line 182: Some components overflow, like the Container image, and seem to be aligned wrong
E.G. I would expect the Container image to be left aligned and take more space since the URL is very likely to be long; I would also expect the button to be aligned on the right
image

client/src/features/dashboardV2/Dashboard.module.scss
lorenzo-cavazzi, line 1: After reviewing the PR, I'm not sure writing a rather verbose CSS file based on `display: grid` leads to good results. Many components seem to be misaligned or overflow. This is especially true when the screen size isn't `xl`.
Wouldn't using the bootstrap grid system be a more straightforward choice here?

client/src/features/dashboardV2/DashboardV2.tsx
```diff
@@ -0,0 +1,349 @@
...
+      </div>
+    );
+  if (error) return <div>Cannot show groups.</div>;
+
+  if (data == null) return <div>No 2.0 groups.</div>;
+
+  return (
+    <div
+      data-cy="dashboard-group-list"
+      className="d-flex flex-column gap-3 mb-sm-2 mb-md-4"
```
lorenzo-cavazzi, line 224: Minor detail: it would be nice to use `cx` when the list is longer. It's already the case in many places, but somewhere is still a longer string.
Iirc we agreed to use cx when there is more than 1 class

tests/cypress/e2e/navV2.spec.ts
lorenzo-cavazzi, line 1: Can we split the dashboard tests into a separate file?

lorenzo-cavazzi (changes requested):
Great to see a new navbar and a new dashboard for Renku 2.0 :rocket: 
I added some inline comments.

After reviewing this, I have concerns about re-using the `rk-dashboard` style here and trying to fit everything into a semi-customized grid system. Renku 1.0 components had a lot of additional properties to handle that we don't have here (yet). Trying to re-use a similar style leads to poorly aligned elements and content overflow.
Perhaps this is a good opportunity to write simple components based on the default Bootstrap grid system that will be easy to expand and maintain.

client/src/features/dashboardV2/DashboardV2.tsx
```diff
@@ -0,0 +1,349 @@
...
+      </Col>
+      <Col xs={6}>
+        <WarnAlert timeout={0} dismissible={false}>
+          <p>
+            Do not do any important work in the Renku 2.0 alpha preview! The
+            alpha is for testing only. We do not guarantee saving and persisting
+            work in the alpha.
+          </p>
+        </WarnAlert>
+      </Col>
```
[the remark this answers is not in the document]
ciyer: Fixed in 5e60c396d924e674d076bfa2edd96a4007b63a2d

client/src/features/rootV2/NavbarV2.tsx
```diff
@@ -33,29 +167,83 @@ export default function NavbarV2() {
...
-          <NavItem className="ms-2">
-            <RenkuNavLinkV2 end to="projects" title="Projects">
-              Projects
-            </RenkuNavLinkV2>
-          </NavItem>
-        </Nav>
+        <NavbarToggler onClick={onToggle} className="border-0">
+          <List className="bi text-rk-white" />
+        </NavbarToggler>
+        <Collapse isOpen={isOpen} navbar>
```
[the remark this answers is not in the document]
ciyer: Fixed in 18aeb6825fcf9827fbfbdabb3dd9e7179b61d750

client/src/features/dashboardV2/DashboardV2Sessions.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  if (error) {
+    return <RtkErrorAlert error={error} />;
+  }
+
+  if (
+    !sessions ||
+    (Object.keys(sessions).length == 0 && Object.keys(v2Sessions).length == 0)
+  ) {
+    return <div>No running sessions.</div>;
```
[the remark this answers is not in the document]
ciyer: Fixed in b26ebf0a3c787b40cab84dd6010072286f93766d

client/src/features/dashboardV2/DashboardV2.tsx
```diff
@@ -0,0 +1,349 @@
...
+
+  if (data == null) return <div>No 2.0 projects.</div>;
+
+  return (
+    <div
+      data-cy="dashboard-project-list"
+      className="d-flex flex-column gap-3 mb-sm-2 mb-md-4"
+    >
+      {data.projects?.map((project) => (
+        <DashboardListElement
```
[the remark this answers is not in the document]
ciyer: Fixed in b26ebf0a3c787b40cab84dd6010072286f93766d

client/src/features/dashboardV2/HelpV2.tsx
ciyer: Fixed in f7a372f353963fa54bce6d9b77670346b5a75184

client/src/features/dashboardV2/DashboardV2Sessions.tsx
ciyer: Fixed in b26ebf0a3c787b40cab84dd6010072286f93766d

client/src/features/dashboardV2/DashboardV2.tsx
ciyer: Fixed in 471429a56d9440618773cbd5118a5eafcd5b52d0

tests/cypress/e2e/navV2.spec.ts
ciyer: Fixed in27dc51146341faac38424775e7805f3f6714b4ec

ciyer:
> After reviewing this, I have concerns about re-using the `rk-dashboard` style here and trying to fit everything into a semi-customized grid system. Renku 1.0 components had a lot of additional properties to handle that we don't have here (yet). Trying to re-use a similar style leads to poorly aligned elements and content overflow. Perhaps this is a good opportunity to write simple components based on the default Bootstrap grid system that will be easy to expand and maintain.

The product owner requested that the Renku 2.0 dashboard look similar to the Renku 1.0 dashboard, so I thought the expedient thing to do was to just re-use the code and styling from the V1 dashboard to the greatest extent possible, but I do see your point, and I agree that the tweaks to the sizing tweaks that were in the V1 dashboard CSS make maintenance and re-usability quite complicated.

In b26ebf0a3c787b40cab84dd6010072286f93766d, I have simplified the CSS, leveraging the standard Bootstrap font and spacing metrics and classes to achieve a layout that is similar in spirit to the V1 dashboard, but with much less complexity and without overriding Bootstrap style properties.

Please take a look at the changes, I think this could be a way forward here.

lorenzo-cavazzi (changes requested):
I like the approach with minimal CSS customization!
Just a couple of additional details. Otherwise, it looks good now :+1: 

* The slug is right-aligned and, when it is too long, is cut on the left side without any visual indication like the 3 dots for the title. Would it make sense to treat it as the title? Or perhaps moving it to a new line since the space would be much wider, as it happens on small screens already (none of my slugs actually fit in the screen unless I go full page)
image
* Some navbar components change size/margins based on the browser width (see gif); these are probably leftovers from the copy/paste since there is enough space all the time and they can always keep the same distances. P.S. Could you make the menu float on the right also on smaller screens as in the V1? On medium screens, it feels like the right menu icon expands an empty since all the content is on the left.
Peek 2024-04-26 10-21
* I noticed the link on the navbar leads to the projects list. Should it now direct the user to the new Dashboard -- `/v2` instead of `/v2/projects`?

client/src/features/dashboardV2/DashboardV2Sessions.tsx
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  return (
+    <div
+      data-cy="dashboard-session-list"
+      className={cx("d-flex", "flex-column", "gap-3", "mb-sm-2", "mb-md-4")}
```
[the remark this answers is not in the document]
ciyer: `gap` does not have an effect on this component because it is not using grid layout. I have removed all the extra margins for simplicity. That can wait until a proper design comes.

client/src/features/dashboardV2/DashboardV2Sessions.tsx
```diff
@@ -0,0 +1,194 @@
...
+          annotations={annotations}
+        />
+      </div>
+      <div
+        className={cx(
+          styles.entityDescription,
+          "d-none",
+          "d-md-block",
+          "cursor-pointer"
+        )}
```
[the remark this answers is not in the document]
ciyer: Yes, `gap-2` will have a similar effect to `m-2` in a component with grid layout.

I didn't invest too much effort in this kind of fine tuning because I expect that there will be a proper design for the dashboard soon, and these kinds of things will be handled there.

client/src/features/dashboardV2/DashboardV2Sessions.tsx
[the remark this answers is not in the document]
ciyer: image

ciyer:
> * The slug is right-aligned and, when it is too long, is cut on the left side without any visual indication like the 3 dots for the title. Would it make sense to treat it as the title? Or perhaps moving it to a new line since the space would be much wider, as it happens on small screens already (none of my slugs actually fit in the screen unless I go full page)

I think this is going to be hard to solve perfectly for all cases. There will always be situations where the space available is too little for the full content, and there will be cases where the start of the content is more important and cases where the end is more important.

For simplicity in this PR, I just put the title on its own line always.

image



> * Some navbar components change size/margins based on the browser width (see gif); these are probably leftovers from the copy/paste since there is enough space all the time and they can always keep the same distances. P.S. Could you make the menu float on the right also on smaller screens as in the V1? On medium screens, it feels like the right menu icon expands an empty since all the content is on the left.



I tried to add margins so that the spacing looked good at a variety of sizes, but I have removed that, since there is no point in making these kinds of optimizations until there is a proper design. There is still a breakpoint at which a small change in the spacing between two of the elements in the menu occurs. I'm not sure why this is, but I think that can be left for the build that implements the final menu.

I have moved the menu to the right side on small viewports.

image


> * I noticed the link on the navbar leads to the projects list. Should it now direct the user to the new Dashboard -- `/v2` instead of `/v2/projects`?

lorenzo-cavazzi (approved):
Lgtm :rocket: 

I have 2 additional nitpick comments, feel free to ignore them if these points were already discussed/decided among the build team:

* The "Renku 2.0" link on the user dropdown leads to the projects list instead of the new Dashboard in `/v2`. We might want to update it.
* The session status disappears on medium and small screens. That seems to be very important information to have, not sure about hiding it
image

ciyer:
> Lgtm 🚀
> 
> I have 2 additional nitpick comments, feel free to ignore them if these points were already discussed/decided among the build team:
> 
> * The "Renku 2.0" link on the user dropdown leads to the projects list instead of the new Dashboard in `/v2`. We might want to update it.

Oh, that link. Yes, this is a good point! I have made this change.

> * The session status disappears on medium and small screens. That seems to be very important information to have, not sure about hiding it

This was not discussed, but there were no precise instructions other than, "Show sessions on the new V2 dashboard as a 'nice-to-have'." My personal feeling is that we should not invest too much in the details of this and wait for a real design to be proposed.

lorenzo-cavazzi (approved):
Lgtm :rocket:

PR #3101, opened 2024-04-08
[ShapeUp] feat: add secrets

### LINKED ISSUE #3100: [ShapeUp] Add secrets to sessions
This [ShapeUp project](https://www.notion.so/Bringing-secrets-to-sessions-7b086846b5ab4d58bdc064342d39c528) aims to add secrets to user sessions. There are 2 changes in the UI:

* Add a page to view/manage user secrets.
* Allow users to inject secrets into sessions as environmental variables or files mounted through k8s secrets.

_Mind that this won't affect sessions 2.0_

lorenzo-cavazzi:
Secrets in the UI -- feature branch

/deploy renku=release-0.52.x

fix #3100

### DIFF
--- a/client/src/features/secrets/SecretEdit.tsx
+++ b/client/src/features/secrets/SecretEdit.tsx
@@ -138,6 +138,7 @@ export default function SecretEdit({ secret }: SecretsEditProps) {
                         errors.value && "is-invalid"
                       )}
                       id="edit-secret-value"
+                      spellCheck="false"
                       type="textarea"
                       {...field}
                     />
--- a/client/src/features/secrets/SecretEdit.tsx
+++ b/client/src/features/secrets/SecretEdit.tsx
@@ -1,0 +1,204 @@
+/*!
+ * Copyright 2024 - Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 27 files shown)


client/src/features/secrets/SecretNew.tsx
```diff
@@ -0,0 +1,243 @@
...
+                  autoComplete="off"
+                  className={cx(
+                    "form-control",
+                    "rounded-0",
+                    "rounded-start",
+                    !showPlainText && styles.blurredText,
+                    errors.value && "is-invalid"
+                  )}
+                  id="new-secret-value"
+                  type="textarea"
```
ciyer, line 163:
```suggestion
                  id="new-secret-value"
                  spellCheck="false"
                  type="textarea"
```

This field should probably not be spellchecked.

image

client/src/features/secrets/SecretEdit.tsx
```diff
@@ -0,0 +1,204 @@
...
+                      autoComplete="off"
+                      className={cx(
+                        "form-control",
+                        "rounded-0",
+                        "rounded-start",
+                        !showPlainText && styles.blurredText,
+                        errors.value && "is-invalid"
+                      )}
+                      id="edit-secret-value"
+                      type="textarea"
```
ciyer, line 141:
```suggestion
                      id="edit-secret-value"
                      spellCheck="false"
                      type="textarea"
```

ciyer (changes requested):
Looks very good and works well! I have a small suggestion to turn off spellcheck in the secrets input.

lorenzo-cavazzi:
@ciyer  I added your suggestions to the code, although not all browsers consider that option
image

ciyer (approved):
LGTM!

leafty:
> @ciyer I added your suggestions to the code, although not all browsers consider that option

Note: It looks like it's added from a chrome extension, not the base browser itself.

lorenzo-cavazzi:
> Note: It looks like it's added from a chrome extension, not the base browser itself.

@leafty That's actually the Firefox browser spellcheck.

On top of that, I also have a browser extension that would take priority and ofc ignore the setting as well :laughing: 

The additional icon on the right side of the field was misleading, it was temporarily off. It shows slightly differently when it's enabled on the page and it finds errors.

image

PR #3102, opened 2024-04-08
chore(client): fix telepresence script

lorenzo-cavazzi:
This fixes an issue in the client telepresence script accidentally introduced here https://github.com/SwissDataScienceCenter/renku-ui/pull/3047


ciyer (approved):
Yes, this is nicer!

PR #3105, opened 2024-04-16
Fix conditional style loading for v1/v2

leafty:


andre-code (approved):
Thanks @leafty I like this solution 🙌

PR #3106, opened 2024-04-16
[ShapeUp] feat: add new secret page

lorenzo-cavazzi:
This PR implements the first part of #3100 , adding a new page to list and manage user secrets.

image

The new page is accessible from the User dropdown menu on the top right.

image

re #3100

/deploy #notest renku=build/secrets-in-sessions renku-data-services=pitch/secret-storage secrets-storage=pitch/secret-storage extra-values=secretsStorage.encryptionKey=vMqdVXcSIwjwONvGiNE2vjzRjmi8NJ8Y9lTB8m3l/Xg=

### DIFF
--- a/client/src/features/secrets/SecretNew.tsx
+++ b/client/src/features/secrets/SecretNew.tsx
@@ -116,7 +116,7 @@ export default function SecretsNew() {
             name="name"
             render={({ field }) => (
               <Input
-                autoComplete="new-password"
+                autoComplete="off"
                 className={cx("form-control", errors.name && "is-invalid")}
                 id="new-secret-name"
                 placeholder="Unique name"
@@ -149,7 +149,7 @@ export default function SecretsNew() {
               name="value"
               render={({ field }) => (
                 <Input
-                  autoComplete="new-password"
+                  autoComplete="off"
                   className={cx(
                     "form-control",
                     "rounded-0",
--- a/client/src/features/secrets/SecretNew.tsx
+++ b/client/src/features/secrets/SecretNew.tsx
@@ -179,16 +179,6 @@ export default function SecretsNew() {
 
   return (
     <>
-      <div className="mb-2">
-        <Button
-          id="new-secret-button"
-          className="btn-outline-rk-green"
-          onClick={toggleModal}
-        >
-          <PlusLg className={cx("bi", "me-1")} />
-          Add New Secret
-        </Button>
-      </div>
       <Modal isOpen={showModal} toggle={toggleModal}>
         <ModalHeader toggle={toggleModal}>Add New Secret</ModalHeader>
         <ModalBody>{modalBody}</ModalBody>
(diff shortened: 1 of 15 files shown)


client/src/features/secrets/SecretNew.tsx
```diff
@@ -0,0 +1,241 @@
...
+        <div className="mb-3">
+          <Label className="form-label" for="new-secret-name">
+            Name
+          </Label>
+          <Controller
+            control={control}
+            name="name"
+            render={({ field }) => (
+              <Input
+                autoComplete="new-password"
```
ciyer, line 119: Not sure this is the best choice for this. Maybe `autocomplete="off"`?

client/src/features/secrets/SecretNew.tsx
```diff
@@ -0,0 +1,241 @@
...
+          <Label className="form-label" for="new-secret-value">
+            Value
+          </Label>
+          <InputGroup>
+            <Controller
+              control={control}
+              name="value"
+              render={({ field }) => (
+                <Input
+                  autoComplete="new-password"
```
ciyer, line 152: In this case, I think we should just turn off autocomplete.
```suggestion
                  autoComplete="off"
```

client/src/features/secrets/SecretsListItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+            "d-flex",
+            "gap-3",
+            "flex-wrap",
+            "align-items-center",
+            "w-100"
+          )}
+        >
+          <span className="fw-bold">{secret.name}</span>
+          <span className="text-rk-text-light my-auto small">
+            Edited {+new Date(secret.modification_date) - +new Date() < 5_000}
```
ciyer, line 49: I do not understand the purpose of this code. I think the expression inside the curly brackets will always evaluate to true (unless the time on the user's computer is messed up or the user has access to a time machine).

ciyer (changes requested):
This look very good! 🚀 I only have a few minor comments about the code.

client/src/features/secrets/SecretNew.tsx
lorenzo-cavazzi: Good point, I'll swap to `autocomplete="off"` since I no longer use a `password` input but a `textarea` and I don't need this workaround anymore.

A little background: before changing the `value` element to a `textarea`, both Firefox and Chrome were ignoring `autocomplete="off"` and suggesting the Renkulab login credentials. According to [this reference on Mdn](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#the_autocomplete_attribute_and_login_fields), I assumed browsers identified the inputs as name and password and override the autocomplete off. The article mentions using `new-password`, which proved effective.

client/src/features/secrets/SecretsListItem.tsx
lorenzo-cavazzi: Ha, that's a leftover, thanks for spotting it!

I had an ugly mechanism to prevent accidentally showing a date slightly in the future after editing a secret (due to how `TimeCaption` is implemented). Adding the modification date to a `key` in an ancestor component works better. I probably forgot to delete this line since it wasn't printing anything anymore

ciyer (approved):
LGTM 🚀

PR #3109, opened 2024-04-22
[ShapeUp] feat: update start session with options

lorenzo-cavazzi:
This PR implements the second part of https://github.com/SwissDataScienceCenter/renku-ui/issues/3100 ,changing the "Start with options" page to include secrets.

re https://github.com/SwissDataScienceCenter/renku-ui/issues/3100

/deploy #notest renku=build/secrets-in-sessions renku-data-services=pitch/secret-storage secrets-storage=pitch/secret-storage extra-values=secretsStorage.encryptionKey=vMqdVXcSIwjwONvGiNE2vjzRjmi8NJ8Y9lTB8m3l/Xg=

### DIFF
--- a/client/src/features/session/components/options/SessionUserSecrets.tsx
+++ b/client/src/features/session/components/options/SessionUserSecrets.tsx
@@ -17,12 +17,20 @@
  */
 
 import cx from "classnames";
+import { useCallback, useEffect, useState } from "react";
 import { Link } from "react-router-dom";
+import { Button, Card, CardBody, Collapse, Input, Label } from "reactstrap";
 
 import { Loader } from "../../../../components/Loader";
 import { User } from "../../../../model/renkuModels.types";
 import useLegacySelector from "../../../../utils/customHooks/useLegacySelector.hook";
 import { Url } from "../../../../utils/helpers/url";
+import ChevronFlippedIcon from "../../../../components/icons/ChevronFlippedIcon";
+import { useGetSecretsQuery } from "../../../secrets/secrets.api";
+import { RtkOrNotebooksError } from "../../../../components/errors/RtkErrorAlert";
+import { setSecretsList, setSecretsPath } from "../../startSessionOptionsSlice";
+import useAppDispatch from "../../../../utils/customHooks/useAppDispatch.hook";
+import useAppSelector from "../../../../utils/customHooks/useAppSelector.hook";
 
 export default function SessionUserSecrets() {
   const secretsUrl = Url.get(Url.pages.secrets);
@@ -36,8 +44,8 @@ export default function SessionUserSecrets() {
         You can select secrets defined in the{" "}
         <Link to={secretsUrl}> User Secrets page</Link> and mount them as files
         in your session.
-        {/* <MountUserSecrets /> */}
       </div>
+      <SessionUserSecretsSection />
     </>
   ) : (
     <div className={cx("form-text", "my-1")}>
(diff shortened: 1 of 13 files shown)


client/src/features/session/components/options/SessionUserSecrets.tsx
```diff
@@ -0,0 +1,239 @@
...
+                          className={cx("form-check-label", "my-auto")}
+                          for={`secrets-session-${secret.name}`}
+                        >
+                          {secret.name}
+                        </Label>
+                      </div>
+                    ))}
+              </div>
+
+              {/* <div className="d-flex gap-2 flex-wrap">
```
ciyer, line 214: Is this a leftover from debugging?

client/src/features/session/components/options/SessionUserSecrets.tsx
```diff
@@ -0,0 +1,239 @@
...
+                {[...sessionOptions.secretsList]
+                  .map((s) => s.name)
+                  .sort((a, b) => {
+                    return a.localeCompare(b); // localCompare is used to ignore case
+                  })
+                  .join(", ")}
+              </span>
+            ) : (
+              <span className="fw-bold">None</span>
+            )}
```
ciyer, line 142:
```suggestion
            <MountedSecrets secretsList={sessionOptions.secretsList} />
```

with 
```
function MountedSecrets({ secretsList }: { secretsList: SessionSecrets[] }) {
  if (secretsList.length < 1) {
    return <span className="fw-bold">None</span>;
  }
  return (
    <span className="fw-bold">
      {[...secretsList]
        .map((s) => s.name)
        .sort((a, b) => {
          return a.localeCompare(b); // localCompare is used to ignore case
        })
        .join(", ")}
    </span>
  );
}
```

client/src/features/session/startSessionOptionsSlice.ts
```diff
@@ -110,6 +113,12 @@ export const startSessionOptionsSlice = createSlice({
     setSessionClass: (state, action: PayloadAction<number>) => {
       state.sessionClass = action.payload;
     },
+    setSecretsList: (state, action: PayloadAction<SessionSecrets[]>) => {
+      state.secretsList = action.payload;
+    },
+    setSecretsPath: (state, action: PayloadAction<string>) => {
+      state.secretsPath = action.payload;
+    },
```
ciyer, line 121: I think it would be better to include a `toggleSecret` action as well. This would be something like:

```suggestion
    },
    toggleSecret(state, action: PayloadAction<SessionSecrets>) => {
          const secret = action.payload;
          const newSecretsList = state.secretsList.map((s) => s.name).includes(secret.name)) ?
            state.secretsList.filter((s) => s.name !== secret.name)
          ) :  [...state.secretsList, secret]));
       state.secretsList = newSecretsList
     },
```

client/src/features/session/components/options/SessionUserSecrets.tsx
```diff
@@ -0,0 +1,239 @@
...
+                          type="checkbox"
+                        />
+                        <Label
+                          className={cx("form-check-label", "my-auto")}
+                          for={`secrets-session-${secret.name}`}
+                        >
+                          {secret.name}
+                        </Label>
+                      </div>
+                    ))}
```
ciyer, line 211: I think this would be easier to read as a separate component

client/src/features/session/components/options/SessionUserSecrets.tsx
```diff
@@ -0,0 +1,239 @@
...
+    (secret: SessionSecrets) => {
+      if (sessionOptions.secretsList.map((s) => s.name).includes(secret.name)) {
+        dispatch(
+          setSecretsList(
+            sessionOptions.secretsList.filter((s) => s.name !== secret.name)
+          )
+        );
+      } else {
+        dispatch(setSecretsList([...sessionOptions.secretsList, secret]));
+      }
```
ciyer, line 94: See the  proposal for a `toggleSecret` action in the slice file
```suggestion
        dispatch(toggleSecret(secret));
        }
```

ciyer (changes requested):
Overall, this looks good! I do have some refactoring suggestions in places in the code where I thought the clarity could be improved.

client/src/features/session/components/options/SessionUserSecrets.tsx
lorenzo-cavazzi: Yes, thanks for spotting it!

client/src/features/session/components/options/SessionUserSecrets.tsx
lorenzo-cavazzi: :+1:

client/src/features/session/components/options/SessionUserSecrets.tsx
lorenzo-cavazzi: I separated this into a new component

client/src/features/session/startSessionOptionsSlice.ts
lorenzo-cavazzi: I'm not sure this is needed. It basically moves the logic of `setSecretsPath` to the slice, but it's used only in one place.
If you don't mind, I would prefer to keep the logic simple there (as with all the other reducers) and keep the toggle-like implementation in the container.

lorenzo-cavazzi:
Thank you for your suggestions @ciyer 
I implemented everything, except the `toggle` logic in the reducer -- see the motivations inline. I can still change it if you think your solution is more convenient.

ciyer (approved):
LGTM! 🚀

PR #3112, opened 2024-04-24
feat(client): add connected services

leafty:
WIP

/deploy #notest renku-data-services=leafty/build-add-connected-services


andre-code (approved):
The code looks good, and everything is working as it should!

PR #3113, opened 2024-04-26
feat: set custom culling times in admin page

Panaetius:
This allows setting custom idle and hibernation times for culling of sessions.

image

/deploy renku=release-0.52.x renku-data-services=main renku-notebooks=custom-culling-times


client/src/features/session/components/options/SessionClassOption.tsx
```diff
@@ -500,6 +519,19 @@ const selectComponents: SelectComponentsConfig<
...
     );
   },
+  GroupHeading: (
+    props: GroupHeadingProps<ResourceClass, false, OptionGroup>
+  ) => {
+    return (
+      <components.GroupHeading {...props}>
+        <span className="text-uppercase me-2">{props.data.label}</span>
+        <span>
+          (Max Idle: {props.data.maxIdle}, Max Stopped:{" "}
```
lorenzo-cavazzi, line 529: I'm not sure this will be clear to the final users, but I don't have much better wording in mind. We use "Pause" instead of "Idle" when referring to sessions.
Perhaps something like "Pause after {duration}, delete after {duration}"?

I would also simplify the time we show by just using the longest duration. E.G: if it's 5 days and 7 hours, it's probably fine to use `5 days`.

client/src/features/session/components/options/SessionClassOption.tsx
lorenzo-cavazzi, line 1: Great that we show feedback to users!
It might look a little cryptic tho, and I'm not really sure how to improve it.


The only idea I have is to add a comment like the following once users pick a specific class, perhaps with better working and a link to the documentation. What do you think?
image

client/src/utils/helpers/DurationUtils.ts
```diff
@@ -46,6 +51,37 @@ export function toHumanDuration({
   return `${rescaled} ${unitStr}`;
 }
 
+/**
+ * Converts a duration-like object to a human-readable string without truncating to the most significant value.
+ * @param duration a Duration instance or a number of seconds
+ * @param units a list of units expected in the output representation, defaults to ["years","weeks","days","hours","minutes","seconds"]
+ * @returns a human-readable string
+ */
+export function toFullHumanDuration(
```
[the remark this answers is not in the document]
ciyer: +1

ciyer (changes requested):
This looks good and works well. Just one minor suggestion to bring the mocked responses in line with the types defined.

client/src/utils/helpers/DurationUtils.ts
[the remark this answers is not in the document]
leafty: Why do we need this? It is inaccurate, especially with the default units listed here. Default units should be `days, hours, minutes, seconds`. Conversions to `weeks, months, years` introduce calendar artefacts and should be done only when there is no chance of confusion.

Converting to a unrestricted amount of units will also be something that can be hard to understand for humans, e.g "1 year 47 weeks 5 hours 25 seconds".

leafty (changes requested):
Relatively small detail, but the update buttons should use the check icon used by the other update buttons:

Screenshot 2024-05-13 at 11 49 25
Screenshot 2024-05-13 at 11 49 15

lorenzo-cavazzi:
Thank you for the reviews!

@ciyer JSON file updated. I also added a test for that duration function as discussed offline.
@leafty Thank you for mentioning it! I changed the icon

client/src/features/admin/UpdateResourcePoolThresholdsButton.tsx
```diff
@@ -0,0 +1,260 @@
...
+  const onSubmit = useCallback(
+    (data: UpdateResourcePoolThresholdsForm) => {
+      updateResourcePool({
+        resourcePoolId: id,
+        idle_threshold: data.idleThreshold
+          ? data.idleThreshold * 60
+          : undefined,
+        hibernation_threshold: data.hibernationThreshold
+          ? data.hibernationThreshold * 60
+          : undefined,
```
leafty, line 116: This feels strange: why do we have the UI talk in minutes when the underlying API is using seconds? The UI code should use seconds as well to avoid any chance of data loss or corruption. I would also suggest to keep the UI with seconds as the main control and have an indicator to show what approximate duration it corresponds to.

leafty (changes requested):
See comment about using minutes instead of seconds.

client/src/utils/helpers/DurationUtils.ts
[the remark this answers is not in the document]
leafty: I don't know why but my previous comment seems to not show up:

> Why do we need this? It is inaccurate, especially with the default units listed here. Default units should be days, hours, minutes, seconds. Conversions to weeks, months, years introduce calendar artefacts and should be done only when there is no chance of confusion.
> 
> Converting to a unrestricted amount of units will also be something that can be hard to understand for humans, e.g "1 year 47 weeks 5 hours 25 seconds".

leafty (commented):
also this

client/src/features/admin/UpdateResourcePoolThresholdsButton.tsx
lorenzo-cavazzi: The first implementation was actually using seconds, but it turned out the backend could not handle that granularity properly because resources scheduled for deletion within a few seconds required more time and were messing up some API responses (little inconsistencies, E.G. paused sessions showing messages like `Paused, will be deleted 15 seconds ago`).

We decided to switch to minutes in the UI since it was a quick solution. Updating the backend APIs would have been best, but I think we can live with this for now.

lorenzo-cavazzi:
@leafty the function `toFullHumanDuration` was implemented cause we needed a way to show durations that was both accurate enough (can't show just the most relevant unit) and compact enough to fit in the space (`5 days 3 hours 27 minutes` is much longer than `5d 3h 27m` and `5d, 3h, 27m` turned out to be confusing for some users).

I don't disagree about some of the longer units like `months` and `years` being a non-optimal choice cause they aren't precise -- while `weeks` are a well-defined duration but possibly less intuitive/immediate than days.
I don't think it's too critical here since it's unlikely admins will be interested in setting months-long durations, but I'll bring this up with the build team with the intention of fixing it.

client/src/features/admin/UpdateResourcePoolThresholdsButton.tsx
leafty: I don't think we can keep it "as is". Please use seconds in the code, which can be shown as minutes in the form controls. Otherwise update the UI code to explicitly mention that the variables are in minutes, e.g. `hibernationThreshold`.

leafty:
> @leafty the function `toFullHumanDuration` was implemented cause we needed a way to show durations that was both accurate enough (can't show just the most relevant unit) and compact enough to fit in the space (`5 days 3 hours 27 minutes` is much longer than `5d 3h 27m` and `5d, 3h, 27m` turned out to be confusing for some users).
> 
> I don't disagree about some of the longer units like `months` and `years` being a non-optimal choice cause they aren't precise -- while `weeks` are a well-defined duration but possibly less intuitive/immediate than days. I don't think it's too critical here since it's unlikely admins will be interested in setting months-long durations, but I'll bring this up with the build team with the intention of fixing it.

From your comment, I do not see why we can't keep it to `"days","hours","minutes","seconds"` as the units. The function should be updated to reflect this.

lorenzo-cavazzi:
> From your comment, I do not see why we can't keep it to `"days","hours","minutes","seconds"` as the units. The function should be updated to reflect this.

Sure, that's my intention! I just want to bring this up with the build team since there was a short discussion on this topic and I don't want to change things last minute without warning

client/src/features/admin/UpdateResourcePoolThresholdsButton.tsx
lorenzo-cavazzi: I guess you are right: I will probably be confused looking at the code 6 months from now.
I updated the variable naming following your suggestion df3f5a29210a168b5f2965413a1b7ea6b64b4f05

lorenzo-cavazzi:
@leafty Switched to using `days` as the default longest duration for `toFullHumanDuration` e7cf4ed705ecccaf5cb830817eafb2eb39ed45de

leafty (dismissed):

leafty (approved):

lorenzo-cavazzi (approved):

PR #3118, opened 2024-05-03
UI changes for auth

ciyer:
/deploy renku=release-0.52.x


lorenzo-cavazzi (changes requested):
I'm reviewing the code right now. As an early comment, I believe the "Keywords" section got accidentally duplicated while merging/rebasing the code.
I'll follow up soon with the full review

image

client/src/features/ProjectPageV2/ProjectPageContent/CodeRepositories/RepositoriesBox.tsx
```diff
@@ -38,7 +41,14 @@ export function CodeRepositoriesDisplay({ project }: { project: Project }) {
           <CodeSquare size={20} className={cx("me-2")} />
           Code Repositories ({project?.repositories?.length})
         </div>
-        <PlusRoundButton handler={toggle} />
+        <AccessGuard
+          disabled={null}
```
lorenzo-cavazzi, line 45: Should the `AccessGuard` component make this property optional instead of passing `null` here?

client/src/features/ProjectPageV2/ProjectPageContent/DataSources/DataSourcesBox.tsx
```diff
@@ -51,7 +54,14 @@ export function DataSourcesDisplay({ project }: { project: Project }) {
           <Database size={20} className={cx("me-2")} />
           Data Sources ({totalStorages})
         </div>
-        <PlusRoundButton handler={toggle} />
+        <AccessGuard
+          disabled={null}
```
lorenzo-cavazzi, line 58: Some as above

client/src/features/ProjectPageV2/ProjectPageContent/ProjectInformation/ProjectInformation.tsx
```diff
@@ -132,6 +165,39 @@ export default function ProjectInformation({ project }: { project: Project }) {
           </div>
         )}
       </div>
+      <div className={cx("border-bottom", "py-3", "text-start", "text-lg-end")}>
+        <div>Keywords ({totalKeywords})</div>
+        {totalKeywords == 0 ? (
+          <MembershipGuard
+            disabled={null}
```
lorenzo-cavazzi, line 172: Same here

client/src/features/ProjectPageV2/ProjectPageContent/Settings/ProjectSettings.tsx
```diff
@@ -175,24 +248,119 @@ export function ProjectSettingsForm({ project }: ProjectSettingsFormProps) {
...
+  );
+}
+
+function ProjectSettingsMetadata({ project }: ProjectPageSettingsProps) {
+  const { userRole } = useProjectAccess({ projectId: project.id });
+  return (
+    <>
+      <h4 className="fw-bold">General settings</h4>
+      <AccessGuard
+        disabled={null}
```
lorenzo-cavazzi, line 297: Same here

client/src/features/ProjectPageV2/ProjectPageContent/Settings/ProjectSettings.tsx
```diff
@@ -175,24 +248,119 @@ export function ProjectSettingsForm({ project }: ProjectSettingsFormProps) {
...
-        </div>
+        <ProjectSettingsMetadata project={project} />
       </div>
-      <div id="delete" className={cx("px-2", "px-md-5", "pt-4")}>
-        <ProjectPageDelete project={project} />
+      <div id="members" className={cx("px-2", "px-md-5", "pt-4")}>
+        <ProjectPageSettingsMembers project={project} />
       </div>
+      <AccessGuard
+        disabled={null}
```
lorenzo-cavazzi, line 356: Same here

lorenzo-cavazzi: Please also check the other occurences. I realize now there are some :grin:

tests/cypress/e2e/projectV2.spec.ts
```diff
@@ -229,29 +272,25 @@ describe("Edit v2 project", () => {
...
+    cy.contains("Connect an existing repository").click();
+    cy.getDataCy("field-group-url").type("https://domain.name/repo3.git");
+    // fixtures.readProjectV2({
+    //   fixture: "projectV2/update-projectV2-repositories.json",
+    //   name: "readPostUpdate",
+    // });
+    // cy.get("button").contains("Update").should("be.visible").click();
+    // cy.wait("@updateProjectV2");
+    // cy.wait("@readPostUpdate");
+    // cy.contains("https://domain.name/repo3.git").should("be.visible");
```
lorenzo-cavazzi, line 288: Should this be removed?

lorenzo-cavazzi (changes requested):
Very nice addition! :rocket: 

Since most of the code was already reviewed separately, I did only a quick pass and everything looks reasonable!
I added a couple of inline comments, very minor changes.

While testing, I noticed the visibility is listed twice on the General Settings tab, and with different formats. Perhaps something mixed up while rebasing?

image

lorenzo-cavazzi (commented):
Here are a few additional comments that can be addressed in separate issues as cooldown activities -- some of them are actually improvements, perhaps not even envisioned in the current build.

* Users can now demote themselves from the `Owner` role to something with less permissions. It would be nice to prevent demoting the last owner; otherwise, projects would un-deletable.
image

* I suggest making users always able to remove themselves from a project they don't own. Otherwise, they could have a role assigned by someone else and be stuck with that project (sounds bad for spamming unwanted content). Here is an example: I can't remove my user from there: https://renku-ci-ui-3118.dev.renku.ch/v2/projects/lorenzo.cavazzi.tech/test-creation

* (cosmetic) The content spacing is a little off; the outer box is in line with the other components, but the internal content adds further spacing. I suggest removing that and adding just the bottom margin to the different sections (or top margins, or using the `gap` logic; whatever solution makes the most sense here). I assume this was evident only after merging the other Project 2.0 changes at the last moment :grin: Not a big deal tbh 
image

client/src/features/ProjectPageV2/ProjectPageContent/CodeRepositories/RepositoriesBox.tsx
leafty: I personally think that it's a good interface to explicitly require a component to be passed, even if it is `null`.

ciyer:
> I'm reviewing the code right now. As an early comment, I believe the "Keywords" section got accidentally duplicated while merging/rebasing the code. I'll follow up soon with the full review

Great catch! Just fixed this.

ciyer:
> While testing, I noticed the visibility is listed twice on the General Settings tab, and with different formats. Perhaps something mixed up while rebasing?

Yes, looks like this happened in the rebase. I have fixed this.

client/src/features/ProjectPageV2/ProjectPageContent/CodeRepositories/RepositoriesBox.tsx
ciyer: I suggest leaving this as it is for the moment; if there is a reason to apply a default, we can easily do this in the future.

ciyer:
> * Users can now demote themselves from the `Owner` role to something with less permissions. It would be nice to prevent demoting the last owner; otherwise, projects would un-deletable.
> * I suggest making users always able to remove themselves from a project they don't own. Otherwise, they could have a role assigned by someone else and be stuck with that project (sounds bad for spamming unwanted content). Here is an example: I can't remove my user from there: https://renku-ci-ui-3118.dev.renku.ch/v2/projects/lorenzo.cavazzi.tech/test-creation

I agree that these are both problems, but I think they need to be fixed in the backend first. The UI should allow any change that is allowed by the backend and should generally prevent the user from making changes that are not allowed by the backend as a UX convenience. 

> * (cosmetic) The content spacing is a little off; the outer box is in line with the other components, but the internal content adds further spacing. I suggest removing that and adding just the bottom margin to the different sections (or top margins, or using the `gap` logic; whatever solution makes the most sense here). I assume this was evident only after merging the other Project 2.0 changes at the last moment 😁 Not a big deal tbh

Yes, this is a good point. I will try to make those cosmetic fixes quickly.

lorenzo-cavazzi:
> I agree that these are both problems, but I think they need to be fixed in the backend first. The UI should allow any change that is allowed by the backend and should generally prevent the user from making changes that are not allowed by the backend as a UX convenience.

I agree the backend should catch up by disallowing unwanted actions, but I don't think this excuses us from making a better UX.
Ofc the backend should support the actions (E.G. no point in adding a UX to let users remove themselves from a project if the backend errors), but I don't see a negative point in preventing a downgrade of the last Owner even if doing it from the backend API works.

tests/cypress/e2e/projectV2.spec.ts
ciyer: I fixed the implementation of this test.

leafty:
> > I agree that these are both problems, but I think they need to be fixed in the backend first. The UI should allow any change that is allowed by the backend and should generally prevent the user from making changes that are not allowed by the backend as a UX convenience.
> 
> I agree the backend should catch up by disallowing unwanted actions, but I don't think this excuses us from giving users a better UX. Ofc the backend should support the actions (E.G. no point in adding a UX to let users remove themselves from a project if the backend errors), but I don't see a negative point in preventing a downgrade of the last Owner even if doing it from the backend API works.

I think you are missing the point: we should not hide the bug in the UI because doing so may hide the existence of the bug from the backend maintainers. The backen

andre-code (dismissed):
Great feature! Is it possible to display the member names in the project information? Currently, it only shows the number of members, not their names.

Screenshot 2024-05-24 at 11 31 52

ciyer:
> Great feature! Is it possible to display the member names in the project information? Currently, it only shows the number of members, not their names.

Sure, I will make a PR to address this.

leafty (approved):
LGTM

leafty (approved):

lorenzo-cavazzi:
> I think you are missing the point: we should not hide the bug in the UI because doing so may hide the existence of the bug from the backend maintainers. The backend needs to have the bug filed before we hide the bug.

I totally get that point :smile:  and it still sounds poor to me.
I don't understand how implementing a poor UX can be the best solution here, especially when the reasoning behind is that the backend allows for it. In this case, it also allows for a better solution -- we don't _need_ any change for that.
I think creating an issue in the backend and communicating with the developers is a much better solution to get the bug addressed than letting users fall for it in the UI.

leafty:
I have filed an issue at last https://github.com/SwissDataScienceCenter/renku-data-services/issues/227

PR #3142, opened 2024-05-24
test(e2e): add project page tests

andre-code:
PR to add e2e project page test.

/deploy renku=release-0.53.x


ciyer (approved):
This is great! For the repository test, there is one that overlaps in https://github.com/SwissDataScienceCenter/renku-ui/pull/3118, and I think it would make sense to consolidate the tests here with those in the `projectsV2.spec.ts`, but I think those issues can be handled later -- having these tests would already be helpful.

client/src/components/buttons/Button.tsx
```diff
@@ -373,6 +373,7 @@ export function PlusRoundButton({
           buttonStyles.PlusIconButton
         )}
         onClick={handler}
+        data-cy="plus-button"
```
lorenzo-cavazzi, line 376: This component already has another `data-cy` property. We could fall back to `"plus-button"` whenever `{dataCy}` isn't provided.

client/src/features/sessionsV2/StartSessionButton.tsx
```diff
@@ -50,6 +50,7 @@ export default function StartSessionButton({
         "px-2"
       )}
       to={startUrl}
+      data-cy="session-btn"
```
lorenzo-cavazzi, line 53: I would add `start` to avoid it being too generic.
Also, you used `-button` in most other properties except the sessions components; let's keep it consistent.

```suggestion
      data-cy="start-session-button"
```

client/src/features/sessionsV2/UpdateSessionLauncherModal.tsx
```diff
@@ -180,6 +180,7 @@ export default function UpdateSessionLauncherModal({
           disabled={result.isLoading || !isDirty}
           type="submit"
           onClick={handleSubmit(onSubmit)}
+          data-cy="edit-session-btn"
```
lorenzo-cavazzi, line 183:
```suggestion
          data-cy="edit-session-button"
```

client/src/features/sessionsV2/components/SessionModals/AddSession.tsx
```diff
@@ -142,7 +142,11 @@ function AddSessionCustomImageModal({
             <XLg className={cx("bi", "me-1")} />
             Cancel
           </Button>
-          <Button disabled={result.isLoading} type="submit">
+          <Button
+            disabled={result.isLoading}
+            type="submit"
+            data-cy="add-launcher-custom-btn"
```
lorenzo-cavazzi, line 148:
```suggestion
            data-cy="add-launcher-custom-button"
```

client/src/features/sessionsV2/components/SessionModals/AddSession.tsx
```diff
@@ -258,6 +262,7 @@ function AddSessionExistingEnvModal({
           disabled={result.isLoading}
           type="submit"
           onClick={handleSubmit(onSubmit)}
+          data-cy="add-session-launcher-btn"
```
lorenzo-cavazzi, line 265:
```suggestion
          data-cy="add-session-launcher-button"
```

tests/cypress/fixtures/cloudStorage/storage-schema-s3.json
lorenzo-cavazzi, line 1: Optional: this is a very long file. Since the tests only set up the `giab` storage, should we strip out parts of the response and keep only what is necessary?

tests/cypress/support/renkulab-fixtures/newSession.ts
```diff
@@ -107,6 +107,40 @@ export function NewSession<T extends FixturesConstructor>(Parent: T) {
...
+        "POST",
+        "/ui-server/api/data/session_launchers",
+        response
+      ).as(name);
+      return this;
+    }
+
+    editLauncher(args?: SimpleFixture) {
+      const { fixture = "", name = "editLauncher" } = args ?? {};
+      const response = { fixture, delay: 2000, statusCode: 201 };
```
lorenzo-cavazzi, line 124: You added a bunch of `delay` in the responses. Any specific reason for that? If yes, please add an inline comment about why they should be delayed.
Otherwise, I suggest removing them. Mind that we have many local Cypress tests that execute quickly and still take 10 minutes in total. We don't  want to slow them down any further

lorenzo-cavazzi (changes requested):
Great to have tests covering the latest features :raised_hands: 

I added a few inline comments, only minor improvements.

Good point from @ciyer .
It would be nice to either have a better distinction for content in the projectV2 and projectPageV2 files, or to have all the content in one single file.

tests/cypress/support/renkulab-fixtures/newSession.ts
andre-code: no needed it anymore. So I removed it

andre-code:
> Great to have tests covering the latest features 🙌
> 
> I added a few inline comments, only minor improvements.
> 
> Good point from @ciyer . It would be nice to either have a better distinction for content in the projectV2 and projectPageV2 files, or to have all the content in one single file.

Good point. I renamed the file and moved the project info check to the projectV2 file to clarify the responsibility of the test in each file.

ciyer (changes requested):
I think this is almost ready to merge! 

Only one question -- shouldn't `tests/cypress/e2e/projectV2setup.ts` be `tests/cypress/e2e/projectV2setup.spec.ts`

ciyer (approved):
LGTM! 🎉

ciyer (approved):
🎉

PR #3153, opened 2024-05-29
fix: improve non existing project errors

### LINKED ISSUE #3152: Show the full 404 page when a project/group is not found/visible to the user
Currently, the missing projects and groups show an error

image

The page should be the 404 currently used for missing v1 projects, similar to https://dev.renku.ch/projects/foo/bar.

lorenzo-cavazzi:
Shows a properly formatted error as it comes from the backend API.
The same style has been applied to the group page.

image

/deploy
fix #3152


leafty (changes requested):
We should show a full page 404. I can have a look at doing it, since it requires routing changes.

lorenzo-cavazzi:
> We should show a full page 404. I can have a look at doing it, since it requires routing changes.

Fine with me :+1: 
If you find the time in this cooldown, we can drop this PR. Otherwise, this PR still improves the current status and we can probably merge it as an intermediate step.

client/src/features/ProjectPageV2/ProjectPageContainer/ProjectPageContainer.tsx
```diff
@@ -48,36 +47,26 @@ function ProjectPageContainerInner({
...
-          <p>Could not retrieve project</p>
-          <RtkErrorAlert error={error} />
+          {error ? (
+            <RtkOrNotebooksError error={error} />
+          ) : (
+            <p>Could not retrieve the project.</p>
+          )}
+          <p>
+            Click here to{" "}
+            <Link to={Url.get(Url.pages.projectV2.list)}>
```
leafty, line 61: Can we avoid `Url.js`?

leafty (approved):

client/src/features/ProjectPageV2/ProjectPageContainer/ProjectPageContainer.tsx
lorenzo-cavazzi: Yes! I'll take the occasion to remove it from a few other V2 places. I'll push it to another PR

PR #3154, opened 2024-05-29
feat: improve membership maintenance UX

ciyer:
* do not allow changes to sole owner of a project
* allow users to always remove themselves from a project

/deploy renku=master


ciyer:
This PR addresses the changes requested in #3156. 
* If there is only one owner in a project, the UI prevents changes being made to that user. 
* A user always has the "Remove" button available if they are a member of a project, regardless of their role (modulo the guard above). If the user is not an Owner, this button will always return an error, since the backend only allows Owners to make membership changes.

lorenzo-cavazzi (changes requested):
Very nice UX improvements! We might need to split them apart since removing your own user isn't supported from the backend yet. When I try to remove my user from a project where I'm not Owner, I get the following error:

image

Would you mind either opening an issue in the backend or chatting with the backend team?

P.S. the error formatting can be improved by using `RtkOrNotebooksError` instead of `RtkErrorAlert` in the `RemoveProjectMemberAccessForm` component.

The code looks good! As a user, I was a little confused by the disabled buttons cause the reason wasn't immediate. Should we add a tooltip explaining why they are disabled? E.G. "Last owner can be removed" and "Only owners can change permissions"

ciyer:
> Very nice UX improvements! We might need to split them apart since removing your own user isn't supported from the backend yet. When I try to remove my user from a project where I'm not Owner, I get the following error:
>
> Would you mind either opening an issue in the backend or chatting with the backend team?

I discussed this with the backend team, and they agreed this would make sense. I added the issue https://github.com/SwissDataScienceCenter/renku-data-services/issues/233. I suggest letting users leave a project in the UI since the backend team said they will fix this (and Renku 2.0 is still considered alpha, so rough edges like this are OK).
 
> P.S. the error formatting can be improved by using `RtkOrNotebooksError` instead of `RtkErrorAlert` in the `RemoveProjectMemberAccessForm` component.

👍 I made this change the the error looks better now.

> The code looks good! As a user, I was a little confused by the disabled buttons cause the reason wasn't immediate. Should we add a tooltip explaining why they are disabled? E.G. "Last owner can't be removed" and "Only owners can change permissions"

image

image

lorenzo-cavazzi (commented):
Works great now :rocket: 

Do you prefer to wait until the backend implements  https://github.com/SwissDataScienceCenter/renku-data-services/issues/233 or split out that part and merge the other changes?

ciyer:
> Works great now 🚀
> 
> Do you prefer to wait until the backend implements [SwissDataScienceCenter/renku-data-services#233](https://github.com/SwissDataScienceCenter/renku-data-services/issues/233) or split out that part and merge the other changes?

I think we can merge this now as is. The backend team said they will implement the change to allow a user to remove themselves -- if there is a period where the button results in an error message, I think it is ok, given that Renku 2.0 is in alpha.

lorenzo-cavazzi (approved):
:rocket:

PR #3155, opened 2024-05-29
chore: remove old reactstrap wrapper

lorenzo-cavazzi:
/deploy


ciyer (approved):
Yes! 🎉

client/src/features/searchV2/components/SearchV2DateFilter.tsx
```diff
@@ -1,9 +1,28 @@
...
+ * limitations under the License.
+ */
+
 import cx from "classnames";
 import { DateTime } from "luxon";
+import { ChangeEvent } from "react";
+import { Input } from "reactstrap";
+
+import { DateFilter } from "../searchV2.types.ts";
+import { DateFilterTypes } from "../../../components/dateFilter/DateFilter.tsx";
```
leafty, line 25: We should not have file extensions on `.ts`, `.tsx`, `.js`, `.jsx` imports.

leafty (changes requested):

client/src/features/searchV2/components/SearchV2DateFilter.tsx
lorenzo-cavazzi: Could you argue a little more about why is that bad?
Btw we have a ton of them, if you don't mind I prefer to address that in a separate PR since it's not related to removing the reactstrap wrapper.

leafty: For consistency, as the usual import statements in our code base do not include file extension for `.ts`, `.tsx`, `.js` and `.jsx` files.

lorenzo-cavazzi: It seems a bit random atm.
If we wish to standardize it, we can add a rule in the coding guidelines and enforce it in the reviews. I suggest discussing this at the next UI standup so we are all on board.

leafty (approved):

PR #3161, opened 2024-05-30
Use Bootstrap icons where possible

ciyer:
* minor: replace old external links file with new version
* minor: use Bootstrap icons in project overview
* minor: use Bootstrap icons in the file views
* minor: use Bootstrap icons in the commits display
* minor: use Bootstrap icons in workflow views
* minor: switch simple external links to use the Bootstrap icon
* minor: replace the Github icon in Renku 1 and 2 help pages with the Bootstrap version

image

/deploy renku=master


client/src/features/project/components/migrations/MigrationHelpers.tsx
```diff
@@ -226,7 +227,7 @@ export function DetailsSection({
   if (titleInfo) {
     const externalLinkStyles = {
       className: "text-rk-white",
-      role: "text",
+      role: "text" as ExternalLinkRole,
```
leafty, line 230: nitpick: using `as const` for string literals works just as well so you don't have to import `ExternalLinkRole`.

```suggestion
      role: "text" as const,
```

client/src/components/ExternalLinks.tsx
```diff
@@ -0,0 +1,262 @@
...
+  to: string;
+  /** The text of the tooltip */
+  tooltip: string;
+}
+/**
+ * IconLink
+ * Internal application link that is shown as a font-awesome icon
+ */
+export function IconLink(props: IconLinkProps) {
+  const ref = useRef(null);
```
leafty, line 192: We should avoid using untyped `useRef()` because it is typed as a `MutableRefObject<null>` as opposed to a `RefObject<HTMLAnchorElement>`.

With `MutableRefObject<null>`, `ref.current` is typed as `null` which prevents you from using it, e.g.

```tsx
  useEffect(() => {
    if (ref.current) {
      const anchor = ref.current;
      function listener(event: Event) {
        console.log({ event });
      }
      anchor.addEventListener("select", listener);
      return () => anchor.removeEventListener("select", listener);
    }
  }, []);
```

With `RefObject`, the `.current` property is marked as `readonly` which is the appropriate behavior.

client/src/components/ExternalLinks.tsx
```diff
@@ -0,0 +1,262 @@
...
+      href={url}
+      target="_blank"
+      rel="noreferrer noopener"
+      className={cx("btn-icon-text", className)}
+    >
+      {icon}
+      {title}
+    </a>
+  );
+}
```
leafty, line 242: There is no need for two functions here (see below).

```suggestion
```

client/src/components/ExternalLinks.tsx
```diff
@@ -0,0 +1,262 @@
...
+  /** The URL to link to */
+  url: string;
+}
+export function ExternalIconLink(props: ExternalIconLinkProps) {
+  return props.tooltip ? (
+    <ExternalIconLinkWithTooltip {...props} />
+  ) : (
+    <ExternalIconLinkWithoutTooltip {...props} />
+  );
+}
```
leafty, line 262: We can handle tooltips just fine in one function here. Also, increase flexibility by using `ReactNode`.

```suggestion
interface ExternalIconLinkProps {
  /** [Optional] Any classes to add, e.g., 'nav-link' or 'dropdown-item' */
  className?: string;
  /** An icon to display */
  icon: ReactNode;
  /** The text to show for the link */
  title: ReactNode;
  /** The text of the tooltip */
  tooltip?: ReactNode;
  /** The URL to link to */
  url: string;
}
export function ExternalIconLink({
  className,
  icon,
  title,
  tooltip,
  url,
}: ExternalIconLinkProps) {
  const ref = useRef<HTMLAnchorElement>(null);

  return (
    <>
      <a
        href={url}
        role="button"
        className={cx("btn-icon-text", "icon-link", className)}
        ref={ref}
        target="_blank"
        rel="noreferrer noopener"
      >
        {icon}
        {title}
      </a>
      {tooltip && (
        <UncontrolledTooltip target={ref}>{tooltip}</UncontrolledTooltip>
      )}
    </>
  );
}
```

client/src/components/ExternalLinks.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+interface ExternalIconLinkProps {
+  /** [Optional] Any classes to add, e.g., 'nav-link' or 'dropdown-item' */
+  className?: string;
+  /** An icon to display */
+  icon: React.ReactNode;
+  /** The text to show for the link */
+  title: string;
```
leafty, line 250: note: this property is a misnomer. It does not correspond to a `<a>` title attribute.

leafty (changes requested):
Thanks for doing this. Small comments on the `ExternalLinks` rewrite.

client/src/components/ExternalLinks.tsx
ciyer: fixed in 855053a498ca8d68089dbd42fc8045b0690a79a9

client/src/components/ExternalLinks.tsx
ciyer: fixed in 855053a498ca8d68089dbd42fc8045b0690a79a9

client/src/components/ExternalLinks.tsx
ciyer: fixed in 979cbb45973a6b7a94fceeb2678d83d352c74258

client/src/components/ExternalLinks.tsx
```diff
@@ -0,0 +1,243 @@
...
+  /** The URL to link to */
+  to: string;
+  /** The text of the tooltip */
+  tooltip: string;
+}
+/**
+ * IconLink
+ * Internal application link that is shown as a font-awesome icon
+ */
+export function IconLink(props: IconLinkProps) {
```
leafty, line 191: TODO in new issue: move this to another file. This component should NOT be inside `ExternalLink` components as it is for **internal** links.

leafty (approved):

client/src/components/ExternalLinks.tsx
```diff
@@ -0,0 +1,243 @@
...
+  url,
+}: ExternalIconLinkProps) {
+  const ref = useRef<HTMLAnchorElement>(null);
+
+  return (
+    <>
+      <a
+        href={url}
+        role="button"
+        className={cx("btn-icon-text", tooltip && "icon-link", className)}
```
leafty, line 230: The `tooltip && "icon-link"` is quite the weird behavior! One would think that the class should be added all the time -> https://getbootstrap.com/docs/5.3/helpers/icon-link.

leafty (approved):

client/src/components/ExternalLinks.tsx
ciyer: It probably it should be, but a goal of this PR was to avoid introducing changes to the appearance of the UI.

PR #3173, opened 2024-06-03
Convert font-awesome icons to bootstrap icons

ciyer:
* refactor: remove unused components

/deploy


client/src/components/Avatar.jsx
leafty, line 1: praise: nice! 👏

leafty (commented):
Thanks for the fix.

I think that we agreed that we should pick the `"fill"` variants of icons when they are available:
* `FileEarmark` -> `FileEarmarkFill`
* `Diagram2` -> `Diagram2Fill`

(@lorenzo-cavazzi for the fill icons)

client/src/components/FileExplorer.tsx
```diff
@@ -205,12 +200,12 @@ class TreeNode extends Component<TreeNodeProps, TreeNodeState> {
   render() {
     const icon = this.props.node.children.length ? (
       this.state.childrenOpen === false ? (
-        <FontAwesomeIcon className="link-primary" icon={faFolder} />
+        <FolderFill className="link-primary" />
```
leafty, line 203: In almost all cases, bootstrap icons need to have the `"bi"` (bootstrap icon) class added so that they align with the surrounding text.

Without `bi`:
Screenshot 2024-06-03 at 13 25 59

WIth `bi`:
Screenshot 2024-06-03 at 13 25 54

Here and everywhere else:
```suggestion
        <FolderFill className={cx("bi", "link-primary")} />
```

leafty (changes requested):
This should be good to go after switching to filled icons and fixing the `"bi"` class name. :rocket:

ciyer:
> I think that we agreed that we should pick the `"fill"` variants of icons when they are available:
> 
> * `FileEarmark` -> `FileEarmarkFill`
> * `Diagram2` -> `Diagram2Fill`

Resolved in 06878da8797b1e8e1c3623bc55fd7c2daa931d32

leafty (approved):
Great, thanks!

PR #3175, opened 2024-06-03
fix: update session badges in the project page of Renku 2.0

### LINKED ISSUE #3174: Session badges are not the same size in Renku 2.0
> very minor thing I noticed - the “session state” bubbles (not sure what they are called) are not of consistent height (27 pixels for “not running” and 24 pixels for “running”) and consequently the separation of the lines is not consistent either

image

leafty:
Fixes #3174.

/deploy #notest


ciyer (approved):
👍

PR #3176, opened 2024-06-03
fix: show a full page 404 when a group is not accessible

leafty:
Details:
* Add a "Error 404 - Group no found" page.
* Small refactoring for routing.

Example: https://renku-ci-ui-3176.dev.renku.ch/v2/groups/no-group

/deploy


ciyer (changes requested):
praise: very nice harmonization of how groups and projects are handled in this respect.

A few things I noticed while testing:

The error page for https://renku-ci-ui-3176.dev.renku.ch/v2/groups/new when not logged in looks off:

image

A blank page is shown if the URL for the group has multiple path components. https://renku-ci-ui-3176.dev.renku.ch/v2/groups/foo/bar

image

ciyer (dismissed):
🎉

ciyer (approved):
👍

PR #3180, opened 2024-06-04
chore: use lodash-es

leafty:
Use `lodash-es` instead of `lodash` to optimize the bundle size. The functionality is exactly the same.

/deploy


lorenzo-cavazzi (approved):
:+1:

PR #3181, opened 2024-06-05
chore(client): move verifyTitleCharacters() to optimize bundle size

leafty:
The `verifyTitleCharacters()` method is the only method which imports `xregexp`. Moving it to a separate file allows for bundle size optimization.

Before:
```
build/assets/index-zwCpWK0N.js                              1,960.44 kB │ gzip: 493.50 kB │ map: 6,098.69 kB
```

After:
```
build/assets/index-TyZG-yHr.js                              1,874.07 kB │ gzip: 456.83 kB │ map: 5,793.17 kB
```

/deploy


lorenzo-cavazzi (approved):
:+1:

PR #3186, opened 2024-06-06
docs: update contribution guidelines

lorenzo-cavazzi:
From our offline discussion earlier today:

* Refresh the Contribute section in the Readme file to include details on commits and releases.
* Deprecate the Changelog file and point to the GitHub Releases page instead.


README.md
```diff
@@ -134,40 +136,61 @@ BREAKING CHANGE: this commit breaks something
...
-with a verb, and the pull request number. This last part is added automatically by GitHub
-when merging a PR.
+```
+build: bump the react version to 18.1 (#2345)
+```
+
+We encourage squashing commits when merging a PR and expanding the body to include
+relevant details. This will make the commit history cleaner and easier to
+understand. You should still keep separate commits whenever it makes
+sense to do so. In that case, you can rebase the commits in the PR before merging.
```
leafty, line 146: > You should still keep separate commits whenever it makes
sense to do so. In that case, you can rebase the commits in the PR before merging.

I don't really see a case for this. If a PR is too large, it should be split into logical parts. A PR should be of one type only (e.g. feat, fix, refactor, etc.).

README.md
```diff
@@ -134,40 +136,61 @@ BREAKING CHANGE: this commit breaks something
...
+will create a new PR bumping the version number wherever needed in the code.
 
-All the other sections are optional, but we encourage you to add a reference to the
-issue you are fixing and to mention any breaking changes.
+We _loosely_ follow the recommendations from [Semantic Versioning](https://semver.org)
+when tagging releases. We don't expose APIs so the distinction between `major` and `minor`
+is subtle. Try to follow these rules:
 
-#### Tag
+- If the release includes only bug fixes or minor changes (i.e. no `feat` nor `perf`
```
leafty, line 160: why `perf`? `perf` should not affect features

leafty (changes requested):
Thanks! Some comments below.

README.md
lorenzo-cavazzi: You are right, I removed `perf`

README.md
lorenzo-cavazzi: That's useful in some cases -- E.G. feature branches adding incremental changes.
It's not that common tho; I rephrased the sentence to make it more obvious it's an exception and having multiple PRs is best when commits are logically separate.

leafty (approved):

PR #3187, opened 2024-06-06
fix: update keywords

andre-code:
PR to fix bug to update project keywords

/deploy

### DIFF
--- a/client/src/features/ProjectPageV2/ProjectPageContent/Settings/ProjectSettings.tsx
+++ b/client/src/features/ProjectPageV2/ProjectPageContent/Settings/ProjectSettings.tsx
@@ -49,8 +49,6 @@ import ProjectPageSettingsMembers from "./ProjectSettingsMembers";
 import { ABSOLUTE_ROUTES } from "../../../../routing/routes.constants";
 import { useProject } from "../../ProjectPageContainer/ProjectPageContainer";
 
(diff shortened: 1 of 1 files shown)


client/src/features/ProjectPageV2/ProjectPageContent/Settings/ProjectSettings.tsx
```diff
@@ -140,6 +140,7 @@ export function ProjectSettingsEditForm({ project }: ProjectPageSettingsProps) {
         if (data[key] !== project[key]) dataToSend[key] = data[key];
       }
       dataToSend.visibility = data.visibility;
+      dataToSend.keywords = data.keywords;
```
leafty, line 143: Simplify the whole function to:

```tsx
  const onSubmit = useCallback(
    (data: ProjectV2Metadata) => {
      if (data.namespace !== project.namespace) setRedirectAfterUpdate(true);
      updateProject({
        "If-Match": project.etag ? project.etag : "",
        projectId: project.id,
        projectPatch: data,
      });
    },
    [project, updateProject]
  );
```

and remove:

```tsx
const projectMetadataStringKeys = ["description", "name", "namespace"] as const;
```

leafty (changes requested):
The original code is using an anti-pattern, please update to a more robust version. See below.

leafty (approved):
LGTM

PR #3190, opened 2024-06-07
refactor: code quality

leafty:
Code quality improvements for the project page in Renku 2.0.

/deploy renku=release-0.53.x


ciyer (approved):
Very nice improvements! When reviewing this PR, I noticed that there are several places where we have code for structuring fields and labels in react-hook-form forms. For example, 

- `client/src/features/projectsV2/fields/DescriptionFormField.tsx`
- `client/src/components/form-field/TextInput.tsx`

We should probably consolidate all of these approaches, but I will make a new issue for that.

PR #3194, opened 2024-06-07
Support testing storage credentials before adding them

ciyer:
# Changes

This PR adds a _Test connection_ button to the "Add Cloud Storage" flow. When configuring cloud storage, the user is first prompted to test the connection before they can continue. If the connection does not work, they are given the option to continue anyway.

image

## Testing

Add a cloud storage to a V1 and V2 project, on each, one public bucket not requiring an access key id/secret (giab), and, if you have access to a non-public one, one of those as well.

/deploy renku=ciyer/release-0.53.2


andre-code (dismissed):
Code looks good! ✨

andre-code (commented):
Everything works as expected!

I noticed you rewrote some components to improve readability in the CloudStorageModal. However, the file is becoming quite large and difficult to manage. Would it make sense to split these components into multiple files, or the new parts added?

ciyer:
> Everything works as expected!
> 
> I noticed you rewrote some components to improve readability in the CloudStorageModal. However, the file is becoming quite large and difficult to manage. Would it make sense to split these components into multiple files, or the new parts added?

Yes, I agree. I think the refactoring would better fit into https://github.com/SwissDataScienceCenter/renku-ui/pull/3209, and I can do it there.

andre-code (approved):
✨ LGTM

PR #3209, opened 2024-06-14
Code refactoring + tests for starting v2 sessions with data sources

ciyer:
/deploy renku=ciyer/release-0.53.2

### DIFF
--- a/client/src/features/sessionsV2/SessionStartPage.tsx
+++ b/client/src/features/sessionsV2/SessionStartPage.tsx
@@ -33,7 +33,7 @@ import ProgressStepsIndicator, {
   StatusStepProgressBar,
   StepsProgressBar,
 } from "../../components/progress/ProgressSteps";
-import { ABSOLUTE_ROUTES } from "../../routing/routes.constants.ts";
+import { ABSOLUTE_ROUTES } from "../../routing/routes.constants";
 import useAppDispatch from "../../utils/customHooks/useAppDispatch.hook";
 
 import type { Project } from "../projectsV2/api/projectV2.api";
@@ -43,8 +43,8 @@ import { useGetProjectSessionLaunchersQuery } from "./sessionsV2.api";
 import { SessionLauncher } from "./sessionsV2.types";
 import startSessionOptionsV2Slice from "./startSessionOptionsV2.slice";
 import useSessionLauncherState from "./useSessionLaunchState";
-import { StartSessionOptionsV2 } from "./startSessionOptionsV2.types.ts";
-import { GetStoragesV2ApiResponse } from "../projectsV2/api/storagesV2.api.ts";
+import { StartSessionOptionsV2 } from "./startSessionOptionsV2.types";
+import { GetStoragesV2ApiResponse } from "../projectsV2/api/storagesV2.api";
 
 interface SessionStartingProps extends StartSessionFromLauncherProps {
   containerImage: string;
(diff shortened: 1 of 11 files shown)


andre-code (commented):
Thank you for cleaning up the code ✨.

I noticed that the part in the StartSessionFromLauncher component where we get the currentSessionClass can be extracted into a custom hook. Feel free to include this change, or I can handle it in the PR for the resource class since I will be working on that section.

I have one suggestion: please include the cloud storage values in the startSessionOptionsV2 object.

Additionally, are you planning to include the refactor of the CloudStorageModal in this PR?

client/src/features/sessionsV2/useSessionLaunchState.ts
```diff
@@ -0,0 +1,160 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License
+ */
+
+import { skipToken } from "@reduxjs/toolkit/query";
+import { useEffect, useMemo } from "react";
+
+import useAppDispatch from "../../utils/customHooks/useAppDispatch.hook.ts";
```
leafty, line 22: Can you remove the `.ts` and `.tsx` extensions at the import sites? Your IDE is probably misconfigured because the convention is to not includedl them in Vite projects.

ciyer: Fixed in 189de2c9d87e5db00a7a0ecee5c0ed63fbfd33f5

ciyer:
> I noticed that the part in the StartSessionFromLauncher component where we get the currentSessionClass can be extracted into a custom hook. Feel free to include this change, or I can handle it in the PR for the resource class since I will be working on that section.

This is a good suggestion. I did not initially do that because I thought it would disrupt your work, but if you are OK with it, I will make this change.

client/src/features/sessionsV2/SessionStartPage.tsx
```diff
@@ -382,3 +144,167 @@ function StartSessionFromLauncher({
...
+    startSessionOptionsV2.sessionClass !== 0 &&
+    !isFetchingOrLoadingStorages
+  )
+    return (
+      <SessionStarting
+        containerImage={containerImage}
+        launcher={launcher}
+        project={project}
+        startSessionOptionsV2={startSessionOptionsV2}
+        storages={storages}
```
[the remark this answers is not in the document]
ciyer: Done in d9fa95a0433332514842ae24c43d55145dede4ed

ciyer:
> I noticed that the part in the StartSessionFromLauncher component where we get the currentSessionClass can be extracted into a custom hook. Feel free to include this change, or I can handle it in the PR for the resource class since I will be working on that section.

I looked at doing this, but on second thought, I think it is better if you make the change. After looking at your PR, I did not understand how best to do this refactoring and keep it easy for you to make your changes at the same time.

> I have one suggestion: please include the cloud storage values in the startSessionOptionsV2 object.

Done.

> Additionally, are you planning to include the refactor of the CloudStorageModal in this PR?

I add this in 09dd2bcc9b78e14369a32fd3ba6af3f95a39b6cd.

andre-code (approved):
LGTM 🚀

PR #3214, opened 2024-06-19
Design changes for Renku 2.0

lorenzo-cavazzi:
Revamp the design, dropping most of the old Bootstrap customization used for v1 in favor of a new minimal set of customizations and dropping a bunch of custom components in favor of standard Bootstrap components.

Mind that this also impacts v1 for a few shared components. E.G. alerts, 404/error pages, ...
Some shared components still need to be re-implemented (E.G. in the project settings page), but there will soon be a follow-up PR.

[Notion page reference](https://www.notion.so/renku/Renku-2-0-Identity-UI-Library-Pitch-8dae80a9a7684767ab5424b6f58dbd4d?pvs=31)

image

/deploy


client/src/styles/renku_bootstrap_customization.scss
```diff
@@ -0,0 +1,99 @@
...
+.card-header {
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6 {
+    font-weight: 600;
+  }
+}
```
leafty, line 99: Use `$headings-font-weight: 600;` instead.

client/src/styles/renku_bootstrap_customization.scss
```diff
@@ -0,0 +1,99 @@
...
+    left: -$spacer;
+    position: absolute;
+    right: -$spacer;
+    top: 0;
+    z-index: -1;
+  }
+  &:hover::before {
+    background-color: $list-group-hover-bg;
+  }
+}
```
leafty, line 87: Do not override the default behavior! Create a new class, e.g. `.list-group-item-action-extended` for which the styles apply. As is, this breaks the behavior of regular flush list groups!

client/src/styles/renku_bootstrap_customization.scss
```diff
@@ -0,0 +1,99 @@
...
+// Adjust list groups spacing
+$list-group-item-padding-y: $spacer;
+.list-group-flush {
+  border-top: 1px solid $list-group-border-color;
+
+  & > .list-group-item {
+    padding-left: 0;
+    padding-right: 0;
+  }
+}
```
leafty, line 71: Same comment here, do not override the default bootstrap behavior! This will most likely break standard bootstrap components.

leafty (commented):
Some early comments about bootstrap icons and overrides.

client/src/components/ExternalLinks.tsx
leafty, line 1: Note: we should probably deprecate this component and use a simpler one (with the icon always on the right and idiomatic props).

client/src/components/Logs.tsx
```diff
@@ -280,29 +283,27 @@ function SessionLogs(props: LogBodyProps) {
   // TODO: Revisit after #1219
   return (
     <>
-      <div className="p-2 p-lg-3 text-nowrap">
+      <div className="text-nowrap mb-3">
```
leafty, line 286:
```suggestion
      <div className={cx("text-nowrap","mb-3")}>
```

client/src/components/Logs.tsx
```diff
@@ -163,14 +166,14 @@ const LogDownloadButton = ({
...
-      color={color ?? "rk-green"}
+      color={color ?? "primary"}
       size={size ?? "s"}
       disabled={!canDownload(logs)}
       onClick={() => {
         save();
       }}
     >
-      <FontAwesomeIcon icon={faSave} />
+      <FileEarmarkArrowDown className={cx("me-2", "text-icon")} />
```
leafty, line 176: We should settle on `me-1` and `ms-1` for icons. This will make it consistent across all uses of icons next to text.

```suggestion
      <FileEarmarkArrowDown className={cx("me-1", "text-icon")} />
```

client/src/components/Logs.tsx
```diff
@@ -125,7 +125,10 @@ const LogTabs = ({ logs }: { logs: Record<string, string> }) => {
           return (
             <TabPane key={`log_${tab}`} tabId={tab}>
               <div className="d-flex flex-column">
-                <pre className="bg-primary text-white p-2 w-100 overflow-hidden log-container border-radius-8">
+                <pre
+                  className="overflow-hidden"
+                  style={{ whiteSpace: "pre-line" }}
```
leafty, line 130: What is the reason for the `pre-line` styling? Log contents should not be changed for presentation.

client/src/components/Logs.tsx
```diff
@@ -280,29 +283,27 @@ function SessionLogs(props: LogBodyProps) {
...
+          color="outline-primary"
           style={{ marginRight: 8 }}
           id="session-refresh-logs"
           onClick={() => {
             fetchLogs(sessionName);
           }}
           disabled={logs.fetching}
         >
-          <FontAwesomeIcon icon={faSyncAlt} /> Refresh logs
+          <ArrowRepeat className={cx("me-2", "text-icon")} /> Refresh logs
```
leafty, line 297:
```suggestion
          <ArrowRepeat className={cx("me-1", "text-icon")} />Refresh logs
```

client/src/components/Logs.tsx
```diff
@@ -377,14 +378,14 @@ const EnvironmentLogsPresent = ({
   return (
     <Modal
       isOpen={!!logs.show}
-      className="bg-body modal-dynamic-width"
+      className="modal-dynamic-width"
```
leafty, line 381: Note: `.modal-dynamic-width` should be deprecated.

client/src/components/buttons/Button.tsx
```diff
@@ -118,6 +114,50 @@ function ButtonWithMenu(props: ButtonWithMenuProps) {
   );
 }
 
+type BButtonWithMenuV2Props = {
+  children?:
```
leafty, line 118: ??? `ReactNode` should be sufficient.

client/src/components/buttons/Button.tsx
```diff
@@ -118,6 +114,50 @@ function ButtonWithMenu(props: ButtonWithMenuProps) {
...
+    | React.ReactNode[]
+    | JSX.Element[]
+    | JSX.Element;
+  className?: string;
+  color?: string;
+  default: React.ReactNode;
+  direction?: "up" | "down" | "start" | "end";
+  disabled?: boolean;
+  id?: string;
+  preventPropagation?: boolean;
```
leafty, line 129: This prop should not be needed (if it is, the code is buggy).

client/src/components/buttons/Button.tsx
```diff
@@ -118,6 +114,50 @@ function ButtonWithMenu(props: ButtonWithMenuProps) {
   );
 }
 
+type BButtonWithMenuV2Props = {
```
leafty, line 117:
```suggestion
interface ButtonWithMenuV2Props {
```

client/src/components/buttons/Button.tsx
```diff
@@ -118,6 +114,50 @@ function ButtonWithMenu(props: ButtonWithMenuProps) {
...
+  className?: string;
+  color?: string;
+  default: React.ReactNode;
+  direction?: "up" | "down" | "start" | "end";
+  disabled?: boolean;
+  id?: string;
+  preventPropagation?: boolean;
+  size?: string;
+};
+export function ButtonWithMenuV2(props: BButtonWithMenuV2Props) {
```
leafty, line 132: Prefer `props` deconstruction (and potentially, default assignments too) here.

client/src/error-boundary/ErrorBoundary.tsx
```diff
@@ -50,33 +51,33 @@ export function AppErrorBoundary({ children }: AppErrorBoundaryProps) {
...
-          >
-            reload the page
-          </a>{" "}
-          or go to the{" "}
-          <a className="btn btn-rk-green mx-1" href="/">
-            Renku home page
-          </a>
-          .
+    <>
+      <style type="text/css">{v2Styles}</style>
```
leafty, line 55: Remove this line, use `Helmet` for this. Also, this probably breaks styles until reload, so I would NOT use it at all.

client/src/features/ProjectPageV2/ProjectPageContent/DataSources/DataSourcesBox.tsx
```diff
@@ -48,48 +55,73 @@ export function DataSourcesDisplay({ project }: { project: Project }) {
...
+        <div
+          className={cx(
+            "align-items-center",
+            "d-flex",
+            "justify-content-between"
+          )}
+        >
+          <div className={cx("align-items-center", "d-flex")}>
+            <h4 className={cx("align-items-center", "d-flex", "mb-0", "me-2")}>
+              <Database className={cx("me-2", "small", "text-icon")} />
```
leafty, line 69: Note: `small` does not do anything when combined with `text-icon` or `bi`.

leafty (commented):
More comments.

client/public/static/public/img/logo-yellow.svg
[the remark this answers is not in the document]
lorenzo-cavazzi: Good catch! I'm not sure how it ended up being committed. I'll revert it

client/src/styles/renku_bootstrap_customization.scss
```diff
@@ -0,0 +1,99 @@
...
+$navy: #01192d;
+$info: #0f7396;
+$enable-shadows: true;
+$shadow-renku: rgba(50, 50, 93, 0.25) 0px 2px 5px -1px,
+  rgba(0, 0, 0, 0.3) 0px 1px 3px -1px;
+
+// Variables and functions
+@import "~bootstrap/scss/functions";
+@import "~bootstrap/scss/variables";
+@import "~bootstrap/scss/variables-dark";
```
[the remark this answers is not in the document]
lorenzo-cavazzi: Bootstrap suggests adding imports in a specific order to simplify overwriting variables ([reference](https://getbootstrap.com/docs/5.3/customize/sass/)).
E.G. if you import the variables before changing the `$primary` color, it won't be picked by all components ([reference](https://getbootstrap.com/docs/5.3/customize/sass/#variable-defaults)).

client/src/styles/renku_bootstrap_customization.scss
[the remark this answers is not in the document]
lorenzo-cavazzi: I see your point. Otoh, if we keep them in one place, it helps not to lose track of the changes targeting a single component.

E.G. for the Nav Tabs, not all the desired customization can be done through the Bootstrap variables. I used variables wherever possible and modified the bottom border by overwriting the classes. Having the following block in one place...

```
// Adjust nav tabs
$nav-tabs-link-active-bg: transparent;
$nav-tabs-link-hover-border-color: transparent;
$nav-tabs-link-active-color: $primary;
$nav-link-color: $body-color;
$nav-link-transition: none;
.nav-tabs .nav-link {
  border: 0 !important;
  border-bottom: 2px solid transparent !important;
}
.nav-tabs .nav-link.active {
  border-bottom: 2px solid $nav-tabs-link-active-color !important;
}
```

... seems better than having it scattered into different files.

What do you think?

client/src/styles/renku_bootstrap_customization.scss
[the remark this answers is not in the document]
leafty: In that case we should follow the docs:
```
// 1. Include functions first (so you can manipulate colors, SVGs, calc, etc)
@import "../node_modules/bootstrap/scss/functions";

// 2. Include any default variable overrides here

// 3. Include remainder of required Bootstrap stylesheets (including any separate color mode stylesheets)
@import "../node_modules/bootstrap/scss/variables";
@import "../node_modules/bootstrap/scss/variables-dark";

// 4. Include any default map overrides here
```

So `functions` import at the top, then variable definitions, then `variables, variables-dark` imports, then map overrides.

client/src/styles/renku_additional_classes.scss
```diff
@@ -0,0 +1,17 @@
...
+// ? other cases.
+
+// align boostrap icons when used inline as/with text
+svg.bi,
+svg.text-icon {
+  display: inline-flex;
+  height: 1em;
+  vertical-align: -0.125em;
+  width: 1em;
+}
```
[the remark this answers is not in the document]
lorenzo-cavazzi: > Why introduce a new class?

You mean, why `text-icon` instead of `bi`? I'm having difficulty finding the exact reference, but while reading through the customization guidelines, I read it's best to qualify the names following the other Bootstrap classes. Since there are many `text-*` classes to position the text, I think following the same approach might be good and easier to remember than `bi`.

> Why is the code not including `fill: currentcolor;`?

The icons from Bootstrap Icons already pick the fill color correctly without forcing that property. We dropped most of the custom icons using `fill` improperly; for the ones using it, we very likely do not want to override the color (E.G. the green frog with yellow eyes)

> Utility classes should NOT restrict the tag they apply to

That's a good point! I'll remove it

client/src/styles/renku_bootstrap_customization.scss
lorenzo-cavazzi: That changes _all_ headers.
It would be best to keep the consistency, but there was a lengthy discussion with the design team about the specific sizing of the card header elements and I hope not to have it again -- this feels like a relatively minor deviation anyway

client/src/styles/renku_bootstrap_customization.scss
lorenzo-cavazzi: Could you expand on this? Why is this broken?

We should not use a different on-hover style for flush style list groups. I don't understand why adding a new class instead of overriding is best here, but I'm open to a longer explanation.

client/src/styles/renku_bootstrap_customization.scss
lorenzo-cavazzi: We use extensively the flush variation of the group item list element. It would be great to have the possibility to change padding only for the flush variation, but we can't. What's wrong with overriding the classes here?

client/src/components/Alert.jsx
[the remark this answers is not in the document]
lorenzo-cavazzi: Yeah, we should do that. Do you mind if I do that in separate PR?

client/src/styles/renku_bootstrap_customization.scss
leafty: This is what happens when a standard `list-group-flush` is used:
Screenshot 2024-07-22 at 16 51 37

As you can see, the styling from the cards used in the Dashboard pollute the global scope for flush lists. A specific SCSS module for list groups inside cards should be created to avoid polluting the global scope.

client/src/components/Logs.tsx
lorenzo-cavazzi: I actually changed all the occurrences I found in 2.0 components to `me-2` or `ms-2` for consistency :facepalm: 
I'm not sure which looks best -- `2` feels better on headers and `1` on buttons or links.
I'm ok with both solutions, but changing it to `1` is quite some work. Should we quickly touch base with the whole team in the next standup and do it in the next round? I also have a different proposal to prevent accidentally showing additional margins on smaller window widths.

client/src/components/Logs.tsx
lorenzo-cavazzi: I agree with you. In this specific case, I just re-applied the relevant style from all the classes I dropped. Removing it resulted in much harder-to-read content.

client/src/components/buttons/Button.tsx
lorenzo-cavazzi: We use that for quick actions on the group list elements we have in many places (projects, sessions, groups, data sources, ...). The whole item is a link to a location. Still, the quick-action button offers a way to do something else (E.G. opening a session or stopping it without going to the corresponding project page and opening the session details).

This seems reasonable to me, but I'm open to different solutions

image

client/src/error-boundary/ErrorBoundary.tsx
lorenzo-cavazzi: That's a good suggestion! Done.
I agree it's not the cleanest solution, but we will drop the old v1 styling in the (not so near :unamused: ) future. Do you have any suggestions for alternative ways of having the two styles co-exist?

client/src/features/ProjectPageV2/ProjectPageContent/DataSources/DataSourcesBox.tsx
lorenzo-cavazzi: `small` changes the font size; it seems effective here.
The design team felt the icons were too big on the card headers on the project page

image

client/src/styles/renku_bootstrap_customization.scss
lorenzo-cavazzi: I see. I reasoned that the flush variation is made _specifically_ for working with parent components ([Bootstrap docs](https://getbootstrap.com/docs/5.3/components/list-group/#flush)); it looks visually off anyway when used anywhere else.

But I see how polluting the global scope is bad and playing with negative margins is a double-edged sword (note that Bootstrap itself uses that mechanism for rows/columns).
If you feel the current solution isn't acceptable, I can make another class as in your previous suggestion

lorenzo-cavazzi:
@leafty Thank you for the review! For every suggestion, I either applied it or added questions/considerations inline.

client/src/components/buttons/Button.tsx
lorenzo-cavazzi: As we discussed offline, we should address this as soon as possible.
I made a separate issue for making a component that simplifies avoiding nesting without forcing re-implementing ad-hoc solutions
https://github.com/SwissDataScienceCenter/renku-ui/issues/3250

client/src/features/ProjectPageV2/ProjectPageContent/DataSources/DataSourcesBox.tsx
leafty: OK, I see.

So some design notes here:
1. The icons are designed to be `1em` when used next to text content (that is being the same size as the surrounding text). Breaking this rule breaks vertical alignment of the icons. In the project page the icons are not nicely aligned with the section titles because of this. I suggest we remove the `small` CSS class. It should not be used for icons.
2. The `h4` title here and elsewhere should not be flex elements (because that doesn't change alignment).

```tsx
<h4 className={cx("mb-0", "me-2")}>
```

client/src/styles/renku_additional_classes.scss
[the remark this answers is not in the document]
lorenzo-cavazzi: After the offline discussion, I removed `text-icon` as suggested.

P.S. Mind that I still haven't added `fill: currentcolor` since it's already embedded in the bootstrap icons.

client/src/components/Logs.tsx
lorenzo-cavazzi: I changed it to always use a space of `1` to maintain consistency everywhere.
I think using whitespaces would be best, but I don't want to include that here -- we can discuss it and see if it makes sense to create a separate issue

client/src/components/Alert.jsx
```diff
@@ -86,13 +85,13 @@ class RenkuAlert extends Component {
...
   getIcon() {
     const icon = {
-      danger: <DangerIcon className="text-danger" />,
-      info: <InfoIcon className="text-info" />,
-      warning: <WarningIcon className="text-warning" />,
-      success: <SuccessIcon className="text-success" />,
+      danger: <ExclamationTriangleFill className={cx(bi, "text-danger")} />,
+      info: <InfoCircleFill className={cx(bi, "text-info")} />,
+      warning: <ExclamationTriangleFill className={cx(bi, "text-warning")} />,
+      success: <CheckCircleFill className={cx(bi, "text-success")} />,
```
leafty, line 91:
```suggestion
      danger: <ExclamationTriangleFill className={cx("bi", "text-danger")} />,
      info: <InfoCircleFill className={cx("bi", "text-info")} />,
      warning: <ExclamationTriangleFill className={cx("bi", "text-warning")} />,
      success: <CheckCircleFill className={cx("bi", "text-success")} />,
```

lorenzo-cavazzi: :pray:

client/src/features/ProjectPageV2/ProjectPageContent/CodeRepositories/CodeRepositoryDisplay.tsx
```diff
@@ -375,57 +345,90 @@ function CodeRepositoryActions({
...
+  //       {canonicalUrl?.hostname && (
+  //         <span data-cy="code-repository-title">{canonicalUrl.hostname}</span>
+  //       )}
+  //       <a href={canonicalUrlStr} target="_blank" rel="noreferrer noopener">
+  //         {title || canonicalUrlStr}
+  //         <BoxArrowUpRight className={cx("bi", "ms-1")} size={16} />
+  //       </a>
+  //     </div>
+  //   </div>
+  // );
```
leafty, line 387: Why is this removed? Can't it work as is?

client/src/features/ProjectPageV2/ProjectPageContent/CodeRepositories/CodeRepositoryDisplay.tsx
```diff
@@ -483,72 +486,54 @@ function RepositoryPermissions({ repositoryUrl }: RepositoryPermissionsProps) {
...
-    "p-2",
-  ];
-
-  const buttonContent = isLoading ? (
-    <Loader className="bi" inline size={16} />
+  const badgeIcon = isLoading ? (
+    <Loader className="me-1" inline size={16} />
   ) : permissions.pull && permissions.push ? (
-    <CheckCircleFill className="bi" />
+    <CircleFill className={cx("me-1", "bi")} />
```
leafty, line 492: Remove all the ternaries here.

client/src/features/ProjectPageV2/ProjectPageContent/CodeRepositories/CodeRepositoryDisplay.tsx
```diff
@@ -483,72 +486,54 @@ function RepositoryPermissions({ repositoryUrl }: RepositoryPermissionsProps) {
...
-    "rounded-circle",
-    "fs-6",
-    "lh-1",
-    "p-2",
-  ];
-
-  const buttonContent = isLoading ? (
-    <Loader className="bi" inline size={16} />
+  const badgeIcon = isLoading ? (
+    <Loader className="me-1" inline size={16} />
```
leafty, line 490: Should the size be smaller here? This should correspond to the font size.

client/src/features/ProjectPageV2/ProjectPageContent/CodeRepositories/CodeRepositoryDisplay.tsx
```diff
@@ -483,72 +486,54 @@ function RepositoryPermissions({ repositoryUrl }: RepositoryPermissionsProps) {
...
-    <XCircleFill className="bi" />
+    <CircleFill className={cx("me-1", "bi")} />
   );
 
-  const buttonClasses = isLoading
-    ? ["border-dark-subtle", "bg-light", "text-dark"]
-    : permissions.pull && permissions.push
-    ? ["border-success", "bg-success-subtle", "text-success"]
+  const badgeText = isLoading
+    ? ""
```
leafty, line 500:
```suggestion
    ? null
```

(see below)

client/src/features/ProjectPageV2/ProjectPageContent/CodeRepositories/CodeRepositoryDisplay.tsx
```diff
@@ -483,72 +486,54 @@ function RepositoryPermissions({ repositoryUrl }: RepositoryPermissionsProps) {
...
-              <XLg className={cx("bi", "me-1")} />
-              Close
-            </Button>
-          </div>
-        </ModalFooter>
-      </Modal>
-    </>
+    <Badge pill className={cx("border", badgeColorClasses)}>
+      {badgeIcon}
+      <span className="fw-normal">{badgeText}</span>
```
leafty, line 518: Avoid rendering empty nodes:

```suggestion
      {badgeText && (<span className="fw-normal">{badgeText}</span>)}
```

client/src/features/ProjectPageV2/ProjectPageContent/CodeRepositories/CodeRepositoryDisplay.tsx
```diff
@@ -588,60 +573,116 @@ function RepositoryPermissionsModalContent({
     () => `${repositoryUrl.replace(/.git$/i, "")}`,
     [repositoryUrl]
   );
+  const canonicalUrl = useMemo(() => {
```
leafty, line 576: Use the `safeNewUrl` method instead.

```tsx
const canonicalUrl = useMemo(
    () => safeNewUrl(canonicalUrlStr),
    [canonicalUrlStr]
  );
```

client/src/features/ProjectPageV2/ProjectPageContent/DataSources/DataSourcesBox.tsx
lorenzo-cavazzi: 1. I see your point; there was a somewhat lengthy discussion with the design team about this, but the alignment looks a little off. I have removed the `small` modifier and I will run this through the design team again :crossed_fingers: 
2. Done

client/src/features/ProjectPageV2/ProjectPageContent/DataSources/DataSourceDisplay.tsx
```diff
@@ -214,79 +183,42 @@ export function DataSourceDisplay({
...
-      <span className={cx("text-truncate")} data-cy="data-storage-name">
-        {storageSensitive.name}
-      </span>
-    </div>
-  );
+  const storageType = storageSensitive?.storage_type ? (
+    <span className="fst-italic" data-cy="data-storage-type">
+      (type: {storageSensitive?.storage_type})
+    </span>
+  ) : null;
```
leafty, line 190: Avoid rendering the space `" "` when this is `null`:

```suggestion
  const storageType = storageSensitive?.storage_type ? (
    <>
      {" "}
      <span className="fst-italic" data-cy="data-storage-type">
        (type: {storageSensitive.storage_type})
      </span>
    </>
  ) : null;
```

client/src/features/ProjectPageV2/ProjectPageContent/DataSources/DataSourceDisplay.tsx
```diff
@@ -214,79 +183,42 @@ export function DataSourceDisplay({
...
+        className={cx("cursor-pointer", "link-primary", "text-body")}
+        onClick={toggleDetails}
       >
-        <DataSourceActions storage={storage} projectId={projectId} />
-      </Col>
-    </Row>
+        <Row className={cx("align-items-center", "g-2")}>
+          <Col>
+            <span>
+              {storageName} {storageType}
```
leafty, line 208:
```suggestion
              {storageName}{storageType}
```

leafty (commented):
Some comments.

Progress: 34/141 files viewed.

client/src/styles/renku_bootstrap_customization.scss
leafty: These adjustments make the flush list work only inside components which use `x` padding with a value of `$spacer`. Other values (e.g. `mx-4`) would also break.

For this reason, I think we could restrict these changes to the flush list to be only for flush lists found inside cards.

client/src/features/ProjectPageV2/ProjectPageContent/CodeRepositories/CodeRepositoryDisplay.tsx
lorenzo-cavazzi: Product/design teams discussed a common solution for showing the source of both code repositories and data sources in a similar way. This (or some variation) will be back in the next PR further changing the design.

client/src/features/ProjectPageV2/ProjectPageContent/CodeRepositories/CodeRepositoryDisplay.tsx
lorenzo-cavazzi: Ah yes! It should be `12` in the badges. Good catch!

client/src/features/projectsV2/fields/AddEntityMemberLookupForm.tsx
```diff
@@ -113,26 +109,27 @@ export default function AddEntityMemberEmailLookupForm({
...
           <XLg className={cx("bi", "me-1")} />
           Close
         </Button>
-        <Button disabled={true} onClick={handleSubmit(onSubmit)} type="submit">
+        <Button
+          color="primary"
+          disabled={true}
+          onClick={handleSubmit(onSubmit)}
+          type="submit"
+        >
```
leafty, line 132: Why assign an `onClick` if this button is disabled?

client/src/features/projectsV2/fields/ProjectVisibilityFormField.tsx
```diff
@@ -85,14 +70,9 @@ export default function ProjectVisibilityFormField<T extends FieldValues>({
...
-                className={cx(
-                  "cursor-pointer",
-                  "d-flex",
-                  "align-items-center",
-                  "gap-1"
-                )}
+                className={cx("cursor-pointer")}
               >
-                Private <Lock size={16} />
+                Private <Lock className="bi" />
```
leafty, line 75:
```suggestion
                Private
                <Lock className={cx("bi", "ms-1")} />
```

Also, why is the icon after the text? Usual placement is before the text for most icons.

client/src/features/projectsV2/list/ProjectV2ListDisplay.tsx
```diff
@@ -183,33 +185,41 @@ function ProjectV2ListProject({ project }: ProjectV2ListProjectProps) {
...
           </p>
           {description && (
-            <p className={cx("mb-2", "card-text")}>{description}</p>
+            <p
+              style={{
+                display: "-webkit-box",
+                overflow: "hidden",
+                WebkitBoxOrient: "vertical",
+                WebkitLineClamp: 3,
+              }}
```
leafty, line 204: We should define re-usable classes or components for this.

client/src/features/projectsV2/new/ProjectV2FormSubmitGroup.tsx
```diff
@@ -44,28 +43,32 @@ export default function ProjectFormSubmitGroup({
...
         onClick={previousStep}
       >
         <ArrowLeft /> Back
       </Button>
       <div>
         {currentStep === 0 && (
-          <Button className="me-1" type="submit">
-            Set Visibility <ArrowRight />
+          <Button color="primary" type="submit">
+            Set visibility <ArrowRight />
```
leafty, line 54:
```suggestion
            Set visibility
            <ArrowRight classname="ms-1" />
```

client/src/features/projectsV2/new/ProjectV2New.tsx
```diff
@@ -66,10 +66,10 @@ function ProjectV2NewHeader({
 }: Pick<NewProjectV2State, "currentStep">) {
   return (
     <>
-      <div className="mb-2">
+      <p>
         V2 Projects let you group together related resources and control who can
-        access them. {"  "} <WipBadge />
-      </div>
+        access them. <WipBadge />
```
leafty, line 71:
```suggestion
        access them.<WipBadge classname="ms-1" />
```

client/src/features/projectsV2/new/ProjectV2NewForm.tsx
```diff
@@ -223,8 +214,12 @@ function ProjectV2NewRepositoryStepForm({
...
             );
           })}
-          <Button data-cy="project-add-repository" onClick={onAppend}>
-            <PlusLg /> repository
+          <Button
+            color="outline-primary"
+            data-cy="project-add-repository"
+            onClick={onAppend}
+          >
+            <PlusLg className="bi" /> Add repository
```
leafty, line 222:
```suggestion
            <PlusLg className={cx("bi", "me-1")} />Add repository
```

client/src/features/projectsV2/show/GroupSimple.types.tsx
```diff
@@ -13,16 +13,14 @@
...
-import ProjectInformation from "./ProjectInformation/ProjectInformation";
-
-export default function ProjectInformationPage() {
-  return (
-    <div className={cx("d-block", "d-lg-none", "d-sm-block", "pt-4")}>
-      <ProjectInformation />
-    </div>
-  );
-}
+export type GroupSimple = {
```
leafty, line 19: Why does this type exist? Can't we use something generated from the API?

client/src/features/projectsV2/show/GroupSimple.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { ABSOLUTE_ROUTES } from "../../../routing/routes.constants";
+import VisibilityIcon from "../../../components/entities/VisibilityIcon";
+import { TimeCaption } from "../../../components/TimeCaption";
+
+interface GroupSimpleProps {
+  className?: string | string[];
+  element: "card-body" | "list-item" | "plain";
+  group: GroupSimple;
+}
+export default function GroupSimple({
```
leafty, line 32: Invalid component name. It cannot use the same name as the type.

client/src/features/projectsV2/show/GroupSimple.tsx
```diff
@@ -0,0 +1,94 @@
...
+import cx from "classnames";
+import { Link, generatePath } from "react-router-dom-v5-compat";
+
+import type { GroupSimple } from "./GroupSimple.types";
+import { ABSOLUTE_ROUTES } from "../../../routing/routes.constants";
+import VisibilityIcon from "../../../components/entities/VisibilityIcon";
+import { TimeCaption } from "../../../components/TimeCaption";
+
+interface GroupSimpleProps {
+  className?: string | string[];
```
leafty, line 28: Only string here for convention.
```suggestion
  className?: string;
```

client/src/features/projectsV2/show/GroupSimple.tsx
```diff
@@ -0,0 +1,94 @@
...
+export default function GroupSimple({
+  className,
+  element,
+  group,
+}: GroupSimpleProps) {
+  const content = (
+    <div
+      className={cx(
+        element === "card-body" && cx("d-flex", "flex-column", "flex-grow-1")
+      )}
```
leafty, line 41: Do not nest `cx` calls!

```suggestion
      className={cx(
        element === "card-body" && ["d-flex", "flex-column", "flex-grow-1"]
      )}
```

client/src/features/projectsV2/show/GroupSimple.tsx
```diff
@@ -0,0 +1,94 @@
...
+      </div>
+    </div>
+  );
+
+  const elementClasses =
+    element === "card-body"
+      ? cx("card-body", "d-flex")
+      : element === "list-item"
+      ? cx("list-group-item", "list-group-item-action")
+      : "";
```
leafty, line 76: Do not nest `cx` call here too.

```suggestion
  const elementClasses =
    element === "card-body"
      ? ["card-body", "d-flex"]
      : element === "list-item"
      ? ["list-group-item", "list-group-item-action"]
      : [];
```

client/src/features/projectsV2/show/ProjectSimple.tsx
```diff
@@ -0,0 +1,100 @@
...
+import VisibilityIcon from "../../../components/entities/VisibilityIcon";
+import { TimeCaption } from "../../../components/TimeCaption";
+
+interface ProjectSimpleProps {
+  className?: string | string[];
+  element: "card-body" | "list-item" | "plain";
+  project: Project;
+}
+
+export default function ProjectSimple({
```
leafty, line 33: The name is a bit confusing. Where is this component supposed to be used? Prefer `ProjectShortHandDisplay` or `ProjectDisplaySmall` or similar.

client/src/features/searchV2/components/SearchV2Results.tsx
```diff
@@ -116,60 +116,60 @@ function SearchV2ResultsContent() {
...
 
 interface SearchV2CardTitleProps {
-  children?: ReactNode;
-  entityType: ReactNode;
-  url: string;
+  name: string;
+  nameUrl: string;
+  namespace: string;
+  namespaceUrl: string;
+  entityType?: "Project" | "Group" | "User";
```
leafty, line 138: This should be imported from the search API, not hard-coded.

client/src/features/searchV2/components/SearchV2Results.tsx
```diff
@@ -116,60 +116,60 @@ function SearchV2ResultsContent() {
...
     </Col>
   );
 }
 
 interface SearchV2CardTitleProps {
-  children?: ReactNode;
-  entityType: ReactNode;
-  url: string;
+  name: string;
+  nameUrl: string;
```
leafty, line 135:
```suggestion
  entityUrl: string;
```

as it is more informative.

leafty (commented):
More comments.

Progress: 85/143 files viewed.

client/src/features/sessionsV2/SessionLauncherForm.module.scss
```diff
@@ -5,16 +5,16 @@
...
-      var(--bs-focus-ring-blur, 0) var(--bs-focus-ring-width)
-      var(--bs-focus-ring-color);
-  }
-}
+//   input:focus-visible + & {
+//     box-shadow: var(--bs-focus-ring-x, 0) var(--bs-focus-ring-y, 0)
+//       var(--bs-focus-ring-blur, 0) var(--bs-focus-ring-width)
+//       var(--bs-focus-ring-color);
+//   }
+// }
```
leafty, line 20: Leftover?

client/src/features/sessionsV2/SessionList/SessionItem.tsx
```diff
@@ -0,0 +1,128 @@
...
+  project,
+  session,
+  toggleSessionDetails,
+}: SessionItemProps) {
+  return (
+    <ListGroupItem
+      action
+      className="cursor-pointer"
+      data-cy="session-launcher-item"
+      onClick={() => toggleSessionDetails()}
```
leafty, line 53:
```suggestion
      onClick={toggleSessionDetails}
```

client/src/features/sessionsV2/SessionView/SessionView.tsx
```diff
@@ -69,67 +71,70 @@ import {
...
-function SessionCard({
-  session,
-  project,
-}: {
-  session: Session;
-  project: Project;
-}) {
+interface SessionCardContentProps {
+  color: string;
+  contentDescription: React.ReactNode;
```
leafty, line 77: I prefer:

```suggestion
  contentDescription: ReactNode;
```

It's shorter and forces to import from `react`, instead of relying on an ambient import.

client/src/features/usersV2/show/UserShow.tsx
```diff
@@ -89,38 +89,38 @@ export default function UserShow() {
...
             </div>
           </div>
-          <p className="fs-4">{`@${username}`}</p>
+          <p className="fst-italic">{`@${username}`}</p>
         </div>
       </div>
 
       <section>
-        <h2 className="fs-4">Personal Projects</h2>
+        <h4>Personal Projects</h4>
```
leafty, line 123: Note: this should probably still be an `h2`. Semantically, it's better to avoid gaps in `hx` uses in a web page.

leafty (commented):
More comments.

Progress: 121/143 files viewed.

tests/cypress/e2e/dashboard.spec.ts
```diff
@@ -256,12 +256,12 @@ describe("dashboard message", () => {
...
-    fixtures.config();
-    visitDashboardPage();
+  // // it("does not display a message when disabled", () => {
+  // //   fixtures.config();
+  // //   visitDashboardPage();
 
-    cy.getDataCy("dashboard-message").should("not.exist");
-  });
+  // //   cy.getDataCy("dashboard-message").should("not.exist");
+  // // });
```
leafty, line 264: ?

client/src/storybook/bootstrap/Accordion.stories.tsx
leafty, line 1: Do we use Accordion?

client/src/features/projectsV2/fields/ProjectVisibilityFormField.tsx
lorenzo-cavazzi: Fair point. Swapped and adjusted for both the Private and Public options

client/src/features/projectsV2/list/ProjectV2ListDisplay.tsx
lorenzo-cavazzi: Good idea. Added here https://github.com/SwissDataScienceCenter/renku-ui/pull/3214/commits/14ba91b22919d9c30234e0ca755cbc2a2e6d81fe

client/src/features/usersV2/show/UserShow.tsx
lorenzo-cavazzi: That's an interesting point. We ended up mainly using h2, h4 and h5. I'm wondering if we should adjust the default header sizing instead to let developers pick headers more intuitively.

client/src/components/Logs.tsx
leafty: Tested: just use the `text-wrap` class from bootstrap to enable text wrapping in the `<pre>` tag.

lorenzo-cavazzi: `text-wrap` seems to have a very different effect. Comper `white-space: pre-line` on the left with `text-wrap` on the right.

image

The best thing would be to improve the backend response; currently, it returns a very long string.
Since `white-space: pre-line` seems to work well enough as a workaround, I would leave this as is since the scope of this redesign effort is already quite broad.

andre-code (commented):
Great to see those changes in place. I appreciate the tremendous effort put into redesigning the app. This is a big PR with many things to consider, and so far, the results are amazing 🤩 👏 💪. Along with the minor comments I've left in the code, I found a few additional issues that would be great to address here or in the next phase of the redesign:

1. Missing top padding in the "Add Data Source" confirmation modal and the "Remove Code Repository" modal.
Screenshot 2024-07-25 at 09 23 07
Screenshot 2024-07-25 at 09 24 27

2. The text buttons combine uppercase and lowercase words. Please standardize them so only the first letter is uppercase.

Screenshot 2024-07-25 at 09 25 14


3. Consider adjusting the off-canvas size for the session view. This off-canvas displays a lot of information, and the reduced size makes it look disorganized and overwhelming.

Screenshot 2024-07-25 at 09 26 04

4. In the session off-canvas, consider including the code repository permission information, as the badge alone is not enough to understand the status.
Screenshot 2024-07-25 at 09 26 49


5. Why does this menu have a dark background?
Screenshot 2024-07-25 at 09 27 40


6. Consider redesigning this; it doesn't look right to display a card inside this list.
Screenshot 2024-07-25 at 10 36 15


7. This modal shows a lot of information. Consider using shorter texts and making the buttons larger to avoid overwhelming the user when selecting an option. Using button-lg makes a big difference.
Screenshot 2024-07-25 at 11 15 38
Screenshot 2024-07-25 at 11 15 23


8. Don’t forget to standardize the use of the user avatar on all pages where the user name is included.
Screenshot 2024-07-25 at 11 16 43


9. I mentioned this in a previous meeting, but as a reminder, please redesign the filter form. Using cards as the result format so far doesn’t help differentiate one component from another.
Screenshot 2024-07-25 at 11 17 08


10. The buttons here are not using the primary and outline buttons as the rest of the app does.
Screenshot 2024-07-25 at 11 22 13

Screenshot 2024-07-25 at 11 22 26


11. Consider removing or moving inside the card the short text "You are currently starting a session...", "You are currently running a session...", "No session is running from...". This text explains the current session status, but being placed just below the title can be mistaken for a label.
Screenshot 2024-07-25 at 13 09 46

Screenshot 2024-07-25 at 13 09 55
Screenshot 2024-07-25 at 13 10 03

client/src/components/ExternalLinks.tsx
lorenzo-cavazzi: Good point!
Here is an issue for doing that -- feel free to add to it https://github.com/SwissDataScienceCenter/renku-ui/issues/3255

client/src/features/usersV2/show/UserShow.tsx
lorenzo-cavazzi: Here is the issue for that https://github.com/SwissDataScienceCenter/renku-ui/issues/3256
I'll be sure to address it in the 2nd round of the design changes

client/src/components/entities/VisibilityIcon.tsx
```diff
@@ -57,7 +58,13 @@ const VisibilityIcon = ({ visibility, className }: VisibilityIconProps) => {
     <>
       <div
         ref={ref}
-        className={`card-visibility-icon d-flex gap-2 align-items-center ${className}`}
+        className={cx(
+          "align-items-center",
+          "card-visibility-icon",
```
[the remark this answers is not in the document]
lorenzo-cavazzi: In places where we use v1 components for v2 content, I've left the v1 classes wherever useful for the v1 pages since they do not affect v2 pages. I would keep it consistent if you don't mind (most v1 components don't actually support `className`).

client/src/components/Alert.jsx
```diff
@@ -86,13 +85,13 @@ class RenkuAlert extends Component {
 
   getIcon() {
     const icon = {
-      danger: <DangerIcon className="text-danger" />,
-      info: <InfoIcon className="text-info" />,
-      warning: <WarningIcon className="text-warning" />,
-      success: <SuccessIcon className="text-success" />,
+      danger: <ExclamationTriangleFill className={cx("bi", "text-danger")} />,
```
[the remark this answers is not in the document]
lorenzo-cavazzi: That's a good point! I'll bring this up with the designer since this changes the visualization quite a lot when it comes to alerts.

lorenzo-cavazzi:
@andre-code thank you for your review!

For all your comments, I've either 1) addressed it or 2) added it to the second round of design changes or 3) added an inline comment (below this message for non-inline comments). Please give it a look.

--------

1, 2: Done

3. > Consider adjusting the off-canvas size for the session view [...]

Currently, all the off-canvas elements have the same width. Making it larger for sessions won't help much -- we have a lot of vertical content, but expanding horizontally won't fit more elements and users will need to scroll anyway. For this reason, I prefer to keep it consistent with all the other off-canvas. We can revise this later if we wish to re-think how we display the content there.

4. > In the session off-canvas, consider including the code repository permission information [...]

The space there seems a little limited to add more info. We could think about opening the Code repositories off-canvas on click. If we do that, we should add the same interaction for the other elements on the page (data source, launcher, ...). I would rather discuss this change with the product team since it will require a bit of work.

5: That's planned for the second round of design changes.

6, 7, 8, 9: Good point! I'll check into those with Jimena during the second round of design changes. Thank you for bringing those up!

10. Done

11. I would rather rework both the "Launched Sessions" and "Session Environment" sections separately; I agree the current output might be a little confusing to the users, but the restyle PR might not be the best place to change the content there.

leafty (changes requested):
Some thing I noted from the deployment:

1. The icon used in the new Alert components should be aligned to the top. If the content of the alert is large, then the icon floats too much in the middle
Screenshot 2024-07-29 at 16 00 53
Screenshot 2024-07-29 at 15 59 59

2. The small avatar should be used for project members:
Screenshot 2024-07-29 at 16 04 13


Otherwise the changes look good, and any more adjustment can be done later.

lorenzo-cavazzi:
@leafty 
For the avatar, there was a discussion and we postponed the decision to the second iteration. I think we should leave this as is since we will address it soon; the alternative would be temporarily removing the avatar also from the other places.

For the alerts, we considered that the content is unlikely to be very long. In principle, I agree with your comment and I will bring this up again, but I would rather address this later since the topic has been discussed already and there are a few other details we would like to address for the alerts (E.G. using non-filled icons and match the color we use for the text)

leafty (approved):

PR #3229, opened 2024-07-03
Allow editing of cloud storage secret credentials

ciyer:
# Changes

- Allow setting and removing credentials on a data source
- Do not prompt for credentials at session launch, if there are stored credentials

# Screenshots

**View credentials**
image

image

**Update credentials**
image

**Clear credentials**
image

# Testing

To test, try the following:
- Create a data source that requires credentials
- Use the "Update Credentials" option to set the credentials (they cannot yet be stored at creation time)
- Launch a session and check that you are not prompted to provide credentials
- Clear the credentials for the data source (currently, you also need to delete the entries from the secrets page, but the backend will be changed to this itself)

/deploy renku-notebooks=pitch/save-storage-secrets renku-data-services=pitch/save-storage-credentials-save-secrets


leafty:
Should the user be given a "Save anyway" option?

Screenshot 2024-07-12 at 13 31 53

client/src/features/sessionsV2/SessionStartCloudStorageSecretsModal.tsx
```diff
@@ -440,19 +559,26 @@ export default function SessionStartCloudStorageSecretsModal({
...
           <CloudStorageConfigurationSecrets
             cloudStorageConfig={cloudStorageConfigs[index]}
+            context={context}
             control={control}
           />
 
           <div className="mt-3">
-            {validationResult.isError ? (
+            {!validationResult.isError ? (
+              <div>&nbsp;</div>
```
leafty, line 568: This existed before this PR, but what?!? This case needs to be removed and no content generated in this case.

client/src/features/sessionsV2/SessionStartCloudStorageSecretsModal.tsx
```diff
@@ -309,6 +424,7 @@ interface SessionStartCloudStorageSecretsModalProps {
     | undefined;
 }
 export default function SessionStartCloudStorageSecretsModal({
+  context = "session",
```
leafty, line 427: non-blocking, create a new issue for this: Adding this attribute starts a slippery slope which leads to spaghetti code later on. This component needs to be split into reusable parts so that each case can be defined cleanly as separate components.

Also, this component is now a misnomer, as it is not exclusively used during session start.

client/src/features/sessionsV2/SessionStartCloudStorageSecretsModal.tsx
```diff
@@ -105,7 +139,8 @@ function CredentialsButtons({
           </span>
         ) : (
           <span>
-            Continue <ChevronRight className="bi" />
+            {context === "session" ? "Continue" : "Test and Save"}{" "}
+            <ChevronRight className="bi" />
```
leafty, line 143:
```suggestion
            {context === "session" ? "Continue" : "Test and Save"}
            <ChevronRight className={cx("bi", "ms-1")} />
```

client/src/features/ProjectPageV2/ProjectPageContent/DataSources/useDataSourceConfiguration.hook.ts
```diff
@@ -0,0 +1,95 @@
...
+import { useMemo } from "react";
+
+import { CLOUD_OPTIONS_OVERRIDE } from "../../../project/components/cloudStorage/projectCloudStorage.constants";
+import { useGetStorageSecretsByV2StorageIdQuery } from "../../../projectsV2/api/projectV2.enhanced-api";
+import { RCloneOption } from "../../../projectsV2/api/storagesV2.api";
+import type { CloudStorageGetRead } from "../../../projectsV2/api/storagesV2.api";
+
+import type { SessionStartCloudStorageConfiguration } from "../../../sessionsV2/startSessionOptionsV2.types";
+
+interface StartSessionFromLauncherProps {
```
leafty, line 28: Misnamed arguments.

```suggestion
interface UseDataSourceConfigurationArgs {
```

client/src/features/ProjectPageV2/ProjectPageContent/DataSources/DataSourceDisplay.tsx
```diff
@@ -182,8 +188,25 @@ export function DataSourceActions({
...
+            className={cx(
+              "d-flex",
+              "align-items-center",
+              "gap-2",
+              "justify-content-start"
+            )}
+            onClick={toggleCredentials}
+            data-cy="data-source-credentials"
+          >
+            <Lock /> Update Credentials
```
leafty, line 201:
```suggestion
            <Lock className={cx("bi", "ms-1")} />
            Update Credentials
```

client/src/features/sessionsV2/SessionStartCloudStorageSecretsModal.tsx
```diff
@@ -52,47 +53,80 @@ import type {
...
+            <Button
+              className={cx("ms-2", "btn-outline-rk-green")}
+              onClick={onSkip}
+            >
+              Skip <SkipForward className={cx("bi", "me-1")} />
+            </Button>
+          </span>
+          <UncontrolledTooltip target={skipButtonRef}>
+            Skip the connection test. At session launch, the storage will try
+            mount
```
leafty, line 96: typo?

```suggestion
            to mount
```

leafty (commented):
some early comments

client/src/features/sessionsV2/SessionStartCloudStorageSecretsModal.tsx
ciyer: I've renamed the component and, though it could still be improved, restructured the code to make it easier to follow in situations where different content is shown based on which context the component is used.

leafty:
> > Should the user be given a "Save anyway" option?
> 
> I thought about this, and I'm not sure it is a good idea. The problem is that there is no feedback given if the storage fails to mount. Because of that, it would be easy to set up a cloud storage where the credentials do not work and not see the fact that the credentials are wrong anymore.

After trying to use credentials, it seems that you can save the credentials anyway during the start session flow. It seems inconsistent to allow saving failing credentials there but not from the data connector edit form.

leafty (changes requested):
It seems that updating credentials is not possible from the data storage panel:

Screenshot 2024-07-15 at 15 10 28

leafty (commented):
Code looks OK otherwise.

ciyer:
> It seems that updating credentials is not possible from the data storage panel:

It is a two step process. You need to clear first, then you can set. This is not ideal and should be improved in follow up work on this issue, but it was ok'd by prod as a shortcut to get this feature to users sooner rather than later.

leafty (approved):
Can be merged in the feature branch :+1:

PR #3233, opened 2024-07-08
fix: do not hide secret value input field

### LINKED ISSUE #3226: User secrets do not get hidden in the input form
On Firefox 127.0.1 on Mac. Renku 0.54.0 on dev.renku.ch

I get the following when I enter a user secret:
image

It seems to occur sometimes. Not always. I think to reproduce it you have to very quickly try to add a secret as soon as the page loads.

When I click the hide/show button after I see the above thing the slight shadow around the text is removed:
image)

Either way the text is always visible and it cannot be hidden.

I think the problem may be that the blurring is somehow rendered behind the text rather than in front?

leafty:
Fixes #3226.

Screenshot 2024-07-09 at 14 42 56
Screenshot 2024-07-09 at 14 43 00

/deploy #notest


ciyer (commented):
I tested this, and it works as advertised, but I have a comment.

This change causes the secrets to interact with my password manager, and I'm not sure that is what we want. Before, the field was treated as just a regular text field, and therefore ignored by the password manager, and I personally prefer that UX. I'm assuming that was the motivation for the initial implementation.

leafty:
> I tested this, and it works as advertised, but I have a comment.
> 
> This change causes the secrets to interact with my password manager, and I'm not sure that is what we want. Before, the field was treated as just a regular text field, and therefore ignored by the password manager, and I personally prefer that UX. I'm assuming that was the motivation for the initial implementation.

In that case we would simply have a text area field and the value would be visible (i.e. we remove the clumsy blur filter). That is how some apps handle secrets.

andre-code (approved):
The code looks good and works as expected. 👌

PR #3238, opened 2024-07-11
feat!: add a convenient way to add an openBIS cloud storage

olloz26:
BREAKING CHANGE: requires changes in the renku-data-services and renku-notebooks repositories

### DIFF
--- a/client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
+++ b/client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
@@ -94,7 +94,6 @@ export default function AddOrEditCloudStorage({
   if (ContentByStep)
     return (
       <>
-        <AddStorageAdvancedToggle state={state} setState={setState} />
         <AddStorageBreadcrumbNavbar state={state} setState={setState} />
         <ContentByStep
           schema={schema}
@@ -124,9 +123,6 @@ export function AddOrEditCloudStorageV2({
   if (ContentByStep)
     return (
       <>
-        <div className={cx("d-flex", "justify-content-end")}>
-          <AddStorageAdvancedToggle state={state} setState={setState} />
-        </div>
         <ContentByStep
           schema={schema}
           state={state}
@@ -271,7 +267,13 @@ interface AddStorageAdvancedForm {
   sourcePath: string;
   configuration: string;
 }
-function AddStorageAdvanced({ storage, setStorage }: AddStorageStepProps) {
+
+function AddStorageAdvanced({
+  setStorage,
+  setState,
+  state,
+  storage,
+}: AddStorageStepProps) {
   const {
     control,
     formState: { errors },
@@ -307,6 +309,7 @@ function AddStorageAdvanced({ storage, setStorage }: AddStorageStepProps) {
 
   return (
     <form className="form-rk-green" data-cy="cloud-storage-edit-advanced">
+      <AddStorageAdvancedToggle state={state} setState={setState} />
       <div className="mb-3">
         <Controller
           name="sourcePath"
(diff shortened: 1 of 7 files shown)


client/src/features/project/components/cloudStorage/projectCloudStorage.types.ts
```diff
@@ -119,6 +119,8 @@ export interface CloudStorageSchema {
   hide?: boolean;
   prefix: string; // ? weird naming; it's the machine readable name
   position?: number;
+  convenientMode?: boolean; // ? Disables the advanced mode
```
lorenzo-cavazzi, line 122: I don't think we should include this; even if it makes little sense for users to mess with the config directly, they might still want to copy/paste the configuration string to pass it to other users to simplify setting up a working solution. I think we should remove it.

client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
```diff
@@ -1043,6 +1065,14 @@ function AddStorageMount({ setStorage, storage }: AddStorageStepProps) {
     setStorage({ ...getValues() });
   };
 
+  const selectedSchema = useMemo(
+    () => getSchemaStorage(schema, !state.showAllSchema, storage.schema),
+    [schema, state.showAllSchema, storage.schema]
+  ).find((s) => s.prefix === storage.schema);
+  if (selectedSchema && selectedSchema.readOnly) {
+    storage.readOnly = true;
+  }
```
lorenzo-cavazzi, line 1074: Is the idea to allow _only_ read-only in every case? How do users change data there?

If that's the case, we should probably improve the UX because, with the current solution, we show a ticked `read-only` box that looks like an element users can interact with, but clicking doesn't do anything. I suggest:

* Making the readOnly input... "readOnly" -- sorry for the pun, I mean that the element `<input id="readonly [...] />` should have something like `<input id="readonly [...] readOnly={selectedSchema.readOnly ?? false} />`
* Show a warning mentioning that there is no way to write with that external storage.

As I write these lines, I wonder why forcing the read-only flag here -- shouldn't that be down to the user's permissions? I assume _some_ users can write.

client/src/features/project/components/cloudStorage/projectCloudStorage.constants.ts
```diff
@@ -62,6 +62,11 @@ export const CLOUD_STORAGE_OVERRIDE = {
         "WebDAV compatible services, including PolyBox and SwitchDrive",
       position: 2,
     },
+    openbis: {
```
lorenzo-cavazzi, line 65: This warning is easy to address: you can add `openbis` to the array `"skipWords"` in the lining configuration in "client/.eslintrc.json"

lorenzo-cavazzi (commented):
Thank you for the PR! Can't wait to try it with a working backend :)

I added a couple of comments inline.

lorenzo-cavazzi:
@olloz26 Sorry for the delay; I'll review this soon.
Do we have a backend for this? Could you point me to the PRs on data-service and renku-notebooks?

P.S. It would be great if you could resolve the conflicts

lorenzo-cavazzi:
@olloz26 any update on this?

lorenzo-cavazzi:
Discussed changes:
* Drop the convenientMode and check whether there are advanced options to show/hide the switch
* Add a "required" label next to the required field and improve the popup message on the Test Connection button whenever required fields are missing
* Rename `readOnly` to `forceReadOnly` and check that also the code for editing existing data connectors considers this setting

lorenzo-cavazzi (approved):
The UI side works well. Great to see this feature getting ready for the release 🙌 

As discussed, let's merge into a feature branch (I've already changed the base) and keep it there until the backend is ready as well.

P.S. when you merge, please squash all the commits into a single commit -- the text looks good already in the title 👍

PR #3248, opened 2024-07-19
feat: add playful 404 and error pages

### LINKED ISSUE #3235: Playful 404 and error pages
# Description

We have [a design proposal](https://www.figma.com/design/2vGPq24rAU15EydWpj94cf/High-fidelity-design?node-id=740-10041) for a new, more playful, 404 page.

Screenshot 2024-07-10 at 21 07 02

Here is the 404 as an SVG image

404-crying-frog

And we have some content for the Error page as well.

Ooops-frog

P.S. mind the buttons and the other visual elements still used outdated colors. Double check the `fill` color is using our new primary color

andre-code:
Playful 404 and error pages

Cases to test:
Cases current Renku
- Not found page https://renku-ci-ui-3248.dev.renku.ch/datf
- Not found dataset  https://renku-ci-ui-3248.dev.renku.ch/datasets/saFxX5caie56yqGhJr6BPdngHywJDQJhUbPg9EN2EMf93XZG18oWeUVJhqPYM6jo7hToDt9YJpA63AVsfc3YgJ7mKGNiAMuZmcZrxM46KRkC3oll
- Not found project https://renku-ci-ui-3248.dev.renku.ch/projects/dandrea.cordoba/dsddsdsdfff
- Error Page

Cases Renku 2.0
- Not found page https://renku-ci-ui-3248.dev.renku.ch/v2/fff
- Not found project https://renku-ci-ui-3248.dev.renku.ch/v2/projects/dalatinrofrau/ml-regression-for-wine-qualityold
- Not found user https://renku-ci-ui-3248.dev.renku.ch/v2/users/dalatinrofrauold
- Not found group https://renku-ci-ui-3248.dev.renku.ch/v2/groups/sdsc-academic-team3
- Error page 

Screenshots
NOT FOUND GROUP

Screenshot 2024-07-26 at 14 21 37
Screenshot 2024-07-26 at 14 21 18

Screenshot 2024-07-26 at 14 20 44
Screenshot 2024-07-26 at 14 20 29

fix #3235 

/deploy


lorenzo-cavazzi (commented):
This looks nice!
A couple of comments before digging deep into the code:

* Some content is flashing on the top left when opening a link like the following (see gif)
https://renku-ci-ui-3248.dev.renku.ch/projects/fakename
Peek 2024-07-29 14-55

* The Error page is blank, except for the footer. I would suggest removing it so that we minimize the probably of crashing the error page itself.
The content is also centered vertically, while it's not for the other error pages; I suggest keeping it consistent, adding only some margin on top
image

* The custom CSS classes should be limited to the bare minimum and avoided wherever unnecessary. E.G. We can use the bootstrap `w-100` instead of `width: 100%;`

lorenzo-cavazzi:
P.S. Sorry I missed that this PR was still a draft :facepalm: 
I hope the comments help anyway

andre-code:
Thanks, @lorenzo-cavazzi, for your helpful suggestions.

Regarding the flashing information, I'm unsure how to address it. The info is the loading message indicating what is being searched for, but since, in the best case, this query is resolved quickly, it shows our 'not found' page in seconds. There isn't enough time to read the loading message. 

So, I am not sure whether to delay the query response or leave it as it is. Do you have a suggestion for this case?
I've already included your other suggestions in the latest commit.

client/src/features/projectsV2/notFound/ProjectNotFound.tsx
```diff
@@ -68,42 +69,52 @@ export default function ProjectNotFound({ error }: ProjectNotFoundProps) {
...
-          </p>
-
-          <div>
-            <Link
-              to={ABSOLUTE_ROUTES.v2.projects.root}
-              className={cx("btn", "btn-outline-primary")}
+    <div className={cx("d-flex", "justify-content-center", "w-100")}>
+      <div className={cx(styles.errorNotFoundContainer, "m-auto")}>
+        <Row>
+          <Col className={cx("p-4", "mt-5")}>
```
lorenzo-cavazzi, line 75: There is an interesting combination of classes here :smile: 

* The custom style interferes with automatic margins forcing the content to a specific width so that on larger screens there is an additional white space on the right.
image
I suggest dropping that div entirely, together with the fixed width, and let `m-auto` do its job.
What's the reason for the fixed width? If you wish to avoid wrapping, you could use `text-nowrap`.
* There is an additional "row-col" layer that feels unnecessary here. Perhaps that's a leftover after trying a multi-column layout?
* There are multiple classes that are adding space to the top. The final result is a space between the top border and the first component that is a combination of `mt-5` + `p-4` + `mt-3`, ending up with a semi-random number of pixels. I suggest avoiding this pattern because it makes keeping spacing consistent throughout the application much harder.
In this case, you could remove the classes from the children components and let the topmost component handle also the space on top.

lorenzo-cavazzi: Please double-check the other pages that implement a similar pattern. This is used in several places in this PR

lorenzo-cavazzi (changes requested):
Great to see these components being standardized! :raised_hands: 

I noticed a few sub-optimal solutions (see inline comments -- and please feel free to justify your changes if you feel I missed the reason).
Even tho they mostly do not impact the users, we should try to avoid bad patterns we used in the past (like prop drilling, inconsistent spaces, unnecessary abstraction layers) so that it's easier to maintain the code in the long term :slightly_smiling_face:

andre-code:
Thanks @lorenzo-cavazzi for your review! I've applied your suggestions and also included the styles for non-authenticated users in v2 when want to create a project or group.

lorenzo-cavazzi:
@andre-code looks good now! There is something off with "oops" images used for the error pages. In my case, they don't show up.

Screenshot_20241015_091835

It's probably just something with the `svg` attributes cause I see the image in my file explorer, but when I try to open it with Firefox it doesn't show up and I see the following message:

> `This XML file does not appear to have any style information associated with it.

client/src/error-boundary/ErrorBoundary.tsx
```diff
@@ -43,44 +44,64 @@ export function AppErrorBoundary({ children }: AppErrorBoundaryProps) {
...
   return (
-    <Sentry.ErrorBoundary onError={onError} fallback={ErrorPage}>
+    <Sentry.ErrorBoundary onError={onError} fallback={fallbackErrorPage}>
       {children}
     </Sentry.ErrorBoundary>
   );
 }
 
-function ErrorPage() {
+const ErrorPage = () => {
```
leafty, line 56: Use the `function` coding style here.

client/src/error-boundary/ErrorBoundary.tsx
```diff
@@ -43,44 +44,64 @@ export function AppErrorBoundary({ children }: AppErrorBoundaryProps) {
     }
   }, []);
 
+  const fallbackErrorPage = <ErrorPage />;
+
   return (
-    <Sentry.ErrorBoundary onError={onError} fallback={ErrorPage}>
+    <Sentry.ErrorBoundary onError={onError} fallback={fallbackErrorPage}>
```
leafty, line 50: I'm not sure I get why you cannot just use `ErrorPage` here.

leafty (changes requested):

leafty (approved):

lorenzo-cavazzi (approved):
Lgtm :rocket:

PR #3257, opened 2024-07-26
feat: Update redirect page styles

andre-code:
PR to apply new design for redirect page.

redirect page

/deploy


leafty (commented):
Imho, we should redirect immediately, removing the need for this page altogether.

andre-code:
Note:
We will wait to determine if the redirect is still necessary. @leafty will work on removing that page in renku-gateway first.

client/public/static/public/theme.css
```diff
@@ -50,43 +50,129 @@ body {
...
 }
 
 .content {
-  background-position: center;
-  background-size: cover;
-  /* calc 100vh - footer height - header height */
   height: calc(100vh - 75px - 70px);
-  background-image: url("./img/404@0.25x.png");
-  color: #d26a98;
+  color: #000;
```
leafty, line 67: nitpick:

```suggestion
  color: black;
```

client/public/static/public/theme.css
```diff
@@ -50,43 +50,129 @@ body {
...
+}
+
+.content {
+  background-image: url("./img/redirect/frog/X-Large_5.png");
+  background-position: center;
+  background-size: inherit;
+  background-repeat: no-repeat;
+  overflow: auto;
+  width: 100%;
+  animation: changeBackgroundXL 5s step-end;
```
leafty, line 152: The animation stops after 5 seconds, but the page may be displayed for longer:

```suggestion
  animation: changeBackgroundXL 5s step-end infinite;
```

leafty (dismissed):
LGTM, some comments here:

leafty (approved):

PR #3258, opened 2024-07-29
feat: display session details in session page Renku 2.0

andre-code:
PR to address small issues in Renku v2

Issues:
- Display Launcher names in session page
- When I go to Resources in a Renku 2.0 session, I see the cheat sheet for the Renku CLI (this should be removed)

What is included in this PR:

- Show the project name and launcher name in the session view navbar title.
- Open a modal with project name and session details when the title in the navbar is clicked.
- Remove the Resources Modal, so only session log modal is displayed on the session page.

session view changes

/deploy

### DIFF
--- a/client/src/features/sessionsV2/ShowSessionPage.tsx
+++ b/client/src/features/sessionsV2/ShowSessionPage.tsx
@@ -476,8 +476,6 @@ function SessionDetails({
             className={cx(
               "d-block",
               "d-lg-flex",
-              "pb-2",
-              "pb-lg-0",
               "gap-2",
               "align-items-center"
             )}
--- a/client/src/features/sessionsV2/ShowSessionPage.tsx
+++ b/client/src/features/sessionsV2/ShowSessionPage.tsx
@@ -18,33 +18,58 @@
 
 import cx from "classnames";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
-import { ArrowLeft, Journals, PauseCircle, Trash } from "react-bootstrap-icons";
+import {
+  ArrowLeft,
+  Box,
+  Briefcase,
+  Clock,
+  Cloud,
+  FileEarmarkText,
+  PauseCircle,
+  Trash,
+} from "react-bootstrap-icons";
 import {
   Link,
   generatePath,
   useNavigate,
   useParams,
 } from "react-router-dom-v5-compat";
-import { Button, UncontrolledTooltip } from "reactstrap";
+import {
+  Button,
+  Modal,
+  ModalBody,
+  ModalHeader,
+  UncontrolledTooltip,
+} from "reactstrap";
 
+import RenkuFrogIcon from "../../components/icons/RenkuIcon";
 import { User } from "../../model/renkuModels.types";
-import { SESSION_TABS } from "../../notebooks/Notebooks.present";
+import { ABSOLUTE_ROUTES } from "../../routing/routes.constants";
 import useLegacySelector from "../../utils/customHooks/useLegacySelector.hook";
 import useWindowSize from "../../utils/helpers/UseWindowsSize";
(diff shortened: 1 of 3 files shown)


client/src/features/sessionsV2/ShowSessionPage.tsx
```diff
@@ -346,3 +364,138 @@ function DeleteSessionBtn({ openDeleteSession }: DeleteSessionBtnProps) {
...
+
+  const projectUrl =
+    project &&
+    generatePath(ABSOLUTE_ROUTES.v2.projects.show.root, {
+      namespace: project?.namespace,
+      slug: project?.slug,
+    });
+
+  if (isLoadingLaunchers || isLoadingProject) return <Loader />;
+  if (launchersError || !launcher) return <i>Session not accessible</i>;
```
lorenzo-cavazzi, line 417: This can also be improved
image

I suggest putting it into a text element (like a `<p>`) styling it to have no bottom margin, white text color, and italics. Optional: add a warning icon at the beginning of the text.

client/src/features/sessionsV2/ShowSessionPage.tsx
```diff
@@ -346,3 +364,138 @@ function DeleteSessionBtn({ openDeleteSession }: DeleteSessionBtnProps) {
...
+  const detailsModal = project && session && projectUrl && (
+    <Modal backdrop="static" centered isOpen={isOpen} size="lg" toggle={toggle}>
+      <ModalHeader toggle={toggle}>Session details {launcher.name}</ModalHeader>
+      <ModalBody>
```
lorenzo-cavazzi, line 421: The spacing is a little messed up here: you always add the class `mb-2` to the divs, but sometimes they contain a single `p` child with a bigger bottom margin (equivalent to `mb-3`) making the class useless, while in other cases the margin is applied but other components add additional vertical margins (see `CommandCopy`).

I suggest:
* Remove the margins within the child components. Depending on the case, you might drop the `div` entirely leaving a `p className="mb-0"`, or drop the `mb-2` on the div, or use properties to remove the margins like in `<CommandCopy noMargin ... />`.
* Handle the spacing between the children at the parent component by adding `d-flex, flex-column, gap-3`

lorenzo-cavazzi (changes requested):
Nice changes!

The code looks overall good, but there are a few details that require adjustments before merging -- see inline comments.

andre-code:
Thank you, @lorenzo-cavazzi, for your review and suggestions. I've applied those changes in the latest commit. They make a lot of sense!

client/src/features/sessionsV2/ShowSessionPage.tsx
```diff
@@ -362,3 +380,153 @@ function DeleteSessionBtn({ openDeleteSession }: DeleteSessionBtnProps) {
...
+          </div>
+          <div
+            className={cx(
+              "d-block",
+              "d-lg-flex",
+              "pb-2",
+              "pb-lg-0",
+              "gap-2",
+              "align-items-center"
+            )}
```
lorenzo-cavazzi, line 483: No need to add additional space on the bottom since the spacing is already handled by the parent component with `gap-3`

```suggestion
            className={cx(
              "d-block",
              "d-lg-flex",
              "gap-2",
              "align-items-center"
            )}
```

lorenzo-cavazzi (changes requested):
Looks nice! Just a tiny detail before merging

client/src/features/sessionsV2/ShowSessionPage.tsx
andre-code: thanks @lorenzo-cavazzi I have removed the unnecessary style

lorenzo-cavazzi (approved):
:rocket:

PR #3259, opened 2024-07-30
refactor: use consistent sizes for side columns

lorenzo-cavazzi:
This makes the size of the side columns consistent throughout the application

image

/deploy


leafty (approved):
✨

PR #3261, opened 2024-07-30
design - search page

lorenzo-cavazzi:
Rework the design on the search page

/deploy


lorenzo-cavazzi:
Reminder to self: this is still missing the entity icons on the search cards

image

client/src/features/searchV2/components/SearchV2Results.tsx
```diff
@@ -104,9 +119,15 @@ function SearchV2ResultsContent() {
...
+            <>
+              {" "}
+              for <span className="fw-bold">{`"${query}"`}</span>
+            </>
+          )}
         </p>
-        <p>You can try another search, or change some filters.</p>
+        {query && <p>You can try another search, or change some filters.</p>}
       </>
     );
```
ciyer, line 132: Suggestion:
```
    return query == null ? (
      <p>No results</p>
    ) : (
      <>
        <p>
          No results for <span className="fw-bold">{`"${query}"`}</span>
        </p>
        <p>You can try another search, or change some filters.</p>
      </>
    );
```

ciyer (changes requested):
There is a problem with the deployment, so I could not try it out in a running environment but played with it in cypress, and I think it looks very good!

I have one minor suggestion for clearer separation of branches in the code, but that is all.

lorenzo-cavazzi:
@ciyer thank you for the review! I re-deployed and it should now be accessible

ciyer (approved):
👍

PR #3263, opened 2024-08-02
fix: unify texts when deleting a session Renku 2.0

andre-code:
PR to remove copy text Inconsistencies referring to stopping/deleting sessions.

/deploy


leafty (approved):
Note: renamed PR to `fix`

PR #3264, opened 2024-08-02
[WIP] feat: set namespace by default in new project form

andre-code:
PR my namespace by default

/deploy #notest


client/src/features/projectsV2/fields/ProjectNamespaceFormField.tsx
```diff
@@ -43,9 +43,10 @@ import {
   useLazyGetNamespacesQuery,
 } from "../api/projectV2.enhanced-api";
 
+import { useDispatch } from "react-redux";
```
leafty, line 46: Use `useAppDispatch` instead (this hook is typed and only accepts valid operations).

client/src/features/projectsV2/fields/ProjectNamespaceFormField.tsx
```diff
@@ -43,9 +43,10 @@ import {
...
+import { NamespaceResponseList } from "../api/namespace.api.ts";
```
leafty, line 47:
```suggestion
import { NamespaceResponseList } from "../api/namespace.api";
```

client/src/features/projectsV2/fields/ProjectNamespaceFormField.tsx
```diff
@@ -292,18 +293,33 @@ function ProjectNamespaceControl(props: ProjectNamespaceControlProps) {
...
+      const userNamespace = namespaces.filter(
+        (namespace) => namespace.namespace_kind === "user"
+      );
+
+      if (userNamespace.length && !value) {
+        onChange(userNamespace[0]);
+      }
+    },
+    [onChange, value]
+  );
```
leafty, line 307: The user's namespace will **always** be returned as the first element (on the first page).

Also, use `find()` on arrays instead of `filter()` when finding an element.

The whole set of code changes can be simplified to:
```tsx
useEffect(() => {
  if (namespacesFirstPage == null) {
    return;
  }
  const userNamespace = namespaces.find(
    (namespace) => namespace.namespace_kind === "user"
  );
  if (userNamespace != null && !value) {
    onChange(userNamespace);
  }
}, [onChange, namespacesFirstPage, value])
```

leafty (changes requested):

andre-code:
Thank you, @leafty, for your review. I've incorporated your suggestions.

client/src/features/projectsV2/fields/ProjectNamespaceFormField.tsx
```diff
@@ -313,19 +322,25 @@ function ProjectNamespaceControl(props: ProjectNamespaceControlProps) {
...
-        namespacesPageResult.currentData.page,
+      hasMore,
       currentRequestId: "",
     });
   }, [
     allNamespaces,
     namespacesPageResult.currentData,
     namespacesPageResult.requestId,
     currentRequestId,
+    onChange,
```
leafty, line 343: I think this is a leftover.

```suggestion
```

leafty (commented):
Looks OK, please check the `e2e` tests.

andre-code:
> Looks OK, please check the `e2e` tests.

sure, done!

ciyer (approved):
LGTM!

PR #3268, opened 2024-08-08
design - update semantic colors

lorenzo-cavazzi:
New semantic colors.

The info color is less bright

image

The `success` color is lighter, with stronger differentiations from the `primary` color

image

/deploy #notest

### DIFF
--- a/client/src/styles/renku_bootstrap_customization.scss
+++ b/client/src/styles/renku_bootstrap_customization.scss
@@ -5,11 +5,12 @@
 @import "./fonts/fonts";
 $font-family-base: "Inter", sans-serif;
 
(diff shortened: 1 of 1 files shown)


leafty (commented):
Code is LGTM, but I have to personally say that the old info blue is nicer to my eyes.

leafty (approved):

lorenzo-cavazzi:
> Code is LGTM, but I have to personally say that the old info blue is nicer to my eyes.

Noted :+1: 

I'm merging this, but we can do a final round on the colors. We plan to change the primary color and start using the secondary color (different from the current one) but that's postponed for now. That might be the best moment to settle on semantic colors.

PR #3269, opened 2024-08-08
design - remove the "card in card" elements when selecting a resource class in the Environments modal

lorenzo-cavazzi:
This PR removes the additional card for the resource class nested in the environment selection modal.

image

/deploy


andre-code (changes requested):
Thanks for removing the card from the resource selector! ✨The new design looks cleaner, but I noticed a couple of challenges that might impact the user experience:

Mobile mode: when the environment list is long enough to require scrolling, the resource selector at the end of the list can easily be missed. This could result in the resource class being completely hidden, making it difficult for users to realize they can select a different resource class.
Screenshot 2024-08-12 at 10 30 51

Desktop Mode: the scrolling behavior on desktop is a bit off. The scroll bar for the modal appears below the modal (as part of the page), when ideally, it should be within the modal itself. Although users can still access the resource class by scrolling to the bottom to reach the save button, it would be more user-friendly to reduce the need for scrolling, especially with longer lists.

select environment list

Perhaps moving the resource class selector to the top would make it more visible and convenient? Or, we could consider placing it back inside the environment card and making the save button always visible, even on desktop. This way, we ensure a smoother user experience across devices.

lorenzo-cavazzi:
Thank you @andre-code for trying with many environments. The resource class selectors might really get lost in that case. I like your proposal to move it to the top. I did that in https://github.com/SwissDataScienceCenter/renku-ui/pull/3269/commits/26e14b260434a88c3669e53796d379db5558c2d6

For the scroll behavior, that's consistent with the other modals to add resources. This was the only different one; as a side effect, the dropdown from the class selector was expanding the height of the modal with a poor UX in some cases (E.G. long list of resource classes and few environments).
Now that the resource class is on top, I think it's not too confusing even when we have a lot of environments. In that case, the modal looks "non-closed" on the bottom and users don't see the buttons, so scrolling down should be intuitive.

andre-code:
Yes, the resource class on top improves the experience, however having to scroll all the list to confirm the selection is not ideal for UX. I'm not sure what you mean by "other modals to add resources," but a good UX example in this case is how the "Add a Data Source" modal works.

scroll in modal

lorenzo-cavazzi:
@andre-code I see your point, but I wonder if the visualization you suggest is better for the UX.
This is how I see the scrollable modal on my screen with the content you added
SCREENo2
Would you guess there are more environments than the ones shown?
It feels more obvious with the non-scrollable modal
SCREEN01

I admit this combination is a little unlucky since the height almost matches the rows :grin: 
And I previously compared the modal with others that are unlikely to grow vertically to the point that it adds a scrollbar (E.G. adding a repository). Your example of the data sources is a better comparison :+1: 
P.S. also the problem with the dropdown I mentioned earlier is mostly irrelevant now that the dropdown for the class is on top and we have more content below.

I made the changes you suggested in https://github.com/SwissDataScienceCenter/renku-ui/pull/3269/commits/6a45fdeb7930f021f277d749f81e1ad8899dd738 ;  even if we probably won't have many environments listed there, the scrollbar feels better than my previous solution :+1:

andre-code (approved):
lgtm 🚀

PR #3275, opened 2024-08-21
Set default to read only in add data source

ciyer:
Change default to read-only when a data source is added. See 

* https://www.notion.so/renku/When-I-add-a-data-connector-the-default-should-be-read-only-with-the-read-write-an-optional-overrid-716f02eb42354f709a1cfb1607a0d447

/deploy renku=ciyer/ui-3275-data-source


leafty (dismissed):
LGTM thanks

leafty (approved):

PR #3277, opened 2024-08-22
feat: add features for GitHub connections

leafty:
WIP

See also: https://github.com/SwissDataScienceCenter/renku-data-services/pull/361


lorenzo-cavazzi:
Included in #3332

PR #3295, opened 2024-09-02
fix: use default storage when launch a session

### LINKED ISSUE #3291: Renku 2.0: Use resource class default disk size when launching sessions
## Description

Currently, when launching a session, the disk value is always set to 1GB, ignoring the default disk value specified in the resource class.

andre-code:
fix #3291 

/deploy renku=release-0.57.0


ciyer (approved):
I tested this and compared it to a different deployment. In the other deployment, I see 4G as the storage size, in this PR, I see 1G as the storage size. I assume this is the only difference I should notice, and if that is the case, this looks good!

If there is something else I should test, let me know, but assuming there is not, I am approving this.

andre-code:
> I tested this and compared it to a different deployment. In the other deployment, I see 4G as the storage size, in this PR, I see 1G as the storage size. I assume this is the only difference I should notice, and if that is the case, this looks good!
> 
> If there is something else I should test, let me know, but assuming there is not, I am approving this.

Thanks for reviewing this, @ciyer. The main improvement is that the default resource class storage size is now used when launching a session, instead of the previously fixed 1GB. Additionally, the resource class selector now shows the default storage size instead of the maximum storage.

PR #3296, opened 2024-09-02
feat: use data-service sessions endpoint

andre-code:
PR to use data-service `/sessions/*` endpoint in Renku V2.

### What's Included
- Launch sessions with a custom or global environment
- Open sessions in the session view
- See the current session state
- Start, resume, hibernate, and shut down a session

### What's Not Included ‼️
- Automatic injection of saved cloud storage secrets
- Displaying an error message when a session has an error

### Pending 
- [ ] Update e2e tests

/deploy renku=feat-jupyter-free-sessions renku-data-services=release-amaltheas-migration amalthea-sessions=main renku-notebooks=master renku-gateway=1.1.0 extra-values=amalthea-sessions.deployCrd=false


client/src/components/LogsV2.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Sessions logs container integrating state and actions V2
+ *
+ * @param {string} name - server name
+ */
+interface EnvironmentLogsPropsV2 {
+  name: string;
+}
+export const EnvironmentLogsV2 = ({ name }: EnvironmentLogsPropsV2) => {
```
leafty, line 34: `function` style -> https://github.com/SwissDataScienceCenter/renku-ui/blob/main/CODING_GUIDELINES.md#r004-include-a-default-export-when-appropriate

client/src/components/LogsV2.tsx
```diff
@@ -0,0 +1,96 @@
...
+      >
+        <div>Logs</div>
+      </ModalHeader>
+      <ModalBody>
+        <div className="mx-2">
+          <SessionLogs fetchLogs={fetchLogs} logs={logs} name={name} />
+        </div>
+      </ModalBody>
+    </Modal>
+  );
```
leafty, line 95: Why is this split into another component here? This code can be in the previous component and that saves all the `interface` and `const =` lines of code.

client/src/features/admin/AddSessionEnvironmentButton.tsx
```diff
@@ -86,13 +86,13 @@ function AddSessionEnvironmentModal({
         addSessionEnvironment({
           container_image: data.container_image,
           name: data.name,
-          default_url: data.default_url.trim() ? data.default_url : undefined,
-          description: data.description.trim() ? data.description : undefined,
+          default_url: data.default_url?.trim() ? data.default_url : undefined,
```
leafty, line 89: For here and everywhere else: should we instead have this?

```suggestion
          default_url: data.default_url?.trim() || undefined,
```

Otherwise we send `untrimmed` strings to the API.

This can be added to a new issue to do later.

client/src/features/sessionsV2/PauseOrDeleteSessionModal.tsx
```diff
@@ -281,12 +273,13 @@ function PauseSessionModalContent({
...
           paused.
         </p>
-        {hibernatedSecondsThreshold > 0 && (
-          <InfoAlert dismissible={false} timeout={0}>
-            Please note that paused session are deleted after{" "}
-            {hibernationThreshold} of inactivity.
-          </InfoAlert>
-        )}
+        {session?.status?.will_hibernate_at &&
+          session?.status?.will_hibernate_at?.length > 0 && (
```
leafty, line 277: No need for a length check on strings:

```suggestion
        {session?.status?.will_hibernate_at && (
```

client/src/features/sessionsV2/PauseOrDeleteSessionModal.tsx
```diff
@@ -281,12 +273,13 @@ function PauseSessionModalContent({
...
+            <InfoAlert dismissible={false} timeout={0}>
+              Please note that paused session are deleted after{" "}
+              {hibernationThreshold} of inactivity.
```
leafty, line 280: This is not an accurate text copy because `session?.status?.will_hibernate_at` refers to this specific session. Also the text refers to `"DELETE"` while the threshold is for `"PAUSE"`.

client/src/features/sessionsV2/PauseOrDeleteSessionModal.tsx
```diff
@@ -257,22 +258,13 @@ function PauseSessionModalContent({
...
-    ? toHumanDuration({ duration })
-    : "a period";
-
+  const now = DateTime.utc();
+  const hibernationThreshold = session?.status?.will_hibernate_at
+    ? toHumanRelativeDuration({
+        datetime: session?.status?.will_hibernate_at,
+        now,
+      })
+    : 0;
```
leafty, line 267: To make things correct, it is preferable to use `<TimeCaption>` instead of computing this. `<TimeCaption>` supports passing class names and the `noCaption` attribute to not have the `time-caption` class.

client/src/features/sessionsV2/session.utils.ts
```diff
@@ -222,3 +224,58 @@ export function isValidJSONStringArray(
...
+    }
+
+    if (parsedUrl.protocol === "http:") {
+      return url.replace("http:", "https:");
+    }
+
+    throw new Error("Unsupported protocol");
+  } catch (error) {
+    if (url && !url.includes("://")) {
+      return `https://${url}`;
```
leafty, line 277: What if this is not a valid URL?

client/src/features/sessionsV2/SessionShowPage/SessionIframe.tsx
```diff
@@ -0,0 +1,72 @@
...
+    : {};
+};
+
+interface SessionIframeProps {
+  height: string;
+  isSessionReady: boolean;
+  session: SessionV2;
+}
+
+export default function SessionIframe({
```
leafty, line 40: New issue: re-factor into existing `<SessionJupyter>` component.

client/src/features/sessionsV2/SessionShowPage/ShowSessionPage.tsx
```diff
@@ -88,12 +85,15 @@ export default function ShowSessionPage() {
     slug: slug ?? "",
   });
 
-  const { data: sessions, isLoading } = useGetSessionsQuery();
+  const { data: sessions, isLoading } = useGetSessionsQuery(undefined, {
+    skip: false,
```
leafty, line 89: ?

leafty (commented):
General note: please use the `function` style for components.

A first batch of comments below:

client/src/features/sessionsV2/components/SessionButton/ActiveSessionButton.tsx
```diff
@@ -577,12 +556,12 @@ function ModifySessionModalContent({
   useEffect(() => {
     const currentSessionClass = resourcePools
       ?.flatMap((pool) => pool.classes)
-      .find((c) => c.id === currentSessionClassId);
+      .find((c) => `${c.id}` == resource_class_id);
```
ciyer, line 559: I think this would be more idiomatic.

```suggestion
      .find((c) => ""+ c.id == resource_class_id);
```

client/src/features/sessionsV2/components/SessionButton/ActiveSessionButton.tsx
```diff
@@ -615,28 +594,26 @@ function ModifySessionModalContent({
...
             disabled={
               isLoading ||
               !resourcePools ||
               resourcePools.length == 0 ||
               isError ||
               currentSessionClass == null ||
-              (currentSessionClassId != null &&
-                currentSessionClassId === currentSessionClass?.id)
+              (resource_class_id != null &&
+                resource_class_id === `${currentSessionClass?.id}`)
```
ciyer, line 616: Since this logic is used in a couple of places:

```suggestion
resourceClassMatchesCurrent(resource_class_id, currentSessionClass)
```

```
function resourceClassMatchesCurrent(resourceClassId: string | undefined, currentSessionClass: string) {
    return (resourceClassId != null && resourceClassId === ""+currentSessionClass.id)
}
```

client/src/features/sessionsV2/components/SessionButton/ActiveSessionButton.tsx
```diff
@@ -646,15 +623,15 @@ function ModifySessionModalContent({
...
           disabled={
             isLoading ||
             !resourcePools ||
             resourcePools.length == 0 ||
             isError ||
             currentSessionClass == null ||
-            (currentSessionClassId != null &&
-              currentSessionClassId === currentSessionClass?.id)
+            (resource_class_id != null &&
+              resource_class_id === `${currentSessionClass?.id}`)
```
ciyer, line 634:
```suggestion
resourceClassMatchesCurrent(resource_class_id, currentSessionClass)
```

client/src/features/sessionsV2/sessionsV2.api.ts
```diff
@@ -110,15 +118,101 @@ const sessionsV2Api = createApi({
       },
       invalidatesTags: ["Launcher"],
     }),
+    getSessions: builder.query<SessionList, void>({
+      query: () => ({ url: "sessions" }),
+      providesTags: (result) =>
+        result
+          ? [
+              ...result.map(({ name }) => ({
```
ciyer, line 126: Question: `name` is used here, but `id` below. Is that correct?

ciyer (changes requested):
Looks very good! I have some small code suggestions and a question about intent, but those are all quite minor. 

We discussed the use of anonymous vs. named functions for the components, and it looks like the project session test is failing, but I'm assuming you are already work on those.

client/src/features/sessionsV2/components/SessionButton/ActiveSessionButton.tsx
leafty: This is a javascript "hack" and should not be used. It uses type coercion so this makes code look weird (see things like `"1"+1 == 11`, etc. `${c.id}` is a valid way to convert a `number` to a `string`.

client/src/features/sessionsV2/sessionsV2.api.ts
leafty: The `name` field is the identifier. Maybe that should be changed in the API?

client/src/features/sessionsV2/components/SessionButton/ActiveSessionButton.tsx
leafty: why not just?
```
resource_class_id === `${currentSessionClass?.id}
```

This is all that's needed here (`resource_class_id` cannot be null-ish and equal a string).

client/src/features/sessionsV2/SessionShowPage/SessionIframe.tsx
andre-code: https://github.com/SwissDataScienceCenter/renku-ui/issues/3366

client/src/features/sessionsV2/sessionsV2.api.ts
andre-code: Yes, the change has been requested. I will implement this in a future PR once the updates are made in the API.

client/src/features/admin/UpdateSessionEnvironmentButton.tsx
```diff
@@ -105,15 +105,11 @@ function UpdateSessionEnvironmentModal({
           environmentId: environment.id,
           container_image: data.container_image,
           name: data.name,
-          default_url: data.default_url.trim() ? data.default_url : "",
-          description: data.description.trim() ? data.description : "",
+          default_url: data.default_url?.trim() ? data.default_url : "",
```
leafty, line 108: It seems this can be done here as well:
```suggestion
          default_url: data.default_url?.trim() || "",
```

client/src/features/dashboardV2/DashboardV2Sessions.tsx
```diff
@@ -1,84 +1,91 @@
...
 import { Col, ListGroup, Row } from "reactstrap";
-
 import { Loader } from "../../components/Loader";
-import { EnvironmentLogs } from "../../components/Logs";
+import EnvironmentLogsV2 from "../../components/LogsV2";
 import { RtkErrorAlert } from "../../components/errors/RtkErrorAlert";
-import { NotebooksHelper } from "../../notebooks";
-import { NotebookAnnotations } from "../../notebooks/components/session.types";
+import { useGetSessionsQuery as useGetSessionsQueryV2 } from "../../features/sessionsV2/sessionsV2.api";
+import "../../notebooks/Notebooks.css";
```
leafty, line 10: Can you keep the CSS imports as the last ones? They semantically different from the other ones, so we should make them stand out. Also keep the associated comment.

leafty (commented):
Should be good to go after these comments.

client/src/features/dashboardV2/DashboardV2Sessions.tsx
```diff
@@ -1,84 +1,111 @@
...
-import { Session } from "../session/sessions.types";
-import { filterSessionsWithCleanedAnnotations } from "../session/sessions.utils";
 import ActiveSessionButton from "../sessionsV2/components/SessionButton/ActiveSessionButton";
 import {
   SessionStatusV2Description,
   SessionStatusV2Label,
 } from "../sessionsV2/components/SessionStatus/SessionStatus";
+import { SessionList, SessionV2 } from "../sessionsV2/sessionsV2.types";
 
-// Required for logs formatting
```
leafty, line 24: Can you keep this comment here? Otherwise there is no indication that removing the `.css` import would break something.

leafty (approved):

PR #3299, opened 2024-09-04
refactor: update Renku 2.0 APIs

leafty:
Update the Renku 2.0 projects API to follow https://github.com/SwissDataScienceCenter/renku-data-services/pull/347.

Also, refactor the API file organization.


leafty:
TODO: re-introduce `"post secrets for session launch"` hack to make this work; add an issue to re-haul state management during session launch, as hacks SHOULD NOT be necessary.

PR #3301, opened 2024-09-09
refactor: separate V1 cloud storage UI elements from V2 data connectors

ciyer:
The old implementation used the `CloudStorageModal` component, which was shared been v1 cloud storages and v2 data sources. These changes introduce a new DataConnectorModal component for use in v2 code and removes v2 references from CloudStorageModal.

/deploy renku=ciyer/ui-3275-data-source


leafty (approved):
Looks good! :clap:

PR #3313, opened 2024-09-12
fix: prevent querying sessions launchers without project id

lorenzo-cavazzi:
When starting a session, the `session_launchers` endpoints is invoked too early when the project id isn't available yet. This prevents it

image

/deploy

### DIFF
--- a/client/src/features/sessionsV2/ShowSessionPage.tsx
+++ b/client/src/features/sessionsV2/ShowSessionPage.tsx
@@ -438,11 +438,11 @@ function SessionDetails({
       </div>
     );
   }
-  if (launchersError || !launcher)
+  if (launchersError || !launcher || !project)
     return (
       <div className={cx("d-flex", "align-items-center")}>
         <p className={cx("text-white", "mb-0")}>
-          <ExclamationTriangle className="bi" /> Session not accessible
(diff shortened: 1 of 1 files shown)


leafty (approved):

andre-code (dismissed):
lgtm

andre-code:
Thanks to @lorenzo for fixing this!

@leafty, could you clarify what you mean by "not well handled"? I see that the buttons displayed are incorrect. Is there anything else we need to address? If so, we should consider handling those issues in a separate PR.

Screenshot 2024-09-13 at 16 08 28

leafty:
> Thanks to @lorenzo for fixing this!
> 
> @leafty, could you clarify what you mean by "not well handled"? I see that the buttons displayed are incorrect. Is there anything else we need to address? If so, we should consider handling those issues in a separate PR.

We can open a new issue for this, but the component doesn't handle properly when its props are null-ish values

lorenzo-cavazzi:
> We can open a new issue for this, but the component doesn't handle properly when its props are null-ish values

That's a fair point. Since we are fixing the component already, I've added a quick solution for handling invalid namespace/slug values.

leafty (approved):
LGTM

Issue #2945, opened 2023-12-14
Further refine the Cloud Storage section


lorenzo-cavazzi:
The last changes to the cloud storage options took longer than expected, leaving behind a few additional optimizations:
- [ ] Refine the options to list to include enumerations (E.G. S3 --> AWS --> Region) as autocomplete elements.
- [ ] Address the optimizations pointed out in the review https://github.com/SwissDataScienceCenter/renku-ui/pull/2908
    - Merge all `useState` into a single RTK slice https://github.com/SwissDataScienceCenter/renku-ui/pull/2908#discussion_r1426482293
    - Do not use additional local states to check the outcome from RTK mutations https://github.com/SwissDataScienceCenter/renku-ui/pull/2908#discussion_r1426485408
    - Improve the logic to watch for changes on the `useForm` fields values https://github.com/SwissDataScienceCenter/renku-ui/pull/2908#discussion_r1426550721
    - Remove unnecessary styles and fall back to bootstrap variables https://github.com/SwissDataScienceCenter/renku-ui/pull/2908#discussion_r1426571311

PR #3315, opened 2024-09-12
feat!: support for data connectors on user and group namespaces

ciyer:
* show/edit data connectors on the group page
* show/edit data connectors to the user page

/deploy renku-data-services=leafty/feat-namespaced-data-connectors-4


client/src/features/projectsV2/api/projectV2.enhanced-api.ts
leafty, line 1: We should not inject into the `projectV2` API. In fact, the PR which updates the API files does split each API cleanly so that each one is handled separately. See: https://github.com/SwissDataScienceCenter/renku-ui/pull/3299.

client/src/features/ProjectPageV2/ProjectPageContent/DataSources/DataSourceModal.tsx
leafty, line 1: It seems that some renaming got reverted here. Names should probably stick to "DataConnector" and the like.

client/src/features/dataConnectorsV2/components/DataConnectorActions.tsx
```diff
@@ -0,0 +1,189 @@
...
+import DataConnectorCredentialsModal from "./DataConnectorCredentialsModal";
+import { ButtonWithMenuV2 } from "../../../components/buttons/Button";
+
+interface DataConnectorDeleteModalProps {
+  dataConnector: DataConnectorRead;
+  isOpen: boolean;
+  onDelete: () => void;
+  toggleModal: () => void;
+}
+function DataConnectorDeleteModal({
```
leafty, line 45: Maybe we can also ask the user to type the slug for deletion? This is what we ask users to do when deleting a project.

client/src/features/dataConnectorsV2/components/DataConnectorModal/DataConnectorModalBody.tsx
```diff
@@ -0,0 +1,572 @@
...
+          help you identify it.
+        </div>
+      </div>
+
+      <div className="mb-3">
+        <Label className="form-label" for="namespace">
+          Owner
+        </Label>
+
+        <Controller
```
leafty, line 328: Why not use the same control as the one used for project creation and edit here?

https://github.com/SwissDataScienceCenter/renku-ui/blob/2945fb9cd21693b3dc52e816ca19a682482412e5/client/src/features/projectsV2/fields/ProjectNamespaceFormField.tsx#L200

client/src/features/dataConnectorsV2/components/DataConnectorModal/index.tsx
```diff
@@ -0,0 +1,521 @@
...
+  }, [createResult, originalToggle, reset, success, validationResult]);
+
+  // Handle unmount
+  useEffect(() => {
+    const cleanup = () => {
+      reset();
+    };
+
+    return cleanup;
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
leafty, line 303: Do not ignore the rules of hooks. Ignoring them produces bugs.

client/src/features/dataConnectorsV2/components/dataConnector.utils.ts
```diff
@@ -0,0 +1,207 @@
...
+  schema?: string;
+  sourcePath?: string;
+};
+
+type DataConnectorOptions = Record<
+  string,
+  string | number | boolean | object | null
+>;
+
+export const EMPTY_DATA_CONNECTOR_FLAT: DataConnectorFlat = {
```
leafty, line 59: Is there really a need for this? `react-hook-form` can handle nested values afaik.

client/src/features/usersV2/show/UserShow.tsx
```diff
@@ -129,6 +133,17 @@ export default function UserShow() {
           }
         />
       </section>
+      <section className="mt-3">
+        <Row>
+          <Col className="order-3" xs={12} xl={8}>
+            <DataConnectorsBox
```
leafty, line 139: As a note, it is very difficult for users to access this page (there is no top-level navigation link for it). Which means it is very difficult to create personal data connectors if we do not have a top-level create menu.

client/src/features/usersV2/show/UserShow.tsx
```diff
@@ -38,6 +40,8 @@ import {
 } from "../../user/dataServicesUser.api";
 import UserAvatar from "./UserAvatar";
 
+// TODO: Add data connectors
```
leafty, line 43:
```suggestion
```

leafty (changes requested):
Thanks!

Some first comments here.

client/src/features/ProjectPageV2/ProjectPageContent/DataSources/DataSourceModal.tsx
ciyer: I reintroduced a DataSourceModal for use on the V2 project pages. This is temporary, until project gets converted to use data connectors, but it is a helpful intermediate step to limit where code changes need to be made.

client/src/features/dataConnectorsV2/components/dataConnector.utils.ts
ciyer: This whole structure is there to maximize compatibility with the existing components for CloudStorage. I agree that the code structure would be better if this were implemented in a "data-connector-native" way, but that would require re-writing further existing components.

leafty (commented):
Looks good, should be good to go with issues fixed.

tests/cypress/e2e/groupV2.spec.ts
```diff
@@ -199,3 +213,147 @@ describe("Edit v2 group", () => {
...
+  });
+
+  it("shows groups", () => {
+    fixtures.readGroupV2().readGroupV2Namespace();
+    cy.contains("List Groups").should("be.visible");
+    cy.contains("test 2 group-v2").should("be.visible").click();
+    cy.wait("@readGroupV2");
+    cy.contains("test 2 group-v2").should("be.visible");
+  });
+});
```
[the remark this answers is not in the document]
ciyer: Good catch -- that was accidentally copied. Fixed in 5e0760d2e05e242567d6259fd3d2ed4a51beb665.

client/src/features/dataConnectorsV2/components/DataConnectorActions.tsx
ciyer: Fixed in 0e58bbb27f44b337c8c3166fe1ee24301b5fdadc.

client/src/features/dataConnectorsV2/components/DataConnectorModal/index.tsx
ciyer: Fixed in e876667a16af6ef150c0499f9a75f949402f845e

leafty: So, this hooks makes the whole form buggy (cannot hit "Test connection"). I wonder why it is needed at all when `reset()` is called when closing the modal with `toggle()`. The whole state in the component is spaghetti code and should use a single source of truth in redux. Note that this does not need to happen in this PR, but the original quality of this code is not up to the standards we should have.

leafty: https://github.com/SwissDataScienceCenter/renku-ui/issues/2945 for reference.

client/src/features/projectsV2/api/projectV2.enhanced-api.ts
ciyer: Fixed in 657286918dee44b94072f6f81921c43f783ad0b8

leafty (approved):

PR #3318, opened 2024-09-16
fix: provide empty auth header

leafty:
Fixes `uiserver` crash looping when Sentry is enabled.

See also: #3319.

/deploy

Tested:
* Running `3.35.0` with Sentry enabled -> crash loop
* Running this version with Sentry enabled -> start and no error logs


andre-code (approved):
lgtm

PR #3323, opened 2024-09-18
feat: data connectors build

ciyer:
feat: data connectors build

/deploy renku-data-services=leafty/build-data-connectors renku-notebooks=leafty/feat-data-connectors


leafty:
There is a bug when starting a session and being prompted for credentials: if I provide credentials and select the "save" option, then there are two POST requests sent to start the session. The second one fails with a `HTTP 500` which is shown as an error to the user. Going back to the project page, the session is starting and is working properly.

image

leafty (approved):
LGTM 👏 This seems to work very well! 🎉

PR #3324, opened 2024-09-18
fix: websockets authentication

olevski:
The target branch is not main.


leafty (commented):
why not include the changes from `main` in `./server`? This hack is NOT how it works with the new API Gateway. (Some of these code files do not exist anymore even)

olevski:
> why not include the changes from main in ./server? This hack is NOT how it works with the new API Gateway. (Some of these code files do not exist anymore even)

I thought that Andrea's branch was up to date with master/main on the ui repo. But I guess that is not the case... I did not know this. I will resolve things with Andrea tomorrow then. I was just looking at the base branch that Andrea is working on.

olevski:
We dont need this now at all.

PR #3329, opened 2024-09-25
feat: handle connected services in the admin panel

lorenzo-cavazzi:
Handle connected services in the admin panel.

image

/deploy

### DIFF
--- a/client/src/features/admin/AddConnectedServiceButton.tsx
+++ b/client/src/features/admin/AddConnectedServiceButton.tsx
@@ -77,8 +77,9 @@ function AddConnectedServiceModal({
       client_id: "",
       client_secret: "",
       display_name: "",
-      scope: "api",
+      scope: "",
       url: "",
+      use_pkce: false,
     },
   });
   const onSubmit = useCallback(
@@ -188,62 +189,24 @@ function ConnectedServiceFormContent({
           control={control}
           name="kind"
           render={({ field }) => (
-            <Input
-              className={cx("form-control", errors.kind && "is-invalid")}
-              id="addConnectedServiceKind"
-              placeholder="Provider kind (gitlab or github)"
-              type="text"
-              {...field}
-            />
+            <>
+              <Input
+                className={cx("form-control", errors.kind && "is-invalid")}
+                id="addConnectedServiceKind"
+                list="addConnectedServiceKindList"
+                placeholder="Provider (gitlab or github)"
+                type="text"
+                {...field}
+              />
+              <datalist id="addConnectedServiceKindList">
+                <option value="gitlab" />
+                <option value="github" />
+              </datalist>
+            </>
           )}
           rules={{ required: true }}
         />
         <div className="invalid-feedback">Please provide a kind</div>
-      </div>
-
-      <div className="mb-3">
-        <Label className="form-label" for="addConnectedServiceClientId">
-          Client ID
-        </Label>
-        <Controller
-          control={control}
-          name="id"
-          render={({ field }) => (
-            <Input
-              className={cx("form-control", errors.id && "is-invalid")}
-              id="addConnectedServiceClientId"
-              placeholder="Client ID"
-              type="text"
-              {...field}
-            />
-          )}
-          rules={{ required: true }}
-        />
-        <div className="invalid-feedback">Please provide an id</div>
-      </div>
-
-      <div className="mb-3">
-        <Label className="form-label" for="addConnectedServiceClientSecret">
-          Client Secret
-        </Label>
-        <Controller
-          control={control}
-          name="client_secret"
-          render={({ field }) => (
-            <Input
-              className={cx(
-                "form-control",
-                errors.client_secret && "is-invalid"
-              )}
-              id="addConnectedServiceClientSecret"
-              placeholder="Client Secret"
-              type="text"
-              {...field}
-            />
-          )}
-          rules={{ required: true }}
-        />
-        <div className="invalid-feedback">Please provide a client secret</div>
       </div>
 
       <div className="mb-3">
@@ -271,27 +234,6 @@ function ConnectedServiceFormContent({
       </div>
 
       <div className="mb-3">
-        <Label className="form-label" for="addConnectedServiceScope">
-          Scope
-        </Label>
-        <Controller
-          control={control}
-          name="scope"
-          render={({ field }) => (
-            <Input
-              className={cx("form-control", errors.scope && "is-invalid")}
-              id="addConnectedServiceScope"
-              placeholder="Scope"
-              type="text"
-              {...field}
-            />
-          )}
-          rules={{ required: true }}
-        />
-        <div className="invalid-feedback">Please provide a scope</div>
-      </div>
-
-      <div className="mb-3">
         <Label className="form-label" for="addConnectedServiceUrl">
           URL
         </Label>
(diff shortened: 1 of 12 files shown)


client/src/features/connectedServices/connectedServices.api.ts
```diff
@@ -32,6 +33,33 @@ const connectedServicesApi = createApi({
...
+          method: "POST",
+          body: {
+            id,
+            kind,
+            client_id,
+            client_secret,
+            display_name,
+            scope,
+            url,
+            use_pkce: false,
```
leafty, line 57: why? this is a genuine setting, even if we (RenkuLab) do not use it. GitLab is compatible with PKCE and the backend works if enabled.

client/src/features/admin/AddConnectedServiceButton.tsx
```diff
@@ -0,0 +1,316 @@
...
+
+      <div className="mb-3">
+        <Label className="form-label" for="addConnectedServiceKind">
+          Kind
+        </Label>
+        <Controller
+          control={control}
+          name="kind"
+          render={({ field }) => (
+            <Input
```
leafty, line 191: Make this a dropdown or a radio select, or any input which only allows the valid values (see api spec).

client/src/features/admin/AddConnectedServiceButton.tsx
```diff
@@ -0,0 +1,316 @@
...
+                "form-control",
+                errors.client_secret && "is-invalid"
+              )}
+              id="addConnectedServiceClientSecret"
+              placeholder="Client Secret"
+              type="text"
+              {...field}
+            />
+          )}
+          rules={{ required: true }}
```
leafty, line 244: Client secret is NOT required. GitLab clients can be configured without a secret and PKCE (though for server deployments, the standard is to use a secret).

client/src/features/admin/AddConnectedServiceButton.tsx
```diff
@@ -0,0 +1,316 @@
...
+          render={({ field }) => (
+            <Input
+              className={cx("form-control", errors.scope && "is-invalid")}
+              id="addConnectedServiceScope"
+              placeholder="Scope"
+              type="text"
+              {...field}
+            />
+          )}
+          rules={{ required: true }}
```
leafty, line 289: Scope is NOT required. GitHub does not use OAuth2 scopes, but instead has fine-grained permissions defined on their backend, which means that the scope is `""` for GitHub clients.

See response from renkulab.io:
```json
{
    "id": "github.com",
    "kind": "github",
    "client_id": "",
    "client_secret": "",
    "display_name": "GitHub.com",
    "scope": "",
    "url": "https://github.com",
    "use_pkce": false
}
```

leafty (commented):
Some comments about the API.

client/src/features/admin/AddConnectedServiceButton.tsx
lorenzo-cavazzi: I wonder if we should use an input with a `datalist` instead. That way, users will still be presented with the supported options but can write any other text. If we add support for anything else, it wouldn't require updating the UI right away.

Screenshot_20240927_105326

leafty: No, this is a bad solution; it just makes errors very likely to happen. Updating the UI should be just re-running code generation on the newer spec, so very low effort.

client/src/features/admin/AddConnectedServiceButton.tsx
lorenzo-cavazzi: Actually, scope is required from the backend. If I don't add it, I get the following error (see screenshot)

From your explanation, I assume I should allow users to leave this empty and pass `""` to the backend.

Screenshot_20240927_111420

leafty: Yes, it needs to be empty. `required` on `react-hook-form` wants a non-empty value.

client/src/features/admin/AddConnectedServiceButton.tsx
lorenzo-cavazzi: > it just makes errors very likely to happen

How so? If you type anything unsupported, you immediately get a clear error and need to fix it before you can add a service.

Anyway, I'm not particularly attached to any solution here. I can change it to a dropdown menu later.

lorenzo-cavazzi:
@leafty Thank you for the early review. I incorporated your suggestions and finished the PR.

P.S. Sorry for rebasing; I changed the target branch to `main` and the history was :face_with_spiral_eyes:

leafty (changes requested):

lorenzo-cavazzi:
@leafty Thank you for the review. I applied your suggestions and fixed the problem with the client secret

client/src/features/admin/UpdateConnectedServiceButton.tsx
```diff
@@ -0,0 +1,178 @@
...
+    },
+    [provider.id, updateProvider]
+  );
+
+  useEffect(() => {
+    if (!result.isSuccess) {
+      return;
+    }
+    toggle();
+  }, [result.isSuccess, toggle]);
```
leafty, line 118: You should also reset the form here. Otherwise the `client secret` field is still populated when opening the form again.

leafty (changes requested):
Code looks good, there is one change to do, see below:

client/src/features/admin/UpdateConnectedServiceButton.tsx
lorenzo-cavazzi: That shouldn't be necessary. After adding a new service, this invokes `toggle()` so that the other `useEffect` will trigger and invoke `result.reset()`. I tried a few combinations to test this.

I can still add a reset here to be safe if you think there are cases that I might be missing.

leafty: Yes, add a form `reset()` in the `useEffect()` just below then 😄 . The `result.reset()` is for RTK Query. You also need to reset the form at the same time when you do a `result.reset()` in almost all cases.

client/src/features/admin/ConnectedServiceFormContent.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+      <div className="mb-3">
+        <Label className="form-label" for="addConnectedServiceClientSecret">
+          Client Secret (optional)
+        </Label>
+        <Controller
+          control={control}
+          name="client_secret"
+          render={({ field }) => (
+            <Input
```
leafty, line 161: Also, this field could be of type `password` with the button to toggle it to `text` like we have for some other hidden values. You can push that to a new issue if you like.

lorenzo-cavazzi: That makes sense; I was undecided whether to use text or password here, but since you also mentioned this I'll switch to `password` -- no need for another PR :slightly_smiling_face:

client/src/features/admin/UpdateConnectedServiceButton.tsx
lorenzo-cavazzi: Oh sorry, I wrote `result.reset()` but I meant `reset()` from the other `useEffect`. When you open the new modal, the previous `client_secret` isn't there.

Anyway, adding a `reset()` here still makes the code more readable (and it makes sense to reset immediately after the successful action) so I added it in d640d8321630d9db9481f20da603778b80714d85.

client/src/features/admin/ConnectedServiceFormContent.tsx
lorenzo-cavazzi: Done in https://github.com/SwissDataScienceCenter/renku-ui/pull/3329/commits/ae3689871eb92bf0e48f789571c56d3a814f5def

leafty:
> I noticed that IDs with spaces are accepted, but editing/deleting those services triggers errors. Any idea what could be wrong? P.S: I added a disabled input to show the ID in the "update" modal.

This is a bug with the backend, I will have to fix it.

leafty (approved):
LGTM Thanks

PR #3340, opened 2024-10-08
temp: restore the admin interface using the legacy manual API file

lorenzo-cavazzi:
Just a temporary deployment; this won't be the solution merged to the feature branch.

/deploy #notest renku-data-services=leafty/richer-github-integration


lorenzo-cavazzi:
Not needed anymore

PR #3346, opened 2024-10-10
feat: support launching sessions with data connectors

ciyer:
/deploy renku-data-services=leafty/build-data-connectors renku-notebooks=leafty/feat-data-connectors


leafty (approved):
Code looks good 👍 . If merged now, we should check that all the tests pass on the feature branch.

PR #3351, opened 2024-10-11
feat: support permissions from the API

### LINKED ISSUE #3350: Update AccessGuard and MembershipGuard to use permissions from the backend
Currently, permissions are computed on the UI for buttons such as "settings" for projects or groups. This is the wrong approach because it duplicates authorization rules in the UI and will fail to be correct when these rules are changed in the backend. Instead, rely on information from the backend.

See: https://github.com/SwissDataScienceCenter/renku-data-services/issues/453.

leafty:
Fixes #3350.

See also: https://github.com/SwissDataScienceCenter/renku-data-services/pull/454, https://github.com/SwissDataScienceCenter/renku-data-services/issues/453.

/deploy renku-data-services=main renku=release-0.59.0


lorenzo-cavazzi (commented):
Nice work!
Just a generic comment: do we want to show a disabled button for users without permissions, or should we hide the button completely instead? The second option seems the most popular on other similar platforms.

image

If we keep the current approach, I'm wondering if we should at least remove the button for non-logged users.

leafty:
> Nice work! Just a generic comment: do we want to show a disabled button for users without permissions, or should we hide the button completely instead? The second option seems the most popular on other similar platforms.
> 
> image
> 
> If we keep the current approach, I'm wondering if we should at least remove the button for non-logged users.

This PR does not change the UI, any such consideration should be for another PR after consultation with the product team.

lorenzo-cavazzi:
I'm ok with keeping the approach on this PR. It changes how it looks for anonymous users (at least for a few buttons, didn't check all of them), but that's not a blocker.

@ciyer feel free to approve the PR if you have reviewed it since I didn't check the code. Otherwise, I'll do it when I'm back next Tuesday

ciyer (approved):
🎉

PR #3353, opened 2024-10-11
minor: fallback on the project id if the namespace is missing

ciyer:
In production there is a case of a project that has no namespace in the search DB, but the ID is correct. This change should make the search display more robust against such projects.

/deploy renku=leafty/improve-cypress-tests


client/src/features/searchV2/components/SearchV2Results.tsx
```diff
@@ -274,10 +274,15 @@ function SearchV2ResultProject({ project }: SearchV2ResultProjectProps) {
...
           slug: namespace?.namespace ?? "",
         });
-  const projectUrl = generatePath(ABSOLUTE_ROUTES.v2.projects.show.root, {
-    namespace: namespace?.namespace ?? "",
-    slug,
-  });
+  const projectUrl =
+    namespace?.namespace != null
+      ? generatePath(ABSOLUTE_ROUTES.v2.projects.show.root, {
+          namespace: namespace?.namespace ?? "",
```
leafty, line 280: I think this should be correct since you test `namespace?.namespace != null`:

```suggestion
          namespace: namespace.namespace,
```

leafty (approved):
LGTM, thanks. One small nitpick below.

leafty (approved):

Issue #3365, opened 2024-10-17
Remove render tests


leafty:
Render tests are currently flaky and do not provide useful coverage. Acceptance tests and cypress tests provide UI tests.

leafty:
closed by #3367

PR #3367, opened 2024-10-17
tests: remove client tests that render (#3365)

ciyer:
Remove the most problematic and least necessary (e.g., covered elsewhere) of the tests that render


client/src/dataset/Datasets.test.jsx
```diff
@@ -1,225 +0,0 @@
...
-              projectsUrl="/projects"
-              selectedDataset="79215657-4319-4fcf-82b9-58267f2a1db8"
-            />
-          </MemoryRouter>
-        </Provider>
-      );
-    });
-  });
-
-  const core_dataset = {
```
leafty, line 91: why don't we keep the code from this line?

client/src/notifications/Notifications.test.jsx
```diff
@@ -1,240 +0,0 @@
...
-        if (keys.includes(key)) subNotifications[key] = notifications[key];
-        return subNotifications;
-      }, {});
-    expect(JSON.stringify(NotificationsInfo)).toBe(
-      JSON.stringify(subNotification)
-    );
-  });
-});
-
-const model = new StateModel(globalSchema);
```
leafty, line 124: Tests above this line are not rendering.

leafty (commented):
Should we try removing *only* tests which have rendering? Some of the removed tests are not performing rendering.

client/src/dataset/Datasets.test.jsx
ciyer: Fixed in 28c101614d3a943b0d765095eccdf6407947781e.

client/src/notifications/Notifications.test.jsx
ciyer: Fixed in 28c101614d3a943b0d765095eccdf6407947781e

client/src/project/Project.test.jsx
leafty, line 1: non-rendering tests

leafty (changes requested):

PR #3369, opened 2024-10-18
feat: show an error if the docker image identifier is invalid

andre-code:
PR to show an error if the docker image identifier is invalid.

Screenshot 2024-10-18 at 17 07 37

Screenshot 2024-10-18 at 17 08 31

Notion ref: https://www.notion.so/renku/Show-an-error-if-the-docker-image-identifier-is-invalid-1190df2efafc809682caee2d8c485285

/deploy renku=feat-jupyter-free-sessions renku-data-services=release-amaltheas-migration amalthea-sessions=main renku-notebooks=master extra-values=amalthea-sessions.deployCrd=false

### DIFF
--- a/client/src/features/sessionsV2/SessionStartPage.tsx
+++ b/client/src/features/sessionsV2/SessionStartPage.tsx
@@ -629,7 +629,7 @@ function ShowContainerImageError({
   projectUrl: string;
 }) {
   if (!("status" in error)) {
-    return false;
+    return null;
   }
 
   return (
--- a/client/src/features/sessionsV2/SessionStartPage.tsx
+++ b/client/src/features/sessionsV2/SessionStartPage.tsx
@@ -16,15 +16,19 @@
  * limitations under the License
  */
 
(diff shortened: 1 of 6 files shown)


client/src/features/sessionsV2/SessionStartPage.tsx
```diff
@@ -567,3 +604,46 @@ export default function SessionStartPage() {
...
+  projectUrl,
+}: {
+  error: FetchBaseQueryError | SerializedError;
+  launcherName: string;
+  steps: StepsProgressBar[];
+  containerImage: string;
+  projectUrl: string;
+}) {
+  if (!("status" in error)) {
+    return false;
```
leafty, line 622: You should not return `false` for a component.

```suggestion
    return null;
```

leafty (commented):
Code looks good. I would like to check the deployment before approving.

leafty (approved):
LGTM :+1:

PR #3370, opened 2024-10-21
feat: use self user API endpoint

### LINKED ISSUE #3223: Consolidate self user info API requests
Information about the user is queried in different ways:

* `/ui-server/api/data/user` -> provides user information from `data-services`
* `/ui-server/api/kc/realms/Renku/protocol/openid-connect/userinfo` -> provides user information from `Keycloak`
* `/ui-server/api/user` -> provides user information from the internal GitLab instance

See https://github.com/SwissDataScienceCenter/renku-data-services/issues/287: we should consolidate the first and second API requests to only query `data-services`. This will be possible once `data-services` provides the `is_admin` field.

Also note that `/ui-server/api/user` should be moved down the App tree to only happen in Renku 1.0 routes in the future.

leafty:
Fixes #3223.


ciyer (approved):
This looks very good! 🎉  I manually tested that basic functionality works, including user access privileges on projects owned by others, but it is hard for me to tell if there are areas that require deeper testing. I will approve, but if there is something that requires a closer look, let me know and I will do that.

PR #3372, opened 2024-10-21
fix: slug field in data connector modal

leafty:
Fixes an issue in the data connector modal where typing a slug by hand would not get taken into account.

/deploy renku=leafty/fix-ui-permissions


ciyer (approved):
🙏

PR #3380, opened 2024-10-23
chore: add async action coding guideline

### LINKED ISSUE #3188: Add a new coding guideline forbidding the use of `unwrap()` as it violates the rule of hooks
Async code in react hooks create a lot of nasty race conditions which are hard to handle correctly. Using `.unwrap()` on RTK query hooks if one of these cases of should be forbidden.

leafty:
Fixes #3188.


lorenzo-cavazzi (approved):
:+1:

PR #3381, opened 2024-10-23
chore: add onclick coding guideline

leafty:
Some counter-examples include the session row in the Renku 2.0 project page, it is not possible to open the session launcher offcanvas using the keyboard at the moment because we have `<li onclick={...}>`.


lorenzo-cavazzi (approved):
:+1:

PR #3388, opened 2024-10-29
fix: namespace setting when current is not in results

leafty:
Fixes an issue with projects in personal namespaces where the current namespace would not be selectable.

Screenshot 2024-10-30 at 16 06 23

The same issue would also exist for new groups which do not appear on the first page of results.

Other fixes:
* Fix focus border on the namespace selector
* Fix label and input id linking. Now clicking on the label focuses on the select element (can then press up or down to select a namespace).

/deploy renku=release-0.60.0


lorenzo-cavazzi (approved):
Lgtm! :rocket:

PR #3391, opened 2024-10-31
fix: remove uses of icon-text class name

leafty:
Fixes icon alignment; the `icon-text` class does not exist. Only `bi` (bootstrap-icon) should be used for icon alignment.

Before (icon too low):
image

After (icon perfectly centered vertically):
image

/deploy #notest


ciyer (approved):
👍

PR #3399, opened 2024-11-13
feat: update project creation

lorenzo-cavazzi:
This PR updates the project and group creation pages, which are now available as modals.
The settings pages have been updated accordingly, but they will undergo further updates in another pitch.

Reference: https://www.notion.so/renku/Clean-up-Project-Group-Creation-0bee33f12b624525a100df79327eb5e7

image

image

/deploy renku=release-0.62.0


client/src/features/ProjectPageV2/ProjectPageContent/Settings/ProjectSettings.tsx
leafty, line 1: Can we avoid the `<div>` bloat? It reads much less nicely in this new version. The addition of `d-flex flex-column gap-3` should already do what you want. All the `*field` components are already wrapped in a `<div>`, so this just add unnecessary `<div>` elements to the DOM tree.

lorenzo-cavazzi: The `*field` components aren't wrapped in `<div>`. The idea here is to leave the positioning/spacing to the component importing them. Removing the divs would result in additional spacing

Screenshot_20241122_204517

leafty: In this case, please update the field components to return fields properly wrapped in a `<div>`. The code will read much nicely when this is the case. E.g. in `client/src/features/projectsV2/fields/NameFormField.tsx`, replace `<>` with `<div>` so that when using the field, we only need to write `<ProjectNameFormField/>` and not `<div><ProjectNameFormField/></div>`.

client/src/features/dashboardV2/DashboardV2.tsx
```diff
@@ -179,18 +177,14 @@ function GroupsDashboard() {
...
-          className={cx(
-            "btn",
-            "btn-outline-primary",
-            "btn-sm",
-            "ms-auto",
-            "my-auto"
-          )}
-          to="/v2/groups/new"
+
+        <CreateGroupButton
```
leafty, line 181: Why are the components named `ProjectV2` but `Group`? Maybe `CreateGroupV2Button` should be used here.

client/src/features/groupsV2/new/CreateGroupButton.tsx
leafty, line 1: I think it would be simpler to use `useState` and have the modal component provided by this component here. The choice of using a slice means that the `CreateGroupButton` component cannot be used independently and needs documentation, that is, that the modal component needs to exist somewhere in the DOM for it to work. In my opinion, this put unnecessary burden on the person who want to use this component.

client/src/features/groupsV2/new/CreateGroupButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+import cx from "classnames";
+import { useCallback } from "react";
+import { useDispatch } from "react-redux";
+import { Button } from "reactstrap";
+
+import { setGroupCreationModal } from "../../projectsV2/new/projectV2New.slice";
+
+export interface CreateGroupButtonProps {
+  children?: JSX.Element | string;
```
leafty, line 27: `ReactNode` is the standard child prop type.

client/src/features/groupsV2/new/GroupNew.tsx
```diff
@@ -0,0 +1,286 @@
...
+import DescriptionFormField from "../../projectsV2/fields/DescriptionFormField";
+import NameFormField from "../../projectsV2/fields/NameFormField";
+import SlugFormField from "../../projectsV2/fields/SlugFormField";
+import {
+  setGroupCreationModal,
+  toggleGroupCreationModal,
+} from "../../projectsV2/new/projectV2New.slice";
+
+export default function GroupNew() {
+  const user = useLegacySelector((state) => state.stateModel.user);
```
leafty, line 55: Directly select `.logged` here.

Also, we should really stop using that legacy selector, because it tells the UI that the user is logged with GITLAB, not with RENKU.

client/src/features/groupsV2/new/GroupNew.tsx
```diff
@@ -0,0 +1,286 @@
...
+        <ModalHeader
+          data-cy="new-group-modal-header"
+          tag="div"
+          toggle={toggleModal}
+        >
+          <h2>Create a new group</h2>
+          <p className={cx("fs-6", "mb-0")}>
+            Groups let you group together related projects and control who can
+            access them.
+          </p>
```
leafty, line 86: I would not put this in the header, it makes the close button ("X") move down and not be presented in the top-right corner. Or maybe we find a solution for the button to be correctly positioned.

client/src/features/groupsV2/new/GroupNew.tsx
```diff
@@ -0,0 +1,286 @@
...
+      navigate(groupUrl);
+      dispatch(setGroupCreationModal(false));
+    }
+  }, [currentSlug, result, dispatch, navigate]);
+
+  const nameHelpText = (
+    <FormText className="input-hint">
+      The URL for this group will be{" "}
+      <span className="fw-bold">
+        renkulab.io/v2/groups/{currentSlug || "<Name>"}
```
leafty, line 168: We lowercase the name:

```suggestion
        renkulab.io/v2/groups/{currentSlug || "<name>"}
```

client/src/features/groupsV2/new/GroupNew.tsx
```diff
@@ -0,0 +1,286 @@
...
+                    </div>
+                  ) : (
+                    errors.slug &&
+                    dirtyFields.slug && (
+                      <div className={cx("d-block", "invalid-feedback")}>
+                        <p className="mb-1">
+                          You can customize the slug only with lowercase
+                          letters, numbers, and hyphens.
+                        </p>
+                      </div>
```
leafty, line 245: It would be nicer to use the error message from `react-hook-form` here. This is done by returning a message string when validation fails. Then the error can be accessed like so: `errors.slug.message`.

client/src/features/projectsV2/fields/DescriptionFormField.tsx
```diff
@@ -31,7 +31,7 @@ export default function DescriptionFormField<T extends FieldValues>({
   name,
 }: GenericFormFieldProps<T>) {
   return (
-    <div className="mb-3">
+    <>
```
leafty, line 34: Keep the `<div>`! 😄 Even better would be to allow a `className` prop if the component needs to be styled.

client/src/features/projectsV2/new/CreateProjectV2Button.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+import cx from "classnames";
+import { useCallback } from "react";
+import { useDispatch } from "react-redux";
+import { Button } from "reactstrap";
+
+import { setProjectCreationModal } from "./projectV2New.slice";
+
+export interface CreateProjectV2ButtonProps {
+  children?: JSX.Element | string;
```
leafty, line 27:
```suggestion
  children?: ReactNode;
```

client/src/features/projectsV2/fields/formField.types.ts
```diff
@@ -6,11 +6,27 @@ import type {
 
 export interface GenericProjectFormFieldProps<T extends FieldValues>
   extends UseControllerProps<T> {
+  compact?: boolean;
```
leafty, line 9: What is compact? Is it really a prop that works for all fields? If not, only add it to the component which needs it.

client/src/features/projectsV2/fields/formField.types.ts
```diff
@@ -6,11 +6,27 @@ import type {
...
   errors: FieldErrors<T>;
+  isDirty?: boolean;
```
leafty, line 11: `isDirty` is available from the form state, do not pass it around.

client/src/features/projectsV2/fields/formField.types.ts
```diff
@@ -6,11 +6,27 @@ import type {
...
+  helpText?: string | JSX.Element;
```
leafty, line 12: ReactNode

client/src/features/groupsV2/new/GroupNew.tsx
```diff
@@ -0,0 +1,286 @@
...
+    (groupPostRequest: GroupPostRequest) => {
+      createGroup({ groupPostRequest });
+    },
+    [createGroup]
+  );
+
+  useEffect(() => {
+    if (result.isSuccess) {
+      const groupUrl = generatePath(ABSOLUTE_ROUTES.v2.groups.show.root, {
+        slug: currentSlug,
```
leafty, line 157: Note: it would probably be safer to use the data from `result`.

client/src/features/projectsV2/fields/SlugFormField.tsx
```diff
@@ -17,44 +17,65 @@
...
+  isDirty,
+  resetFunction,
   name,
-}: GenericFormFieldProps<T>) {
+}: GenericFormFieldPropsWithReset<T>) {
+  const content = (
+    <Controller
+      control={control}
+      name={name}
+      render={({ field }) => {
```
leafty, line 40: Get the value of `isDirty` here.

```suggestion
      render={({ field, fieldState }) => {  // then use fieldState.isDirty
```

Or
```suggestion
      render={({ field, fieldState: { isDirty } }) => {
```

leafty (commented):
Some more comments here 👇 .

The PR is overall good, but form code needs to be improved.

client/src/features/dashboardV2/DashboardV2.tsx
lorenzo-cavazzi: I reused the naming convention for the existing components; it made sense since we have no concept of v1 group and the suffix might be redundant.
If you think it's better to add the V2 label, I'll do that to the other Group components too.

client/src/features/groupsV2/new/GroupNew.tsx
lorenzo-cavazzi: Good point, I'll use the new endpoint. Could check if this makes sense? https://github.com/SwissDataScienceCenter/renku-ui/pull/3399/commits/d337040b9c82731860d194bf7ae1a63510048483

P.S: I opened this issue to update the other v2 components
https://github.com/SwissDataScienceCenter/renku-ui/issues/3409

client/src/features/groupsV2/new/GroupNew.tsx
lorenzo-cavazzi: The description was a product team requirement, and putting it in the modal body makes it less visible.

I agree that the close button is positioned weirdly, but that's already the case in a few other modals.
We can open a separate issue to address that for all the modals if that makes sense
Screenshot_20241126_103426

client/src/features/groupsV2/new/GroupNew.tsx
lorenzo-cavazzi: Not sure I get this comment. That's only for the user's convenience to show a placeholder while the name isn't filled in. They see the upper case `Name` on the label, I don't think making it lowercase improves the UX

image

client/src/features/groupsV2/new/GroupNew.tsx
lorenzo-cavazzi: Good point :+1:  https://github.com/SwissDataScienceCenter/renku-ui/pull/3399/commits/f6b8ebc46c1ccfa8e471a679f21fa246eb9abfa0

client/src/features/rootV2/RootV2.tsx
```diff
@@ -67,9 +80,49 @@ export default function RootV2() {
...
+  // Project creation
+  const showProjectModal = useCallback(() => {
+    dispatch(setProjectCreationModal(true));
+  }, [dispatch]);
+  useEffect(() => {
+    if (searchParams.get("createProject") != null) {
+      showProjectModal();
+      removeSearchParams("createProject");
+    }
+  }, [removeSearchParams, searchParams, showProjectModal]);
```
leafty, line 119: I don't understand what is the point of doing this. Just remove the old routes and set-up the modals from the top nav bars.

leafty: I we need some form of URL handling, have a look at the `useLocationHash` hook and how it is used to control some offcanvas elements.

client/src/features/projectsV2/fields/DescriptionFormField.tsx
lorenzo-cavazzi: Yeah, that makes sense. Returning unwrapped components isn't common.
I made that to keep the positioning flexible for a first implementation that I later ditched. Going back to wrapping fields in divs is the best option.

https://github.com/SwissDataScienceCenter/renku-ui/pull/3399/commits/080758bcdaa7e882851f3b6a9c4933a6f8bfb65e

client/src/features/projectsV2/fields/formField.types.ts
lorenzo-cavazzi: Done :+1:  https://github.com/SwissDataScienceCenter/renku-ui/pull/3399/commits/fd135713dacb6d12c831f92290dda74869efa62e

client/src/features/projectsV2/fields/formField.types.ts
lorenzo-cavazzi: True. The `isDirty` name was very misleading; fixed in  https://github.com/SwissDataScienceCenter/renku-ui/pull/3399/commits/fd135713dacb6d12c831f92290dda74869efa62e

client/src/features/groupsV2/new/GroupNew.tsx
lorenzo-cavazzi: Good catch! https://github.com/SwissDataScienceCenter/renku-ui/commit/e4a6ec2995f4fa778f6b94913efbfc8962da4ef0

client/src/features/projectsV2/fields/SlugFormField.tsx
lorenzo-cavazzi: :+1:  https://github.com/SwissDataScienceCenter/renku-ui/pull/3399/commits/13f253c06b28b0ea5959aadcffed75e1fc0f2227

client/src/features/rootV2/RootV2.tsx
lorenzo-cavazzi: > Just remove the old routes and set-up the modals from the top nav bars.

We need to open the modals from other places too, like the Dashboard and, in the future, the groups (and possibly other pages).

I'm happy to discuss this offline to explain better the details and understand the alternative solution you are proposing

lorenzo-cavazzi:
@leafty thank you for the review!

I either updated the code as suggested or dropped an inline content. Most points are solved or trivial to solve after a confirmation (E.G: `Group` vs `GroupV2`).
The one outstanding is "how to treat routes" / "where to put modals". I'm happy to talk about it offline to explain the requirements better and understand your proposed solution.

client/src/features/rootV2/RootV2.tsx
leafty: The modals open perfectly fine from any page when using the top navbar. And you can also include the button+modal on any page anywhere, it already works. There is no need for the query param solution.

leafty: If you use the usual `useState` and button+modal component approach used elsewhere, then you can add that component anywhere without issue.

client/src/features/dashboardV2/DashboardV2.tsx
lorenzo-cavazzi: As discussed offline, let's keep Group components without V2

client/src/features/groupsV2/new/GroupNew.tsx
lorenzo-cavazzi: As discussed offline, I opened a separate issue to address that for all the modals: https://github.com/SwissDataScienceCenter/renku-ui/issues/3410

client/src/features/groupsV2/new/GroupNew.tsx
lorenzo-cavazzi: Done

client/src/features/groupsV2/new/GroupNew.tsx
lorenzo-cavazzi: Fixed here https://github.com/SwissDataScienceCenter/renku-ui/pull/3399/commits/f8c7a301a79eb1fcba33d3019928d530b0774485
It seems to work fine now

client/src/features/projectsV2/fields/NameFormField.tsx
```diff
@@ -51,9 +52,12 @@ export default function NameFormField<T extends FieldValues>({
...
       <div className="invalid-feedback">Please provide a name</div>
-      <FormText id={`${entityName}NameHelp`} className="input-hint">
-        The name you will use to refer to the {entityName}.
-      </FormText>
+      {helpText && typeof helpText === "string" && (
+        <FormText id={`${entityName}-help`} className="input-hint">
+          {helpText}
+        </FormText>
+      )}
+      {helpText && typeof helpText !== "string" && <>{helpText}</>}
```
leafty, line 60: Can we avoid doing a special case here? Also, the association breaks in the second case (no `aria-describedby`).

```suggestion
      {helpText ? (
        <FormText id={`${entityName}-help`}>
          {helpText}
        </FormText>
      ) : (
        <FormText id={`${entityName}NameHelp`}>
          The name you will use to refer to the {entityName}.
        </FormText>
      )}
```

Note: we should NOT use the `input-hint` class since it comes from Renku 1.0 css files which pollute the global CSS namespace. For reference, see `client/src/styles/renku_additional_classes.scss` where only a couple of CSS classes are added to the global v2 styles.

It is OK to define a `input-hint` class in a CSS/SCSS module to be imported widely, as the corresponding classes would be prefixed.

client/src/features/projectsV2/fields/ProjectNamespaceFormField.tsx
```diff
@@ -264,12 +258,13 @@ export default function ProjectNamespaceFormField<T extends FieldValues>({
             /^(?!.*\.git$|.*\.atom$|.*[-._][-._].*)[a-zA-Z0-9][a-zA-Z0-9\-_.]*$/,
         }}
       />
-      <div className="invalid-feedback">
-        A project must belong to a namespace.
-      </div>
-      <FormText id={`${entityName}NamespaceHelp`} className="input-hint">
-        The user or group namespace where this project should be located.
-      </FormText>
+      <div className="invalid-feedback">A project must belong to a owner.</div>
```
leafty, line 261:
```suggestion
      <div className="invalid-feedback">A project must belong to an owner.</div>
```

client/src/features/projectsV2/fields/ProjectSlugFormField.tsx
```diff
@@ -19,19 +19,25 @@
 import type { FieldValues } from "react-hook-form";
 
 import SlugFormField from "./SlugFormField";
-import type { GenericProjectFormFieldProps } from "./formField.types";
+import type { SlugProjectFormFieldProps } from "./formField.types";
```
leafty, line 22: Define the type here.

client/src/features/projectsV2/fields/ProjectVisibilityFormField.tsx
```diff
@@ -22,66 +22,77 @@ import type { FieldValues } from "react-hook-form";
...
-                {...field}
-                value="private"
-                checked={field.value === "private"}
-              />
-              <Label
-                for="project-visibility-private"
-                className="cursor-pointer"
+      <div data-cy="project-visibility-controller">
+        <Controller
+          aria-describedby="projectVisibilityHelp"
```
leafty, line 39: This does nothing. The component is not rendered as an HTML tag.

client/src/features/projectsV2/fields/ProjectVisibilityFormField.tsx
```diff
@@ -22,66 +22,77 @@ import type { FieldValues } from "react-hook-form";
...
+          control={control}
+          name={name}
+          render={({ field }) => (
+            <>
+              <ButtonGroup id="project-visibility">
+                <Input
+                  type="radio"
```
leafty, line 46: Add the `aria-describedby` here and to the other input. It is fine to refer to the same `aria-describedby` component.

client/src/features/projectsV2/fields/ProjectVisibilityFormField.tsx
```diff
@@ -22,66 +22,77 @@ import type { FieldValues } from "react-hook-form";
...
-                data-cy="project-visibility-private"
-                id="project-visibility-private"
-                {...field}
-                value="private"
-                checked={field.value === "private"}
-              />
-              <Label
-                for="project-visibility-private"
-                className="cursor-pointer"
+      <div data-cy="project-visibility-controller">
```
leafty, line 37: Do we really need the `<div>` with the `data-cy`? The presence of the form input can be checked with the label, e.g. `cy.contains("label", "Visibility").should("be.visible")` and then  `data-cy` properties can be given to each `input` element.

client/src/features/projectsV2/fields/formField.types.ts
```diff
@@ -7,10 +7,26 @@ import type {
...
   extends UseControllerProps<T> {
   entityName: string;
   errors: FieldErrors<T>;
+  helpText?: React.ReactNode;
+}
+
+export interface SlugFormFieldProps<T extends FieldValues>
+  extends GenericFormFieldProps<T> {
+  compact?: boolean;
+  countAsDirty?: boolean;
```
leafty, line 23: What is `countAsDirty`? It looks like a hack because react-hook-form is being mis-used.

client/src/features/rootV2/RootV2.tsx
```diff
@@ -140,8 +149,11 @@ function GroupsV2Routes() {
     <Routes>
       <Route
         path={RELATIVE_ROUTES.v2.groups.new}
-        element={<LazyGroupV2New />}
+        element={
+          <Navigate to={`${ABSOLUTE_ROUTES.v2.root}#createGroup`} replace />
```
leafty, line 153:
```suggestion
          <Navigate to={{ pathname: ABSOLUTE_ROUTES.v2.root, hash: "create-group" } replace />
```

Note: Ideally these would be handled as effects with `navigate()`.

client/src/features/groupsV2/new/GroupNew.tsx
```diff
@@ -0,0 +1,285 @@
...
+                      "btn",
+                      "btn-link",
+                      "p-0",
+                      "text-decoration-none"
+                    )}
+                    data-cy="group-slug-toggle"
+                    onClick={toggleCollapse}
+                    type="button"
+                  >
+                    Customize group URL <ChevronDown className="bi" />
```
leafty, line 210: nitpick: the `<ChevronFlippedIcon>` component can be used instead. It has a `flipped` property (e.g. `fipped={isCollapseOpen}`) to indicate that the collapse is open.

client/src/features/groupsV2/new/GroupNew.tsx
```diff
@@ -0,0 +1,285 @@
...
+                        "mb-0"
+                      )}
+                    >
+                      <span>renkulab.io/v2/groups/</span>
+                      <SlugFormField
+                        compact={true}
+                        control={control}
+                        entityName="group"
+                        errors={errors}
+                        countAsDirty={dirtyFields.slug && dirtyFields.name}
```
leafty, line 227: Note that this is not working at the moment on the deployment: the reset button does not appear when it I type a custom slug.

leafty (commented):
NOTE: the modals with useLocationHash are broken on the search page. Let's make an issue for this. The `useSearchParams` hook may be interfering with it on the search page.

client/src/features/projectsV2/fields/ProjectSlugFormField.tsx
lorenzo-cavazzi: We generally encourage having a `.types` file for types/interfaces. I don't think defining the type here is better than the current solution

client/src/features/projectsV2/fields/ProjectVisibilityFormField.tsx
lorenzo-cavazzi: This mechanism is very convenient. Perhaps there will never be a page where `cy.contains("label", "Visibility").should("be.visible")` will test true based on another entity's visibility, but we had similar cases in the past and `data-cy` helps simplifying the logic in the Cypress tests.

client/src/features/projectsV2/fields/formField.types.ts
lorenzo-cavazzi: Mis-used?
The UX has been optimized to avoid showing useless interactions; on some pages, it's easier to integrate this component seamlessly by considering other component's states. I don't think that logic should be embedded here.

client/src/features/groupsV2/new/GroupNew.tsx
lorenzo-cavazzi: :pray:

client/src/features/groupsV2/new/GroupNew.tsx
lorenzo-cavazzi: Perhaps the name wasn't filled in?
I updated the logic a little to make it more consistent. The button should show when:
* The URL is invalid
* The name isn't blank

and in no other case.

lorenzo-cavazzi:
> NOTE: the modals with useLocationHash are broken on the search page. Let's make an issue for this. The `useSearchParams` hook may be interfering with it on the search page.

@leafty Should we use a query parameter instead of the hash? We would have an additional useless network request, which isn't a big deal, and it should work fine on the search page.

If you prefer the hash solution (I agree it's really neat) and already have an idea of what could be the conflict, I can either open the issue or, if it's quick enough, try to fix it in this PR.

leafty:
> > NOTE: the modals with useLocationHash are broken on the search page. Let's make an issue for this. The `useSearchParams` hook may be interfering with it on the search page.
> 
> @leafty Should we use a query parameter instead of the hash? We would have an additional useless network request, which isn't a big deal, and it should work fine on the search page.
> 
> If you prefer the hash solution (I agree it's really neat) and already have an idea of what could be the conflict, I can either open the issue or, if it's quick enough, try to fix it in this PR.

I highly suspect it would not work, given how the search page works. That's why fixing it later is probably fine, because creating a new project or group from the search page is not the biggest use case.

lorenzo-cavazzi:
@leafty Ok, sounds good :+1: 

I opened a new issue https://github.com/SwissDataScienceCenter/renku-ui/issues/3415 and added a simple workaround to avoid having broken buttons on the navbar https://github.com/SwissDataScienceCenter/renku-ui/pull/3399/commits/0bc7bfe845f1041fd73216fe31ee31365d2f8de0

client/src/features/groupsV2/new/GroupNew.tsx
leafty: Shouldn't I see the reset button in this case? I filled in the name of the project and edited the slug. What if I want to reset its value now?

image

client/src/features/projectsV2/fields/NameFormField.tsx
```diff
@@ -51,9 +52,7 @@ export default function NameFormField<T extends FieldValues>({
         rules={{ required: true, maxLength: 99 }}
       />
       <div className="invalid-feedback">Please provide a name</div>
-      <FormText id={`${entityName}NameHelp`} className="input-hint">
-        The name you will use to refer to the {entityName}.
-      </FormText>
+      {helpText}
```
leafty, line 55: You should keep the `<FormText>` component here, otherwise the association with the input field is broken.

```suggestion
      {helpText && <FormText id={`${entityName}-help`}>{helpText}</FormText>}
```

leafty (commented):
Should be the last small things to fix.

client/src/features/groupsV2/new/CreateGroupButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+  dataCy,
+}: CreateGroupButtonProps) {
+  const [, setHash] = useLocationHash();
+  const openModal = useCallback(() => {
+    setHash(groupCreationHash);
+  }, [setHash]);
+
+  return (
+    <Button
+      className={cx(className)}
```
leafty, line 45: ?
```suggestion
      className={className}
```

client/src/features/projectsV2/fields/SlugPreviewFormField.tsx
```diff
@@ -0,0 +1,101 @@
...
+                <p className="mb-0">
+                  Mind the URL will be updated once you provide a name.
+                </p>
+              </div>
+            ) : (
+              errors.slug &&
+              dirtyFields.slug && (
+                <div className={cx("d-block", "invalid-feedback")}>
+                  <p className="mb-1">
+                    {errors?.slug?.message?.toString() ?? ""}
```
leafty, line 91: You need to provide a default error message here, otherwise it is confusing for users.

Example:
Screenshot 2024-12-13 at 11 29 29

client/src/features/projectsV2/fields/SlugFormField.tsx
```diff
@@ -17,44 +17,76 @@
...
+        return (
+          <InputGroup>
+            <div className="input-group-text">
+              <small>{url}</small>
+            </div>
+            <Input
+              aria-describedby={`${entityName}SlugHelp`}
+              className={cx(
+                "form-control",
+                errors.slug && isDirty && "is-invalid",
```
leafty, line 50: This prevents showing an error when the slug is empty

```suggestion
                errors.slug && "is-invalid",
```

client/src/features/projectsV2/fields/formField.types.ts
```diff
@@ -7,10 +7,36 @@ import type {
...
   extends UseControllerProps<T> {
   entityName: string;
   errors: FieldErrors<T>;
+  helpText?: React.ReactNode;
+}
+
+export interface SlugFormFieldProps<T extends FieldValues>
+  extends GenericFormFieldProps<T> {
+  compact?: boolean;
+  url: string;
```
leafty, line 23: nitpick: this should be optional

client/src/features/projectsV2/fields/formField.types.ts
```diff
@@ -7,10 +7,36 @@ import type {
...
+  slug: string;
+  label: string;
+  entityName: "project" | "group";
+  dirtyFields: Partial<
+    Readonly<{
+      name?: boolean | undefined;
+      namespace?: boolean | undefined;
+      slug?: boolean | undefined;
+    }>
+  >;
```
leafty, line 41: Can be made into a new issue: You do not need to pass down all these fields. Actually, the whole form state can be obtained from `control` only.

client/src/features/projectsV2/fields/SlugFormField.tsx
```diff
@@ -17,44 +17,76 @@
...
+            )}
+          </InputGroup>
+        );
+      }}
+      rules={{
+        required: true,
+        maxLength: 99,
+        pattern: {
+          message:
+            "You can customize the slug only with lowercase letters, numbers, and hyphens.",
```
leafty, line 76: This message is incorrect or not complete:

Screenshot 2024-12-13 at 11 14 47

Hyphen as the first character is not allowed (first character needs to be alphanum).

Dots (`"."`) are allowed.

Ending in `.git` or `.atom` is not allowed.

client/src/features/projectsV2/fields/SlugFormField.tsx
```diff
@@ -17,44 +17,76 @@
...
+              >
+                <ArrowCounterclockwise className="bi" />
+              </Button>
+            )}
+          </InputGroup>
+        );
+      }}
+      rules={{
+        required: true,
+        maxLength: 99,
```
leafty, line 73: It would be nice to provide a message here.

client/src/features/projectsV2/fields/SlugPreviewFormField.tsx
```diff
@@ -0,0 +1,101 @@
...
+    <div>
+      <FormText>
+        The URL for this {entityName} will be{" "}
+        <span className="fw-bold">
+          {url}
+          {slug || "<name>"}
+        </span>
+      </FormText>
+      <button
+        className={cx("btn", "btn-link", "p-0", "text-decoration-none", "ms-2")}
```
leafty, line 51: Nitpick: `ms-2` does not work when this wraps.

Screenshot 2024-12-13 at 11 16 02

client/src/features/projectsV2/new/CreateProjectV2Button.tsx
```diff
@@ -0,0 +1,53 @@
...
+  dataCy,
+}: CreateProjectV2ButtonProps) {
+  const [, setHash] = useLocationHash();
+  const openModal = useCallback(() => {
+    setHash(projectCreationHash);
+  }, [setHash]);
+
+  return (
+    <Button
+      className={cx(className)}
```
leafty, line 45:
```suggestion
      className={className}
```

leafty (commented):
Looks good, there are some minor issues left 👇

client/src/features/projectsV2/fields/SlugFormField.tsx
```diff
@@ -70,10 +70,13 @@ export default function SlugFormField<T extends FieldValues>({
       }}
       rules={{
         required: true,
-        maxLength: 99,
+        maxLength: {
+          value: 99,
+          message: "This slug exceed maxLength.",
```
leafty, line 75: This does not sound useful for users.

```suggestion
          message: "The slug must not exceed 99 characters.",
```

client/src/features/projectsV2/fields/SlugFormField.tsx
```diff
@@ -70,10 +70,13 @@ export default function SlugFormField<T extends FieldValues>({
...
+        },
         pattern: {
           message:
-            "You can customize the slug only with lowercase letters, numbers, and hyphens.",
+            "Slug can include lowercase letters, numbers, dots ('.'), and hyphens, but must start with a letter or number and cannot end with '.git' or '.atom' or start with a hyphen.",
```
leafty, line 79:
```suggestion
            "A valid slug can include lowercase letters, numbers, dots ('.'), hyphens ('-') and underscores ('_'), but must start with a letter or number and cannot end with '.git' or '.atom'.",
```

leafty (commented):
If needed, the copy for error messages should be checked with a native English speaker from the product team.

andre-code:
> If needed, the copy for error messages should be checked with a native English speaker from the product team.

Thanks, copy approved by Product team

PR #3400, opened 2024-11-14
Fix several minor data-connector bugs

ciyer:
* minor: fix DOM structure issue
* minor: use dropdown z-index for namespace select menu
* fix: only save credentials when there are sensitive values

/deploy renku=feat-jupyter-free-sessions renku-data-services=main amalthea-sessions=main renku-notebooks=master extra-values=amalthea-sessions.deployCrd=false

### DIFF
--- a/client/src/features/projectsV2/fields/ProjectNamespaceFormField.tsx
+++ b/client/src/features/projectsV2/fields/ProjectNamespaceFormField.tsx
@@ -176,12 +176,6 @@ function NamespaceSelector({
       isClearable={false}
       isSearchable={false}
       isLoading={isFetchingMore}
-      styles={{
-        menu: (base: Record<string, unknown>) => ({
-          ...base,
-          zIndex: 1000,
-        }),
-      }}
     />
   );
 }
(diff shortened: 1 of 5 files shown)


client/src/features/projectsV2/fields/ProjectNamespaceFormField.tsx
```diff
@@ -176,6 +176,12 @@ function NamespaceSelector({
       isClearable={false}
       isSearchable={false}
       isLoading={isFetchingMore}
+      styles={{
+        menu: (base: Record<string, unknown>) => ({
+          ...base,
+          zIndex: 1000,
+        }),
+      }}
```
leafty, line 184: Why not add `z-2` to the `menu` class in `selectClassNames`?

ciyer: I believe the correct z-index for things that behave like drop-down in Bootstrap is 1000: https://getbootstrap.com/docs/5.3/extend/approach/#overlay-components

leafty: In this case, please add the styling to the `.module.scss` file using `--#{$prefix}dropdown-zindex` or `$zindex-dropdown` (https://getbootstrap.com/docs/5.3/components/dropdowns/#css).

PR #3402, opened 2024-11-18
feat: ask for user secrets on session launch

leafty:
Part of #3412, merging into `build/session-secrets`.

Ask logged-in users to provide missing session secrets at launch. The interrupt happens first, that is before the interrupt for data connector credentials.

image

/deploy #notest renku=build/session-secrets renku-data-services=leafty/session-secrets-5


ciyer (approved):
This looks good! I found a stray `TODO` without a description, and I was not sure if that has been resolved already. I think it makes sense to either remove the comment or add some details about what is still outstanding, but since this is a PR that is part of a larger build, I think these changes can be merged and that can be fleshed out in the follow up PR.

client/src/features/sessionsV2/SessionStartPage.tsx
```diff
@@ -545,6 +554,20 @@ function StartSessionFromLauncher({
   if (sessionStarted)
     return <SessionStarting launcher={launcher} project={project} />;
 
+  // TODO
```
leafty, line 557:
```suggestion
```

PR #3406, opened 2024-11-21
fix: reset state of data connectors on modal close

ciyer:
After closing the data connector modal, some state would be retained, causing bugs.

## Testing

The bugs would occur when a second data connector was created immediately after the first. For example, one data connector was created, and then a second data connector was started, but canceled (e.g., by closing the dialog), the second would get created anyway. Check that these situations do not happen anymore

/deploy renku=release-0.61.0

### DIFF
--- a/client/src/features/ProjectPageV2/ProjectPageContent/DataConnectors/ProjectDataConnectorsBox.tsx
+++ b/client/src/features/ProjectPageV2/ProjectPageContent/DataConnectors/ProjectDataConnectorsBox.tsx
@@ -101,12 +101,14 @@ function ProjectDataConnectorBoxContent({
           )}
         </CardBody>
       </Card>
-      <ProjectConnectDataConnectorsModal
-        isOpen={isModalOpen}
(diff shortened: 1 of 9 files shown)


andre-code (commented):
The code looks great and does the job well!

I just noticed that when editing a data connector, it’s not possible to change its visibility. Is that how it’s supposed to work?

Also, when editing a data connector I found the message “project was not linked” a bit confusing, what exactly does it mean?

Screenshot 2024-11-25 at 10 19 03

ciyer:
> The code looks great and does the job well!
> 
> I just noticed that when editing a data connector, it’s not possible to change its visibility. Is that how it’s supposed to work?
> 
> Also, when editing a data connector I found the message “project was not linked” a bit confusing, what exactly does it mean?

Thanks for pointing these out! They are both bugs which I have been able to fix.

andre-code (dismissed):
issues solved! 🚀

andre-code (approved):
lgtm

PR #3407, opened 2024-11-22
feat: add new dashboard layout

andre-code:
PR to add new dashboard layout

/deploy renku=andrea/update-dashboard-layout renku-data-services=main


client/public/dashboardHeader1080x280.png
leafty, line 1: Why do we have all of these assets? Can't we use a single SVG file for this?

client/src/features/dashboardV2/DashboardV2.tsx
```diff
@@ -166,49 +354,210 @@ function ProjectList() {
...
+        >
+          <h3 className={cx("text-center", "mb-0")}>
+            {userInfo.first_name} {userInfo.last_name}
+          </h3>
+        </Link>
+        <p className="mb-0">
+          <Link to={userPageUrl} className="link-primary">
+            @{userInfo.username ?? "unknown"}
+          </Link>
+        </p>
```
leafty, line 450: Why make every element a link here? This feels confusing to me, as they all go to the same location.

Screenshot 2024-12-09 at 13 12 35

It seems wrong to allow this to happen (green text in the avatar):
Screenshot 2024-12-09 at 13 13 00

I would suggest to keep only one link on the `@username` element and use the `stretch-link` utility to make the whole card clickable. See: https://getbootstrap.com/docs/5.3/helpers/stretched-link/.

client/src/features/sessionsV2/sessionsV2.api.ts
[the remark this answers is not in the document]
andre-code: we can create an issue for this

andre-code:
Thanks @leafty for the review! I’ve made the changes you suggested and added the slug to the project list on the dashboard.

client/src/features/dashboardV2/DashboardV2.tsx
```diff
@@ -15,117 +15,306 @@
...
+          media="(min-width: 992px)"
+          srcSet="/dashboardHeader1920x280.png"
+        />
+        <source
+          media="(min-width: 768px)"
+          srcSet="/dashboardHeader1080x280.png"
+        />
+        <source
+          media="(min-width: 576px)"
+          srcSet="/dashboardHeader750x200.png"
```
leafty, line 132: There is an issue with the perceived luminance of this image:

Screenshot From 2024-12-11 09-31-56

The tree foilage looks much darker with this image compared to the other ones.

(This may be due to a [Moiré effect](https://en.wikipedia.org/wiki/Moir%C3%A9_pattern) because of the dots)

client/src/features/sessionsV2/sessionsV2.api.ts
```diff
@@ -81,6 +82,18 @@ const sessionsV2Api = createApi({
...
     }),
+    getProjectSessionLauncher: builder.query<
+      SessionLauncher,
+      GetProjectSessionLauncherParams
+    >({
+      query: ({ id }) => {
+        return {
+          url: `/session_launchers/${id}`,
+        };
+      },
```
leafty, line 93: Do you need to override the query?

leafty (commented):
There are some issues which need to be resolved:
1. Footer card consistency (see comments below)
2. Some small SCSS / bootstrap classnames changes (see comments below)
3. Avatars not using proper initials (see here)

You do not need to resolve the issues about the cover image on the Dashboard, but we probably need to have a technical discussion about this.

The avatars are not correct on the Dashboard page:
Screenshot From 2024-12-11 09-24-12

See the correct versions here:
Screenshot From 2024-12-11 09-24-20
Screenshot From 2024-12-11 09-24-27

I have to say that part of the issue is that the Avatar component should be updated and handle rendering initials based on the entity type (user or group) so that mistakes are not common.

andre-code:
Thanks @leafty  for your review, the reason for calculating the avatar based on the slug/username instead of the initials is to ensure consistency across the app.

Right now, we fetch the list of projects and data connector and get the namespace slug, but not the namespace name. This can result in situations where, for example, a group named Potato shows initials "PO" on the dashboard, but a project under the namespace displays initials "SP" instead.

This approach is temporary. Once the projects API includes the namespace name, we can switch to using the initials derived from the name to create the avatar instead of relying on the slug.

leafty:
> Thanks @leafty for your review, the reason for calculating the avatar based on the slug/username instead of the initials is to ensure consistency across the app.
> 
> Right now, we fetch the list of projects and data connector and get the namespace slug, but not the namespace name. This can result in situations where, for example, a group named Potato shows initials "PO" on the dashboard, but a project under the namespace displays initials "SP" instead.
> 
> This approach is temporary. Once the projects API includes the namespace name, we can switch to using the initials derived from the name to create the avatar instead of relying on the slug.

I will make a PR to have the avatar component fetch the required data if needed. I don't think we need to update the API.

leafty:
> Thanks @leafty for your review, the reason for calculating the avatar based on the slug/username instead of the initials is to ensure consistency across the app.

Also, this is wrong, since the existing pages use the initials (which is much more appropriate) than the namespace slug.

client/src/features/sessionsV2/sessionsV2.api.ts
andre-code: Could you provide more details or clarify this comment?

andre-code:
> > Thanks @leafty for your review, the reason for calculating the avatar based on the slug/username instead of the initials is to ensure consistency across the app.
> 
> Also, this is wrong, since the existing pages use the initials (which is much more appropriate) than the namespace slug.


I understand the confusion. The changes required for the group and user pages are addressed in the other PRs. Here, you’ll see the avatar functioning consistently across all pages. https://renku-ci-ui-3433.dev.renku.ch/

client/src/features/sessionsV2/sessionsV2.api.ts
leafty: Why do you override the `query`? It should be correctly generated.

leafty: Oh I see it is not generated. Well, all the renku 2.0 should be generated.

leafty (approved):
Should be OK to merge now. Good work.

leafty (changes requested):
Actually, should be good to go once the footer cards are fixed.

leafty (approved):

client/src/features/dashboardV2/DashboardV2.module.scss
```diff
@@ -0,0 +1,43 @@
...
+  &:hover {
+    background-color: var(--bs-primary-bg-subtle) !important;
+  }
+}
+
+.FooterCard {
+  a svg {
+    transition: all 0.3s linear;
+  }
+  a:hover svg {
```
leafty, line 40: Small detail, but this would be nice:

```suggestion
  a:hover svg, a:focus svg {
```

Reference: https://webaim.org/techniques/hypertext/link_text#hover

client/src/features/dashboardV2/DashboardV2.tsx
```diff
@@ -574,12 +568,6 @@ function SessionsDashboard() {
           </h4>
           <Badge>{totalSessions}</Badge>
         </div>
-        {totalSessions > 0 && (
```
leafty: Did you intend to remove this?

leafty (approved):

Issue #3412, opened 2024-11-27
[ShapeUp] Session Secrets in Renku 2.0


leafty:
Part of https://github.com/SwissDataScienceCenter/renku/issues/3832.

Content:
1. #3398
2. #3402

PR #3414, opened 2024-11-28
feat: make the session secret mount location configurable

leafty:
Part of #3412, merging into `build/session-secrets`.

Update the "Session Secrets" section to allow for the mount location to be configured.

Screenshot 2024-12-10 at 14 11 38

/deploy #notest renku=build/session-secrets renku-data-services=build/session-secrets


andre-code (approved):
It works really well! The code looks great, the interfaces are clean, and the UX is excellent. Great job, @leafty!

PR #3421, opened 2024-12-03
feat: add sftp storage in shortlist DC

andre-code:
PR to add sftp in DC storage list
/deploy renku-data-services=main renku-notebooks=andrea/add-polybox-switchdrive-config


ciyer (approved):
Works as advertised!

PR #3393, opened 2024-11-07
feat: form for copying a project (initial version)

ciyer:
Minimal interface for copying projects. To test:

- Go to any project
- Click the "Copy project" button
- Look at the result
- See that you can navigate back to the original project

/deploy

PR #3427, opened 2024-12-05
feat!: allow copying projects (#3393)

ciyer:
# Testing
- copy a regular project
- break the copy link
- mark a project as a template and see that banners are shown
  - when viewing a project that is a template, but for which you are not an owner
  - when viewing a project that is a template, where you are an owner
   
/deploy renku=release-0.63.0

### DIFF
--- a/client/src/components/PrimaryAlert.module.scss
+++ b/client/src/components/PrimaryAlert.module.scss
@@ -1,0 +1,9 @@
+@import "~bootstrap/scss/functions";
+@import "~bootstrap/scss/variables";
+@import "~bootstrap/scss/variables-dark";
+@import "../styles/renku_bootstrap_customization.scss";
+
+.primaryAlert {
+  --bs-primary-bg-subtle: #{tint-color($primary, 90%)};
+  max-height: 50vh;
+}
--- a/client/src/components/PrimaryAlert.module.scss
+++ b/client/src/components/PrimaryAlert.module.scss
@@ -5,5 +5,4 @@
 
 .primaryAlert {
   --bs-primary-bg-subtle: #{tint-color($primary, 90%)};
-  max-height: 50vh;
 }
--- a/client/src/components/buttons/Button.tsx
+++ b/client/src/components/buttons/Button.tsx
@@ -27,7 +27,13 @@ import { faSyncAlt } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import cx from "classnames";
 import { Fragment, ReactNode, useRef, useState } from "react";
-import { ArrowRight, ChevronDown, Pencil, PlusLg } from "react-bootstrap-icons";
+import {
+  ArrowRight,
+  ChevronDown,
+  Pencil,
+  PlusLg,
+  ThreeDotsVertical,
+} from "react-bootstrap-icons";
 import { Link } from "react-router-dom";
 import {
   Button,
@@ -114,7 +120,7 @@ function ButtonWithMenu(props: ButtonWithMenuProps) {
   );
 }
 
-interface BButtonWithMenuV2Props {
+interface ButtonWithMenuV2Props {
   children?: React.ReactNode;
   className?: string;
   color?: string;
@@ -125,7 +131,9 @@ interface BButtonWithMenuV2Props {
   preventPropagation?: boolean;
   size?: string;
 }
-export function ButtonWithMenuV2({
+export const ButtonWithMenuV2 = SplitButtonWithMenu;
+
+export function SplitButtonWithMenu({
   children,
   className,
   color,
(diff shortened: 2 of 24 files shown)


client/src/components/PrimaryAlert.module.scss
```diff
@@ -0,0 +1,9 @@
+@import "~bootstrap/scss/functions";
+@import "~bootstrap/scss/variables";
+@import "~bootstrap/scss/variables-dark";
+@import "../styles/renku_bootstrap_customization.scss";
+
+.primaryAlert {
+  --bs-primary-bg-subtle: #{tint-color($primary, 90%)};
+  max-height: 50vh;
```
leafty, line 8: The other alerts do not have this CSS rule. Maybe it should not be here by default.

client/src/components/buttons/Button.tsx
```diff
@@ -165,6 +173,44 @@ export function ButtonWithMenuV2({
...
+  direction,
+  disabled,
+  id,
+  preventPropagation,
+  size,
+}: Omit<ButtonWithMenuV2Props, "default">) {
+  // ! Temporary workaround to quickly implement a design solution -- to be removed ASAP #3250
+  const additionalProps = preventPropagation
+    ? { onClick: (e: React.MouseEvent) => e.stopPropagation() }
+    : {};
```
leafty, line 189: If you are not using this button on interactive elements, then you do not need this workaround. Interactive elements should not be nested in the DOM tree (e.g. having <button> as a descendant of <a>).

leafty (approved):
This looks good! 👏 

Some small comments below, but they do not need to be resolved immediately.

For the tests, it may be necessary to delete the deployment, delete the PVCs and re-deploy. It it unfortunately quite common for the KG to have database issues when deployments are a bit old.

andre-code (dismissed):
It works as expected, great work! 🎉
There’s a small visual issue: the space between the badge (number) and the "copy" text feels a bit awkward. While it’s not urgent, it would be better if the space between the badge and the text wasn’t underlined.

Screenshot 2025-01-06 at 10 50 37

client/src/features/ProjectPageV2/ProjectPageHeader/ProjectTemplateInfoBanner.tsx
```diff
@@ -0,0 +1,159 @@
...
+      isOpen={isOpen}
+      toggle={toggle}
+      size="lg"
+      centered
+    >
+      <ModalHeader toggle={toggle}>
+        <span className="fw-normal">{title} </span>
+        {project.namespace}/{project.slug}
+      </ModalHeader>
+      <ModalBody className={cx("overflow-y-scroll", styles.modalBody)}>
```
leafty, line 62: This breaks the rounded corners, see screenshot.

Screenshot 2025-01-06 at 11 12 38

To fix this, you can use the `<ScrollableModal>` component which has a scrollable body. (`client/src/components/modal/ScrollableModal.tsx`).

client/src/components/buttons/Button.tsx
ciyer: Fixed in a4a6efa3ea3fbe8354b109d94020ba41b5c51246

client/src/components/PrimaryAlert.module.scss
ciyer: Fixed in 63fa0009d3e88ffdc309ceda5e702f08178ff8f5

ciyer:
> It works as expected, great work! 🎉 There’s a small visual issue: the space between the badge (number) and the "copy" text feels a bit awkward. While it’s not urgent, it would be better if the space between the badge and the text wasn’t underlined.
> 
> Screenshot 2025-01-06 at 10 50 37

Fixed in a195420b5cc58a0eb6a2b64f060a927c881512ca

client/src/features/ProjectPageV2/ProjectPageHeader/ProjectTemplateInfoBanner.tsx
ciyer: Fixed in c86fcc9f6a73812cc3ac90a140cf7589af09cd61. Out of curiosity, which browser did you see this on? It looks correct in Chrome/Mac

ciyer:
Thanks for the reviews, everyone. I have made the requested changes.

client/src/features/ProjectPageV2/ProjectPageHeader/ProjectTemplateInfoBanner.tsx
```diff
@@ -0,0 +1,161 @@
...
```
leafty, line 63:
```suggestion
      <ModalBody className={styles.modalBody}>
```

The scrollbar should appear only if necessary (screenshots taken when `overflow-y-scroll` is removed):
Screenshot 2025-01-06 at 11 56 01
image

ciyer: Fixed in 44455c28e59968780314cc8c2bcddaa7c33649ac

leafty (approved):

PR #3433, opened 2024-12-09
feat: add user page design

andre-code:
Apply new user page design

/deploy renku=release-0.62.0 renku-data-services=main


client/src/features/usersV2/show/UserShow.tsx
```diff
@@ -82,66 +92,67 @@ export default function UserShow() {
...
-              />
-              <ItsYouBadge username={username} />
-            </div>
-          </div>
-          <p className="fst-italic">{`@${username}`}</p>
-        </div>
+  const information = (
+    <div className={cx("d-flex", "flex-column", "gap-3")}>
+      <div>
+        <p className={cx("align-items-center", "d-flex", "gap-2", "mb-0")}>
```
leafty, line 98: Just use `me-2` on the Icon, all these classes are not needed except `mb-0`.

Also, I don't think using a `p` element is appropriate here since the semantic flow is broken. The whole `Identifer: <username>` should be found inside a single `p` if possible. Otherwise just use `<div>` and `<span>`.

client/src/features/usersV2/show/UserShow.tsx
```diff
@@ -82,66 +92,67 @@ export default function UserShow() {
...
-              namespace={username}
-              namespaceKind="user"
-              pageParam="data_connectors_page"
-            />
-          </Col>
-        </Row>
-      </section>
+  return (
+    <ContainerWrap className="position-relative">
+      <EntityWatermark type="user" />
```
leafty, line 109: This may be broken just like for groups.

leafty (commented):
The design of the group and user pages do not look consistent, see the watermark here:

Screenshot 2024-12-17 at 11 48 28
Screenshot 2024-12-17 at 11 48 33

The alignment for the user watermark is not the same and there is no horizontal bar.

andre-code:
@leafty The margin is part of the icon's SVG itself, and I couldn't find a suitable style to remove it. In this case, the solution would be to use a custom image instead, which I recommend to address in a separate PR.

leafty:
> @leafty The margin is part of the icon's SVG itself, and I couldn't find a suitable style to remove it. In this case, the solution would be to use a custom image instead, which I recommend to address in a separate PR.

You could manually adjust the position for this case.

PR #3438, opened 2024-12-13
fix: validate container images

### LINKED ISSUE #3436: Properly validate docker images specified in environments
Currently there's no validation for docker images in custom environments, allowing users to e.g. input ` renku/renku:latest` (notice the space in the beginning) and other invalid image references, which then fail on launch.

https://github.com/SwissDataScienceCenter/renku-data-services/pull/574 has added validation in the backend and invalid docker references will now be rejected. The UI should consider doing some validation at the time of input to provide a nicer user experience.

leafty:
Fixes #3436.

/deploy


lorenzo-cavazzi (approved):
Lgtm :rocket:

PR #3443, opened 2024-12-17
fix: make modals work on the search page

leafty:
The modals for project and group creation now work on the search page.

Also, fix some issues with the navbar in Renku 2.0.

/deploy renku=release-0.62.0


client/src/features/rootV2/NavbarV2.tsx
```diff
@@ -55,59 +55,22 @@ function NavbarItemPlus() {
...
-              className="dropdown-item"
-              dataCy="navbar-group-new"
-              color="link"
-            >
-              Group
-            </CreateGroupButton>
-          )}
-        </DropdownItem>
+      <DropdownMenu end>
+        <DropdownItem onClick={toggleNewProject}>Project</DropdownItem>
```
lorenzo-cavazzi, line 72: Using the `CreateProjectV2Button` element should spare us handling the hashes here

lorenzo-cavazzi (approved):
Looks great! Should we reuse the project and group creation buttons here?
I've seen they haven't been used properly in the dashboard too, so I can handle that in a separate PR too

client/src/features/rootV2/NavbarV2.tsx
leafty: It was removed because it created a button within a button, which is not acceptable ->

```
validateDOMNesting(...): <button> cannot appear as a descendant of <button>.

<button tabindex="0" role="menuitem" class="p-0 dropdown-item"><button type="button" data-cy="navbar-project-new" class="dropdown-item btn btn-link">Project</button></button>
```

lorenzo-cavazzi (approved):

lorenzo-cavazzi (approved):

PR #3444, opened 2024-12-17
feat: update dashboard header images

andre-code:
PR to update dashboard header images in v2

/deploy


andre-code:
note: pending final imagenes

andre-code:
Fixed in https://github.com/SwissDataScienceCenter/renku-ui/pull/3474

PR #3463, opened 2024-12-23
feat: configure disk storage for Renku 2.0 sessions

### LINKED ISSUE #3451: Allow users to customize the amount of storage when starting a session
Currently, users can't define how much storage a session should have; it takes the default from the dropdown menu.
We should add the slider as we have it in Renku 1.0

Screenshot_20241219_122624

leafty:
Closes #3451.

Add an option to configure disk storage in 3 locations:
* Add launcher
* Modify launcher's resource class
* Custom launch

**Breaking change**: requires update to `renku-data-services`.

**Add launcher**
Screenshot 2024-12-23 at 15 30 04

**Modify launcher's resource class**
image

**Custom launch**
image

**Offcanvas warning**
This can happen if the disk storage was set through the API or if the admin changed the quotas on the resource pools.
image


andre-code (changes requested):
The session with the custom disk size work as expected, however I noticed a few things that could be improved:

- The disk size option isn’t visible until you change the resource class selector. It would be better if it was always visible when a resource class is selected. It only happens when the form is use for first time.
Screenshot 2025-01-06 at 15 05 06


- If the disk size exceeds the available limit, the input shows an error, but the error message isn’t displayed. We should make sure the error message is shown when this happens.
Screenshot 2025-01-06 at 14 50 07


- When trying to delete the current value with the keyboard, the toggle gets triggered instead, so it’s not possible to delete the value. This needs to be fixed so users can delete the value without triggering the toggle.
change disk


Also, I’m curious—why add the toggle to show the input instead of just keeping the input always visible? It could make the experience smoother for the user.

leafty:
> Also, I’m curious—why add the toggle to show the input instead of just keeping the input always visible? It could make the experience smoother for the user.

The question is: how to handle the difference between keeping the value unset (and thus using the default value from the resource class) and setting the value equal to the default for the resource class?

lorenzo-cavazzi:
After thinking about this, I wonder if we need to distinguish between default and manually set.
Once the session is started, it doesn't matter if the value was manually set or taken from the default. I agree with Andrea that leaving the slider always visible simplifies the UX.

leafty:
> After thinking about this, I wonder if we need to distinguish between default and manually set.

Yes, because the default value can be changed by an admin, while a value specifically set here will only be changed by the project members.

lorenzo-cavazzi:
> Yes, because the default value can be changed by an admin, while a value specifically set here will only be changed by the project members.

True, but would that make any difference to users with running sessions? Even if an admin changes the default, the running sessions won't be affected immediately. And we don't want to accidentally resize the disk while patching as well.
It seems safer (and easier) to use the default for the first input and not automatically modify that as long as the session lives

leafty:
Also, storing the value when not touched by users would prevent admins from "upgrading" the session classes. Right now, if an admin upgrades the default storage to be 4GB instead of 1GB, then all users without a value set would benefit from the upgrade. With your solution, no user would get the automatic upgrade to session storage.

lorenzo-cavazzi:
I totally agree with this:

> I think it would be good to merge a way to configure disk storage instead of insisting upon a better way to configure storage (you're free to patch it afterwards).

I might have lost some bits here and there; I thought my suggestions implied _less_ work, not more. And I'm afraid of the possible negative consequences of downsizing. Anyway, let's try to merge the changes and, if needed, we can adjust details later

leafty:
> And I'm afraid of the possible negative consequences of downsizing.

Why? This cannot affect any existing session.

andre-code (approved):
The code looks great and works as expected. It's intuitive, and the error messages are clear and helpful. Great job!

PR #3467, opened 2025-01-03
fix: redirect to newest group slug

### LINKED ISSUE #3385: Handle redirect to current slug
When a slug is updated in Renku 2.0, the old slug can still be used for GET queries (until it is claimed). In such cases, the UI should handle updating the browser URL to the new one.

Example:
1. Create a group, with slug `example-group`.
2. Update the group with a new slug `example-new-group`.
3. Manually load the group page with the old slug `example-group`.

The UI should change the URL to `example-new-group` but does not at the moment. Note that this should happen for groups, users, projects, etc.

leafty:
Fixes #3385.

Example:
* https://renku-ci-ui-3467.dev.renku.ch/v2/groups/test-group-hello -> redirects to https://renku-ci-ui-3467.dev.renku.ch/v2/groups/test-group
* https://renku-ci-ui-3467.dev.renku.ch/v2/projects/test-group-hello/test-project -> redirects to https://renku-ci-ui-3467.dev.renku.ch/v2/projects/test-group/test-project

/deploy renku=release-0.63.0


andre-code (commented):
If I got it right, when a project’s namespace slug changes, it should redirect to the new one, right?

I found an example where, after changing a group slug, the project still opens with the old namespace slug. Because of this, the namespace dropdown in the project settings shows up empty and the list include the current and old namespace slug. Does this sound like something that needs fixing?


if I open the link as anonymous, it redirects as expected.
 
url slug

leafty:
@andre-code this is because of 304 caching not being correct. I will try to make a fix in the backend.

andre-code (approved):
Lgtm 🚀

PR #3473, opened 2025-01-08
test: use port 3001 instead of 3000 for Cypress

lorenzo-cavazzi:
Currently, we use the same default port for both Cypress and telepresence. There is nothing wrong with that, but it makes running both at the same time a little more tricky.

This PR switches to using a different port for Cypress (3001 instead of 3000)

/deploy

### DIFF
--- a/.github/workflows/test-and-ci.yml
+++ b/.github/workflows/test-and-ci.yml
@@ -123,7 +123,7 @@ jobs:
         with:
           browser: chrome
           start: npm run client:preview
-          wait-on: http://localhost:3000/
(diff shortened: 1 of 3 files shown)


leafty (changes requested):
This PR is not necessary, and is actually having webpack run twice serving the same UI which could cause issues. If you are already running a UI, you can run tests by using the `cypress` script instead of the `e2e` script.

lorenzo-cavazzi:
@leafty That's a good point. I can use `cypress:headless` instead of `e2e:headless` in my workflow instead of going for the suggested changes

PR #3474, opened 2025-01-08
feat: apply early access changes

andre-code:
PR to apply early access changes

- Use the pink logo.
- Change the badge text to "Early Access."
- Modify the "Connected Services" link to "Integrations."
- Always show "Renku 2.0" in the menu.
- Add the Early Access banner.
- Change the background image in Dashboard v2.

/deploy


client/public/dashboardV2header.png
lorenzo-cavazzi, line 1: The image seems higher than necessary, with some empty space at the bottom. Since it is scaled with the current solution, should we remove the bottom empty space and use an SVG instead?

P.S. About the scaling, the dots are small on smaller screens; while enlarging the window, they get big, then smaller again, and then gradually bigger (see gif). Is that intended? I think we should either scale them gradually or fix two/three sizes, but not mix the two styles (low priority).
Peek 2025-01-09 10-14

client/src/components/earlyAccessBanner/EarlyAccessBanner.module.scss
```diff
@@ -0,0 +1,7 @@
+.EarlyAccessBannerBtn {
+  &:hover,
+  &:active {
+    background-color: var(--bs-rk-green) !important;
```
lorenzo-cavazzi, line 4: We should use the hover color instead

client/src/components/earlyAccessBanner/EarlyAccessBanner.tsx
```diff
@@ -0,0 +1,154 @@
...
+    >
+      <Container
+        className={cx(
+          "w-100",
+          "d-flex",
+          "justify-content-between",
+          "align-items-center"
+        )}
+      >
+        <div className="d-none d-sm-block d-lg-none">
```
lorenzo-cavazzi, line 100:
```suggestion
        <div className={cx("d-none", "d-sm-block", "d-lg-none")}>
```

client/src/components/earlyAccessBanner/EarlyAccessBanner.tsx
```diff
@@ -0,0 +1,154 @@
...
+          <img
+            src={themeAssets[theme].dotsImgUrl}
+            className={cx("img-fluid", "object-fit-cover")}
+            style={{ width: "83px", height: "50px" }}
+            alt="Small image"
+          />
+        </div>
+        <div className="d-none d-lg-block">
```
lorenzo-cavazzi, line 108:
```suggestion
        <div className={cx("d-none", "d-lg-block")}>
```

client/src/components/earlyAccessBanner/EarlyAccessBanner.tsx
```diff
@@ -0,0 +1,154 @@
...
+            "flex-column",
+            "flex-md-row"
+          )}
+        >
+          <img
+            src={themeAssets[theme].logoImgUrl}
+            alt="Renku 2.0"
+            height="25"
+          />
+          <h5 className="mt-1 mb-0">EARLY ACCESS!</h5>
```
lorenzo-cavazzi, line 130:
```suggestion
          <h5 className={cx("mt-1", "mb-0")}>EARLY ACCESS!</h5>
```

lorenzo-cavazzi (changes requested):
Great to see the final steps towards early access! :star_struck: 

The code looks good, just a couple of minor details (see inline).
For the dots image on the dashboard, feel free to ignore the part on scaling/resizing if changing it is time-consuming.

client/src/components/earlyAccessBanner/EarlyAccessBanner.module.scss
andre-code: Could you clarify what you mean?

lorenzo-cavazzi: In Renku V1, that's the button's primary green color (#009568), not the hover color (#007a6c). That is usually imported as `$rk-green-hover`

client/public/dashboardV2header.png
lorenzo-cavazzi: As discussed offline, let's keep this for later

andre-code:
Thanks, @lorenzo-cavazzi , for the review!
I’ve updated the button in the banner with the correct color and made a small adjustment requested by the PO to align the banner's location with the main navigation. Let me know if there’s anything else!

lorenzo-cavazzi (approved):
Lgtm! :rocket:

PR #3477, opened 2025-01-10
fix: show all data connector password fields

### LINKED ISSUE #3476: Fix data connector password fields
Data connector options are created using the verbose data from the `storage_schema` endpoint and a few UI customizations from the constant `CLOUD_OPTIONS_OVERRIDE`. The `examples` property was used to create "soft" dropdown (inputs with datalist), but recent changes used the same property to adjust the `friendlyName` and `help` text for the providers.
As a side effect, some options aren't displayed correctly anymore -- here is an example from S3 / AWS
Screenshot_20250109_150400

We should instead use another field that doesn't collide with the backend, following what we are doing with `friendlyName` and `position`, and revert the changes creating the collisions (E.G. on `PasswordOptionItem` the bits on `const inputName = option.filteredExamples?.length > 0 && [...]`; on `AddStorageOptions` the bits on `<OptionTruncatedText text={ inputType !== "dropdown" && o.filteredExamples?.length > 0 [...]`; please look for other similar changes from the same PR)

andre-code:
This change introduces the CLOUD_OPTIONS_PROVIDER_OVERRIDE constant, enabling option overrides based on schema and provider. It replaces the need for passing the customisation using the example parameters.

Before / after
Screenshot 2025-01-10 at 10 37 16

fix #3476 

/deploy


client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
lorenzo-cavazzi, line 1: Changes should also be reverted in the `help` section to prevent some fields from showing the help text from an example instead of the option.

I think it's enough to swap in `<OptionTruncatedText text={o.help} />` instead of the current
```
<OptionTruncatedText
              text={
                inputType !== "dropdown" && o.filteredExamples?.length > 0
                  ? o.filteredExamples[0]?.help
                  : o.help
              }
            />
```

But give it a try to be sure

image

client/src/features/project/components/cloudStorage/projectCloudStorage.constants.ts
```diff
@@ -206,6 +174,44 @@ export const CLOUD_OPTIONS_OVERRIDE = {
   },
 } as Record<string, Record<string, Partial<CloudStorageSchemaOptions>>>;
 
+export const CLOUD_OPTIONS_PROVIDER_OVERRIDE = {
```
lorenzo-cavazzi, line 177: The solution looks good :+1: 
Could you add an example in the test file? :pray:

lorenzo-cavazzi (changes requested):
Nice solution! :+1: 
There is another code change necessary to show the correct help text. We can then merge it :rocket:

andre-code:
thanks @lorenzo-cavazzi for your review,
I have added a unit test for the new constant and updated the e2e tests to ensure that the password values are displayed correctly. Let me know if there's anything else you'd like me to address!

PR #3483, opened 2025-01-15
fix: use placeholders for data connectors

lorenzo-cavazzi:
This PR reverts the data connector option default values to the previous behavior.
It also prevents showing objects converted to strings as placeholders.

/deploy

### DIFF
--- a/client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
+++ b/client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
@@ -465,7 +465,12 @@ function PasswordOptionItem({
               id={option.name}
               type={showPassword ? "text" : "password"}
               className={cx("form-control", "rounded-0", "rounded-start")}
-              placeholder={option.convertedDefault?.toString() ?? ""}
+              placeholder={
+                option.convertedDefault &&
+                option.convertedDefault?.toString() !== "[object Object]"
+                  ? option.convertedDefault?.toString()
+                  : ""
+              }
               onChange={(e) => {
                 field.onChange(e);
                 onFieldValueChange(option.name, e.target.value);
@@ -530,7 +535,12 @@ function PasswordOptionItemStored({
             readOnly={true}
             type="text"
             className={cx("form-control", "rounded-0", "rounded-start")}
-            placeholder={option.convertedDefault?.toString() ?? ""}
+            placeholder={
+              option.convertedDefault &&
+              option.convertedDefault?.toString() !== "[object Object]"
+                ? option.convertedDefault?.toString()
+                : ""
+            }
             onChange={() => {}}
           />
         )}
@@ -580,7 +590,8 @@ function InputOptionItem({
                 type={inputType}
                 className="form-control"
                 placeholder={
-                  option.convertedDefault
+                  option.convertedDefault &&
+                  option.convertedDefault?.toString() !== "[object Object]"
                     ? option.convertedDefault?.toString()
                     : inputType === "dropdown"
                     ? option.filteredExamples[0].value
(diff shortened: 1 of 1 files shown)


client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
```diff
@@ -465,7 +465,12 @@ function PasswordOptionItem({
               id={option.name}
               type={showPassword ? "text" : "password"}
               className={cx("form-control", "rounded-0", "rounded-start")}
-              placeholder={option.convertedDefault?.toString() ?? ""}
+              placeholder={
+                option.convertedDefault &&
+                option.convertedDefault?.toString() !== "[object Object]"
```
leafty, line 470: Is there no type information to prevent this? This would also display unusable information when `typeof option.convertedDefault === "function"` or when `typeof option.convertedDefault === "symbol"`.

lorenzo-cavazzi: I couldn't find a good rule to generalize.
There are convoluted objects as defaults, but never functions nor symbols

leafty: Why not use `typeof` then? It would be more robust than checking string representation.

andre-code (approved):
Thanks for fixing this! The change makes sense, the default values now work as expected, and I've reviewed and tested the connection for both personal and shared Polybox, and it's still working well.

client/src/features/project/components/cloudStorage/AddOrEditCloudStorage.tsx
leafty: Please fix now, there are too many cases when this would fail, e.g. on array objects.

leafty: You just have to replace the current code with a `typeof x === "object"` test.

lorenzo-cavazzi: It does not work. `option.convertedDefault` is a string already, it would test false for `typeof x === "object"`.
You can try it yourself

lorenzo-cavazzi: This is an example. Your suggestion does not work here.
I agree we could find better solutions, but they require deeper changes. Can we keep those for later?

image

leafty: Ok, but I'm not gonna appreciate it if this is not fixed asap.

lorenzo-cavazzi: The data quality we get from the `storage_schema` endpoint isn't great, and that's coming straight from `Rclone` (I guess multiple contributions from different people ended up in various attributes/formats for default values).
Harmonizing them all isn't worth the effort since we use them as placeholders, so a function deals with the most common defaults and converts them to numbers/booleans/strings.
Improving the situation here would require dealing with this issue in the first place. If you feel that's important, you are welcome to improve the code.

leafty: This is very different from understanding that comparing to `[Object object]` is not robust, because many objects have a different representation. The issue is that the `convertedDefault` variable is not robust and the code which creates it needs to be updated unless we want to deal with more bugs in the future.

lorenzo-cavazzi: I understand very well the difference between testing `myObject.toString() === [object Object]"` and `typeof myObject === "object"`. And I understand the solution for `convertedDefault` isn't the best.
But I do not understand the request to convert a quick bugfix PR into overhauling the logic to handle non-standard values for cases that never appear.
Again, feel free to improve the logic whenever you have time. That's _always_ welcome for every part of the code.

leafty: Because replacing buggy code with another buggy code is not going to be sustainable. Doing `=== "[object Object]"` is a big code smell which indicates that there is a problem with the code and that should not happen. At the very least it should be tracked and solved quickly.

PR #3492, opened 2025-01-17
Upgrade ck-editor

ciyer:
/deploy


ciyer:
We have decided to move away from CK editor: closing.

PR #3506, opened 2025-01-24
chore: run UI client without telepresence

olevski:
All you need to do is `RENKU_BASE_URL=https://dev.renku.ch npm run start` and the ui will talk to the backend from that deployment.

Note that the session cookies issued by the gateway are tied to the host or something like that. So when you run on localhost you cannot actually log in. But we can add (or there already is) a setting on the gateway that will make it issue weaker cookies in dev. Then one can just run `/deploy values=gateway.weakCookies=true` in a PR and point your local ui client at that deployment and everything should work. Will do just a bit more testing to see that that is the case.

But I can already look at the renku home page and even complete the login flow wihtout doing anything on the gateway. But for now the session cookie is not there.

### DIFF
--- a/server/src/routes/apis.ts
+++ b/server/src/routes/apis.ts
@@ -36,16 +36,16 @@ import { getUserData } from "./helperFunctions";
 const proxyMiddleware = createProxyMiddleware({
   // set gateway as target
   target: config.deployment.gatewayUrl,
-  changeOrigin: true,
+  changeOrigin: false,
   proxyTimeout: config.server.proxyTimeout,
   timeout: config.server.proxyTimeout,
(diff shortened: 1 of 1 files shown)


leafty:
@olevski I will have a look 👀 .

olevski:
@leafty this is still wip

leafty:
OK, I am not sure how this can work without being limited by CORS, Secure Context,  secure cookies, etc.

olevski:
> OK, I am not sure how this can work without being limited by CORS, Secure Context, secure cookies, etc.

Yeah that is what I am trying to figure out rn.

PR #3509, opened 2025-01-27
dev: auto-detect telepresence context

leafty:
In `client/`, you can now do:

```bash
$ ./run-telepresence.sh

Detected current PR: 3491
Exchanging k8s deployments for the following context/namespace: 
renku-dev/renku-ci-ui-3491
The deployment URL is expected to be: 
https://renku-ci-ui-3491.dev.renku.ch
[...]
```

The script assumes that the GitHub CLI is available as `gh`, but this can be overriden with `GH_CLI="github" ./run-telepresence.sh`.
Also, the script does not crash if the CLI is not available or if the current branch does not correspond to a PR.

### DIFF
--- a/client/run-telepresence.sh
+++ b/client/run-telepresence.sh
@@ -99,6 +99,15 @@ then
   done
 fi
 
+# Try to guess the namespace using the GitHub CLI
+GH_CLI="${GH_CLI:-gh}"
+PR_NUMBER=$($GH_CLI pr view --json number --jq .number || true)
+if [[ $PR_NUMBER ]]
+then
+  echo "Detected current PR: ${PR_NUMBER}"
+  NAMESPACE=$PR_NUMBER
+fi
+
 # Get the namespaces when not set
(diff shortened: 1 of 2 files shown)


lorenzo-cavazzi (changes requested):
The change makes sense and has limited impact on users without the GitHub cli installed (I assume it will print an error; perhaps swallowing that would be cleaner).

We don't need to execute that code when users provide a `NAMESPACE`. I would move the code inside the conditional `if [[ ! $NAMESPACE ]]`

andre-code:
I agree with @Lorenzo about moving this code if a NAMESPACE doesn’t exist. Do you think it makes sense to add this to the telepresence script in the server folder? Even though we don’t work there much yet, it feels like it should be an option too.

andre-code (approved):
Great addition! It remains flexible for other cases when I want to set up a different PR than the one recognized by Git CLI

PR #3513, opened 2025-01-28
refactor: use API code generators for sessions v2

leafty:
Merging into the `build/session-env-builders` feature branch.

This refactor updates the sessions v2 API to use generated code from the API spec. This is done to allow for the API files to be easily updated after changes from the backend services.

/deploy


ciyer (approved):
Looks good! This makes things easier to work with, and the split of the API between sessions and session launchers makes things more logical as well.

leafty:
> Looks good! This makes things easier to work with, and the split of the API between sessions and session launchers makes things more logical as well.

The split exists because the API spec itself is split between two files as well.

PR #3515, opened 2025-01-29
fix: add allowed hosts to vite config

leafty:
See:
* https://github.com/vitejs/vite/security/advisories/GHSA-vg6x-rcgg-rjx6
* https://vite.dev/config/server-options.html#server-allowedhosts


lorenzo-cavazzi (approved):
:+1:

PR #3516, opened 2025-01-29
fix: correct alert test in maintenance.spec.ts

leafty:
Fixes a flakey test in `maintenance.spec.ts` introduced by the "Early access" banner.

### DIFF
--- a/tests/cypress/e2e/maintenance.spec.ts
+++ b/tests/cypress/e2e/maintenance.spec.ts
@@ -74,7 +74,7 @@ describe("display the status page", () => {
     fixtures.config().versions().userNone().getStatuspageInfo();
     cy.visit("/");
     cy.wait("@getStatuspageInfo");
-    cy.get(".alert").should("not.exist");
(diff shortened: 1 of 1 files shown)


tests/cypress/e2e/maintenance.spec.ts
```diff
@@ -74,7 +74,7 @@ describe("display the status page", () => {
     fixtures.config().versions().userNone().getStatuspageInfo();
     cy.visit("/");
     cy.wait("@getStatuspageInfo");
-    cy.get(".alert").should("not.exist");
+    cy.get(".alert").contains("Ongoing incident").should("not.exist");
```
leafty, line 77: Note: the test can be made to fail all the time with:

```
    cy.contains("Try it out").should("be.visible");
    cy.get(".alert").should("not.exist");
```

This ensures that Cypress will have the whole landing page rendered when testing that the Incident banner exists.

lorenzo-cavazzi (approved):
Lgtm :rocket:

PR #3525, opened 2025-02-04
feat: create data connector in project

olevski:
This is the initial implementation.

Things that are missing:
- no tests
- the way that data connectors in projects are displayed is not right
- the backend is still a bit wonky and will not support everything (i.e. launching sessions fails)

/deploy renku-data-services=feat-add-project-as-dc-owner-pt2

### DIFF
--- a/client/src/features/projectsV2/api/namespace.enhanced-api.ts
+++ b/client/src/features/projectsV2/api/namespace.enhanced-api.ts
@@ -1,0 +1,12 @@
+import {
+  NamespaceKind as NamespaceKindOrig,
+  NamespaceResponse as NamespaceResponseOrig,
+} from "./namespace.api.ts";
+
+export type NamespaceKind = NamespaceKindOrig | "project";
+export interface NamespaceResponse
+  extends Omit<NamespaceResponseOrig, "namespace_kind"> {
+  namespace_kind: NamespaceKind;
+}
+export type NamespaceResponseList = NamespaceResponse[];
+export type GetNamespacesApiResponse = NamespaceResponseList;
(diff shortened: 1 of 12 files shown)


client/src/features/projectsV2/fields/ProjectNamespaceFormField.tsx
```diff
@@ -249,6 +251,7 @@ export default function ProjectNamespaceFormField<T extends FieldValues>({
               id={`${entityName}-namespace`}
               inputId={`${entityName}-namespace-input`}
               onChange={(newValue) => field.onChange(newValue?.slug)}
+              project={undefined}
```
lorenzo-cavazzi, line 254:
```suggestion
```
This line is unnecessary since the `project` prop is not mandatory.

client/src/features/projectsV2/fields/ProjectNamespaceFormField.tsx
lorenzo-cavazzi, line 1: I'm surprised this component named "_Project_ NamespaceFormField" is already used for Data Connectors namespaces. I was tempted to suggest creating another component, or at least renaming and moving it to make it clear it's not only for Project namespaces.
We can keep it as is for now. Still, if the logic to extract Data connector namespaces gets more complicated later in the build, we might consider separating it so that we have less code for each entity (project vs data connector).

lorenzo-cavazzi (commented):
The PR changes make overall sense so far :+1:  Some early comments inline.

In short, there is a better way to add types from the updated APIs from other repos branches, and there is some less-than-ideal prop drilling here -- tho the components you are touching here weren't the most straightforward and your solution looks acceptable in this case.

client/src/features/projectsV2/api/namespace.enhanced-api.ts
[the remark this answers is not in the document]
olevski: They are not changed in the backend. But I think I will have to update them in the backend. For now I am kind of forcing the project namespace in a few places. And that is why I need this. I will remove this when it is fully integrated in the backend. I think I will be able to do this during the build.

olevski:
This change is part of the following stack:

- #3525 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

lorenzo-cavazzi (approved):
Lgtm! :rocket:

PR #3511, opened 2025-01-27
feat: add app params for image builders

leafty:
Configure the UI for session image builds based on the helm chart values.

* If `IMAGE_BUILDERS_ENABLED==true`, then the "Create from code" option for session launchers is available.
* If `IMAGE_BUILDERS_ENABLED==false`, then the "Create from code" option for session launchers is not shown to users.

Deployments:
* Feature enabled -> https://renku-ci-ui-3511.dev.renku.ch/
* Feature disabled -> https://renku-ci-ui-3538.dev.renku.ch/

/deploy #notest renku=build/session-env-builders renku-data-services=kpack-resources renku-notebooks=leafty/shipwright-buildrun-cache extra-values=dataService.imageBuilders.enabled=true,dataService.imageBuilders.pushSecretName=flora-docker-secret,dataService.imageBuilders.buildRunRetentionAfterFailedSeconds=86400,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/flora-dev/

PR #3529, opened 2025-02-13
feat: get build status and trigger builds

leafty:
PR stack:
* (this) #3529
  * #3534
    * #3536
      * #3511

Update the UI for session environments created from code repositories.

* Get the build status and show it in the session offcanvas
* Allow users with edit permissions on the project to trigger builds and to cancel in-progress builds.

image

/deploy #notest renku=build/session-env-builders renku-data-services=kpack-resources renku-notebooks=leafty/shipwright-buildrun-cache extra-values=dataService.imageBuilders.enabled=true,dataService.imageBuilders.pushSecretName=flora-docker-secret,dataService.imageBuilders.buildRunRetentionAfterFailedSeconds=86400,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/flora-dev/

### DIFF
--- a/client/src/features/sessionsV2/SessionView/EnvironmentCard.tsx
+++ b/client/src/features/sessionsV2/SessionView/EnvironmentCard.tsx
@@ -16,20 +16,24 @@
  * limitations under the License.
  */
 
+import { skipToken } from "@reduxjs/toolkit/query";
 import cx from "classnames";
 import { ReactNode } from "react";
 import {
+  Bricks,
   CircleFill,
   Clock,
   Globe2,
   Link45deg,
-  Tools,
 } from "react-bootstrap-icons";
 import { Badge, Card, CardBody, Col, Row } from "reactstrap";
 
+import { Loader } from "../../../components/Loader";
+import { RtkOrNotebooksError } from "../../../components/errors/RtkErrorAlert";
 import { ErrorLabel } from "../../../components/formlabels/FormLabels";
 import { toHumanDateTime } from "../../../utils/helpers/DateTimeUtils";
 import type { SessionLauncher } from "../api/sessionLaunchersV2.api";
+import { useGetEnvironmentsByEnvironmentIdBuildsQuery as useGetBuildsQuery } from "../api/sessionLaunchersV2.api";
 import { BUILDER_IMAGE_NOT_READY_VALUE } from "../session.constants";
 import { safeStringify } from "../session.utils";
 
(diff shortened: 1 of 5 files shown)


client/src/features/sessionsV2/SessionView/EnvironmentCard.tsx
```diff
@@ -179,6 +220,32 @@ function CustomBuildEnvironmentValues({
...
+        ) : (
+          <div className="d-block">
+            <label className={cx("text-nowrap", "mb-0", "me-2")}>
+              Last build status:
+            </label>
+            <span>
+              <BuildStatusBadge status={lastBuild.status} />
+            </span>
+          </div>
+        )}
```
lorenzo-cavazzi, line 247: Question: would it make sense to keep the "label + span" pattern to have more consistency? E.G. always `<label>Last build</label>` and then "loading" | "no builds yet" | badge.
Of course, it doesn't apply in the case of an error.

lorenzo-cavazzi (commented):
Nice addition!
And the code is already great! I dropped a couple of suggestions/questions inline, only minor details.

client/src/features/sessionsV2/SessionView/EnvironmentCard.tsx
```diff
@@ -163,6 +171,39 @@ function CustomBuildEnvironmentValues({
...
+  );
+
+  const lastBuild = builds?.at(0);
+
+  sessionLaunchersV2Api.endpoints.getEnvironmentsByEnvironmentIdBuilds.useQuerySubscription(
+    lastBuild?.status === "in_progress"
+      ? { environmentId: environment.id }
+      : skipToken,
+    {
+      pollingInterval: 1_000,
```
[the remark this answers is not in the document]
leafty: Why? The backend uses the kubernetes cache so it cannot overload the kubernetes API/etcd.

client/src/features/sessionsV2/SessionView/EnvironmentCard.tsx
leafty: I am not sure I feel a need for consistency here. For example, we can have this:
image

This does not look out-of-place to me.

Also, on another note, `<label>` outside of a `<form>` should be forbidden. `<label>` is an HTML element made to label form inputs and not other things. In this case the `<dl>` element would be the most fitting (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl).

lorenzo-cavazzi (approved):

PR #3549, opened 2025-03-03
refine: indicate the resource pool name in the session offcanvas

### LINKED ISSUE #3539: Indicate the resource pool in the session offcanvas
> On [renkulab.io](http://renkulab.io/), the fact that the resource pool is not noted here makes it annoying to understand which class exactly is being used. The fact we give similar names to classes in different pools does not help. Maybe we can also list the pool here?

Image

See: https://swiss-data-science.slack.com/archives/C03E66DHUG3/p1740130937234799

leafty:
Closes #3539.

/deploy #notest

### DIFF
--- a/client/src/features/session/components/SessionsList.tsx
+++ b/client/src/features/session/components/SessionsList.tsx
@@ -314,6 +314,7 @@ function SessionRowProject({ annotations }: SessionRowProjectProps) {
 }
 
 export interface SessionLauncherResources {
+  poolName?: string;
   name?: string;
   cpu: number;
   memory: number;
@@ -343,7 +344,7 @@ export function SessionRowResourceRequests({
             <span className="fw-bold">
               {value} {(key === "memory" || key === "storage") && "GB "}
             </span>
-            {key !== "name" && key}
+            {key !== "name" && key !== "poolName" && key}
           </span>
           {entries.length - 1 === index ? " " : " | "}
         </span>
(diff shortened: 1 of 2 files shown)


lorenzo-cavazzi (approved):
The change itself looks good :+1: 

I wonder if we should style the content better, or add a label to indicate what is what. It's hard to understand what's `default | default` here -- ofc it can be other names, but it's still not intuitive.
Screenshot_20250304_134004

We can either create a follow-up issue (and involve the designer), or change it here if you have an idea.

lorenzo-cavazzi:
This would already be better 
Screenshot_20250304_134551

Or perhaps this
Screenshot_20250304_134608

leafty:
I see. I will have to change the logic of the component to do this, but that should be nicer anyway.

client/src/features/session/components/SessionsList.tsx
```diff
@@ -330,22 +331,44 @@ export function SessionRowResourceRequests({
   if (!resourceRequests) {
     return null;
   }
-  const entries = Object.entries(resourceRequests);
-  if (entries.length == 0) {
+  // const entries = Object.entries(resourceRequests);
```
lorenzo-cavazzi, line 334:
```suggestion
```

lorenzo-cavazzi (changes requested):
I like the solution you picked! :rocket: 
Tiny little change before merging

client/src/features/session/components/SessionsList.tsx
leafty: Oh thanks 🤦‍♀️

lorenzo-cavazzi (approved):

PR #3553, opened 2025-03-05
minor: show descriptive text in the appropriate style

ciyer:
Descriptive text like, "Define interactive environments...," "Connect code repositories...," etc.  should be displayed in a specific style.

image

image

image

/deploy

### DIFF
--- a/client/src/styles/bootstrap/_custom_bootstrap_variables2.0.scss
+++ b/client/src/styles/bootstrap/_custom_bootstrap_variables2.0.scss
@@ -53,6 +53,7 @@ $rk-success: #4d7c0f;
 $rk-danger: #991b1b;
 $rk-warning: #854d0e;
 
+$rk-text-describe: mix(#16192c, $rk-white, 75%);
 $rk-info-text: #1e40af; // contrast ratio 7:1 for blue bg (#cfe3ea)
 $rk-success-text: #1a2e05;
 $rk-danger-text: #450a0a;
--- a/client/src/styles/bootstrap/_custom_bootstrap_variables2.0.scss
+++ b/client/src/styles/bootstrap/_custom_bootstrap_variables2.0.scss
@@ -53,7 +53,6 @@ $rk-success: #4d7c0f;
 $rk-danger: #991b1b;
 $rk-warning: #854d0e;
 
-$rk-text-describe: mix(#16192c, $rk-white, 75%);
 $rk-info-text: #1e40af; // contrast ratio 7:1 for blue bg (#cfe3ea)
 $rk-success-text: #1a2e05;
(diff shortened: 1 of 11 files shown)


client/src/styles/bootstrap/_custom_bootstrap_variables2.0.scss
leafty, line 1: This file should not be edited, it is not part of the Renku 2.0 style stack. It is only imported into `client/src/components/buttons/Buttons.module.scss` and should most likely be removed in the future.

I understand this can be confusing, but I think the refactoring of the Renku 2.0 styles forgot to remove this file. The file which should be changed is `client/src/styles/renku_bootstrap_customization.scss`.

leafty (changes requested):
The new color needs to be added in `client/src/styles/renku_bootstrap_customization.scss`: https://github.com/SwissDataScienceCenter/renku-ui/blob/07c67f9ece10e3847665474e298ae9b2be5aaf7b/client/src/styles/renku_bootstrap_customization.scss#L23-L25

With this change in the style colors, you can then have the color added to any DOM node with `className="text-description"` (if `description` is the name of the color).

This will also simplify the changes, as no new `.scss` file needs to be added and imported.

client/src/styles/bootstrap/_custom_bootstrap_variables2.0.scss
```diff
@@ -53,6 +53,7 @@ $rk-success: #4d7c0f;
 $rk-danger: #991b1b;
 $rk-warning: #854d0e;
 
+$rk-text-describe: mix(#16192c, $rk-white, 75%);
```
lorenzo-cavazzi, line 56: Should we use a simpler color definition? My browser shows it as `#505361`.

Question: I'm wondering if we should consider picking an existing gray shade instead of introducing a new color. That would better match the existing color scheme. It seems hard to derive a generic rule here: why use a unique color only for some description? When should devs use it? We would need to add it to the storybook colors page too

lorenzo-cavazzi: No worries. And thanks for the reference!
In this case, the relevant part is the Bootstrap classes; it says `text-muted`, which should map to `text-body-secondary`.
@ciyer We can avoid introducing new colors or CSS files and apply the `text-body-secondary` to those descriptions

client/src/styles/bootstrap/_custom_bootstrap_variables2.0.scss
ciyer: After discussion, we have decided to use a style that already exists.

leafty (approved):
LGTM, great! 👏 🎉

PR #3557, opened 2025-03-06
refactor: cleanup unused searchV2 slice

lorenzo-cavazzi:
These should be just leftovers.

/deploy


leafty:
??? This code is in active use by the search page.

lorenzo-cavazzi:
Ehm... Yes, it is. My bad :facepalm: 
I was doing some (unrelated) cleanup for projects and groups and accidentally removed the searchV2 slice, then thought `buildSearchQuery` wasn't used anywhere.

PR #3563, opened 2025-03-07
chore: update issues template

lorenzo-cavazzi:
This matches better what we suggest in the ["Contribute" wiki page](https://github.com/SwissDataScienceCenter/renku-ui/wiki/Contribute).


leafty (approved):
Looks good.

PR #3564, opened 2025-03-07
chore: update all actions to use ubuntu-24.04

leafty:
This PR is similar to https://github.com/SwissDataScienceCenter/renku/pull/3938.

/deploy renku=build/session-env-builders


ciyer (approved):
Yes, makes sense to stay on the current LTS version.

PR #3565, opened 2025-03-07
build: use current LTS version of NodeJS

leafty:
See release schedule: https://nodejs.org/en/about/previous-releases.


lorenzo-cavazzi (approved):
Lgtm :+1:

PR #3566, opened 2025-03-07
build: update pretty-quick

ciyer:
A bug in how pretty-quick finds the git directory was fixed in 3.3.0:

  https://github.com/prettier/pretty-quick/pull/183


client/package.json
leafty, line 1: Can you update it everywhere? 🙏

ciyer: You are right! Done in 38b5ebc8f42bb09f3acc035f868f122ee9a1ae32

leafty (approved):

PR #3570, opened 2025-03-10
minor: move add member button to right

### LINKED ISSUE #3562: Add Members ➕ button in non-standard location
The plus button on sections is usually in the top right corner. In the settings for both the project and group pages, in the Members section, the plus button is next to the header text instead.

Image

ciyer:
image

image

Fix #3562

/deploy


leafty (approved):
LGTM

Issue #3559, opened 2025-03-07
The create project modal is broken if shown on a project's settings page


leafty:
https://github.com/user-attachments/assets/01fd7f0f-00ca-48a5-8280-0084bfd99104

PR #3571, opened 2025-03-10
Prevent confusion between project create and project settings forms

ciyer:
Fix # #3559
/deploy


client/src/features/projectsV2/fields/ProjectVisibilityFormField.tsx
leafty, line 1: This fix should be propagated to all the fields in the form, not just visibility.

leafty (commented):
Code is good 👍 , it would be nice to make this fix for all the fields in the projects forms.

client/src/features/ProjectPageV2/ProjectPageHeader/ProjectCopyModal.tsx
```diff
@@ -132,7 +134,12 @@ export default function ProjectCopyModal({
...
             </div>
           )}
-          <ProjectNameFormField control={control} errors={errors} name="name" />
+          <ProjectNameFormField
+            control={control}
+            errors={errors}
+            formId={formId}
+            name="name"
+          />
           <ProjectNamespaceFormField
```
leafty, line 143: The same thing should be done here too, see screenshot:

image

leafty (commented):
Great, I think the namespace input should also have it too, see below.

leafty (approved):
LGTM

PR #3573, opened 2025-03-10
fix: do not set z-index for session actions

### LINKED ISSUE #3569: Dashboard session listing z-index issue
## Description

The session action menu is rendered under the action button.

Image

## Reproduce

Start **multiple** sessions (the issue will not be visible with just one session running), then go to the dashboard and click on the menu of the first session.

leafty:
Closes #3569.

image

/deploy renku=build/session-env-builders


ciyer (approved):
LGTM!

PR #3587, opened 2025-03-12
ci: do not run legacy tests by default

lorenzo-cavazzi:
We can now disable running legacy acceptance tests by default

/deploy


leafty (approved):
LGTM

leafty:
Actually, `v2` tests do not run at all!

PR #3480, opened 2025-01-14
refactor(client): some migration step for `react-router@v6`

leafty:
Affected routes:

* `/projects/all`
* `/projects/*`

See:
* #3028
* #3041
* #3151
* #3466

/deploy

PR #3466, opened 2025-01-03
refactor(client): some migration step for `react-router@v6`

leafty:
Setup CompatRoute for dataset routes.

* `/projects/<project>/datasets/*`

See:
* #3028
* #3041
* #3151

/deploy

PR #3179, opened 2024-06-03
refactor: some migration step for `react-router@v6`

leafty:
Refactors all the routes under `/projects` to use `react-router-dom-v5-compat`.

See:
* #3028
* #3041
* #3151
* #3466
* #3480

/deploy #legacy

PR #3151, opened 2024-05-28
refactor(client): some migration step for `react-router@v6`

leafty:
Setup CompatRoute for some top-level routes.

Setup CompatRoute for:
* `/search`
* `/inactive-kg-projects`
* `/projects`
* `/projects/starred`
* `/projects/all`
* `/projects/new`

See:
* #3028
* #3041

/deploy renku=master

PR #3589, opened 2025-03-13
refactor: some migration step for `react-router@v6`

leafty:
Use the `Link` component from `react-router-dom-v5-compat` almost everywhere.

See:
* #3028
* #3041
* #3151
* #3466
* #3480
* #3179

Saved action run: https://github.com/SwissDataScienceCenter/renku-ui/actions/runs/13894378941

/deploy #legacy renku=release-0.66.0


ciyer (approved):
This is a good step in this direction!

PR #3605, opened 2025-03-20
feat: default namespace when creating a project from group page

### LINKED ISSUE #3550: Creating a new project from the group page defaults to the user namespace
This is tested from `release-0.65.0`

To reproduce:
- create group
- go on the group page
- click the `+` sign to  create a project in the group
- the modal that opens to create the project has the user namespace as the default - but it should be the group, because it was triggered from the group page

Another issue - probably related - is that when you click on the button to create a project from the group page you get navigated to the home page (i.e. the users dashboard) and then the modal to create the project in the group opens. So it could be that this navigation to the home page away from the group page as the modal opens that is causing trouble.

Image

ciyer:
Fix #3550 

/deploy


client/src/features/projectsV2/list/ProjectV2ListDisplay.tsx
```diff
@@ -221,7 +223,8 @@ function AddButtonForGroupNamespace({ namespace }: { namespace: string }) {
             "ms-auto",
             "my-auto"
           )}
-          to="/v2/projects/new"
+          data-cy="group-create-project-button"
+          to={{ hash: PROJECT_CREATION_HASH }}
```
leafty, line 227:
```suggestion
          to={{ hash: PROJECT_CREATION_HASH, search: location.search }}
```

with:

```tsx
const location = useLocation();
```

client/src/features/projectsV2/list/ProjectV2ListDisplay.tsx
```diff
@@ -245,7 +248,7 @@ function AddButtonForUserNamespace({ namespace }: { namespace: string }) {
           "ms-auto",
           "my-auto"
         )}
-        to="/v2/projects/new"
+        to={{ hash: PROJECT_CREATION_HASH }}
```
leafty, line 251:
```suggestion
        to={{ hash: PROJECT_CREATION_HASH, search: location.search }}
```

ciyer:
> Looks good overall! Just noticed one thing, if you're on the Group Settings page and click "+ Project" from the main menu, it defaults to the user namespace instead of the group. Might be worth checking if the user is in a group context and setting that as the default.

Thanks for pointing this out. This has been fixed.

client/src/utils/customHooks/useLocationSearchParams.hook.ts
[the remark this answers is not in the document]
leafty: But clearing the hash is something that should happen when it comes to URL and browser history semantics.

How come that would be possible anyway?!? Changing the search params means clicking on a pagination link, which is under the modal, so that is not allowed.

Please remove the hook file.

leafty (changes requested):
Remove the hook file. It will cause confusion for future changes: which hook is the correct one?

I am going to be firm on this: the added hook file is not OK.

client/src/utils/customHooks/useLocationSearchParams.hook.ts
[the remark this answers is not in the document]
ciyer: Removed in 81e752bbbb682e1b27192eb825dd0831d3652f7c. The problem that the custom hook fixed is now gone -- I suspect that the bug was fixed in `react-router`, and we are benefiting from the dependency upgrade (this PR was started against the earlier react-router version).

client/src/features/projectsV2/new/ProjectV2New.tsx
```diff
@@ -111,6 +111,10 @@ export default function ProjectV2New() {
 function ProjectV2CreationDetails() {
   const [createProject, result] = usePostProjectsMutation();
   const navigate = useNavigate();
+  const groupMatch = useMatch(ABSOLUTE_ROUTES.v2.groups.show.root);
+  const groupSettingsMatch = useMatch(ABSOLUTE_ROUTES.v2.groups.show.settings);
```
leafty, line 115: You can replace this with a splat route, it will match any group route, even if we change them in the future:
```suggestion
  const groupMatch = useMatch(`${ABSOLUTE_ROUTES.v2.groups.show.root}/*`);
```

We could also add: `ABSOLUTE_ROUTES.v2.groups.show.splat = "/v2/groups/:slug/*"` as a constant.

See: https://reactrouter.com/start/declarative/routing#splats

ciyer: This is a good suggestion. I have implemented it this way in 437f8523daad1715ede914b7806e198cd7b10a58.

leafty (approved):

PR #3609, opened 2025-03-24
release 3.60.1

olevski:
build: release 3.60.1


leafty (changes requested):
Can we fix the version numbers? We can easily yank the tags and releases from GH.

olevski:
@leafty sure. Here is a new PR https://github.com/SwissDataScienceCenter/renku-ui/pull/3610

PR #3613, opened 2025-03-24
feat: add migrate project in Renku 2.0

andre-code:
This PR adds a way for users to move their old (legacy) projects to the new Renku 2.0 platform.

What’s included:

- A banner that appears for users with legacy projects, with a button to check which projects can be migrated.
- A form to select and migrate a project directly from the UI, with options to customize the title or slug.
- An option to dismiss the banner if the user isn’t interested at the moment.

This should make it easier for users to upgrade their work to Renku 2.0 with just a few clicks, without leaving the current Renku 2.0 interface.

migrate a project

/deploy renku=release-0.68.0 renku-data-services=main


ciyer (changes requested):
The UI looks good and works well! It's nice how things are shared between V1 and V2 as well.

As discussed, I think the code would be better structured if the state for certain components was managed in the component itself instead of its parent.

ciyer (approved):
Works very well now!

PR #3616, opened 2025-03-27
Use a standard component for pagination

lorenzo-cavazzi:
This PR removes the dependency `react-js-pagination` using a standard solution based on Bootstrap 5.
This will fix the following 2 high-level vulnerabilities:

https://github.com/SwissDataScienceCenter/renku-ui/security/dependabot/24
https://github.com/SwissDataScienceCenter/renku-ui/security/dependabot/20

/deploy

### DIFF
--- a/client/src/components/Pagination.tsx
+++ b/client/src/components/Pagination.tsx
@@ -54,9 +54,9 @@ export default function Pagination({
         onChange={onPageChange}
         innerClass={className}
         // Some defaults for the styling
-        itemClass={"page-item"}
-        linkClass={"page-link"}
-        activeClass={"page-item active"}
+        itemClass="page-item"
+        linkClass="page-link"
+        activeClass="page-item active"
         hideFirstLastPages={false}
         hideDisabled
       />
--- a/client/src/components/Pagination.tsx
+++ b/client/src/components/Pagination.tsx
@@ -17,7 +17,7 @@
  */
 
 import cx from "classnames";
-import ReactPagination from "react-js-pagination";
+import { PaginationItem, PaginationLink } from "reactstrap";
 
 interface PaginationProps {
   className?: string;
@@ -47,18 +47,12 @@ export default function Pagination({
 
   return (
     <div className={cx("d-flex", "align-items-center", "flex-column")}>
-      <ReactPagination
+      <CustomPagination
         activePage={currentPage}
-        itemsCountPerPage={perPage}
-        totalItemsCount={totalItems}
-        onChange={onPageChange}
         innerClass={className}
-        // Some defaults for the styling
-        itemClass="page-item"
-        linkClass="page-link"
-        activeClass="page-item active"
-        hideFirstLastPages={false}
-        hideDisabled
+        itemsCountPerPage={perPage}
+        onChange={onPageChange}
+        totalItemsCount={totalItems}
       />
       {showDescription && totalInPage && (
         <ExtraInfoPagination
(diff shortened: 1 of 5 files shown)


client/src/components/Pagination.tsx
```diff
@@ -72,6 +66,133 @@ export default function Pagination({
...
+interface CustomPaginationProps {
+  activePage: number;
+  innerClass?: string;
+  itemsCountPerPage: number;
+  maxPages?: number;
+  onChange: (pageNumber: number) => void;
+  totalItemsCount: number;
+}
+
+function CustomPagination({
```
leafty, line 78: Rename this component to `PaginationNav`, as this is what the component should render.

`CustomPagination` is not a helpful name here.

client/src/components/Pagination.tsx
```diff
@@ -72,6 +66,133 @@ export default function Pagination({
   );
 }
 
+interface CustomPaginationProps {
+  activePage: number;
+  innerClass?: string;
```
leafty, line 71: nitpick:
```suggestion
  innerClassName?: string;
```
because the standard prop is called `className`.

client/src/components/Pagination.tsx
```diff
@@ -47,18 +47,12 @@ export default function Pagination({
 
   return (
     <div className={cx("d-flex", "align-items-center", "flex-column")}>
-      <ReactPagination
+      <CustomPagination
```
leafty, line 50: Optimization: I think it would make sense to wrap this component inside `memo`.

See: https://react.dev/reference/react/memo

client/src/components/Pagination.tsx
```diff
@@ -72,6 +66,133 @@ export default function Pagination({
...
+      <PaginationElement
+        ariaLabel="last page"
+        content="⟫"
+        onClick={() => onChange(totalPages)}
+        key="last"
+      />
+    );
+  }
+
+  return <ul className={cx(innerClass)}>{pages}</ul>;
```
leafty, line 163: These links should be presented in a `nav` element:

```suggestion
  return <nav aria-label={ariaLabel}><ul className={innerClass}>{pages}</ul></nav>;
```

with `ariaLabel` being a prop and populated with a useful label, e.g. "data connectors pages"

Also, do not use `cx` for a single element, it is a no-op.

client/src/components/Pagination.tsx
```diff
@@ -72,6 +66,133 @@ export default function Pagination({
...
+}
+
+interface PaginationElementProps {
+  ariaLabel?: string;
+  content: React.ReactNode;
```
leafty, line 168: Rename to `children` and use good old JSX/TSX syntax.

```suggestion
  children?: React.ReactNode;
```

client/src/components/Pagination.tsx
```diff
@@ -72,6 +66,133 @@ export default function Pagination({
...
+  extraClass?: string;
```
leafty, line 169: Use a standard prop name when possible:

```suggestion
  className?: string;
```

client/src/components/Pagination.tsx
```diff
@@ -72,6 +66,133 @@ export default function Pagination({
...
+  key: string | number;
```
leafty, line 170: Never use `key` as a prop!

See: https://react.dev/learn/rendering-lists#why-does-react-need-keys
> Note that your components won’t receive key as a prop. It’s only used as a hint by React itself. If your component needs an ID, you have to pass it as a separate prop: `<Profile key={id} userId={id} />`.

client/src/components/Pagination.tsx
```diff
@@ -72,6 +66,133 @@ export default function Pagination({
...
+}
+function PaginationElement({
+  ariaLabel = "",
+  content,
+  extraClass = "",
+  key,
+  onClick,
+}: PaginationElementProps) {
+  return (
+    <PaginationItem key={key} className={extraClass}>
```
leafty, line 181:
```suggestion
    <PaginationItem className={extraClass}>
```

client/src/components/Pagination.tsx
```diff
@@ -72,6 +66,133 @@ export default function Pagination({
...
+      <PaginationLink
+        aria-label={ariaLabel}
+        href="#"
+        onClick={(e) => {
+          e.preventDefault();
```
leafty, line 186: Note: this is not great 😞 
I think we should have as a TODO issue for later to make sure that we replace this non-standard javascript pagination with `Link` elements from `react-router` which have the proper `?page=` search param. Having fake `#` links will mess up DOM parsing, especially for assistive technologies.

leafty (changes requested):
Thanks for removing this old dependency.

Code comments below.

client/src/components/Pagination.tsx
leafty: Also, it does not work when users do "open in new tab" actions, the new tab will not render the correct page.

client/src/components/Pagination.tsx
lorenzo-cavazzi: :facepalm: 
I initially made a function to generate partial content and then switch to a normal JSX element but forgot to swap the content argument into a children element.

client/src/components/Pagination.tsx
lorenzo-cavazzi: Added the `<nav aria-label` part.

> Also, do not use cx for a single element, it is a no-op.

Removed in https://github.com/SwissDataScienceCenter/renku-ui/pull/3616/commits/8a126c1084f964d5fbce31ecf998a385a224e263

Wouldn't `cx` still do its job of sanitizing the string? E.G. handling `undefined`? I added that cause we have legacy v1 code using the element.

client/src/components/Pagination.tsx
lorenzo-cavazzi: Yeah, that's not ideal. I added a comment.
Since the current solution isn't worse than the one being replaced, I suggest we keep that for now and open a follow-up issue.

client/src/components/Pagination.tsx
leafty: `cx` is used to compute the correct `className` prop, it does not do cleanup or sanitation.

client/src/components/Pagination.tsx
```diff
@@ -72,6 +70,143 @@ export default function Pagination({
...
+});
+
+interface PaginationElementProps {
+  ariaLabel?: string;
+  children: React.ReactNode;
+  className?: string;
+  onClick: () => void;
+}
+function PaginationElement({
+  ariaLabel = "",
```
leafty, line 188:
```suggestion
  ariaLabel,
```

client/src/components/Pagination.tsx
```diff
@@ -72,6 +70,143 @@ export default function Pagination({
...
+  children,
+  className = "",
```
leafty, line 190:
```suggestion
  className,
```

client/src/components/Pagination.tsx
```diff
@@ -72,6 +70,143 @@ export default function Pagination({
...
+  innerClassName?: string;
+  itemsCountPerPage: number;
+  maxPages?: number;
+  onChange: (pageNumber: number) => void;
+  totalItemsCount: number;
+}
+
+const PaginationNav = memo(function PaginationNav({
+  activePage,
+  ariaLabel = "Page navigation",
```
leafty, line 85: nitpick:

```suggestion
  ariaLabel = "page navigation",
```

client/src/components/Pagination.tsx
```diff
@@ -72,6 +70,143 @@ export default function Pagination({
...
+    );
+  }
+
+  // Page number buttons
+  for (let pageNumber = startPage; pageNumber <= end; pageNumber++) {
+    pages.push(
+      <PaginationElement
+        ariaLabel={`page ${pageNumber.toString()} of ${totalPages}`}
+        onClick={() => onChange(pageNumber)}
+        className={pageNumber === activePage ? "active" : ""}
```
leafty, line 140: This is where you use `cx`:
```suggestion
        className={cx(pageNumber === activePage && "active")}
```

client/src/components/Pagination.tsx
```diff
@@ -72,6 +70,143 @@ export default function Pagination({
...
+  return (
+    <nav aria-label={ariaLabel}>
+      <ul className={innerClassName}>{pages}</ul>
+    </nav>
+  );
+});
+
+interface PaginationElementProps {
+  ariaLabel?: string;
+  children: React.ReactNode;
```
leafty, line 183: nitpick:
```suggestion
  children?: React.ReactNode;
```

client/src/components/Pagination.tsx
```diff
@@ -72,6 +70,143 @@ export default function Pagination({
...
+        ⟨
+      </PaginationElement>
+    );
+  }
+
+  // Page number buttons
+  for (let pageNumber = startPage; pageNumber <= end; pageNumber++) {
+    pages.push(
+      <PaginationElement
+        ariaLabel={`page ${pageNumber.toString()} of ${totalPages}`}
```
leafty, line 138:
```suggestion
        ariaLabel={`page ${pageNumber} of ${totalPages}`}
```

client/src/components/Pagination.tsx
```diff
@@ -72,6 +70,143 @@ export default function Pagination({
...
+
+  // First page button: only show if the first page is not already shown
+  if (startPage !== 1) {
+    pages.push(
+      <PaginationElement
+        ariaLabel="first page"
+        onClick={() => onChange(1)}
+        key="first"
+      >
+        ⟪
```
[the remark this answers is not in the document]
lorenzo-cavazzi: Oh, interesting. It renders quite differently on my browser
Screenshot_20250328_114728
Using the Bootstrap icons makes the footprint much bigger, but I guess there aren't many alternatives -- using text like "first" or "last" won't be better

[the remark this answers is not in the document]
lorenzo-cavazzi: Screenshot_20250328_115305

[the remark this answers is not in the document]
lorenzo-cavazzi: Fixed in https://github.com/SwissDataScienceCenter/renku-ui/pull/3616/commits/1b60cd04b90189dd5e598cc6ba80c1d5537ea2be
Considering the footprint of the chevron icons, playing with padding improves the visualization -- I tried this solution https://github.com/SwissDataScienceCenter/renku-ui/pull/3616/commits/a1448b2a3ab0484524ae7e446984be66a00e956e
Screenshot_20250328_120300

client/src/components/Pagination.tsx
lorenzo-cavazzi: https://github.com/SwissDataScienceCenter/renku-ui/issues/3622

PR #3630, opened 2025-04-02
fix: ensure project as a namespace is populated when creating a data connector

leafty:
/deploy


olevski:
This fixes the problem. The only thing is that we can have repeating namespaces in the list now.

For example lets say that you are creating a dc in group35 which is on the second page of the namespace listing.

When you go to select the namespace for the dc you will see group35 selected by default. But once you load the second page of restults from the dropdown (which includes group35) then you will have 2 gruop35 options in the list. 1 at the top that was the default and another in the bottom. But we can live with this for now.

lorenzo-cavazzi (approved):
Lgtm! :rocket:

PR #3632, opened 2025-04-03
Support for launcher env variables

ciyer:
/deploy renku=release-0.68.0  renku-data-services=main

### DIFF
--- a/client/src/features/sessionsV2/SessionView/EnvVariablesModal.tsx
+++ b/client/src/features/sessionsV2/SessionView/EnvVariablesModal.tsx
@@ -1,0 +1,373 @@
+/*!
+ * Copyright 2025 - Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import cx from "classnames";
+import { useCallback, useEffect, useMemo, useState } from "react";
+import { CheckLg, Plus, XLg } from "react-bootstrap-icons";
+import {
+  Controller,
+  useFieldArray,
+  useForm,
+  type Control,
+  type FieldErrors,
+  type UseFieldArrayRemove,
+  type UseFormRegister,
+} from "react-hook-form";
+import {
+  Button,
+  Form,
+  Modal,
+  ModalBody,
+  ModalFooter,
+  ModalHeader,
+} from "reactstrap";
+
+import { SuccessAlert } from "../../../components/Alert";
+import { RtkErrorAlert } from "../../../components/errors/RtkErrorAlert";
+import { Loader } from "../../../components/Loader";
+import type {
+  SessionLauncher,
+  SessionLauncherPatch,
+} from "../api/sessionLaunchersV2.api";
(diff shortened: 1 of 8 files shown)


client/src/features/sessionsV2/SessionView/EnvVariablesModal.tsx
```diff
@@ -0,0 +1,318 @@
...
+          control={control}
+          {...register(`envVariables.${index}.value`)}
+          render={({ field }) => (
+            <Input
+              className={cx(error?.value && "is-invalid")}
+              placeholder="value"
+              type="text"
+              data-cy={`env-variables-input_${index}-value`}
+              {...field}
+              rules={{
```
andre-code, line 153: I think this should be defined in the Controller's rules rather than on the Input itself, right?

andre-code (commented):
The code looks clean! I couldn't test the feature properly because the deployment has an issue reading the launcher.

One suggestion: it would be nice to add a check to prevent duplicate environment variable names, maybe some validation to ensure a variable isn't added more than once?

andre-code:
More comments:

In the side panel view, it would be nice to show a short message like "No ENV variables set" instead of displaying an empty card.
Screenshot 2025-04-17 at 10 21 49


The UX can be improved when there are no ENV variables yet. It would be better to show a single input row with just the placeholders, ready for the user to enter a new variable. Why is there a default value set in this case?

Screenshot 2025-04-17 at 10 24 24

Also, if the user removes all the input fields, the update should reflect that by clearing all ENV variables. Currently, it doesn't seem to update properly when all are removed.

env variables

ciyer:
Thanks for these comments!


> In the side panel view, it would be nice to show a short message like "No ENV variables set" instead of displaying an empty card.


This is how it looks now:

image


> The UX can be improved when there are no ENV variables yet. It would be better to show a single input row with just the placeholders, ready for the user to enter a new variable. Why is there a default value set in this case?

I do not think it is a good idea to insert input rows for the user. Instead, I am showing a message:

image

> Also, if the user removes all the input fields, the update should reflect that by clearing all ENV variables. Currently, it doesn't seem to update properly when all are removed.

Oh, that was a bug in the backend, but it has now been fixed.

client/src/features/sessionsV2/SessionView/EnvVariablesModal.tsx
```diff
@@ -0,0 +1,365 @@
...
+  return (
+    <div>
+      <Button
+        data-cy="add-env-variable-button"
+        color="outline-secondary"
+        size="sm"
+        className="me-2"
+        onClick={onAddEnvVariable}
+      >
+        <Plus className={cx("bi")} />
```
andre-code, line 105: Please remove cx if we use only one class

client/src/features/sessionsV2/SessionView/EnvVariablesModal.tsx
```diff
@@ -0,0 +1,365 @@
...
+        </div>
+      </div>
+      <div>
+        <Button
+          data-cy={`env-variables-input_${index}-remove`}
+          color="outline-danger"
+          onClick={onRemove}
+          size="sm"
+        >
+          <XLg className={cx("bi")} />
```
andre-code, line 209: Same here, remove cx if it only uses one class.

client/src/features/sessionsV2/SessionView/EnvVariablesModal.tsx
```diff
@@ -0,0 +1,365 @@
...
+              value: 500,
+            },
+          }}
+        />
+        <div className="invalid-feedback">
+          {error?.value?.message ?? "Please input valid value."}
+        </div>
+      </div>
+      <div>
+        <Button
```
andre-code, line 203: Consider disabling this button when it's not allowed to remove the row (e.g. when there's only one row left)

andre-code (changes requested):
Looks great! The experience is much smoother now, thanks for the changes. Just dropped a couple of small suggestions.

client/src/features/sessionsV2/SessionView/EnvVariablesModal.tsx
ciyer: I've changed the behavior of the modal so that an empty row is only added to the form once. The row can be deleted, and then a message is shown that there are no environment variables defined.

andre-code (approved):
LGTM 🚀

PR #3634, opened 2025-04-04
fix: find project data connectors when linking

olevski:

This fixes the issue with linking data connectors currently on prod.


client/src/features/ProjectPageV2/ProjectPageContent/DataConnectors/ProjectConnectDataConnectorsModal.tsx
leafty, line 1: Why are you doing manual dispatch in this file? There is no reason to not use the hooks for querying data services here.

leafty (changes requested):

client/src/features/ProjectPageV2/ProjectPageContent/DataConnectors/ProjectConnectDataConnectorsModal.tsx
olevski: I have no clue why. That is what the code was like before so I just changed what is being called. I did not add that manual dispatch in the first place. I am not very familiar with react query so I assumed that the manual dispatch was needed.

leafty (approved):

PR #3636, opened 2025-04-04
build: add script to automate API spec retrieval

leafty:
Add a new script to automatically update `src/features/usersV2/api/users.openapi.json`. More spec files will be supported soon.

Run:
```bash
$ npm run update-api:users
```


lorenzo-cavazzi (approved):
Nice addition!

P.S. It would be great to allow overwriting the target release from the console by passing a different release/ref

PR #3640, opened 2025-04-08
[tmp] migration second part demo

andre-code:
PR to test migration projects
/deploy renku-data-services=andrea/add-user-preferences-for-dismiss-project-migration


andre-code:
temporary deployment for testing

PR #3643, opened 2025-04-11
fix: import the v2 component for unavailable sessions

lorenzo-cavazzi:
Import the correct V2 `SessionUnavailable` component for handling unavailable sessions. Previously, we where showing the v1 buttons using the v1 URLs

/deploy


ciyer (approved):
👍

PR #3644, opened 2025-04-14
Pass through env variables specified in query parameters

ciyer:
# Testing

This PR allows overriding environment variables by passing in values as query parameters. To test, create a project and add a session launcher. In the sidebar for the launcher, you can find and copy the session start link.

1. No variable defined

If an override is provided for a variable that is not defined in the launcher, the session will not start. Take the session start link and add a query parameter (e..g., `?FOO=value`). When following this link you should get an error message:

image

2. Variable defined

Add an environment variable to the session launcher. Then provide a different value as a query parameter to the session start link. In this case, the session should start, and the environment should have the variable defined and set to the value provided in the query parameter.

/deploy renku=release-0.68.0 renku-data-services=main

### DIFF
--- a/client/src/features/sessionsV2/SessionStartPage.tsx
+++ b/client/src/features/sessionsV2/SessionStartPage.tsx
@@ -193,8 +193,29 @@ function SaveCloudStorage({
   );
 }
 
+function envVariableOverrides(
+  envVariables: StartSessionFromLauncherProps["launcher"]["env_variables"],
+  searchParams: URLSearchParams
+) {
+  if (envVariables == null) return [];
+  const envVariableNames = envVariables.map((env) => env.name);
+  const envVariableOverrides = envVariableNames.map((name) => {
+    const value = searchParams.get(name);
+    if (value == null) return null;
+    return {
+      name,
+      value,
+    };
+  });
+  return envVariableOverrides.filter((env) => env != null) as {
+    name: string;
+    value: string;
+  }[];
+}
+
 function SessionStarting({ launcher, project }: StartSessionFromLauncherProps) {
   const [steps, setSteps] = useState<StepsProgressBar[]>([]);
+  const [searchParams] = useSearchParams();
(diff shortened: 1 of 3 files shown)


leafty (changes requested):
I don't think this system is future-proof. If ANYTHING ELSE needs to customizable using search parameters, then it becomes impossible to do so with this. E.g. how would you make the resource class customizable? The resource class is already parameterizable at launch, so using a search param for it seems natural. By just using `ENV_KEY=ENV_VALUE`, then we cannot have `resource_class=<c>` in the search parameters.

I think it will be much more preferable to add a prefix to the env var, e.g. `env_MY_KEY=MY_VALUE` to encode for `MY_KEY=VALUE`. This way we keep the space open for more search parameters, e.g. `resource_class=x`, `include_data_connectors=[a,b]`, etc.

ciyer:
> I don't think this system is future-proof. If ANYTHING ELSE needs to customizable using search parameters, then it becomes impossible to do so with this. E.g. how would you make the resource class customizable? The resource class is already parameterizable at launch, so using a search param for it seems natural. By just using `ENV_KEY=ENV_VALUE`, then we cannot have `resource_class=<c>` in the search parameters.
> 
> I think it will be much more preferable to add a prefix to the env var, e.g. `env_MY_KEY=MY_VALUE` to encode for `MY_KEY=VALUE`. This way we keep the space open for more search parameters, e.g. `resource_class=x`, `include_data_connectors=[a,b]`, etc.

This is an interesting use case that is not implemented in this PR, but could be supported later. 

We have reserved names that start with "renku": users cannot declare environment variables that start with this string, but we can use them ourselves. For the use case you have described, we could use the search parameter "RENKU_RESOURCE_CLASS" (or "RENKU_DATA_CONNECTOR", etc.) to handle these situations.

leafty (approved):
LGTM

Please fix the commit/PR title before merging: `feat: pass through...`

PR #3645, opened 2025-04-14
fix: use higher uid/gid in docker image

olevski:
This makes it so that the ui images can run with arbitrary uid/gid.

Also the nginx image is already setup to run with a bunch of things being in `/tmp`. The image has this in its `/etc/nginx.conf` file:

```
$ cat /etc/nginx/nginx.conf

worker_processes  auto;

error_log  /var/log/nginx/error.log notice;
pid        /tmp/nginx.pid;


events {
    worker_connections  1024;
}


http {
    proxy_temp_path /tmp/proxy_temp;
    client_body_temp_path /tmp/client_temp;
    fastcgi_temp_path /tmp/fastcgi_temp;
    uwsgi_temp_path /tmp/uwsgi_temp;
    scgi_temp_path /tmp/scgi_temp;
```

This is in line with the docs about running a non-privileged image here: (see running as non root)
https://hub.docker.com/_/nginx

To test locally:
```
cd client
docker build -t test .
docker run -ti --rm -p 8080:8080 -u 100000:100000 test
```
/deploy

### DIFF
--- a/client/Dockerfile
+++ b/client/Dockerfile
@@ -21,42 +21,29 @@ COPY .storybook /app/.storybook
 RUN npm run storybook-build -- -o storybook-static
 
 FROM nginxinc/nginx-unprivileged:1.27-alpine
+ARG UID=65543
+ARG GID=65543
 
-COPY --from=builder /app/build /usr/share/nginx/html
-COPY --from=builder /app/storybook-static /usr/share/nginx/html/storybook
 COPY nginx.vh.default.conf /etc/nginx/conf.d/default.conf
-COPY docker-entrypoint.sh /app/docker-entrypoint.sh
-COPY scripts/generate_sitemap.sh /app/scripts/generate_sitemap.sh
+
+USER $UID:$GID
+COPY --from=builder --chown=$UID:$GID --chmod=a+rwx /app/build /tmp/nginx/html
+COPY --from=builder --chown=$UID:$GID --chmod=a+rwx /app/storybook-static /tmp/nginx/html/storybook
+COPY docker-entrypoint.sh /tmp/docker-entrypoint.sh
+COPY scripts/generate_sitemap.sh /tmp/scripts/generate_sitemap.sh
 
 # Set up the config files written by docker-entrypoint
-USER root
-RUN touch /usr/share/nginx/html/config.json
-RUN chmod a+r /usr/share/nginx/html/config.json
-RUN chown nginx /usr/share/nginx/html/config.json
+RUN touch /tmp/nginx/html/config.json && \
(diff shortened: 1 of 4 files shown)


client/Dockerfile
```diff
@@ -21,42 +21,29 @@ COPY .storybook /app/.storybook
...
-RUN chown nginx /usr/share/nginx/html/privacy-statement.md
-
-RUN touch /usr/share/nginx/html/terms-of-use.md
-RUN chmod a+r /usr/share/nginx/html/terms-of-use.md
-RUN chown nginx /usr/share/nginx/html/terms-of-use.md
-
-USER nginx
+USER $UID:$GID
+COPY --from=builder --chown=$UID:$GID --chmod=a+rwx /app/build /tmp/nginx/html
+COPY --from=builder --chown=$UID:$GID --chmod=a+rwx /app/storybook-static /tmp/nginx/html/storybook
```
leafty, line 31: I don't think `a+w` is acceptable, use `rwxr-xr-x` please.

client/Dockerfile
```diff
@@ -21,42 +21,29 @@ COPY .storybook /app/.storybook
...
+COPY docker-entrypoint.sh /tmp/docker-entrypoint.sh
+COPY scripts/generate_sitemap.sh /tmp/scripts/generate_sitemap.sh
```
leafty, line 33: Why not keep it in `/app/scripts/generate_sitemap.sh`?

client/Dockerfile
```diff
@@ -21,42 +21,29 @@ COPY .storybook /app/.storybook
...
 
+# Set up the config files written by docker-entrypoint
+RUN touch /tmp/nginx/html/config.json && \
+	touch /tmp/nginx/html/robots.txt && \
+	touch /tmp/nginx/html/sitemap.xml && \
+	touch /tmp/nginx/html/privacy-statement.md && \
+	touch /tmp/nginx/html/terms-of-use.md && \
+	chmod a+rw -R /tmp/nginx/html
```
leafty, line 41: Same here, `rw-r--r--` permissions.

client/docker-entrypoint.sh
```diff
@@ -16,7 +16,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-export NGINX_PATH=/usr/share/nginx/html
+export NGINX_PATH=/tmp/nginx/html
```
leafty, line 19: Why not `chown /usr/share/nginx/html`?

leafty (commented):
General comment: before the `nginx` process starts, everything inside the `html` and config folders should be made read-only. There is no reason to let files be modified since they are all static at that point.

PR #3660, opened 2025-04-25
refactor: pagination

### LINKED ISSUE #3622: Use standard Link for the Pagination element
We recently swapped the component used for Pagination with a solution based on standard Bootstrap 5 elements (https://github.com/SwissDataScienceCenter/renku-ui/commit/eb4353cfe4bd743ad0cae2c7245a2a639064d505).

Clicking on the links currently uses a non-optimal solution based on `event.preventDefault()` that doesn't allow smooth navigation ([reference](https://github.com/SwissDataScienceCenter/renku-ui/blob/main/client/src/components/Pagination.tsx#L211-L215)). We should improve that with a solution based on Link

leafty:
Closes #3622.

Updates the pagination component to use real `Link` components. This means that Crtl+click works on pagination elements now.

Note: some v1 components are not migrated with this PR and use `PaginationLegacy`.

/deploy #legacy renku=release-0.68.0 renku-data-services=main


lorenzo-cavazzi (approved):
Code looks good :+1: 
Thanks for bringing back the navigation

PR #3666, opened 2025-04-28
fix: set nginx client server version to 1.27.4

olevski:
If we stick with `1.27` then we get newer versions and the latest one (which I think is 1.27.5) does not work because of permissions issue on `/var/nginx.pid`

The problem is this:
```
To fix nginx: [emerg] open() "/var/run/nginx.pid" failed (13: Permission denied), you have to specify a valid pid location by adding the line pid /tmp/nginx.pid; at the top level of your config. NOTE: NGINX 1.27.5 will complain about permissions for /run/nginx.pid due to a policy change for this path.
```

As described in the readme here: https://github.com/nginx/docker-nginx-unprivileged/blob/main/README.md#troubleshooting-tips

### DIFF
--- a/client/Dockerfile
+++ b/client/Dockerfile
@@ -20,7 +20,7 @@ COPY .storybook /app/.storybook
 
 RUN npm run storybook-build -- -o storybook-static
 
-FROM nginxinc/nginx-unprivileged:1.27-alpine
+FROM nginxinc/nginx-unprivileged:1.27.4-alpine
 
 COPY --from=builder /app/build /usr/share/nginx/html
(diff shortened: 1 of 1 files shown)


leafty (approved):

lorenzo-cavazzi (changes requested):
It seems to deploy correctly on 1.27.5

https://github.com/SwissDataScienceCenter/renku-ui/actions/runs/14725600062/job/41327464482?pr=3667

olevski:
So it seems we cannot reproduce this. It must have been some weird case or accident.

PR #3679, opened 2025-05-12
refactor: set max-height to 75vh on all modals

### LINKED ISSUE #3544: Limit max-height to 75vh in all modals
Note: should we make this standard for all modal?
Can be made into an issue.

_Originally posted by @leafty in https://github.com/SwissDataScienceCenter/renku-ui/pull/3478#discussion_r1975336547_

leafty:
Closes #3544.

Set max-height to `75vh` on the `<ScrollableModal>` component and use this component everywhere.

/deploy #legacy renku=release-0.68.0

### DIFF
--- a/client/src/components/Logs.tsx
+++ b/client/src/components/Logs.tsx
@@ -16,10 +16,10 @@
  * limitations under the License.
  */
 
+import cx from "classnames";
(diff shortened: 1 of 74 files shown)


leafty:
Note: will try also another approach for this issue.

PR #3694, opened 2025-05-16
fix: warn admins about editing global data connectors

### LINKED ISSUE #3693: Add a warning that admins should avoid editing a global data connector
While global datasets can be edited by admins, they should only be edited in special cases.

Also, the UI should reflect which fields are editable.

leafty:
Closes #3693.

WIP

/deploy #notest


ciyer (dismissed):
I could not test the exact scenario that this PR addresses, but I checked that this does not change behavior for normal users, and that part works fine.

ciyer (approved):
LGTM

PR #3697, opened 2025-05-16
chore: remove launcher breadcrumb

andre-code:
This PR removes the breadcrumb navigation from the "Add Session Launcher" form.

The form consists of only two steps, so using a breadcrumb for navigation is unnecessary and adds visual clutter. To improve the user experience, a simple "Back" button provides a more intuitive and focused way to navigate between the steps.

Screenshot 2025-05-16 at 11 12 09

/deploy


ciyer (approved):
LGTM!

PR #3704, opened 2025-05-22
fix: parse doi reference on the data connector view

lorenzo-cavazzi:
Use a standard DOI reference to avoid showing specific flavours used once the global connector was created the first time.

See: https://github.com/SwissDataScienceCenter/renku/issues/4025.

/deploy renku=leafty/update-search-tests


leafty (approved):
LGTM, thanks 🙏

PR #3714, opened 2025-06-04
feat: create images for both amd64 and arm64

sgaist:
## Describe your changes

This PR implements building and publishing images for both amd64 and arm64 architecture.

This make renku-ui follow other packages such as renku-data-services that are available for both platforms.

### DIFF
--- a/.github/workflows/test-and-ci.yml
+++ b/.github/workflows/test-and-ci.yml
@@ -150,6 +150,10 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
       - name: set up environment variables
         run: |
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi (approved):
Tested here https://github.com/SwissDataScienceCenter/renku-ui/actions/runs/15440410504/job/43456799838?pr=3715
It looks good :+1: 

The one thing to notice is that the time to build & push images went from 3-4 minutes (E.G: [3.59.1](https://github.com/SwissDataScienceCenter/renku-ui/actions/runs/15297755721/job/43030968121) or [3.59.0](https://github.com/SwissDataScienceCenter/renku-ui/actions/runs/15214705334/job/42798188824)) to a whooping ~30 minutes. Is that to be expected? Are we building and pushing just what we need?

leafty:
@lorenzo-cavazzi these are the supported platforms, not the ones being used for builds.

lorenzo-cavazzi:
Oh, thanks for clarifying. I'd say we merge then

PR #3724, opened 2025-06-10
feat: redirect autostart links to migrated projects

ciyer:
Session autostart links to legacy projects that have been migrated redirect the user to the new project (instead of starting the session).

To test, create a legacy project and an autostart link to that project. When used, the autostart link should bring you into a (legacy) session. Then migrate the project. Now the autostart link should bring you to the V2 project.

/deploy

### DIFF
--- a/client/src/features/session/components/StartNewSession.tsx
+++ b/client/src/features/session/components/StartNewSession.tsx
@@ -145,10 +145,7 @@ export default function StartNewSession() {
     return (
       <>
         <BackButton />
-        <CheckingMigrationRedirect
-          isFetchingMigrations={isFetchingMigrations}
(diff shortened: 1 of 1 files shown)


client/src/features/session/components/StartNewSession.tsx
```diff
@@ -187,6 +217,51 @@ interface LocationState {
   fromLanding?: boolean;
 }
 
+interface CheckingMigrationRedirectProps {
+  isFetchingMigrations: boolean;
```
leafty, line 221: Unused prop

```suggestion
```

leafty (commented):
Code looks alright, one comment about unused props:

PR #3735, opened 2025-06-18
feat: add a search page for groups

lorenzo-cavazzi:
_Work in progress_

/deploy renku=lorenzo/groups-search


lorenzo-cavazzi:
> Is this working? I created a project in a group and the group search page is empty.

I missed the proper reference to renku-data-services, I understood it was released already but it's not.
I just added `renku=release-2.5.0`

leafty:
Going to the group search page adds two entries in the browser history and not one, so you have to do "back" twice to get back. It should behave as normal and let the user go back.

client/.eslintrc.json
```diff
@@ -126,6 +126,7 @@
           "craco",
           "crosshair",
           "dagre",
+          "dataconnector",
```
leafty, line 129: Note: it doesn't seem needed for the moment.

client/src/features/searchV2/api/searchV2Api.api.ts
```diff
@@ -34,6 +34,7 @@ export const searchV2Api = searchV2GeneratedApi.injectEndpoints({
         url: "/search/query",
         params,
       }),
+      keepUnusedDataFor: 10,
```
leafty, line 37: Not needed, you should always query when the user clicks "search".
```suggestion
```

client/src/features/groupsV2/search/GroupSearchQueryInput.tsx
leafty, line 1: Is it really needed to make a new component when `SearchV2Bar` exists? Also, it should follow the same component naming convention (GroupSearchBar).

client/src/features/groupsV2/search/GroupSearchQueryInput.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const { control, register, handleSubmit, setFocus } = useForm<SearchBarForm>({
+    defaultValues: { query },
+  });
+
+  // focus search input when loading the component
+  useEffect(() => {
+    setFocus("query");
+  }, [setFocus]);
+
+  const onClick = useCallback(() => {
```
leafty, line 44: To make this action have an explicit name (clicking the clear button):
```suggestion
  const onClear = useCallback(() => {
```

leafty (commented):
It really seems that some bugs seen here (e.g. the double history entry) would be avoided if the existing searchv2 redux slice was used instead of re-implementing the whole search stack. The state of the search slice can be adapted to support the features of the group search page.

client/src/features/groupsV2/search/groupSearch.hook.ts
```diff
@@ -0,0 +1,35 @@
...
+export function useGroupSearch(ignoredParams?: string[]) {
+  const [searchParams] = useSearchParams();
+  const { group } = useGroup();
+
+  const params = useMemo(
+    () => generateQueryParams(searchParams, group.slug, ignoredParams),
+    [group.slug, ignoredParams, searchParams]
+  );
+
+  return useGetSearchQueryQuery({ params });
```
leafty, line 34: You should use the lazy query here instead and **always** hit the search API when the user clicks "search".

lorenzo-cavazzi:
@leafty thank you for the comments! Here are the changes:
- https://github.com/SwissDataScienceCenter/renku-ui/pull/3735/commits/7f2685463b978dee732603240f350648b01ef192: fix the navigation history by properly setting query params.
- https://github.com/SwissDataScienceCenter/renku-ui/pull/3735/commits/9c172e9f1d1db648622d7c1d91990f09b138d4a3: fix the query logic. I agree that clicking on the Search button should _always_ query. I kept a short debounce on params changes to avoid unnecessary queries (E.G. accidental double-clicks on keywords or filters, or fast browsing forth/back).
- https://github.com/SwissDataScienceCenter/renku-ui/pull/3735/commits/bceb0998abc1e978bbacfa7b1d770e603e5dc420: addresses all the other points

In insight, I can only agree with your comment at https://github.com/SwissDataScienceCenter/renku-ui/pull/3735#pullrequestreview-3023508097 . The initial implementation envisioned a different approach to dynamically change all filters based on the previous query, making it impractical to re-use the Search components. A bunch of constraints gradually aligned the query logic much more closely to the current Search page. Time is tight now to merge the logic together (and I still hope to move to the dynamic filters soon :crossed_fingers: ).

P.S. there are a couple of minor changes from the Product side that I'm going to push soon -- it's mostly visual details.

leafty:
@lorenzo-cavazzi I would expect that all search pages use the same logic in the future. Having the logic in a redux slice makes that very easy to re-use in the different search pages. If we don't do that now, then we should document it so that we refactor the code for search to always use the same logic and components.

For reference:
* Global search page -> should use the new search logic in the future 
* Group search page -> nice new search
* User search page -> to be implemented

client/src/features/groupsV2/search/GroupV2Search.tsx
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import cx from "classnames";
+import { useLayoutEffect } from "react";
```
leafty, line 20: Code smell: this is used for something that does not require layout attributes. It means the hook logic is probably buggy or that the hook doesn't implement correct cleanup.

Use `useEffect()` and update the hook if necessary.

leafty (changes requested):

lorenzo-cavazzi:
@leafty I fixed (hopefully) all the double-navigations cases and dropped `useLayoutEffect`. I added some notes about the components intended future

lorenzo-cavazzi:
@leafty since I was at it, I fixed the highlighting issue in https://github.com/SwissDataScienceCenter/renku-ui/pull/3735/commits/e2cc5036383df34369922c42c5236f6cb61a6955
I also fixed a tiny problem when updating keywords in projects -- if you dropped a keyword, no "update" was possible (see https://github.com/SwissDataScienceCenter/renku-ui/pull/3735/commits/b46d49cfa913c3fadff24c5786802d107a759828).

Could you re-approve?

PR #3738, opened 2025-06-19
fix: update head text query in SessionIframe.tsx

### LINKED ISSUE #3737: TypeError: Cannot read properties of null (reading 'querySelector')
Sentry Issue: [RENKU-UI-8](https://renku.sentry.io/issues/47050545/?referrer=github_integration)

```
TypeError: Cannot read properties of null (reading 'querySelector')
  at checkIframe (../../src/features/sessionsV2/SessionShowPage/SessionIframe.tsx:53:43)
```

leafty:
Closes #3737.

According to TypeScript's DOM types, `contentDocument.head` should always exist but in practice we can sometimes see a "Cannot read properties of null (reading 'querySelector')" error.


ciyer (approved):
👍

PR #3744, opened 2025-06-26
feat: add design tokens in Storybook

andre-code:
PR to add design tokens in Storybook

/deploy renku=release-2.3.0

### DIFF
--- a/client/src/storybook/bootstrap/DesignTokens/Fonts.stories.tsx
+++ b/client/src/storybook/bootstrap/DesignTokens/Fonts.stories.tsx
@@ -1,0 +1,352 @@
+import cx from "classnames";
+import React, { useState } from "react";
+import { Meta, StoryObj } from "@storybook/react";
+import { Copy, CopyIcon } from "~/storybook/bootstrap/utils.tsx";
+
+const fontTokens = {
+  typography: {
+    "body-font-family": { value: '"Inter", sans-serif', px: "" },
+    "body-font-size": { value: "1rem", px: "16px" },
+    "body-font-weight": { value: "400", px: "" },
+    "body-line-height": { value: "1.5", px: "" },
+  },
+  fontSizes: {
+    "fs-1": { value: "2.5rem", px: "40px", type: "fontSizes" },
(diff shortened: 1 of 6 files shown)


lorenzo-cavazzi (commented):
I'm going through this PR and I have two main comments:

- Design tokens: I like the level of detail! That should cover most of the needs of the design team. However, using static strings for the style attributes we wish to capture isn't the best approach. We would need to _manually_ update those strings if we update anything on Bootstrap. We can check the rendered elements instead using `getComputedStyle` 
- Molecules: Great to include more examples and longer descriptions wherever necessary (Cards are a good example) but sections like "Layout & Spacing" or "Typography" seem unnecessary (E.G. we don't need to say that card's background are white, nor that the class `.bg-white` is applied -- which isn't the case) and hard to maintain in the long run (not extracted automatically). 

In the reference pitch there was something about exporting to Figma but I don't see from here how it is done.

client/src/storybook/bootstrap/DesignTokens/Fonts.stories.tsx
```diff
@@ -0,0 +1,284 @@
...
+import { Copy, CopyIcon, pxToRem } from "~/storybook/bootstrap/utils";
+
+const fontTokens = {
+  typography: {
+    "body-font-family": "The primary typeface for body text",
+    "body-font-size": "Base font size for body copy",
+    "body-font-weight": "Default font weight",
+    "body-line-height": "Default line height for readability",
+  },
+  fontSizes: ["fs-1", "fs-2", "fs-3", "fs-4", "fs-5", "fs-6"],
```
lorenzo-cavazzi:
```suggestion
  fontSizes: ["fs-1", "fs-2", "fs-3", "fs-4"],
```

We can drop 5 and 6 since they are the same size of 3 and 4 and we generally don't use them anymore -- there are leftovers ofc, but most of them are on legacy pages.

lorenzo-cavazzi (changes requested):
Very nice! 👏 

Just a couple of small changes before we merge.

lorenzo-cavazzi (approved):
Lgtm! 🚀

PR #3759, opened 2025-07-11
fix: websocket problems when legacy is disabled

ciyer:
extra-values=enableV1Services=false

/deploy renku=ciyer/no-legacy extra-values=enableV1Services=false


leafty (approved):
how does the fix work?

ciyer:
> how does the fix work?

This is the code from `App.jsx` where the web socket is initialized:

```
    let webSocketUrl = props.client.uiserverUrl + "/ws";
    if (webSocketUrl.startsWith("http"))
      webSocketUrl = "ws" + webSocketUrl.substring(4);
```

You will see why the webSocketUrl was incorrect when the field on the `client` object is called `uiServerUrl`.

PR #3760, opened 2025-07-16
fix: prevent overflow problems with setting default resource class

lorenzo-cavazzi:
This PR prevents the overflow problem described here https://github.com/SwissDataScienceCenter/renku-ui/discussions/3754

Mind that switching to a non-scrolling modal maintains consistency with the custom resource class selection modal, which can be triggered when launching a new session.

image

/deploy renku=release-2.5.0


leafty (approved):
LGTM

PR #3764, opened 2025-07-18
feat: add v2 documentation links

lorenzo-cavazzi:
Add links to the documentation

Screenshot_20250718_162843

### DIFF
--- a/client/src/features/dashboardV2/HelpV2.tsx
+++ b/client/src/features/dashboardV2/HelpV2.tsx
@@ -17,7 +17,13 @@
  */
 import cx from "classnames";
 import { useContext } from "react";
-import { ChatSquareDots, Github, JournalText } from "react-bootstrap-icons";
+import {
+  ChatSquareDots,
+  FileEarmarkText,
+  Github,
+  JournalText,
+  Share,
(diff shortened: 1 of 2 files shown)


client/src/features/dashboardV2/HelpV2.tsx
```diff
@@ -158,6 +172,55 @@ function HelpGetting() {
   );
 }
 
+function HelpDocumentation() {
+  return (
+    <div>
+      <p>Here is our documentation for RenkuLab.</p>
+      <Row className="g-3">
+        <Col xs={12} md={6}>
+          <Card>
```
leafty, line 181: Maybe add `h-100` to make the cards look even:

Screenshot 2025-07-21 at 11 16 15

leafty (commented):
I think it would make more sense to keep only one tab for help and documentation. Having to click more to access the documentation is wasting the user's time.

client/src/features/dashboardV2/HelpV2.tsx
lorenzo-cavazzi: `h-100` makes sense :100: 
Especially with the longer text on the Community Portal section request from the Product team.

PR #3766, opened 2025-07-22
feat: add support for git revision and context dir options to image builds

leafty:
Add "advanced repository settings" when creating a session environment from code.

Details:
* The "git revision" field allows users to specify a git revision (branch, tag or commit) to build from.
* The "context directory" field allows users to specify the relative path to a folder to build from.
* Both fields can be reset by setting them to the empty string (select all then backspace).

See also: https://github.com/SwissDataScienceCenter/renku-data-services/pull/933

Deployment PR: https://github.com/SwissDataScienceCenter/renku/pull/4102

Add `repository_revision` and `context_dir` fields to `build_parameters` (session environments built from code).

image

### DIFF
--- a/client/src/App.jsx
+++ b/client/src/App.jsx
@@ -80,9 +80,7 @@ function CentralContentContainer({ user }) {
           index
           element={
             user.logged ? (
-              <ContainerWrap fullSize={true}>
-                <LazyRootV2 />
-              </ContainerWrap>
+              <LazyRootV2 />
             ) : (
               <div className="w-100">
                 <LazyAnonymousHome />
--- a/client/src/features/dashboardV2/HelpV2.tsx
+++ b/client/src/features/dashboardV2/HelpV2.tsx
@@ -26,10 +26,7 @@ import {
 } from "react-bootstrap-icons";
 import { Route, Routes } from "react-router";
 import { Card, CardBody, CardHeader, Col, Nav, NavItem, Row } from "reactstrap";
-import {
-  ExternalDocsLink,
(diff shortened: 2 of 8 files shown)


client/src/features/sessionsV2/components/SessionForm/BuilderEnvironmentFields.tsx
```diff
@@ -93,11 +94,14 @@ export default function BuilderEnvironmentFields({
     </WarnAlert>
   ) : (
     <div className={cx("d-flex", "flex-column", "gap-3")}>
-      <CodeRepositorySelector
-        name="repository"
-        control={control}
-        repositoriesDetails={repositoriesDetails}
-      />
+      <div>
```
lorenzo-cavazzi, line 97: Should we add a `gap-1` to have at least minimal spacing?

client/src/features/sessionsV2/components/SessionForm/CodeRepositoryAdvancedSettings.tsx
```diff
@@ -0,0 +1,118 @@
...
+            "d-flex",
+            "flex-column",
+            "gap-1",
+            "ms-1",
+            "ps-2",
+            "border-3",
+            "border-start",
+            "border-dark-subtle"
+          )}
+        >
```
lorenzo-cavazzi, line 68: Note: This is a nice approach to logically group together the content of a collapsible element. We should think about making a standard component for re-using that in other places too

PR #3771, opened 2025-07-22
refactor: make the getting help section clickable as links

lorenzo-cavazzi:
This makes the whole Card element clickable in the "Getting Help" section.

Peek 2025-07-22 15-09

/deploy renku=release-2.5.0

### DIFF
--- a/client/src/features/dashboardV2/HelpV2.tsx
+++ b/client/src/features/dashboardV2/HelpV2.tsx
@@ -86,6 +86,28 @@ function HelpNav({ statuspageId }: HelpNavProps) {
   );
 }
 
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi:
> Nice! I wonder if we should also add a slight primary background color when hovering.

That would be consistent with what we do in other components.
I want to add reusable classes to simplify adopting the same approach in other places. I couldn't settle on a clean solution tho -- this is my current attempt https://github.com/SwissDataScienceCenter/renku-ui/pull/3773

lorenzo-cavazzi:
@leafty I accidentally pushed an additional commit here and then removed it -- could you reapprove? :pray:

PR #3773, opened 2025-07-22
refactor: add background on hover on the help page cards

lorenzo-cavazzi:
/deploy extra-values=enableV1Services=false

### DIFF
--- a/client/src/features/dashboardV2/HelpV2.module.scss
+++ b/client/src/features/dashboardV2/HelpV2.module.scss
@@ -3,5 +3,5 @@
 @import "~bootstrap/scss/variables-dark";
 
 .linkBgAction:hover {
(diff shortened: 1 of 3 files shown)


client/src/features/dashboardV2/HelpV2.module.scss
```diff
@@ -0,0 +1,7 @@
+@import "~bootstrap/scss/functions";
+@import "~bootstrap/scss/variables";
+@import "~bootstrap/scss/variables-dark";
+
+.linkBgAction:hover {
+  background-color: var(--bs-tertiary-bg) !important;
```
leafty, line 6:
```suggestion
  background-color: var(--#{$prefix}tertiary-bg) !important;
```

leafty (changes requested):
otherwise ok

PR #3781, opened 2025-07-29
refactor: update session startup messages

lorenzo-cavazzi:
This PR updates the text we show while starting sessions to use a more generic 
"session service" definition instead of the technical word  "container"

/deploy
fix: https://www.notion.so/renku/Handle-autoscaling-well-Show-node-provisioning-status-errors-2220df2efafc80cb9d77f9b87a736c3e?source=copy_link#23a0df2efafc809081ade89056fa09fd


leafty (approved):
LGTM, but one of the "services" is the session itself. 🤷‍♀️

PR #3821, opened 2025-08-22
feat: show strip url path option

olevski:
Adds the option to strip the url path prefix for a custom user session.

/deploy


leafty (approved):

olevski:
Thank you @leafty for fixing the issues! :pray:

PR #3828, opened 2025-08-29
Redirect visitors to V1 projects if a new URL has been registered

ciyer:


client/src/features/legacy/NoLegacySupportForProjects.tsx
```diff
@@ -0,0 +1,410 @@
...
+      </Row>
+    </div>
+  );
+}
+
+function ProjectNoLegacyMigrateStep1() {
+  const location = useLocation();
+  const locationPath = location.pathname;
+  const gitlabProjectPath = locationPath.replace("/projects/", "");
+  const gitlabProjectUrl = `https://gitlab.renkulab.io/${gitlabProjectPath}`;
```
leafty, line 210: Are we showing this alert only for repositories hosted on `gitlab.renkulab.io`?  If so then we the alert should be matched for projects hosted there only.

leafty (commented):
I am not sure I understand the intention of this alert: are we showing this only on renkulab.io or are we showing this elsewhere too, but in this case users may not be at risk of losing their data. ❓

client/src/features/legacy/NoLegacySupportForProjects.tsx
ciyer: The alert is shown when Renku Legacy is configured as *not enabled* and a user somehow comes to a legacy project URL. Because legacy is not enabled, it is not possible to determine where the project is hosted -- we can only assume that it is on `gitlab.renkulab.io`.

leafty:
Should you move to Renku 2.0? Yes

Should you move the repository? This depends on whether or not the hosting GItLab instance will shut down. Only gitlab.renkulab.io is the instance which will shut down.

leafty (changes requested):

leafty (dismissed):

client/src/features/legacy/NoLegacySupportForProjects.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { useEffect, useState } from "react";
+
+import NoLegacySupport from "./NoLegacySupport";
+import NoLegacySupportForProjectsRenkulabIo from "./NoLegacySupportForProjectsRenkulabIo";
+
+export default function NoLegacySupportForProjects() {
+  const [isRenkulabIo, setIsRenkulabIo] = useState(false);
+  useEffect(() => {
+    const hostname = window.location.hostname;
+    if (hostname === "renkulab.io" || hostname.endsWith(".renkulab.io")) {
```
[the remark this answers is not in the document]
ciyer: Done in 43974de4c5e1095621cca2ea6e46e44f8403232a

PR #3837, opened 2025-09-03
feat: update to latest connected services API spec

leafty:
See: https://github.com/SwissDataScienceCenter/renku-data-services/pull/996


ciyer (approved):
LGTM!

PR #3869, opened 2025-10-16
refactor: remove ui-server routing where unnecessary

### LINKED ISSUE #3325: Update API URLs to remove `/ui-server`
Update client-side APIs where possible, e.g. `/ui-server/api/data`-> `/api/data`.

lorenzo-cavazzi:
These seem to be the last bits using the UI server routing when unnecessary.
Mind that the UI server's last stronghold is the WebSocket channel, which still relies on session status change notifications.

/deploy
fix #3325

### DIFF
--- a/client/src/features/dataConnectorsV2/components/DataConnectorModal/DataConnectorModalFooter.tsx
+++ b/client/src/features/dataConnectorsV2/components/DataConnectorModal/DataConnectorModalFooter.tsx
@@ -345,7 +345,7 @@ function DataConnectorCreateFooter({
           <RtkOrNotebooksError error={actionError} />
         </div>
       )}
-      <div className={cx("d-flex", "flex-grow-1")}>
+      <div className={cx("d-flex", "me-auto", "my-auto")}>
         <AddStorageBreadcrumbNavbar
           state={cloudStorageState}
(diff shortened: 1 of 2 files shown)


client/src/components/navbar/NavBarItems.tsx
```diff
@@ -286,9 +287,14 @@ export function RenkuToolbarItemUser({
...
-  const uiserverURL = params.UISERVER_URL;
-  const redirect_url = encodeURIComponent(params.BASE_URL);
+  const redirectUrl = encodeURIComponent(params.BASE_URL);
+  const logoutLocation = `/auth/logout?redirect_url=${redirectUrl}`;
+
   const isLegacyEnabled = params.LEGACY_SUPPORT.enabled;
+  const logoutBase = useMemo(
+    () => (isLegacyEnabled ? params.UISERVER_URL : params.GATEWAY_URL),
+    [isLegacyEnabled, params.UISERVER_URL, params.GATEWAY_URL]
+  );
```
leafty: The logout location is always `/api/auth/logout`. You don't need to handle two different cases here. Just always use `/api/auth/logout`.

leafty (changes requested):

PR #3872, opened 2025-10-16
refactor: avoid gaps when using headers

### LINKED ISSUE #3256: Avoid gaps in the headers and keep consistency
Headers (intended as  `<hx>` components) [should avoid gaps](https://html.spec.whatwg.org/multipage/sections.html#headings-and-outlines) and be used sequentially.

Even after the first round of restyling (https://github.com/SwissDataScienceCenter/renku-ui/pull/3214), we didn't do a good enough job because we ended up skipping a few elements (mostly `h1` and `h3`) instead of changing the size appropriately.
We should adjust that by using headers in a semantically valid way, changing the size through bootstrap variables to fit the Design team's requirements.

lorenzo-cavazzi:
This PR adjusts the header sizes so that it will feel more natural to use h1 -> h2 -> h3 instead of the current h2 -> h4 -> h5 + fs.

Most pages are identical, with the new header elements and font-size utilities adjusted to the new values. Some pages had minor adjustments, either because they were using non-standard sizes (E.G. the old h3) or margins/spaces were slightly off.

Mind that the anonymous landing page had bigger changes because it had _much larger_ text than the other pages. The approach has been kept and the text is still larger, but adapted to the new sizes so that the header hierarchy is respected, and the standard text looks one size larger than on other pages.

/deploy
fix #3256

### DIFF
--- a/client/src/features/sessionsV2/SessionList/SessionLauncherCard.tsx
+++ b/client/src/features/sessionsV2/SessionList/SessionLauncherCard.tsx
@@ -75,8 +75,7 @@ export default function SessionLauncherCard({
   const environment = launcher?.environment;
   const imageBuildersEnabled =
     params?.IMAGE_BUILDERS_ENABLED ?? DEFAULT_APP_PARAMS.IMAGE_BUILDERS_ENABLED;
-  const isCodeEnvironment =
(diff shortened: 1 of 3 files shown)


leafty (approved):
LGTM

PR #3873, opened 2025-10-17
feat: allow data connectors to be skipped when launching sessions

leafty:
See: https://github.com/SwissDataScienceCenter/renku-data-services/pull/1065.

The data-service API is changed to allow not mounting data connectors at session launch. With this change, the "Skip" action now means "do not mount this data connector", which is what Renku should do in most cases (trying to mount without credentials or with invalid credentials leads to session nodes being banned e.g. from SFTP hosts).

Updated modal:
Screenshot 2025-10-17 at 13 32 44
Screenshot 2025-10-17 at 13 33 02

Skipping will not mount:
Screenshot 2025-10-17 at 13 35 04


client/scripts/update_api_spec.js
```diff
@@ -24,7 +24,7 @@ import { parseDocument } from "yaml";
 
 const GH_BASE_URL = "https://raw.githubusercontent.com";
 const DATA_SERVICES_REPO = "SwissDataScienceCenter/renku-data-services";
-const DATA_SERVICES_RELEASE = "main";
+const DATA_SERVICES_RELEASE = "leafty/feat-1009-allow-dc-skip";
```
lorenzo-cavazzi: Remember to revert before updating

leafty: I would merge the backend PR first

lorenzo-cavazzi (approved):
🚀

PR #3877, opened 2025-10-20
fix: replace broken docs link in admin panel

mohammad-alisafaee:
This PR temporarily fixes a broken docs link in the admin panel by replacing it with a link to an older version of the docs. The new docs currently doesn't include the target page. Since it was only a single link I didn't add a new `rdt...()` function and just added a new link.


client/src/utils/constants/Docs.js
```diff
@@ -48,6 +48,8 @@ const Links = {
   EPFL: "https://www.epfl.ch/en/",
   ETHZ: "https://ethz.ch/en.html",
   RENKU_BLOG: "https://blog.renkulab.io",
+  RENKU_2_ADMIN_HOW_TO_GUIDE_INCIDENTS:
```
leafty: Add this to [NewDocs.ts](https://github.com/SwissDataScienceCenter/renku-ui/blob/main/client/src/utils/constants/NewDocs.ts).

PR #3906, opened 2025-11-17
feat: use react-select to pick group member role

ciyer:
# Summary

Use a react-select component to set the role for group members. This allows us to show more detailed information about the differences between the roles.

image

/deploy


lorenzo-cavazzi (changes requested):
Nice addition 🚀 

Minor change: I suggest we use the standard styling/positioning for dropdown menus, with the label above the input. It looks a little off since there is already another dropdown in the same modal (on the left is the proposal, on the right the current implementation)

Image

ciyer:
> Minor change: I suggest we use the standard styling/positioning for dropdown menus, with the label above the input.]


I agree, it looks better when using the our standard.

image

PR #3928, opened 2025-12-04
feat: show a warning when about to shut down a session

lorenzo-cavazzi:
Shows a modal to warn users when a session is flagged as inactive by Amalthea and set for termination.
Clicking the button will reset the counter.

Reference on Notion: https://www.notion.so/renku/Explicit-session-culling-28c0df2efafc800d92dcd2d9bf5f98b6

image

/deploy amalthea-sessions=0.23.0 renku-data-services=main extra-values=enableInternalGitlab=false

### DIFF
--- a/client/src/features/sessionsV2/api/computeResources.generated-api.ts
+++ b/client/src/features/sessionsV2/api/computeResources.generated-api.ts
@@ -653,6 +653,7 @@ export type RemoteConfigurationFirecrest = {
 export type RemoteConfiguration = RemoteConfigurationFirecrest;
 export type IdleThreshold = number;
 export type HibernationThreshold = number;
+export type HibernationWarningPeriod = number;
 export type RuntimePlatform = "linux/amd64" | "linux/arm64";
 export type ResourcePoolWithIdFiltered = {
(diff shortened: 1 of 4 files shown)


client/src/features/sessionsV2/SessionShowPage/ShowSessionPage.tsx
```diff
@@ -99,6 +117,78 @@ export default function ShowSessionPage() {
...
+        setShowPauseWarningModal(true);
+      }
+    };
+
+    // Either show it immediately or set a timeout
+    if (nextNotificationDelay <= 0) {
+      showWarningIfNeeded();
+      return;
+    }
+    const id = window.setTimeout(showWarningIfNeeded, nextNotificationDelay);
```
ciyer: question: is there a reason to call this function as a method of `window` rather than just `setTimeout`?

ciyer (dismissed):
LGTM!=

leafty (changes requested):
Please run:
```bash
npm run update-api
npm run generate-api
npm run format
```

To update the API specs to the corresponding backend feature.

PR #3947, opened 2025-12-22
chore: update documentation link (data connectors from repo)

lorenzo-cavazzi:
/deploy


ciyer (changes requested):
It looks like the link to the documentation is incorrect.

Issue #3959, opened 2026-01-08
Add a link to the search page when clicking on a keyword


lorenzo-cavazzi:
Keywords should link to the global search page with a query parameter `q=keyword:"<keyword_value>"`.

Mind that we have a specific visualization for whenever users can edit/remove keywords. We don't want a link to be added in that case.

<img width="814" height="570" alt="Image" src="https://github.com/user-attachments/assets/cfa917c3-309a-4caf-8f8f-a38ddbf30697" />

PR #3960, opened 2026-01-08
minor: make keyword badges links (#3959)

ciyer:
# Description

Keywords in the project and data connector pages when clicked navigate to the search page for that keyword.

## Screenshots

image

image

/deploy

### DIFF
--- a/client/src/components/keywords/KeywordBadge.tsx
+++ b/client/src/components/keywords/KeywordBadge.tsx
@@ -19,7 +19,7 @@
 import cx from "classnames";
 import { useMemo } from "react";
 import { XCircle } from "react-bootstrap-icons";
-import { generatePath, Link } from "react-router";
+import { Link } from "react-router";
 
 import { ABSOLUTE_ROUTES } from "~/routing/routes.constants";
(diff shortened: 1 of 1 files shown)


client/src/components/keywords/KeywordBadge.tsx
```diff
@@ -47,8 +52,16 @@ export default function KeywordBadge({
...
     </button>
   ) : null;
+  const linkPath = useMemo(() => {
+    if (searchKeyword == null || searchKeyword.length < 1) return null;
+    const searchPath = generatePath(ABSOLUTE_ROUTES.v2.search);
+    return (
+      `${searchPath}?` +
+      new URLSearchParams({ q: `keyword:"${searchKeyword}"` }).toString()
+    );
+  }, [searchKeyword]);
```
leafty: `react-router` can make a link based on path and search:


```suggestion
  const search = useMemo(() => {
    if (searchKeyword == null || !searchKeyword) return null;
    return (
      "?" + new URLSearchParams({ q: `keyword:"${searchKeyword}"` }).toString()
    );
  }, [searchKeyword]);
```

PR #3968, opened 2026-01-09
refactor: do not throw in hooks (#3817)

ciyer:
/deploy


leafty (commented):
I'm not sure I understand what is the improvement we get from this change. The Logs component still uses async methods in hooks which should be avoided and will still cause issues upon HTTP 500.

PR #3981, opened 2026-01-22
chore: remove legacy components - sessions+notebooks

lorenzo-cavazzi:
Remove sessions and related components.
Mind that v2 sessions are still fetched in response to messages from the WebSocket component.

/deploy

### DIFF
--- a/client/src/components/Logs.tsx
+++ b/client/src/components/Logs.tsx
@@ -33,6 +33,7 @@ import {
 
 import { getSessionStatusStyles } from "~/features/sessionsV2/components/SessionStatus/SessionStatus";
 import { SessionV2 } from "~/features/sessionsV2/sessionsV2.types";
+import { NotebookAnnotations } from "~/notebooks/notebooks.types";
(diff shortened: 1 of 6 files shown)


client/src/components/Logs.tsx
```diff
@@ -48,6 +46,31 @@ import { ErrorAlert } from "./Alert";
 import { Loader } from "./Loader";
 import ScrollableModal from "./modal/ScrollableModal";
 
+interface NotebookAnnotations {
```
leafty: Can we keep this type in a separate source file?

client/src/features/project/components/ProjectSettings.tsx
```diff
@@ -36,7 +35,7 @@ export default function ProjectSettings({ ...props }: ProjectSettingsProps) {
         <Col key="content" sm={12} md={10} data-cy="settings-container">
           <Routes>
             <Route path="/" element={<ProjectSettingsGeneral {...props} />} />
-            <Route path="sessions" element={<ProjectSettingsSessions />} />
+            <Route path="sessions" element={null} />
```
leafty: Just remove the route then.

PR #3999, opened 2026-01-29
feat: use group search in global search

andre-code:
PR to refactor and enhance the Search page.

### Changes

- Adapted GroupSearch into the global Search flow.

- Added User and Group visualization in Global Search, including counts of how many projects and data connectors are associated with a specific user or group.

- Moved search logic from the Group folder to the Search folder and renamed components from `GroupSearch*` to `Search*`.

- Added a Date filter.

- Fixed query/URL synchronization issues to ensure filters and results stay in sync.

/deploy renku=andrea/update-search-page renku-data-services=main

### DIFF
--- a/client/src/features/searchV2/Search.tsx
+++ b/client/src/features/searchV2/Search.tsx
@@ -17,29 +17,16 @@
  */
 
 import cx from "classnames";
-import { Navigate, useSearchParams } from "react-router";
 import { Col, Row } from "reactstrap";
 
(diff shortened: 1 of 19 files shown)


leafty (approved):
LGTM, very nice! 👏 

One small comment, not blocking: if I type a search filter, it stays in the search bar but if I reload the page, it disappears from the search bar. The filter state is correct but it feels a bit weird.

PR #4002, opened 2026-01-29
feat: add support for data connectors using OAuth 2.0 credentials

leafty:
_Merging into a feature branch_

Add support for using integrations with data connectors.

Deployed at: https://flora.dev.renku.ch/

Details:
* Google Drive and Dropbox can work with Renku integrations
* When the corresponding integration is not configured:
    Screenshot 2026-01-30 at 10 02 41
* When the integration is available and the user needs to connect:
    Screenshot 2026-01-30 at 10 02 54
* When everything is ready:
    Screenshot 2026-01-30 at 10 03 04
* On the project page, show badges showing if the data connector is working for the user. If the badge is not green, the data connector will not be mounted in sessions.
    Screenshot 2026-02-04 at 10 38 53

### DIFF
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -138,6 +138,7 @@
           "dismissible",
           "dockerfile",
           "dockerignore",
+          "dropbox",
           "dropdown",
           "dropzone",
           "enum",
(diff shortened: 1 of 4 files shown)


lorenzo-cavazzi (approved):
Lgtm! 👏

PR #4005, opened 2026-01-29
feat: show data connector options on listings

lorenzo-cavazzi:
Show the actions drodpwn menu on data connection list items as we do for the other project entities.

image

/deploy #extra-values=enableInternalGitlab=false renku=lorenzo/tests-dc-update


lorenzo-cavazzi:
P.S. This requires an update to the acceptance tests. I'll do that as the last step, after we agree on the solution.

PR #4010, opened 2026-02-06
build: use react-route integration for Sentry

leafty:
Update the Sentry packages to use the `react-router` integration.

Details:
* Client-side Sentry is initialized in the root route
* Server-side Sentry is initialized before anything else in production
* Server-side Sentry is initialized with react-router in development

/deploy extra-values=ui.client.sentry.enabled=true,ui.client.sentry.environment=renku-ci-ui-4010,ui.client.sentry.dsn=https://b8aaa5810d2743f281350c0944c38ec8@o4509039310995456.ingest.de.sentry.io/4509044025131088

### DIFF
--- a/client/react-router.config.ts
+++ b/client/react-router.config.ts
@@ -6,7 +6,8 @@ export default {
   appDirectory: "src",
   ssr: true,
 
-  // TODO: configure this maybe?
+  // TODO: configure Sentry integration for source maps
+  // TODO: Reference: https://docs.sentry.io/platforms/javascript/guides/react-router/manual-setup/#step-3-add-readable-stack-traces-with-source-maps-optional
   // Sentry release hook
   // buildEnd: async (
   //   {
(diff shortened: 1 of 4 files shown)


client/react-router.config.ts
```diff
@@ -1,6 +1,24 @@
...
+  //     reactRouterConfig: reactRouterConfig,
+  //     buildManifest: buildManifest
+  //   }
+  // ) => {
+  //   await sentryOnBuildEnd({
+  //     viteConfig: viteConfig,
+  //     reactRouterConfig: reactRouterConfig,
+  //     buildManifest: buildManifest
+  //   });
+  // }
```
lorenzo-cavazzi: Was this something temporary? If not, could you please add a longer explanation comment (why is this here? if you pick it from the sentry docs, a link would be great)

lorenzo-cavazzi (commented):
Lgtm! I left a comment suggesting either removing or adding a description for a couple of commented-out sections.

P.S: I got this when redirected back to the application after signing up with a new user. I haven't seen this message elsewhere recently. Not sure if it's something worth investigating
Screenshot_20260209_120403

PR #4020, opened 2026-02-12
refactor: remove legacy client

leafty:
Remove the legacy API client from renku-ui

/deploy


lorenzo-cavazzi (commented):
`client` is really going away!? 🎉

lorenzo-cavazzi (approved):

PR #4023, opened 2026-02-13
fix(member-list): prevent overflow and add +n avatar indicator

andre-code:
/deploy renku=andrea/update-search-page renku-data-services=main


lorenzo-cavazzi (changes requested):
Looks good! A couple of tiny adjustments before merging

lorenzo-cavazzi:
P.S: probably not a real-world scenario, but having _many_ members might result in a very long text. I wonder if it makes sense to cut the tooltip text after a number of chars (100? 200?) and add "..."

Screenshot_20260220_100814

PR #4030, opened 2026-02-17
fix: use public repository to build from code

### LINKED ISSUE #4029: Create from code should be restricted to public repositories
See also: https://github.com/SwissDataScienceCenter/renku/issues/4064

It seems that the UI for "Create from code" has been updated to allow any repository with pull permission to be used for "Create from code" while the build system only allows public repositories to be used.

leafty:
Closes #4029.

Backend PR: https://github.com/SwissDataScienceCenter/renku-data-services/pull/1214

Private code repositories cannot be selected to "Create from code":
Screenshot 2026-02-17 at 14 43 31

/deploy renku-data-services=main extra-values=dataService.imageBuilders.enabled=true


leafty:
@lorenzo-cavazzi I had to make a small code change to account for the new types

PR #4040, opened 2026-02-20
chore: remove some unused components

lorenzo-cavazzi:
This PR touches error handling too

/deploy renku=leafty/ui-ssr-update extra-values=ui.client.resources.limits.memory=300Mi,ui.client.resources.requests.memory=300Mi


client/src/components/buttons/Button.tsx
```diff
@@ -134,9 +121,7 @@ interface ButtonWithMenuV2Props {
   size?: string;
   isDisabledDropdownToggle?: boolean;
 }
-export const ButtonWithMenuV2 = SplitButtonWithMenu;
-
-export function SplitButtonWithMenu({
+export function ButtonWithMenuV2({
```
leafty: Component names are a bit confusing: which component should we use?

PR #4049, opened 2026-02-25
refactor: update logs display components

### LINKED ISSUE #4042: Refreshing Session Logs Does Not Maintain The Selected Container
I know the way we handle logs is not ideal and there are many things  that can be done to improve things. Especially on the backend.

But one thing will be helpful.

Basically when you select a specific container (not the one selected by default) and then click "Refresh logs" the selected container switches back to the one selected by default.

So it would be nice if the selected container is maintained between refreshes.

There could be cases (in case of restarts) when this may not be possible. Then we can fall back to the default. But when the same session is just starting for a while or failing and nothing has changed. We should not reset the container selection when viewing session logs.

To reproduce:
- launch a session
- wait for more than 1 container to show up when viewing logs
- select the non-default container for the logs
- click refresh logs
- the selected container is not maintained - it always reverts to the same defaul

leafty:
Closes #4042.

Refactor the logs display component to use RTK Query hooks and fix some issues like forced scrolling and forced tab switching.

<code>
/deploy extra-values=dataService.imageBuilders.enabled=true,dataService.imageBuilders.strategyName=renku-buildpacks-v3,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/renku-build/,dataService.imageBuilders.nodeSelector.renku.io/node-purpose=user,dataService.imageBuilders.tolerations[0].effect=NoSchedule,dataService.imageBuilders.tolerations[0].key=renku.io/dedicated,dataService.imageBuilders.tolerations[0].operator=Equal,dataService.imageBuilders.tolerations[0].value=user
</code>

### DIFF
--- a/client/src/features/logsDisplay/LogsModal.tsx
+++ b/client/src/features/logsDisplay/LogsModal.tsx
@@ -397,7 +397,13 @@ function useDownloadLogs(
     return () => {
       ignore = true;
     };
-  }, [name, refetch, renkuToastDanger, renkuToastSuccess]);
+  }, [
+    downloadQueryTrigger,
+    name,
+    refetch,
+    renkuToastDanger,
+    renkuToastSuccess,
+  ]);
 
   return [isDownloading, triggerDownload];
 }


lorenzo-cavazzi (changes requested):
Code looks great! Thanks for fixing the component!
Just a couple of minor comments

PR #4053, opened 2026-02-26
UI support for remote session on Runai

ciyer:
# Description

Changes to the admin console to allow setting up Run:AI as a remote backend.

## Limitations

The Run:AI remote backend is currently just a proof-of-concept. It works, but with limitations. Launching a session from the Renkulab UI will start a Run:AI workspace, but it will not be displayed in the Renkulab UI, because the tunneling/port-forwarding does not work. Also, the git repositories are not cloned into the Run:AI workspace.

## Testing

Set up a resource pool for Run:AI, targeting the SDSC Run:AI cluster, for example (though this should work with any Run:AI cluster).

Create an "Access Key" for your user in the user settings. The information to configure the remote session needs to be passed in Environment Variables. Create a session launcher, set one of the Runai resources classes as the default. Then set the following environment variables:

* RSC_AUTH_KIND: client_credentials
* RSC_RUNAI_PROJECT: [your Run:AI project]
* RSC_AUTH_RUNAI_CLIENT_ID: [the client id of the access key]
* RSC_AUTH_RUNAI_CLIENT_SECRET: [the secret of the access key]

/deploy renku=release-2.16.0 amalthea-sessions=main renku-data-services=main

### DIFF
--- a/client/eslint.config.mjs
+++ b/client/eslint.config.mjs
@@ -211,6 +211,7 @@ const jsRules = {
         "reusability",
         "rproj",
         "rstudio",
+        "runai",
         "scala",
         "scrollable",
         "selectautosuggest",
--- a/client/src/features/admin/AddResourcePoolButton.tsx
+++ b/client/src/features/admin/AddResourcePoolButton.tsx
@@ -109,11 +109,16 @@ function AddResourcePoolModal({ isOpen, toggle }: AddResourcePoolModalProps) {
       clusterId: "",
       remote: {
         enabled: false,
-        kind: "firecrest",
-        providerId: "",
-        apiUrl: "",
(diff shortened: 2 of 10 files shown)


leafty (commented):
Notes:
* please add a description
* I would like to review and merge the backend pieces first

client/.prettierignore
```diff
@@ -1 +1,5 @@
 build/
+public/config.json
+# Do not format generated files
+*generated-api.ts
+.react-router/types
```
leafty: We need to keep formatting the generated files: they are checked into git and the formatting makes them easier to read.

```suggestion
public/config.json
.react-router/types
```

PR #4076, opened 2026-03-11
fix: silence Sass warnings from bootstrap

leafty:


ciyer (approved):
Great that you are fixing this!

PR #4077, opened 2026-03-11
feat: add meta tags for group and user pages

leafty:
Use SSR for group and user pages to set meta tags. Also, use specific social cards for projects, users and groups.

/deploy renku=release-2.16.0


lorenzo-cavazzi (approved):
Lgtm! 🚀

PR #4079, opened 2026-03-11
chore: cleanup unused exports and files

lorenzo-cavazzi:
More cleanup

/deploy


client/src/utils/Utils.test.jsx
```diff
@@ -27,110 +27,10 @@ import { describe, expect, it } from "vitest";
...
-  formatBytes,
-  getEntityImageUrl,
-  parseINIString,
-  refreshIfNecessary,
   slugFromTitle,
-  splitAutosavedBranches,
 } from "./helpers/HelperFunctions";
 import { verifyTitleCharacters } from "./helpers/verifyTitleCharacters.utils";
 
-describe("Ini file parser", () => {
```
leafty: Why remove this test? It is most likely used for `rclone` config validation.

PR #4083, opened 2026-03-17
Apply usage quotas to user sessions

ciyer:
Use this again once the notifications have been merged into the feature branch
renku-data-services=build/resource-quotas

/deploy renku-data-services=build/resource-quotas-notification #extra-values=dataService.dataTasks.enableResourceRequestTracking=true


client/src/components/progress/ProgressSteps.tsx
```diff
@@ -70,8 +71,9 @@ interface ProgressStepsIndicatorProps {
   style: ProgressStyle;
   title: string;
   description: string;
-  status: StepsProgressBar[];
   moreOptions?: React.ReactNode;
+  usageAvailability?: React.ReactNode;
```
leafty: Naming anti-pattern: this name is not generic.


```suggestion
  extraDescription?: React.ReactNode;
```

Picking `extraDescription` as it is a node to be displayed below the description.

client/src/features/dashboardV2/DashboardV2Sessions.tsx
```diff
@@ -218,6 +231,11 @@ function DashboardSession({ session }: DashboardSessionProps) {
       <div className={cx(styles.sessionButton, "position-absolute")}>
         <ActiveSessionButton
           className="my-auto"
+          usageAvailable={{
+            hours: userLauncherClass?.usage_available,
+            percentage: userLauncherClass?.usage_available_percentage,
```
leafty: Note: This is a bit weird from an API point of view. Why not get `usage_available` and `usage_total` and compute the ratio on the client side? With the percentage API, the user (or client) doesn't know the total compute-hours allocated.

ciyer: If you know the available number of hours and what percentage of the total this is, you can compute the total number of hours.

client/src/components/progress/ProgressSteps.tsx
ciyer: Fixed in 839badba467570deb940110c4004e6ca5f96e63a

client/src/features/dashboardV2/DashboardV2Sessions.tsx
leafty: Yes, but you will get floating point errors. Compared to having floating point errors on the percentage, having the exact amount of compute hours allocated (total) is a much better solution.

ciyer: Fixed in 4c1cf842cfdca440edbcafdcaafaa9e85ed249a6

tests/cypress/fixtures/sessions/sessionsV2Paused.json
leafty, line 1: Note: the UI uses "paused", so we should name it `sessionsV2Paused.json`

client/src/features/sessionsV2/session.utils.tsx
```diff
@@ -419,3 +422,38 @@ export function isImageCompatibleWith(
...
+          Usage quota for this resource pool <strong>has been reached</strong>
+        </span>
+      </div>
+    );
+
+  return (
+    <div className={cx("d-flex", "align-items-center", "gap-2")}>
+      <Stopwatch fontSize={16} className="flex-shrink-0" />
+      <span>
+        <strong>{usageAvailableHours}h of compute time</strong> until quota is
```
leafty: Maybe this line should also use `usageAvailableString()`.

client/src/features/sessionsV2/session.utils.tsx
```diff
@@ -419,3 +422,38 @@ export function isImageCompatibleWith(
...
+}
+
+export function UsageAvailable({
+  usageAvailableHours,
+}: {
+  usageAvailableHours: number;
+}) {
+  if (usageAvailableHours <= 0)
+    return (
+      <div className={cx("d-flex", "align-items-center", "gap-2")}>
```
leafty: Why use a `div` with `flex` here? Does it not work to just return `<>` or a `<span>`?

```
return (<><Alarm fontSize={16} className={cx("bi", "me-2")} />Usage...</>);
```

client/src/features/sessionsV2/components/SessionButton/ActiveSessionButton.tsx
```diff
@@ -72,17 +74,220 @@ import {
...
+        )}
+        {isUserLoggedIn ? "Pause" : "Delete"}
+      </Button>
+    </>
+  );
+}
+
 interface ActiveSessionButtonProps {
   className?: string;
+  usageAvailable: SessionLauncherResourceUsageAvailable;
```
leafty: Why not pass the `resourceClass` here?


```suggestion
  resourceClass: ResourceClassWithIdFiltered;
```

client/src/features/sessionsV2/SessionList/SessionCard.tsx
```diff
@@ -99,6 +113,11 @@ export default function SessionCard({ project, session }: SessionCardProps) {
           >
             <div>
               <ActiveSessionButton
+                usageAvailable={{
```
leafty: See comment about using `resourceClass` in the prop: you would just pass `resourceClass={userLauncherClass}`.

client/src/features/sessionsV2/api/computeResources.openapi.json
leafty, line 1: Comment about the API: I don't understand why the usage values are returned only for the `GET /resource_pool` endpoint. Would it not make sense to include that in all other resource pool and resource class endpoints?

leafty (approved):
:+1:

client/src/features/sessionsV2/session.utils.tsx
ciyer: Fixed in b49635dbd8dd714e6657f6994c966d45590eb279

client/src/features/sessionsV2/session.utils.tsx
ciyer: Fixed in 53620c2848f7d42f4fdafb25af67022aede532eb

client/src/features/sessionsV2/components/SessionButton/ActiveSessionButton.tsx
ciyer: Fixed in 1553e1b636f8de0add7f08ca5a2f8a0d7431896f

client/src/features/sessionsV2/SessionList/SessionCard.tsx
ciyer: Fixed in 1553e1b636f8de0add7f08ca5a2f8a0d7431896f

PR #4130, opened 2026-04-15
fix: markdown table rendering on ToS and privacy policy pages

wesjdj:
Before:
SCR-20260415-nezi

After:
SCR-20260416-iybu

### DIFF
--- a/client/src/components/markdown/Markdown.tsx
+++ b/client/src/components/markdown/Markdown.tsx
@@ -111,6 +111,14 @@ export default function Markdown({
       className={styles.renkuMarkdown}
     >
       <ReactMarkdown
+        components={{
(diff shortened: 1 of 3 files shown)


client/src/components/markdown/Markdown.tsx
leafty, line 114: We can add:

```suggestion
      <ReactMarkdown
        components={{
          table: ({ className, ...props }) => (
            <table
              className={cx("table", "table-striped", className)}
              {...props}
            ></table>
          ),
        }}
```

Image

leafty (commented):
Great, just one thing to make the table look nice 👇

leafty (approved):

PR #4133, opened 2026-04-15
feat: enable build from private repositories

sgaist:
## Describe your changes

Users who have pull access to a repository can now use them for building custom images.

The UI changes reflect the new possibility but is likely lacking information to the users such as using a different registry to store the resulting image.

## Links and tickets

Depends on https://github.com/SwissDataScienceCenter/renku-data-services/pull/1250
Should be merged after https://github.com/SwissDataScienceCenter/renku/pull/4428 is done


leafty (commented):
1. How do we convey to the user that the image is going to be public or private?
2. I don't think the resulting image is _usable_ at the moment, so do we wait until using a private image from harbor is working?

sgaist:
@leafty 

1) Good point, I think this should be discussed more thoroughly with our designer friends. My main goal with this initial implementation was to make it possible to use the feature from the UI and not just through the API.
2) This indeed requires https://github.com/SwissDataScienceCenter/renku/pull/4428 to be merged first.

I'll check and fix the conflict over there and fix the description here.

sgaist:
Beside the renku PR listed above, https://github.com/SwissDataScienceCenter/renku-data-services/pull/1265 is required for sessions to be able to pull the image.

leafty (commented):
Offcanvas information needs to be updated, and the badge on the session launcher as well.

Image

sgaist:
@leafty I think this particular issue comes from the backend not the UI.

The API to check session images does not handle the private build part yet.

leafty:
> @leafty I think this particular issue comes from the backend not the UI.
> 
> The API to check session images does not handle the private build part yet.

Yes, but if the image is not accessible (because the user does not have permissions), then this message is still not correct for images built from a private repository.

leafty (commented):
@sgaist Maybe you need to run `npm ci` in `client/`. The eslint and formatter may have been updated in `main`.

leafty (approved):

PR #4134, opened 2026-04-15
adjust main entities page

lorenzo-cavazzi:
This PR updates the header on the user, group, and projects pages.
It updates a few other details related to these entities.

Here is the Figma reference; consider the first row only (Data Connector page not included) https://www.figma.com/design/8e1SX4SC8qqL5QQnEvzQis/Standardize-off-canvases---full-pages?node-id=377-54&p=f&t=nQiuZoxZRN7Os8NP-0

/deploy renku=lorenzo/data-connector-modal


client/src/components/entityWatermark/EntityWatermark.tsx
```diff
@@ -17,29 +17,41 @@
...
 
 import styles from "./entityWatermark.module.scss";
 
 interface EntityWatermarkProps {
-  type: "user" | "group";
+  className?: string;
+  type: "group" | "user" | "project";
 }
-export function EntityWatermark({ type }: EntityWatermarkProps) {
+export function EntityWatermark({ className, type }: EntityWatermarkProps) {
```
leafty, line 28: Why don't we keep the positioning trick inside this component? Now, you have to remember to use TWO components instead of one. And I see that it's always the same combination used in the `{User,Group,Project}Layout` component.

leafty (changes requested):

client/src/features/ProjectPageV2/ProjectPageHeader/ProjectAutostartRedirectBanner.tsx
leafty, line 1: Why is this removed?

lorenzo-cavazzi: It wasn't shown anymore, we removed the legacy components and now we only show the "data removal deadline" banner even with the `autostart` parameter

leafty: No, we show that component, e.g. https://renkulab.io/projects/mltox/adore-modeling?autostart=1

client/src/components/entityWatermark/EntityWatermark.tsx
lorenzo-cavazzi: The plan is to use a watermark for offcanvas too, without the `position-aboslute` class. I'll just need to add a `size` prop. I see how it looks unnecessary for this PR. Do you want me to change it back and add it again the offcanvas PR later?

client/src/features/ProjectPageV2/ProjectPageHeader/ProjectAutostartRedirectBanner.tsx
leafty: I think you are confusing components. This is shown when users are redirected from registered links from legacy projects.

lorenzo-cavazzi: Oh! I couldn't find a way to be redirected to `autostartRedirect=true`. Thanks for explaining, I'll add that back

client/src/components/entityWatermark/EntityWatermark.tsx
leafty: Even if you have something different for offcanvas, you should not make it a guess-work for users of the component. They should import and use a single component. If the offcanvas is not using a `position: relative > position:absolute` trick for overflow, then add props to the component or make another component for it.

client/src/features/ProjectPageV2/ProjectPageHeader/ProjectAutostartRedirectBanner.tsx
lorenzo-cavazzi: Fixed here :pray: https://github.com/SwissDataScienceCenter/renku-ui/pull/4134/commits/b3a603aeb93b68212500d67b3846e5750b017b78

client/src/components/entityWatermark/EntityWatermark.tsx
lorenzo-cavazzi: ~I'm not sure I follow. Is this what you are suggesting? https://github.com/SwissDataScienceCenter/renku-ui/pull/4134/commits/2271e068e4479938ba47cb52df2c1b11fde8c104~

lorenzo-cavazzi: Oh sorry, I guess you meant to remove the separate placeholder! Now I get it :facepalm: :sweat_smile: 

That makes sense! Here we go https://github.com/SwissDataScienceCenter/renku-ui/pull/4134/commits/b7be2567d6f2a157a7dbf88c6a1154139f2b2eff

client/src/features/ProjectPageV2/ProjectPageHeader/ProjectAutostartRedirectBanner.tsx
leafty: Yes, now the pages just have to import and use one simple component, like it was before for some.

leafty (approved):
LGTM

PR #4145, opened 2026-04-21
feat: add a data connector page

lorenzo-cavazzi:
This PR adds a standalone Data Connector page that can be opened from the data connectors side panels.

**Added elements**

- New full page for Data Connectors with SSR for linking DC to social media. For now, the page can be accessed on the Offcanvas by clicking on the icon to enlarge the visualization, next to the `X` close button .
- Settings page opening modals to apply the changes.

**Missing elements**

- There are a few places where the new DC page can be linked. Since the PR is big already, those changes to (a bunch of) other pages will be added later separately.
- There is an overlap between the new Cards in the DC overview and the elements used in the offcanvas. Updating the offcanvas will be in a separate PR. Mind that the new element can easily fit in there and the old ones will be removed.
- I'll add Cypress tests specific to this new page in a separate PR

image

image

/deploy renku=release-2.17.0

### DIFF
--- a/client/src/features/dataConnectorsV2/components/DataConnectorActions.tsx
+++ b/client/src/features/dataConnectorsV2/components/DataConnectorActions.tsx
@@ -70,30 +70,30 @@ import { getDataConnectorScope } from "./dataConnector.utils";
 import DataConnectorCredentialsModal from "./DataConnectorCredentialsModal";
 
 interface DataConnectorRemoveModalProps {
-  dataConnector: DataConnectorRead;
+  dataConnector?: DataConnectorRead | null;
   dataConnectorLink?: DataConnectorToProjectLink;
   isOpen: boolean;
   onDelete: () => void;
   toggleModal: () => void;
 }
 
-function DataConnectorRemoveDeleteModal({
+export function DataConnectorRemoveDeleteModal({
   dataConnector,
   onDelete,
   toggleModal,
   isOpen,
 }: DataConnectorRemoveModalProps) {
   const { permissions, isLoading: isLoadingPermissions } =
-    useDataConnectorPermissions({ dataConnectorId: dataConnector.id });
+    useDataConnectorPermissions({ dataConnectorId: dataConnector?.id });
 
   const {
     data: dataConnectorLinks,
     isLoading: isLoadingLinks,
     isError: isLoadingLinksError,
   } = useGetDataConnectorsByDataConnectorIdProjectLinksQuery({
-    dataConnectorId: dataConnector.id,
+    dataConnectorId: dataConnector?.id ?? "",
   });
-  const [deleteDataConnector, { isLoading, isSuccess }] =
+  const [deleteDataConnector, { error, isLoading, isSuccess }] =
     useDeleteDataConnectorsByDataConnectorIdMutation();
 
   const [typedName, setTypedName] = useState("");
@@ -111,9 +111,9 @@ function DataConnectorRemoveDeleteModal({
   }, [isOpen, isSuccess, onDelete]);
   const onDeleteDataCollector = useCallback(() => {
     deleteDataConnector({
-      dataConnectorId: dataConnector.id,
+      dataConnectorId: dataConnector?.id ?? "",
     });
-  }, [deleteDataConnector, dataConnector.id]);
+  }, [deleteDataConnector, dataConnector?.id]);
 
   return (
     <Modal size="lg" isOpen={isOpen} toggle={toggleModal} centered>
@@ -144,7 +144,7 @@ function DataConnectorRemoveDeleteModal({
                       possible that it is used in some projects.
                     </p>
                     <p>
-                      Please type <strong>{dataConnector.slug}</strong>, the
+                      Please type <strong>{dataConnector?.slug}</strong>, the
                       slug of the data connector, to confirm.
                     </p>
                     <Input
@@ -179,7 +179,7 @@ function DataConnectorRemoveDeleteModal({
                       )}
                     </p>
                     <p>
-                      Please type <strong>{dataConnector.slug}</strong>, the
+                      Please type <strong>{dataConnector?.slug}</strong>, the
                       slug of the data connector, to confirm.
                     </p>
                     <Input
@@ -195,6 +195,12 @@ function DataConnectorRemoveDeleteModal({
             userPermissions={permissions}
           />
         )}
+        {error && (
+          <RtkOrDataServicesError
+            className={cx("mb-0", "mt-3")}
+            error={error}
+          />
+        )}
       </ModalBody>
       <ModalFooter>
         <div className="d-flex justify-content-end">
@@ -208,9 +214,8 @@ function DataConnectorRemoveDeleteModal({
               <Button
                 color="danger"
                 className={cx("float-right", "ms-2")}
-                disabled={isLoading || typedName !== dataConnector.slug.trim()}
+                disabled={isLoading || typedName !== dataConnector?.slug.trim()}
                 data-cy="delete-data-connector-modal-button"
-                type="submit"
                 onClick={onDeleteDataCollector}
               >
                 {isLoading ? (
@@ -274,10 +279,10 @@ function DataConnectorRemoveUnlinkModal({
     if (!linkId) return;
 
     unlinkDataConnector({
-      dataConnectorId: dataConnector.id,
+      dataConnectorId: dataConnector?.id ?? "",
       linkId,
     });
-  }, [unlinkDataConnector, dataConnector.id, linkId]);
+  }, [unlinkDataConnector, dataConnector?.id, linkId]);
 
   return (
     <Modal size="lg" isOpen={isOpen} toggle={toggleModal} centered>
@@ -307,7 +312,7 @@ function DataConnectorRemoveUnlinkModal({
                     <>
                       <p>
                         Are you sure you want to unlink the data connector{" "}
-                        <strong>{dataConnector.slug}</strong> from the project{" "}
+                        <strong>{dataConnector?.slug}</strong> from the project{" "}
                         <strong>
                           {projectNamespace}/{projectSlug}
                         </strong>
@@ -340,7 +345,7 @@ function DataConnectorRemoveUnlinkModal({
                     className={cx("float-right", "ms-2")}
                     disabled={isLoadingUnlink}
                     data-cy="delete-data-connector-modal-button"
-                    type="submit"
+                    type="button"
                     onClick={onDeleteDataCollector}
                   >
                     {isLoadingUnlink ? (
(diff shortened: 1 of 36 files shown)


client/src/features/dataConnectorsV2/components/DataConnectorActions.tsx
```diff
@@ -208,9 +214,8 @@ function DataConnectorRemoveDeleteModal({
               <Button
                 color="danger"
                 className={cx("float-right", "ms-2")}
-                disabled={isLoading || typedName !== dataConnector.slug.trim()}
+                disabled={isLoading || typedName !== dataConnector?.slug.trim()}
                 data-cy="delete-data-connector-modal-button"
-                type="submit"
```
andre-code, line 213: could we add `type="button"` here for consistency with the unlink button in `DataConnectorRemoveUnlinkModal`?

client/src/features/dataConnectorsV2/show/DataConnectorShow.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+function DataConnectorInformationWrapper() {
+  const ctx = useNamespaceContext();
+  const { kind } = ctx;
+  const dataConnector = ctx.kind === "dataConnector" ? ctx.dataConnector : null;
+
+  // ? Not that any of this should ever happen... Hence the return null.
+  if (!ctx || kind !== "dataConnector" || !dataConnector) return null;
+
+  return <DataConnectorInfoBox dataConnector={dataConnector} headerTag="h2" />;
```
andre-code, line 36: question: `headerTag` is only used with the value "h2" (matching the default). Since this flexibility isn't being used yet, would it make sense to hardcode tag="h2" on <CardHeader> and drop the prop for now?

client/src/features/dataConnectorsV2/components/DataConnectorInfoBox.tsx
```diff
@@ -0,0 +1,330 @@
...
+    <>
+      {nonCredentialFields.map((fieldName) => {
+        const title =
+          fieldName == "provider" && hasAccessMode
+            ? "Mode"
+            : capitalize(fieldName);
+        const value =
+          dataConnector.storage.configuration[fieldName]?.toString() ?? "";
+        return (
+          <>
```
andre-code:
```suggestion
         <InfoEntry title={title}>{value}</InfoEntry>
```

client/src/features/dataConnectorsV2/components/DataConnectorProjectsBox.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { DataConnectorRead } from "../api/data-connectors.api";
+import useDataConnectorProjects from "./useDataConnectorProjects.hook";
+
+interface DataConnectorProjectsBoxProps {
+  dataConnector: DataConnectorRead;
+  headerTag?: "h2" | "h3" | "h4";
+}
+export default function DataConnectorProjectsBox({
+  dataConnector,
+  headerTag = "h2",
```
andre-code, line 16: same suggestion to remove headerTag prop

client/src/features/dataConnectorsV2/components/DataConnectorCredentialsBox.tsx
```diff
@@ -0,0 +1,101 @@
...
+  dataConnector,
+  headerTag = "h2",
+}: DataConnectorCredentialsBoxProps) {
+  // Sensitive fields
+  const sensitiveFields = dataConnector.storage.sensitive_fields
+    ? dataConnector.storage.sensitive_fields.map((f) => f.name)
+    : [];
+  const anySensitiveField = Object.keys(
+    dataConnector.storage.configuration
+  ).some((key) => sensitiveFields.includes(key));
```
andre-code:
```suggestion
  + const sensitiveFieldNames = useMemo(
+   () => dataConnector.storage.sensitive_fields?.map((f) => f.name) ?? [],
+   [dataConnector.storage.sensitive_fields]
+ );
+ const anySensitiveField = useMemo(
+   () => Object.keys(dataConnector.storage.configuration)
+         .some((key) => sensitiveFieldNames.includes(key)),
+   [dataConnector.storage.configuration, sensitiveFieldNames]
+ );
```

client/src/features/dataConnectorsV2/components/DataConnectorCredentialsBox.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  // Fields requiring credentials and their status
+  const credentialFieldDefinitions = useMemo(
+    () =>
+      getCredentialFieldDefinitions({
+        storage: dataConnector.storage,
+        sensitive_fields: dataConnector.storage.sensitive_fields,
+      }),
+    [dataConnector]
```
andre-code:
```suggestion
    [dataConnector.storage, dataConnector.storage.sensitive_fields]
```

andre-code (commented):
The page looks great!


I notice the close (×) and expand (↗) icons have inconsistent visual weight (the lines). Maybe we need to consider for this case use a different icon for close?

Also consider to add a tooltip in the expand icon, like "Open in full page"

Image




We should consider adding a label in the page header. The icon alone doesn't feel sufficient. Previously, the project image helped distinguish the page from others, and users and groups benefit from having the avatar for that same reason.
We could reuse the label already used in the side panels and apply it consistently across all pages (users, groups, data connectors, projects). What do you think?

Image

Image

client/src/features/dataConnectorsV2/components/DataConnectorView.tsx
```diff
@@ -143,14 +154,21 @@ export default function DataConnectorView({
...
             data-cy="data-connector-view-back-button"
             data-bs-dismiss="offcanvas"
             onClick={toggleView}
           ></button>
+          <Link
+            className="link-secondary"
+            data-cy="data-connector-standalone-page-link"
+            to={dataConnectorStandaloneLink}
+          >
+            <ArrowsAngleExpand />
```
leafty: Accessibility, required for icon buttons.

```suggestion
            <ArrowsAngleExpand />
            <span className="visually-hidden">Open in full page</span>
```

lorenzo-cavazzi:
> I notice the close (×) and expand (↗) icons have inconsistent visual weight (the lines). Maybe we need to consider for this case use a different icon for close?
Also consider to add a tooltip in the expand icon, like "Open in full page"

Yeah that's a good point, the two buttons are visually too different and a tooltip might help here. I was reluctant to swap out the standard close button from Bootstrap, even because the expand/close icons from Bootstrap don't have the same visual footprint :facepalm: 

I went for this solution, using the font size classes to find a good visual compromis https://github.com/SwissDataScienceCenter/renku-ui/pull/4145/commits/ff361abb145682842a7eb20b9f1ce9f939835627
Wdyt?
image

If you like it, I can create a shared component to reuse in the other off-canvas as well, and update the close button in other places to use a single style everywhere.

client/src/features/dataConnectorsV2/components/DataConnectorView.tsx
lorenzo-cavazzi: Good point, included here https://github.com/SwissDataScienceCenter/renku-ui/commit/ff361abb145682842a7eb20b9f1ce9f939835627
Mind that I added more changes to make the 2 icons visually more similar -- see https://github.com/SwissDataScienceCenter/renku-ui/pull/4145#issuecomment-4396733538

lorenzo-cavazzi:
> We could reuse the label already used in the side panels and apply it consistently across all pages (users, groups, data connectors, projects). What do you think?

EDIT: thank you for the suggestion! The textual name is nice; here is a solution to introduce it on smaller screens when the big icon is hidden https://github.com/SwissDataScienceCenter/renku-ui/pull/4145/commits/45594541e34e3f1756d86f558ef86eebbbea340a

image


~I'll discuss it with the build team. The full pages shouldn't be too difficult to distinguish perhaps it's best to have some visual cue always there~

lorenzo-cavazzi:
P.S: I noticed in your screenshots that the Icons were overlapping to the titles. That got accidentally broken in a recent commit :facepalm: 
Here is the fix https://github.com/SwissDataScienceCenter/renku-ui/pull/4145/commits/e18c08188af603d0c80474bb85fe3d6bce9a0fe2

image

client/src/features/dataConnectorsV2/show/DataConnectorShow.tsx
lorenzo-cavazzi: I'm replacing the older components on the Data Connectors side panel where I only need a different header tag. It will be in a separate PR; if you don't mint, I'll leave it as is since I'm going to use it soon.

client/src/features/dataConnectorsV2/components/DataConnectorCredentialsBox.tsx
lorenzo-cavazzi: Even just `[dataConnector.storage]` works :slightly_smiling_face: https://github.com/SwissDataScienceCenter/renku-ui/pull/4145/commits/fe0fb302421c2cfecd06626363ade43a9a77b0e6

client/src/features/dataConnectorsV2/components/DataConnectorCredentialsBox.tsx
lorenzo-cavazzi: :+1:  https://github.com/SwissDataScienceCenter/renku-ui/pull/4145/commits/fe0fb302421c2cfecd06626363ade43a9a77b0e6

client/src/features/dataConnectorsV2/components/DataConnectorInfoBox.tsx
lorenzo-cavazzi: :+1: https://github.com/SwissDataScienceCenter/renku-ui/pull/4145/commits/e1f97a5db89b8dd02dba83b8868ac5bc484e9ed2

client/src/features/dataConnectorsV2/components/DataConnectorActions.tsx
lorenzo-cavazzi: I made all those buttons `type=submit` https://github.com/SwissDataScienceCenter/renku-ui/pull/4145/commits/1f89ba48323d0a93ec6ada1382be96343fdaec7d

lorenzo-cavazzi:
Mind there was a bug when deleting data connectors that would prevent redirecting properly. This shoulds fix it https://github.com/SwissDataScienceCenter/renku-ui/pull/4145/commits/d2074d1da3b6bc6651de03a77b046229c8d931f8

@andre-code thank you for the review, I should have addressed all your points

andre-code (approved):
lgtm

lorenzo-cavazzi:
Tests here https://github.com/SwissDataScienceCenter/renku-ui/pull/4162

PR #4146, opened 2026-04-22
feat: add search in add connection modal

andre-code:
PR to add search functionality in add connection modal.

Screenshot 2026-04-28 at 16 52 22

/deploy renku=release-2.17.0


client/src/features/connectedServices/ConnectedServicesPage.tsx
```diff
@@ -487,50 +514,83 @@ function AddIntegrationModal({
...
-              {visibleProviders.map(({ provider, connection }) => (
-                <ListGroupItem
-                  key={provider.id}
-                  action={true}
-                  data-cy="provider-item"
+            <div className="mb-4">
+              <Label for="add-integration-search-input">Search</Label>
+              <InputGroup>
+                <Input
+                  className="lg"
```
lorenzo-cavazzi: `lg` is not changing the size here. You meant to add `input-group-lg` to the parent? The large input really looks gigantic and we don't use it in other places. I would keep it as is an remove the class
```suggestion
```

client/src/features/connectedServices/ConnectedServicesPage.tsx
```diff
@@ -487,50 +514,83 @@ function AddIntegrationModal({
...
+              <span className={cx("small", "text-muted")}>
+                No integrations found for &quot;{userSearchInput.trim()}&quot;.
+              </span>
+            ) : (
+              <ListGroup
+                className={cx(
+                  "bg-white",
+                  "rounded-3",
+                  "border",
+                  "border-rk-green"
```
lorenzo-cavazzi: That's a legacy class. Leftover from a copy-paste?
```suggestion
```

client/src/features/connectedServices/ConnectedServicesPage.tsx
```diff
@@ -457,16 +462,38 @@ function AddIntegrationModal({
 }: AddIntegrationModalProps) {
   const [showAllIntegrations, setShowAllIntegrations] = useState(false);
   const isListExpanded = isOpen && showAllIntegrations;
+  const searchInputRef = useRef<HTMLInputElement | null>(null);
+  const [userSearchInput, setUserSearchInput] = useState("");
```
lorenzo-cavazzi, line 464: Should we reset this value when closing (or opening) the modal? At the moment, the input is preserved, which is a little annoying after connecting, because the modal opens showing nothing

lorenzo-cavazzi (changes requested):
Great to be able to filter the potentially long list! I really like the whole Integrations page 🚀 

Just a couple of minor details before merging.

client/src/features/connectedServices/ConnectedServicesPage.tsx
andre-code: Good catch, thanks! I have added `resetModalState` to clear the input on close.

lorenzo-cavazzi (approved):
Lgtm! 🚀

PR #4147, opened 2026-04-22
refactor: react-router route for index page

leafty:
Move the index (`/`) page to a proper react-router route.

/deploy renku=lorenzo/data-connector-modal


lorenzo-cavazzi (approved):
🚀

PR #4150, opened 2026-04-23
Apply usage quotas

ciyer:
# Description

  * feat: present usage quota information at session launch time
  * feat: show usage quota information when viewing session launcher details
  * feat: show usage quota for running or paused sessions
  * feat: show usage quota during session start
  * feat: prevent resume if usage quota is used 

Use the following deploy string when ready 

/deploy #extra-values=dataService.dataTasks.enableResourceRequestTracking=true

## Screenshots

image

image

image

image


olevski:
@ciyer I hope you dont mind that I edited the description so that this is deployed for the demo tomorrow. I will check if it is in a state that is demoable. If not we can skip the demo.

lorenzo-cavazzi (commented):
This looks very nice! 👏 
Does this require tagging again `data-services`? I see it points to `main`.

------

I have a couple of minor comments; it's likely too late to change now and they can be something for the cooldown, but I wonder if they were discussed with the build group.

* The "remaining quota" information shows up when starting the sessions. It might raise some questions, especially from users who have been using sessions for a while and now see that prominent message. Should we add a link to the documentation? Some like "Learn more about usage quota"?
* A session reaching the quota won't immediately shut down. I'm not sure about the delay, but users might see the message for some time and be able to close it and keep working. Should we phrase it closer to "the session will be paused soon" so that people won't think it's potentially a stale message and realize it might (will?) go down soon? Image

ciyer:
> This looks very nice! 👏 Does this require tagging again `data-services`? I see it points to `main`.

Probably. :) 

> I have a couple of minor comments; it's likely too late to change now and they can be something for the cooldown, but I wonder if they were discussed with the build group.
> 
> * The "remaining quota" information shows up when starting the sessions. It might raise some questions, especially from users who have been using sessions for a while and now see that prominent message. Should we add a link to the documentation? Some like "Learn more about usage quota"?

There was a decision to introduce usage quota, but not yet enforce them. There is supposed to be some communication about this, but it probably would be nice to also include some links to documentation in the UI, as you suggest. 

> * A session reaching the quota won't immediately shut down. I'm not sure about the delay, but users might see the message for some time and be able to close it and keep working. Should we phrase it closer to "the session will be paused soon" so that people won't think it's potentially a stale message and realize it might (will?) go down soon?

That message text is specified in the backend. It should be fixed there.

leafty (dismissed):

lorenzo-cavazzi (approved):

PR #4157, opened 2026-05-04
fix: address rough edges for images built from a private repository

leafty:
Add basic messages for users who cannot launch a session built from a private code repository.

Screenshot 2026-05-04 at 11 45 36


sgaist (approved):
Looks good !
Thanks

PR #4162, opened 2026-05-07
test: fix cypress test for DC

lorenzo-cavazzi:
This PR adds new tests for the standalone data connector page.
It moves some other DC specific tests to the new spec file.

/deploy


client/src/features/dataConnectorsV2/components/DataConnectorModal/DataConnectorModalBody.tsx
```diff
@@ -454,8 +454,9 @@ function DataConnectorMount({ dataConnector }: AddOrEditDataConnectorProps) {
                     }}
                   />
                   <Label
-                    for="data-controller-visibility-public"
                     className={cx("btn", "btn-outline-primary")}
+                    data-cy="data-controller-visibility-public-html"
+                    for="data-controller-visibility-public"
```
leafty, line 459: Remark: you could select based on the `for=` value.

tests/cypress/e2e/dataConnector.spec.ts
```diff
@@ -0,0 +1,1172 @@
...
+      .versions()
+      .userTest()
+      .listNamespaceV2()
+      .dataServicesUser({
+        response: {
+          id: "0945f006-e117-49b7-8966-4c0842146313",
+          username: "user-1",
+          email: "user1@email.com",
+        },
+      })
```
leafty: You should not use both `.userTest()` and `dataServicesUser()` at the same time as they provide the same fixture. It looks like `.dataServicesUser()` is a legacy fixture we should stop using and remove.

leafty (changes requested):

tests/cypress/e2e/dataConnector.spec.ts
lorenzo-cavazzi: Good point. I dropped `dataServicesUser` and use `userTest` everywhere https://github.com/SwissDataScienceCenter/renku-ui/pull/4162/commits/0b1dc9dc2ae0f4723c09afc8c3f7d8731a442d8a

tests/cypress/support/renkulab-fixtures/namespaceV2.ts
```diff
@@ -338,16 +338,14 @@ export function NamespaceV2<T extends FixturesConstructor>(Parent: T) {
       return this;
     }
 
-    readUserV2Namespace(args?: UserNamespaceV2Args) {
+    readGenericNamespace(args?: GenericNamespaceArgs) {
```
leafty: Why the rename? This is: read a user's namespace, not read a group's namespace.

lorenzo-cavazzi: You can read a data connector namespace with it -- which I'm doing now

lorenzo-cavazzi: It _can_ be used for group namespaces as well, there is no difference in the intercepted API (unless I'm missing something)

leafty: This is about which type of namespace is RETURNED, not from where you read it.

tests/cypress/support/renkulab-fixtures/user.ts
```diff
@@ -87,9 +87,25 @@ export function User<T extends FixturesConstructor>(Parent: T) {
       this.userTest({ user, dataServiceUser });
       return this;
     }
+
+    getUsersUser(args?: GetUserArgs) {
```
leafty: `getUsersUser` is confusing

```suggestion
    getAnyUser(args?: GetUserArgs) {
```

tests/cypress/support/renkulab-fixtures/namespaceV2.ts
leafty: There is a `readGroupNamespace` fixture just above which returns a group's namespace.

tests/cypress/support/renkulab-fixtures/user.ts
lorenzo-cavazzi: Yeah that's a better naming, thank you :+1:  https://github.com/SwissDataScienceCenter/renku-ui/pull/4162/commits/5d060c5b41a6c3ef9325d2e1a2c64b91779e80e3

tests/cypress/support/renkulab-fixtures/namespaceV2.ts
lorenzo-cavazzi: `readGroupV2Namespace` is another duplicate that we probably don't need. They intercept exactly the same path `/api/data/namespaces/{some_slug}` and accept the same arguments, but the name given to the argument for `some_slug` is different.

Would you prefer that I restore the previous naming and add a 3rd similar function to catch the data connector namespaces? I can do it if you think it's better, but it would be a 3rd duplicate.
P.S. I see it might be convenient in some cases, like  `readProjectV2Namespace` that reads namespace and name separately and composes it, but I don't think that should force us to have different naming for all the other cases.

leafty: ??? There is no such thing as a data connector namespace.

leafty: > readGroupV2Namespace is another duplicate that we probably don't need. They intercept exactly the same path /api/data/namespaces/{some_slug} and accept the same arguments, but the name given to the argument for some_slug is different.

The point is that the group version returns data for a group namespace and the user version returns data for a user namespace.

At the top level there are only group namespaces and user namespaces, there is no such thing as a data connector namespace.

lorenzo-cavazzi: `readUserV2Namespace` restored here https://github.com/SwissDataScienceCenter/renku-ui/pull/4162/commits/ae353d888e0da9190a84653bef0c437c387dba49

leafty (approved):

tests/cypress/e2e/projectV2.spec.ts
```diff
@@ -312,7 +312,7 @@ describe("Edit v2 project", () => {
     cy.get("a[title='Settings']").should("be.visible").click();
     // Fetch the second page of namespaces
     cy.wait("@listNamespaceV2");
-    cy.wait("@readUserV2Namespace");
+    cy.wait("@readGenericNamespace");
```
leafty:
```suggestion
    cy.wait("@readUserV2Namespace");
```

leafty (approved):

PR #4166, opened 2026-05-11
build: upgrade prettier

leafty:
Upgrade to the latest version of `prettier` which is required to then upgrade the API code generation.


lorenzo-cavazzi (approved):
👍

PR #4171, opened 2026-05-15
feat: add job launcher in launchers

andre-code:
PR to add job launchers in Lauchers before called Sessions.

/deploy  amalthea-sessions=main renku-data-services=eikek/session-launcher-type #no-tests


andre-code:
duplicated

PR #4181, opened 2026-05-22
chore: update API specs

leafty:


lorenzo-cavazzi (approved):
Thank you for updating this 🙏

lorenzo-cavazzi:
P.S. Is this the reason for the DC test to fail? https://github.com/SwissDataScienceCenter/renku-data-services/issues/1325 Should we wait until fixed?

leafty:
> P.S. Is this the reason for the DC test to fail? [SwissDataScienceCenter/renku-data-services#1325](https://github.com/SwissDataScienceCenter/renku-data-services/issues/1325) Should we wait until fixed?

Yes, there was a fix pushed to renku-data-services which fixed a DC API but then broke another DC API. There is no reason to wait.

PR #4184, opened 2026-05-22
feat: export data to the EnviDat platform

lorenzo-cavazzi:
Add support for exporting to the EnviDat platform.

P.S. Mind that testing the full flow isn't supported in this PR

image

image

/deploy renku=build/envidat


andre-code (changes requested):
Looks great! My only suggestion would be to disable all the fields in the edit form for this deposit provider. That way, users won't make changes only to realize later that those values can't actually be updated.

lorenzo-cavazzi:
> Looks great! My only suggestion would be to disable all the fields in the edit form for this deposit provider. That way, users won't make changes only to realize later that those values can't actually be updated.

Good suggestion! I changed that here https://github.com/SwissDataScienceCenter/renku-ui/pull/4184/commits/5f4d30dda6dff3e3f40cccd8352fd3715af57936

andre-code (approved):
lgtm ✅

PR #4200, opened 2026-06-02
refactor: change the unavailable session message

lorenzo-cavazzi:
Users complained about supposedly disappearing sessions that were in fact still running but not visible because the user was logged out.

With this PR, we show different messages on the inaccessible session page when users are logged out, suggestion to log in first. Hopefully, this should help.

P.S. Mind that we now show a prominent modal when a user's session expires and needs to log in again, so this change is a backup in case that won't work or users accidentally close the modal too fast.

/deploy renku=release-2.18.0


lorenzo-cavazzi:
Screenshot_20260602_133552

client/src/features/sessionsV2/SessionShowPage/SessionUnavailable.tsx
```diff
@@ -34,30 +39,44 @@ export default function SessionUnavailable() {
...
+        <InfoAlert timeout={0} dismissible={false}>
+          You can {link}.
+        </InfoAlert>
+      ) : (
+        <WarnAlert timeout={0} dismissible={false}>
+          <p>
+            You are not logged in. If you have an account, you should{" "}
+            {loginLink} and try again.
+          </p>
+          <p className="mb-0">Otherise, you can {link}.</p>
```
lorenzo-cavazzi:
```suggestion
          <p className="mb-0">Otherwise, you can {link}.</p>
```

client/src/features/sessionsV2/SessionShowPage/SessionUnavailable.tsx
```diff
@@ -34,30 +39,44 @@ export default function SessionUnavailable() {
...
-      </Alert>
+      {loggedIn ? (
+        <InfoAlert timeout={0} dismissible={false}>
+          You can {link}.
+        </InfoAlert>
+      ) : (
+        <WarnAlert timeout={0} dismissible={false}>
+          <p>
+            You are not logged in. If you have an account, you should{" "}
+            {loginLink} and try again.
```
leafty: Have the `{loginLink}` code here. Why? If `loggedIn` is true then we do not display that component but we still store it in memory. Doing `const component = (<a>...</a>)` is useful only if the same component is re-used.

lorenzo-cavazzi: :+1: 
https://github.com/SwissDataScienceCenter/renku-ui/pull/4200/commits/1be0fa1c62a32845a0e9ab7d6176c596b972abdd

leafty (approved):

PR #4203, opened 2026-06-03
feat: avoid Google Translate plugin crashes

### LINKED ISSUE #4202: Cannot Create projects when translation to Chinese in turned on in the browser
Seen on renku 2.17.2

To replicate:
- go to renkulab.io
- turn on translation to chiense (simplified)
- try to create a project
- it fails

image

Reported by @bethcg seen during a workshop with many chinese students who used translation software.

I looked at the API requests, the project gets successfully created in the backend.

I suspect this may be happening with many other languages - probably more likely on the ones that use non-latin alphabets.

I tested this on a brave browser.

lorenzo-cavazzi:
Avoid crashing the browser when Google Translate does its own thing with the DOM

/deploy renku=release-2.18.0
fix https://github.com/SwissDataScienceCenter/renku-ui/issues/4202


client/src/utils/helpers/domMutation.utils.ts
leafty, line 1: We should keep the original implementation from React.

1. Add a comment like this:
    ```tsx
    // Reference: https://github.com/facebook/react/issues/11538
    // This fixes React crashing when the DOM is manipulated by extensions, e.g. Google Translate
    ```
2. Name the method something like `fixExternalDOMMutationsCrashes()`
3. Use the implementation from https://github.com/facebook/react/issues/11538#issuecomment-417504600. As far as we know, it's the best implementation of the fix (from the React team).
4. Instead of logging to `console` we can also consider sending `warning` or `info` to Sentry.

client/src/root.tsx
```diff
@@ -47,12 +47,15 @@ import NotFound from "./not-found/NotFound";
...
 import { CONFIG_JSON } from "./utils/.server/config.constants";
 import { validatedAppParams } from "./utils/context/appParams.utils";
+import removeChildGuard from "./utils/helpers/removeChildGuard";
 import { initClientSideSentry } from "./utils/helpers/sentry/utils";
 import { makeMeta, makeMetaTitle } from "./utils/meta/meta";
 
 import "./styles/renku_bootstrap.scss";
 import "./utils/bootstrap/bootstrap.client";
 
+removeChildGuard();
```
leafty: This runs in the server too, so this is not the right place for this method.

leafty (changes requested):
The fix-method needs to run at the top of the client init script. At the moment the client entrypoint is hidden, so here is how to do it properly:

1. Run `$ npx react-router reveal` in `client/`:
    ```
    Entry file entry.client created at src/entry.client.tsx.
    Entry file src/entry.server.tsx already exists.
    ```
2. Open the new file `src/entry.client.tsx` and edit it:

```tsx
import { startTransition, StrictMode } from "react";
import { hydrateRoot } from "react-dom/client";
import { HydratedRouter } from "react-router/dom";
import { fixExternalDOMMutationsCrashes } from "./utils/...";

// Prevent crashes from browser extensions, e.g. Google Translate
fixExternalDOMMutationsCrashes();

startTransition(() => {
  hydrateRoot(
    document,
    <StrictMode>
      <HydratedRouter />
    </StrictMode>,
  );
});
```

client/src/entry.client.tsx
```diff
@@ -0,0 +1,17 @@
...
+  // Prevent crashes from browser extensions like Google Translate
+  fixExternalDOMMutationsCrashes();
+
+  hydrateRoot(
+    document,
+    <StrictMode>
+      <HydratedRouter />
+    </StrictMode>,
+  );
+});
```
leafty: If the method in `startTransition()` is executed more than once, then maybe this is better:

```suggestion
// Prevent crashes from browser extensions like Google Translate
fixExternalDOMMutationsCrashes();

startTransition(() => {
  hydrateRoot(
    document,
    <StrictMode>
      <HydratedRouter />
    </StrictMode>,
  );
});
```

We can check if it's better outside of the `startTransition()`.

client/src/utils/helpers/domMutation.utils.ts
```diff
@@ -0,0 +1,29 @@
...
+    Node.prototype.removeChild = function <T extends Node>(child: T): T {
+      // Translate moved it, detach from its real parent.
+      if (child.parentNode !== this) {
+        if (child.parentNode) {
+          child.parentNode.removeChild(child);
+        }
+        return child;
+      }
+      return originalRemoveChild.call(this, child) as T;
+    };
```
leafty, line 28: I think the one you have here creates a potential infinite loop by calling `child.parentNode.removeChild(child);`.

Here is the original implementation from React:

```suggestion
    Node.prototype.removeChild = function(child) {
    if (child.parentNode !== this) {
      if (console) {
        console.error('Cannot remove a child from a different parent', child, this);
      }
      return child;
    }
    return originalRemoveChild.apply(this, arguments);
  }
```

client/src/entry.client.tsx
leafty: Sorry, I am not sure if it's best to call before `startTransition` or inside it 😅

client/src/utils/helpers/domMutation.utils.ts
leafty: Note: to still call `originalRemoveChild()` when `child.parentNode != null`, then maybe we can have:

```ts
if (child.parentNode) {
  originalRemoveChild.apply(child.parentNode, arguments)
}
```

lorenzo-cavazzi: I found the other version at a linked reference with an explanation of why it's better, but I can't find it now -- I should have added the link in the first place.
But let's play safe, I haven't seen complaints with the original version in the thread, so we can stick to that.
Not sure if we want to log this anywhere.

P.S: Linter complains without the additional types, so I'll leave them
https://github.com/SwissDataScienceCenter/renku-ui/pull/4203/commits/9e23b4a23efeb75a40aa8b7a8d8ea7a723dd6a87

client/src/entry.client.tsx
lorenzo-cavazzi: It should run only once, let's try to leave it outside https://github.com/SwissDataScienceCenter/renku-ui/pull/4203/commits/5509d5d2a02887a1bf2ea682065ecd462ada2ff6

client/src/utils/helpers/domMutation.utils.ts
leafty: :+1: No problem if you find the link to an alternative version. And having types is fine.

lorenzo-cavazzi: > Note: to still call `originalRemoveChild()` when `child.parentNode != null`, then maybe we can have:
> 
> ```ts
> if (child.parentNode) {
>   originalRemoveChild.apply(child.parentNode, arguments)
> }
> ```

Sorry I missed that comment. Wdyt about the current version after https://github.com/SwissDataScienceCenter/renku-ui/commit/9e23b4a23efeb75a40aa8b7a8d8ea7a723dd6a87?
If you think it can be improved, feel free to push the changes directly

client/src/utils/helpers/domMutation.utils.ts
```diff
@@ -0,0 +1,33 @@
...
+      newNode: T,
+      referenceNode: Node | null,
+    ): T {
+      if (referenceNode && referenceNode.parentNode !== this) {
+        // Reference node was moved. Do not throw, append instead
+        return originalInsertBefore.call(this, newNode, null) as T;
+      }
+      return originalInsertBefore.call(this, newNode, referenceNode) as T;
+    };
+  }
```
leafty, line 48: Same situation here. Original code below:

```suggestion
     Node.prototype.insertBefore = function(newNode, referenceNode) {
    if (referenceNode && referenceNode.parentNode !== this) {
      if (console) {
        console.error('Cannot insert before a reference node from a different parent', referenceNode, this);
      }
      return newNode;
    }
    return originalInsertBefore.apply(this, arguments);
  }
```

lorenzo-cavazzi: Ouch, ofc :facepalm: 
https://github.com/SwissDataScienceCenter/renku-ui/pull/4203/commits/4f11f6810dc629e2e8a46f17b93fd5024d40936a

leafty (dismissed):
LGTM, one non-blocking comment below.

client/src/utils/helpers/domMutation.utils.ts
```diff
@@ -0,0 +1,40 @@
+// Guards against React crashes when the DOM is manipulated by extensions like Google Translate
+// ? Reference: https://github.com/facebook/react/issues/11538#issuecomment-417504600
+export function fixExternalDOMMutationsCrashes() {
```
leafty, line 9: Note: I have seen implementations use a variable on `window` to ensure this is executed only once, e.g.

```ts
if (typeof window === 'object' && window.__renku_appliedDOMFix == null) {
  // ...body...
  window.__renku_appliedDOMFix = true;
}
```

lorenzo-cavazzi: That's a good suggestion to prevent accidental double run. Worst case, it's an unnecessary safeguard
https://github.com/SwissDataScienceCenter/renku-ui/pull/4203/commits/ffa76319ee62dbb43da3de6265c14877d0590086

leafty (approved):
LGTM, tested on chrome: I crash in dev with translate and with this PR I can use translate and it seems to not crash.

PR #4207, opened 2026-06-04
build: move compression library from dev to prod

lorenzo-cavazzi:
/deploy renku=release-2.18.0


client/package.json
leafty, line 1: Note: maybe we also bring back `"@types/compression": "^1.8.1",`?

leafty (dismissed):

leafty (approved):

PR #4213, opened 2026-06-09
fix?: setup rtk listeners to allow using refetchOnFocus and the like

lorenzo-cavazzi:
We use RTK features like `refetchOnFocus` that don't work properly because we don't invoke `setupListeners`.

This PR is an attempt at fixing it.

/deploy renku=release-2.18.0


leafty (approved):

leafty:
Seems to force re-fetching when going to a hibernated tab.

PR #4215, opened 2026-06-09
fix: reset data connector useForm values on edit

### LINKED ISSUE #4208: Data connector edit form with stale values
Data Connector edit form bug:
1. You need a project with at least two data connectors that you can edit
2. Open the first edit form by clicking "Edit" on the project page (not the offcanvas)
3. Close the form
4. Open the other DC edit form (same, click "Edit" on the project page)
Observe that the form is filled in with wrong values.

lorenzo-cavazzi:
Properly reset when closing the Modal

/deploy renku=release-2.18.0

fix https://github.com/SwissDataScienceCenter/renku-ui/issues/4208

### DIFF
--- a/client/src/features/dataConnectorsV2/components/DataConnectorModal/DataConnectorModalBody.tsx
+++ b/client/src/features/dataConnectorsV2/components/DataConnectorModal/DataConnectorModalBody.tsx
@@ -17,7 +17,7 @@
  */
 
 import cx from "classnames";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 import { Eye, Globe, Lock, Pencil, PlusLg } from "react-bootstrap-icons";
 import { Controller, useForm } from "react-hook-form";
 import {
@@ -68,6 +68,7 @@ import DataConnectorSaveCredentialsInfo from "./DataConnectorSaveCredentialsInfo
 
 interface AddOrEditDataConnectorProps {
   dataConnector?: DataConnectorRead | null;
+  isOpen?: boolean;
   project?: Project;
   storageSecrets: DataConnectorSecret[];
   switchMode?: () => void;
@@ -237,6 +238,7 @@ function DataConnectorMount({ dataConnector }: AddOrEditDataConnectorProps) {
     formState: { dirtyFields, errors, touchedFields },
     setValue,
     getValues,
+    reset,
     watch,
   } = useForm<DataConnectorMountForm>({
     mode: "onChange",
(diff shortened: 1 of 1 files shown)


client/src/features/dataConnectorsV2/components/DataConnectorModal/DataConnectorModalBody.tsx
```diff
@@ -68,6 +68,7 @@ import DataConnectorSaveCredentialsInfo from "./DataConnectorSaveCredentialsInfo
 
 interface AddOrEditDataConnectorProps {
   dataConnector?: DataConnectorRead | null;
+  isOpen?: boolean;
```
leafty: Is this used?

client/src/features/dataConnectorsV2/components/DataConnectorModal/DataConnectorModalBody.tsx
```diff
@@ -237,7 +238,7 @@ function DataConnectorMount({ dataConnector }: AddOrEditDataConnectorProps) {
     formState: { dirtyFields, errors, touchedFields },
     setValue,
     getValues,
-    watch,
+    reset,
```
leafty, line 240: I don't think that you should remove `watch` and replace it with `useWatch`. They will both end up using the same implementation code from react-hook-form. `useWatch()` is useful so that you can pass only `control` down to children components (and not 10 props from `useForm`). Since you call `useForm()` and `watch()` here, you can keep using `watch` and you don't need to call `useWatch`.

lorenzo-cavazzi: I agree `watch` is more convenient in some situations, but here it's pretty much the same and the linter has started complaining about using `watch` https://react.dev/reference/eslint-plugin-react-hooks/lints/incompatible-library so I thought about replacing it with `useWatch`
image

client/src/features/dataConnectorsV2/components/DataConnectorModal/DataConnectorModalBody.tsx
lorenzo-cavazzi: Ouch, leftover :facepalm: 
https://github.com/SwissDataScienceCenter/renku-ui/pull/4215/commits/9c24f76843241d70bf3eb2598ac2e54e1da100a4

client/src/features/dataConnectorsV2/components/DataConnectorModal/DataConnectorModalBody.tsx
leafty: So the linter is specifically about `react-compiler`, which we are not using, so it's not the most important warning for now.

The react-hook-form issue is here -> https://github.com/react-hook-form/react-hook-form/issues/12298

It seems that using `useWatch({ control })` is correct with our version of RHF. Also, `useController()` may be more preferable than `<Controller>`.

leafty (approved):

client/src/features/dataConnectorsV2/components/DataConnectorModal/DataConnectorModalBody.tsx
lorenzo-cavazzi: OK. Thank you for sharing the link, it's good to know.

PR #4222, opened 2026-06-12
refactor: launcher actions

andre-code:
Refactors launcher buttons into shared session and job components with one readiness hook so the card and side panel use the same actions and adds cypress tests for launcher actions.

**AI disclosure:** Used Cursor as a coding assistant for parts of this PR (autocomplete/refactor suggestions). All code was written/reviewed and is fully understood by me; not all of it is AI-generated — Cursor was used selectively as a dev tool.

/deploy renku=feat/add-jobs amalthea-sessions=main renku-data-services=main extra-values=dataService.imageBuilders.enabled=true,dataService.imageBuilders.strategyName=renku-buildpacks-v3,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/renku-build/,dataService.imageBuilders.nodeSelector.renku.io/node-purpose=user,dataService.imageBuilders.tolerations[0].effect=NoSchedule,dataService.imageBuilders.tolerations[0].key=renku.io/dedicated,dataService.imageBuilders.tolerations[0].operator=Equal,dataService.imageBuilders.tolerations[0].value=user


client/src/features/permissionsV2/permissions.types.ts
```diff
@@ -21,3 +21,7 @@ export type RequestedPermission = "write" | "delete" | "change_membership";
 export type Permissions = {
   [key in RequestedPermission]: boolean;
 };
+
+export type ProjectPermissions = Permissions & {
+  isLoadingPermissions: boolean;
+};
```
leafty: This is weird. Why is it called ProjectPermissions when it is generic?

client/src/features/ProjectPageV2/utils/useProjectPermissions.hook.ts
```diff
@@ -43,13 +43,18 @@ export default function useProjectPermissions({
     }
   }, [fetchPermissions, isUninitialized, projectId]);
 
+  const isLoadingPermissions = isLoading || isUninitialized;
```
leafty: Logic error:


```suggestion
  const isLoadingPermissions = isLoading || (projectId && isUninitialized);
```

You can see above that we trigger fetching data when `projectId && isUninitialized` => this is when the projectId is set and the API fetch has not been triggered yet.

If you do not check for `projectId` (not null, not empty), then `isUninitialized` will be true forever.

client/src/features/sessionsV2/components/launcherActions/job/JobPanelSubmit.tsx
leafty, line 1: Misnomer, the parent component is an Offcanvas. Use `JobOffcanvasSubmit.tsx`.

client/src/features/sessionsV2/components/launcherActions/job/JobCardActions.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+  const displayBuildActions =
+    isCodeEnvironment &&
+    write &&
+    (resolvedUseOldImage || lastBuild?.status !== "succeeded");
+
+  const submitTooltip =
+    resolvedUseOldImage && containerImage?.accessible !== false
+      ? `Launch ${categoryDefinition.text.inline} using an older image`
+      : undefined;
```
leafty: Move this below L76 since they are not used before.

client/src/features/sessionsV2/components/launcherActions/job/JobLauncherActions.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+  if (isLoadingPermissions) {
+    return <CheckingLauncherButton />;
+  }
+
+  if (!write) {
+    return <JobPanelSubmit launcher={launcher} useOldImage={useOldImage} />;
+  }
+
+  const defaultAction = (() => {
```
leafty: You should probably `useMemo()` here.

client/src/features/sessionsV2/components/launcherActions/job/JobPanelSubmit.tsx
```diff
@@ -0,0 +1,55 @@
...
+    useOldImage,
+  });
+
+  const permissions = useProjectPermissions({ projectId: launcher.project_id });
+
+  const tooltip = isLaunchButtonDisabled
+    ? `No image available. ${
+        permissions.write
+          ? "Run the Build action"
+          : "Contact the project administrator "
```
leafty: Project administrator?

client/src/features/sessionsV2/components/launcherActions/job/JobSubmitButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+  tooltip?: string;
+}
+
+export default function JobSubmitButton({
+  className,
+  disabled,
+  launcherId,
+  tooltip,
+}: JobSubmitButtonProps) {
+  const targetId = jobSubmitButtonTargetId(launcherId);
```
leafty: ? You don't need this. Use a `ref` to the button and give it as a target for the tooltip.

client/src/features/sessionsV2/components/launcherActions/job/JobSubmitButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  return (
+    <>
+      <span id={targetId}>
```
leafty: No need for this extra span (bad HTML practice).


```suggestion
```

client/src/features/sessionsV2/components/launcherActions/job/JobSubmitButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <Button
```
leafty: You can have a ref to a `<Button>`, the prop is called `innerRef` with `reactstrap` components.

client/src/features/sessionsV2/components/launcherActions/session/SessionCardActions.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const shouldShowPrimaryLaunchAction =
+    displayLaunchSession &&
+    (!isExternalImageEnvironment || containerImage?.accessible);
+
+  const hasPrimaryLaunchTarget =
+    !isLoadingContainerImage &&
+    displayLaunchSession &&
+    (displayBuildActions || shouldShowPrimaryLaunchAction);
+
+  const defaultAction = (() => {
```
leafty: useMemo

client/src/features/sessionsV2/components/launcherActions/job/JobCardActions.tsx
```diff
@@ -0,0 +1,130 @@
...
+    return defaultAction;
+  }
+
+  return (
+    <ButtonWithMenuV2
+      color="primary"
+      default={defaultAction}
+      preventPropagation
+      size="sm"
+      isDisabledDropdownToggle={false}
```
leafty: ? Why specify the default value?

```suggestion
```

client/src/features/sessionsV2/components/launcherActions/session/SessionCardActions.tsx
```diff
@@ -0,0 +1,191 @@
...
+        <UncontrolledTooltip target={sessionLaunchLinkTargetId(launcher.id)}>
+          Cannot launch more than 1 session per session launcher.
+        </UncontrolledTooltip>
+      ) : resolvedUseOldImage &&
+        containerImage?.accessible !== false &&
+        hasPrimaryLaunchTarget ? (
+        <UncontrolledTooltip target={sessionLaunchLinkTargetId(launcher.id)}>
+          Launch {categoryDefinition.text.inline} using an older image
+        </UncontrolledTooltip>
+      ) : null}
```
leafty: 1. Bad practice: you should render the tooltip next to the target component so that you can link them up using `ref` and not hard-to-track DOM element IDs.
2. Why render two tooltips when the only thing to change is the tooltip content. You can do:
    ```tsx
    const tooltipContent = hasSession && hasPrimaryLaunchTarget ? (<>Content...</>) : containerImage?.accessible !== false && hasPrimaryLaunchTarget ? (<>Other content...</>) : undefined;
    if (tooltipContent) { ...render tooltip... }
    ```

client/src/features/sessionsV2/components/launcherActions/session/SessionPanelActions.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const permissions = useProjectPermissions({ projectId: launcher.project_id });
+
+  const isDisabled = isLaunchButtonDisabled || !!hasSession;
+
+  const launchButtonDisableReason = hasSession
+    ? "Cannot launch more than 1 session per session launcher."
+    : `No image available. ${
+        permissions.write
+          ? "Run the Build action"
+          : "Contact the project administrator "
```
leafty: Project administrator?

client/src/features/sessionsV2/components/launcherActions/shared/CheckingLauncherButton.tsx
```diff
@@ -17,36 +17,14 @@
...
-          event.stopPropagation(); // TODO: implement action when submit a job in other PR
-        }}
-        size="sm"
-        disabled={disabled}
-      >
-        <Send className={cx("bi", "me-1")} />
-        Submit
-      </Button>
-    </>
+    <Button color="outline-primary" className={cx("disabled")} size="sm">
```
leafty:
```suggestion
    <Button color="outline-primary" className="disabled" size="sm">
```

But also: why not set the `disabled` prop then?

client/src/features/sessionsV2/components/launcherActions/shared/CheckingLauncherButton.tsx
```diff
@@ -17,36 +17,14 @@
...
+      <Loader size={12} inline /> Checking launcher
```
leafty:
```suggestion
      <Loader className="me-1" size={12} inline />Checking launcher
```

client/src/features/sessionsV2/components/launcherActions/shared/ShowLauncherDetailsButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const launcherHash = useMemo(() => `launcher-${launcherId}`, [launcherId]);
+  const toggleLauncherView = useCallback(() => {
+    setHash((prev) => {
+      const isOpen = prev === launcherHash;
+      return isOpen ? "" : launcherHash;
+    });
+  }, [launcherHash, setHash]);
+
+  return (
+    <span id={`open-panel-btn-${launcherId}`}>
```
leafty: Bad practice: unnecessary span.


```suggestion
```

For tooltips, use `ref`, `innerRef`, or set the ID on the component itself (avoid if possible)

client/src/features/sessionsV2/useLauncherEnvironmentReadiness.hook.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  const isLastBuildRunning = lastBuild?.status === "in_progress";
+
+  const hasValidImage =
+    !isCodeEnvironment ||
+    lastBuild?.status === "succeeded" ||
+    containerImage?.accessible === true ||
+    useOldImage;
+
+  const showSubmitJob = !(isBuildInProgress && !hasValidImage && !useOldImage);
```
leafty: This is not readable:


```suggestion
  const showSubmitJob = !isBuildInProgress || hasValidImage || useOldImage;
```

client/src/features/sessionsV2/useLauncherEnvironmentReadiness.hook.ts
```diff
@@ -0,0 +1,151 @@
...
+  type ImageCheckResponse,
+} from "./api/sessionsV2.api";
+import { getLauncherEnvironmentFlags } from "./launcherEnvironment.utils";
+
+interface UseLauncherEnvironmentReadinessArgs {
+  launcher: SessionLauncher;
+  enabled?: boolean;
+  lastBuild?: Build;
+  lastSuccessfulBuild?: Build;
+  useOldImage?: boolean;
```
leafty: Q: Are there cases when we use `useOldImage={false}`?

leafty (commented):
General note, AI usage: please disclose AI use. See https://github.com/ghostty-org/ghostty/blob/main/AI_POLICY.md as a general starting point for guidance.

client/src/features/sessionsV2/components/launcherActions/job/JobPanelSubmit.tsx
andre-code: I would like to add an action that users can take if they want to launch the session and job. If they don't have the required access, they should be instructed to contact the project owner or manager.

leafty: That is not an actionable suggestion anyway: in Renku there is no way for users to contact other users inside or outside Renku. The only way this can work is if the two users already know each other.

The actual action users can take is to clone the project and re-build images from there.

client/src/features/permissionsV2/permissions.types.ts
andre-code: This type is really Permissions plus a loading flag from the hook, not the API's ProjectPermissions. I'll rename it to something like PermissionsWithLoadingState and keep it close to the hook so the generic permissions.types.ts stays API-aligned.

client/src/features/sessionsV2/components/launcherActions/job/JobPanelSubmit.tsx
leafty: The simplest course of action here is to remove that copy text and keep a TODO item for cooldown. There is no concept of "project administrator" and the suggestion (contacting a Renku user) is not actionable.

client/src/features/sessionsV2/useLauncherEnvironmentReadiness.hook.ts
andre-code: No, it is no longer needed, so I've removed it.

client/src/features/sessionsV2/components/launcherActions/job/JobSubmitButton.tsx
andre-code: Hi @leifty, I updated JobSubmitButton to avoid the previous span + id tooltip target workaround.
Instead, the button now uses aria-disabled and guards the action in handleClick (it returns early when the action should be disabled).
This keeps the element interactive enough for the tooltip to work, while still preventing submission when the action is not allowed.

Let me know what you think.

client/src/features/sessionsV2/components/launcherActions/session/SessionCardActions.tsx
andre-code: yes, I moved the tooltip logic next the button.

andre-code:
Hi @leafty , thanks for reviewing the PR. I’ve added the AI disclosure and simplified the logic by reusing more of the hook, with a single launcher action per type (job or session).
It’s ready for your review again.

leafty (approved):
LGTM

PR #4223, opened 2026-06-12
fix popover

lorenzo-cavazzi:
fix popover

### DIFF
--- a/client/src/features/sessionsV2/components/SessionForm/AdvancedSettingsFields.tsx
+++ b/client/src/features/sessionsV2/components/SessionForm/AdvancedSettingsFields.tsx
@@ -30,6 +30,7 @@ import { FormText, Input, Label } from "reactstrap";
 import { InputType } from "reactstrap/types/lib/Input";
 
 import LazyMarkdown from "~/components/markdown/LazyMarkdown";
+import Markdown from "~/components/markdown/Markdown";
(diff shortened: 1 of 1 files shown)


client/src/features/sessionsV2/components/SessionForm/AdvancedSettingsFields.tsx
```diff
@@ -75,7 +76,7 @@ function FormFieldLabel<T extends FieldValues>({
       </Label>
       {info && (
         <MoreInfo>
-          <LazyMarkdown>{info}</LazyMarkdown>
+          <Markdown>{info}</Markdown>
```
leafty, line 79: Why would the non-lazy version do anything here?

lorenzo-cavazzi: If the `Markdown` component hasn't loaded yet, the popover appears near the top of the screen. I guess the positioning cannot be correctly computed for the lazy component. The following times, Markdown is available and the popover shows properly.

I've tried forcing an additional refresh on the component but it didn't work well. If you have a more elegant solution, you are welcome to push it.

606384548-a6ca17f8-694b-401f-895f-be3ef6b0b0ba

leafty (approved):

PR #4241, opened 2026-06-23
chore: update form-data

leafty:
See: https://github.com/SwissDataScienceCenter/renku-ui/security/dependabot/469

/deploy


lorenzo-cavazzi (approved):
Lgtm!

PR #4243, opened 2026-06-24
feat(SessionLauncher): add repo access status pill for built environment

sgaist:
## Describe your changes

This PR adds a repository permissions badge to the session launcher card for build environments.

This allows the user to know whether there is something to do as having the image built and ready and access to the repository used to build it can be out of sync.

This is usefull especially when a project involves multiple repositories.

/deploy #notest extra-values=dataService.imageBuilders.enabled=true,dataService.imageBuilders.strategyName=renku-buildpacks-v3,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/renku-build/,dataService.imageBuilders.nodeSelector.renku.io/node-purpose=user,dataService.imageBuilders.tolerations[0].effect=NoSchedule,dataService.imageBuilders.tolerations[0].key=renku.io/dedicated,dataService.imageBuilders.tolerations[0].operator=Equal,dataService.imageBuilders.tolerations[0].value=user


leafty:
@sgaist I will try to review this soon.

leafty (commented):
Note: the client is not compiling at the moment. Maybe some types have changed with the merge of the jobs feature.l

sgaist:
> Note: the client is not compiling at the moment. Maybe some types have changed with the merge of the jobs feature.l

AFAICS, theses types haven't changed. It seems the latest changes seems to have broken the narrowing.

sgaist:
@leafty thanks for the fix !

leafty (commented):
As a general comment, I don't think adding a new pill is what we want for the session launcher card. This adds too much visual weight and will give mixed signals to users.

For example, this is what I get building an image from a public project:

Image

The image is perfectly usable without "requesting integration". I think a better way to tell users if they can start a session or not is to update the logic in the existing pill. I can make a PR soon for doing this.

sgaist:
I agree it does not make sense to show it for public repositories. I thought I added a conditional to show the new pill only for "non-public" repositories but it got lost somewhere along the way.

leafty (commented):
From the typescript compiler:

```
Error: src/features/sessionsV2/components/BuildStatusComponents.tsx(92,16): error TS2339: Property 'result' does not exist on type 'Build'.
  Property 'result' does not exist on type 'BuildCommonPart & BuildNotCompletedPart'.
```

client/src/features/sessionsV2/components/BuildStatusComponents.tsx
```diff
@@ -83,37 +81,51 @@ export function BuildStatusBadge({
...
-        : buildStatus === "in_progress"
-          ? "Build in progress"
-          : buildStatus === "cancelled"
-            ? "Build cancelled"
-            : buildStatus === "succeeded"
-              ? "Build succeeded"
-              : "Build failed";
+  const { data, isLoading } = useGetRepositoryQuery({
+    url: build.result?.repository_url,
+  });
```
leafty: I think you can query this only if the build is successful


```suggestion
  const { data, isLoading } = useGetRepositoryQuery(
    build.status === "succeeded"
      ? {
          url: build.result?.repository_url,
        }
      : skipToken,
  );
```

sgaist: You're right. I thought this was taken care of one level above but it just ensure that the build object exists not it success state.

leafty (approved):
LGTM

PR #4245, opened 2026-06-26
feat: add job styles, cancel job modal, launcher job hash

andre-code:
PR to add job styles in the project page, launcher side panel, and dashboard.
Add a cancel job modal for jobs that are in progress (starting or running).
Add launcher job hash in sessionHash to manage opening job details in the session panel from the job listing.

/deploy renku=feat/add-jobs amalthea-sessions=main renku-data-services=main extra-values=dataService.imageBuilders.enabled=true,dataService.imageBuilders.strategyName=renku-buildpacks-v3,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/renku-build/,dataService.imageBuilders.nodeSelector.renku.io/node-purpose=user,dataService.imageBuilders.tolerations[0].effect=NoSchedule,dataService.imageBuilders.tolerations[0].key=renku.io/dedicated,dataService.imageBuilders.tolerations[0].operator=Equal,dataService.imageBuilders.tolerations[0].value=user


client/src/utils/helpers/DurationUtils.ts
leafty, line 1: Can we remove the "short" format? I do not see where in the pitch this is explicitly asked for. This adds a change we can do without unless there is a specific product design for it.

client/src/utils/helpers/DurationUtils.ts
```diff
@@ -145,14 +173,20 @@ export function toShortHumanDuration({
...
  * @param now the current instant
+ * @param format use short unit labels (e.g. "3min ago") or long labels (e.g. "3 minutes ago")
+ * @param includeSuffix when false, returns only the duration (e.g. "3 minutes") without "ago" or "from now"
  * @returns a human-readable string
  */
 export function toHumanRelativeDuration({
   datetime: datetime_,
   now: now_,
+  format = "long",
+  includeSuffix = true,
```
leafty: This is dangerous. How do we differentiate between the past and the future? We should instead have:
`removeSuffix: "past" | "future" | false` and remove the suffix in the correct condition.

leafty (changes requested):

client/src/features/sessionsV2/SessionStyles.constants.ts
```diff
@@ -102,6 +125,15 @@ export const SESSION_TITLE = {
   [SESSION_STATES.SUCCEEDED]: "Session succeeded (TBD)",
   default: "Unknown status",
 };
+export const JOB_TITLE = {
+  [SESSION_STATES.RUNNING]: "My running job",
+  [SESSION_STATES.STARTING]: "Starting job...",
+  [SESSION_STATES.STOPPING]: "Dismissing my job...",
+  [SESSION_STATES.HIBERNATED]: "Paused job",
+  [SESSION_STATES.FAILED]: "Errored job", //eslint-disable-line spellcheck/spell-checker
```
leafty: Ignoring spellchecker on user facing messages?


```suggestion
  [SESSION_STATES.FAILED]: "Failed job",
```

client/src/features/sessionsV2/SessionStyles.constants.ts
```diff
@@ -112,3 +144,13 @@ export const SESSION_TITLE_DASHBOARD = {
   [SESSION_STATES.SUCCEEDED]: "Session succeeded (TBD)",
   default: "Unknown status",
 };
+
+export const JOB_TITLE_DASHBOARD = {
```
leafty: Why is this different?

client/src/features/sessionsV2/sessionStopIntent.slice.ts
leafty, line 1: Why do we need this slice? The "intent" is just a function of the session status.

client/src/utils/helpers/DurationUtils.ts
andre-code: This isn't explicitly mentioned in the pitch, but it was part of the design work done during this build. I don't see why a separate build would be needed for this change, could you pls explain your reasoning?

client/src/utils/helpers/DurationUtils.ts
andre-code: The component using the time value should decide whether to show the suffix. It's a yes/no choice, not something in between.
For example: when a session is hibernated, the time is in the future, so showing the suffix makes sense, it's clearly needed there. But for a running job, the time is in the past, and we show it as something like 'Running for 16s' instead. So here, not showing the suffix is also correct.
If a running job ever showed a future time, that would be a bug, and the suffix still wouldn't make sense.

client/src/features/sessionsV2/SessionStyles.constants.ts
andre-code: yes, I know. It is  a product requirement to explicitly display that message in this case.

client/src/features/sessionsV2/sessionStopIntent.slice.ts
andre-code: We want to display the correct follow-up action after the user clicks Cancel or Dismiss on a job. Cancel applies only when a job is in progress (starting or running), and Dismiss applies when the job is done. However, the status during both actions is the same: 'stopping'. So the idea is to somehow track the right action in order to display the correct message  'cancelling' or 'dismissing' in the status badge and button text.In a follow-up task, this should be added to the data-service, but as a workaround, this would be the solution

client/src/utils/helpers/DurationUtils.ts
leafty: The issue is that we should use standard ways of displaying time and this code is not achieving that. There must be a non-breaking space between digits and units (no space is not a valid display of time).
Also, we should validate that the short units are valid like "w" and "mo".

Given the complexity. I would rather not have the change here.

client/src/utils/helpers/DurationUtils.ts
leafty: But then you are not guarding against the potentially buggy conditions which is why I commented here. Either we should have a separate React component to display durations (e.g. `<DurationCaption>`) or we have to handle the relative duration being in wrong side of past vs future.

client/src/features/sessionsV2/sessionStopIntent.slice.ts
leafty: You introduce complex code (slice, custom API response handling, etc.) to correctly display a job row that should disappear within seconds. I think the cost of the code is not worth it. We can have a common phrase once the user has clicked "Dismiss" or "Cancel": "Removing my job...". This will show only for a couple of seconds.

client/src/features/sessionsV2/SessionStyles.constants.ts
andre-code: No difference there, this is is actually unused for jobs. It's a leftover from a previous version, I'll clean it up.

client/src/utils/helpers/DurationUtils.ts
andre-code: ok, I added your code suggestion.

fyi

category | status | config | currently (example)
-- | -- | -- | --
session | running | false | Launched 1 hour ago
job | running | past | Running for 16 seconds
session | starting | false | Launching since 3 minutes ago
job | starting | false | Starting since 3 minutes ago
session | hibernated (will_delete_at) | future | Session will be deleted in 2 hours
job | hibernated (will_delete_at) | future | Job will be dismissed in 2 hours
session | hibernated (will_hibernate_at) | false | Paused session for 2 hours ago
job | hibernated (will_hibernate_at) | false | Paused for 2 hours ago
session | hibernated (no timestamp) | — | Paused session …
job | hibernated (no timestamp) | — | Paused …
session | failed | false | Error in session (created 1 hour ago)
job | failed | false | Errored 1 hour ago
session | succeeded | false | Completed 2 hours ago
job | succeeded | false | Completed 2 hours ago

The UI will still visibly highlight an inconsistency if the time doesn't match the expected direction (past/future), but that's already the case in the current setup so I'll leave that to be fixed in a separate issue outside of this feature, and we can figure out what to do in those cases then.

client/src/features/sessionsV2/sessionStopIntent.slice.ts
andre-code: You're right, the complexity wasn't worth it for something this short-lived. I've removed the stop intent specific logic.

andre-code:
Thanks, @leafty all your suggestions have been addressed

leafty (approved):
LGTM

PR #4249, opened 2026-07-01
fix: indicate repeated queries to downstream

leafty:
Detects and marks repeated browser requests to break Sentry distributed tracing.

Issue: we have over-subscribed Sentry tracing span caused by the UI repeatedly querying the same endpoints (`GET /api/data/sessions` and `GET /api/data/alerts`). Sentry's `reactRouterTracingIntegration()` sets up distributed tracing by injecting Sentry tracing headers when the UI calls `fetch()` (from RTK-Query operations). The Sentry middleware does not offer a way to control when the tracing headers are added to a request.

Solution: we add a custom header `Renku-Repeated-Request: true` to indicate repeated requests so that the gateway can break off Sentry distributed tracing.

See also: https://github.com/SwissDataScienceCenter/renku-gateway/pull/878

/deploy renku-gateway=master extra-values=ui.client.sentry.enabled=true,ui.client.sentry.environment=renku-ci-ui-4249,ui.client.sentry.sampleRate=0.1,gateway.sentry.enabled=true,gateway.sentry.environment=renku-ci-ui-4249,gateway.sentry.sampleRate=0,dataService.sentry.enabled=true,dataService.sentry.environment=renku-ci-ui-4249,dataService.sentry.sampleRate=0.001,ui.client.sentry.dsn=https://b8aaa5810d2743f281350c0944c38ec8@o4509039310995456.ingest.de.sentry.io/4509044025131088,gateway.sentry.dsn=https://b4747210a4b6a37a4ccc33beee2dbea7@o4509039310995456.ingest.de.sentry.io/4509513007693904,dataService.sentry.dsn=https://5ec3c75875797b564a9f4654d5fc46d5@o4509039310995456.ingest.de.sentry.io/4509044015169616


client/src/utils/helpers/sentry/utils.ts
```diff
@@ -55,6 +57,56 @@ export function initClientSideSentry(params: AppParams) {
...
+  window.fetch = function wrappedFetch(
+    input: RequestInfo | URL,
+    init?: RequestInit,
+  ): Promise<Response> {
+    if (!isGatewayOrigin(input, GATEWAY_ORIGIN)) {
+      return origFetch(input, init);
+    }
+
+    const span = Sentry.getActiveSpan();
+    const traceId = span?.spanContext().traceId ?? "unknown";
```
andre-code, line 82: When there is no active Sentry span, all requests share traceId = "unknown", right?
Is it expected that unrelated requests can accumulate in the same counter and cross the repeated threshold?

If this behavior is intended, can we document it in code? If not, we should narrow these two paths to avoid false repeated-request tagging.

client/src/utils/helpers/sentry/utils.ts
```diff
@@ -55,6 +57,56 @@ export function initClientSideSentry(params: AppParams) {
...
+      input instanceof Request ? input.method : (init?.method ?? "GET");
+    const key = `${method.toUpperCase()}|${url.toString()}`;
+    const count = (requestCounts.get(key) ?? 0) + 1;
+    requestCounts.set(key, count);
+
+    if (count <= REPEATED_REQUEST_THRESHOLD) {
+      return origFetch(input, init);
+    }
+
+    const headers = new Headers(
```
andre-code, line 100: Can you confirm the intended precedence between init.headers and Request.headers?
My small worry is that we might end up ignoring `init.headers` in some cases, which could accidentally drop auth or content headers. I'd love to confirm the expected behavior for these three cases:

- `fetch(url, { headers: { Authorization: ..., Content-Type: ... } })`
- `fetch(new Request(url, { headers: ... }), { headers: ... })`
- `fetch(new Request(url), { headers: ... })`

client/src/utils/helpers/sentry/utils.ts
leafty: I am not sure I understand what is the issue. The counter is re-set when a new transaction is started (usually client-side navigation or UI interaction) so that we capture the requests linked to that new transaction.

If a pageload is not sampled or if a transaction is closed for idleness, then the trace ID will not be defined and we use `unknown`. This is still desired behavior: we want to stop downstream services (gateway and data services) from sampling repeated polling requests with the same parent trace ID.

client/src/utils/helpers/sentry/utils.ts
leafty: Fixed precedence: `init.header` always win if it is present. Then if `input` is an instance of `Request`, `fetch` will use the headers from `input.headers`.
Also to note that there is NO MERGING: `init.headers` win and erase any headers in `input.headers`.

You can check in your browser's console:
```js
let req = new Request("/")
req.headers.set("foo", "bar")

window.fetch(req) // will have the foo: bar header
window.fetch(req, { method: "HEAD" }) // will have the foo: bar header
window.fetch(req, { headers: {} }) // will NOT have the foo: bar header
window.fetch(req, { headers: { hello: "world" } }) // will NOT have the foo: bar header (will have hello: world)
```

andre-code (approved):
LGTM

PR #4252, opened 2026-07-02
chore: remove unused github action

leafty:
cc @lorenzo-cavazzi : I think we do not use this project board, so we probably can remove this github action from the repository.


lorenzo-cavazzi (approved):
Yes! Thank you

PR #4253, opened 2026-07-02
feat: add `infer` and `none` frontends to custom environments

SalimKayal:
/deploy #notest extra-values=dataService.imageBuilders.enabled=true,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/renku-build/,dataService.imageBuilders.nodeSelector.renku.io/node-purpose=user renku-data-services=salimkayal/feat/add-support-for-inferred-frontend-and-no-implicit-frontend renku=salimkayal/chore/update-buildpacks-version

### DIFF
--- a/client/src/features/sessionsV2/session.constants.tsx
+++ b/client/src/features/sessionsV2/session.constants.tsx
@@ -154,6 +154,22 @@ export const BUILDER_FRONTENDS = [
     /* eslint-enable spellcheck/spell-checker */
   },
   {
+    /* eslint-disable spellcheck/spell-checker */
+    value: "infer",
+    label: "Infer Frontend",
+    description:
+      "Infer frontend from dependencies, Currently only Marimo is supported.",
+    /* eslint-enable spellcheck/spell-checker */
+  },
+  {
+    /* eslint-disable spellcheck/spell-checker */
+    value: "none",
+    label: "None",
+    description:
+      "Do not inject a frontend. Please define the entrypoint in a Procfile if needed.",
+    /* eslint-enable spellcheck/spell-checker */
+  },
+  {
     value: "rstudio",
     label: "RStudio",
     description: "Web-based integrated development environment for R.",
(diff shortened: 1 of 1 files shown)


SalimKayal:
the problem is that `marimo` forces `target="_blank"` on the links for notebooks. So opening a notebook escapes the ui.

client/src/features/sessionsV2/session.constants.tsx
```diff
@@ -158,12 +158,28 @@ export const BUILDER_FRONTENDS = [
     label: "RStudio",
     description: "Web-based integrated development environment for R.",
   },
+  {
+    /* eslint-disable spellcheck/spell-checker */
+    value: "infer",
+    label: "Infer Frontend",
+    description:
+      "Infer frontend from dependencies, Currently only Marimo is supported.",
```
leafty: Note: I wonder if the user-facing phrasing should be "auto-detect"?

leafty (commented):
Code looks OK 👍

leafty (commented):
From Laura: update to the copy text:

Auto-Detect
Auto-detect the frontend from dependencies. Currently only Marimo is supported.

Custom / No Frontend
Renku does not add a frontend. For a custom frontend, define an entrypoint in a Procfile.

SalimKayal:
should the description for auto-detect add 

Falls back to Custom / No Frontend.

?

what about the marimo behavior wrt escaping the iframe ? Is this ok for now ? I am trying to implement the behavior in marimo currently and will propose a PR if maintainers deem it fit for marimo.

leafty:
I think it's ok if it either crashes or if it goes to no frontend if auto-detect doesn't work. We can describe more in the docs.

For the new tab behavior, I think we can't do much, so we just let marimo open new tabs, as things still work.

SalimKayal:
ok done.

SalimKayal:
and asked in a discussion in marimo whether there would be an interest in a PR for that 
https://github.com/marimo-team/marimo/discussions/8475

leafty (approved):

leafty:
Note: we can also change the job frontend to `none` after this.

PR #4258, opened 2026-07-08
feat: add UI for project storage in project page

boucherjeanphi:
- New btn "Add project storage" in `DataConnectorModal`
- Project storage appears below the list of data connnectors in `ProjectDataConnectorsBox`
- Side panel for Project storage
- Add / edit & delete project storage only available for project owners
- Update Shutdown session content to include project storage
- Custom icon for Renku storage

Screenshot from 2026-08-05 14-52-28
Screenshot from 2026-08-05 14-51-30

/deploy renku=feat-add-project-storage renku-data-services=feature/project-storage extra-values=dataService.projectStorage.enabled=true,dataService.projectStorage.storageClass=azurefile


client/src/features/dataConnectorsV2/api/data-connectors.enhanced-api.ts
```diff
@@ -268,6 +269,18 @@ const enhancedApi = injectedApi.enhanceEndpoints({
...
+    },
+    postDataConnectorsStorage: {
+      invalidatesTags: ["ProjectStorage"],
+    },
+    deleteDataConnectorsStorageByStorageId: {
+      invalidatesTags: ["ProjectStorage"],
+    },
+    patchDataConnectorsStorageByStorageId: {
+      invalidatesTags: ["ProjectStorage"],
+    },
```
lorenzo-cavazzi: That's the way to go 👍 
Tip: you could even make the cache smarter by considering the IDs. As of now, any operation invalidates _all_ storages, while E.G. patching a storage could invalidate only that storage's cache. I think for this case we can keep this code without overcomplicating

client/src/features/ProjectPageV2/ProjectPageContent/DataConnectors/ProjectConnectDataConnectorsModal.tsx
```diff
@@ -245,10 +274,75 @@ export function ProjectConnectDataConnectorModeSwitch({
         <PlusLg className={cx("fs-3", "me-1")} />
         Create a data connector
       </Label>
+
+      {storageAllowData && projectStorage && projectStorage.length == 0 && (
```
lorenzo-cavazzi: Nitpick: you can condense `projectStorage && projectStorage.[...]` into `projectStorage?.[...]`.
Also, prefer `===` in TypeScript since `==` allows for type conversions with some unexpected results (`===` does what `==` does in most of the other programming languages`)

```suggestion
      {storageAllowData && projectStorage?.length === 0 && (
```

client/src/features/ProjectPageV2/ProjectPageContent/DataConnectors/ProjectConnectDataConnectorsModal.tsx
```diff
@@ -245,10 +274,75 @@ export function ProjectConnectDataConnectorModeSwitch({
...
+  toggle,
+}: ProjectConnectDataConnectorsModalProps) {
+  useEffect(() => {
+    if (!isOpen) {
+      return;
+    }
+  }, [isOpen]);
+
+  return (
+    <ModalBody data-cy="data-connector-search-body" toggle={toggle}>
```
lorenzo-cavazzi: That seems like a copy/paste leftover. We use `data-cy` for Cypress tests, better _not_ to reuse existing tags

```suggestion
    <ModalBody data-cy="data-connector-add-storage-body" toggle={toggle}>
```

client/src/features/ProjectPageV2/ProjectPageContent/ProjectStorage/ProjectStorageForm.tsx
```diff
@@ -0,0 +1,265 @@
...
+  projectStorage?: ProjectStorage;
+  toggle: () => void;
+}
+
+export default function ProjectStorageForm({
+  projectId,
+  namespace,
+  projectStorage,
+  toggle,
+}: ProjectConnectDataConnectorsFormProps) {
```
lorenzo-cavazzi: Nitpick: not a strict rule, but we try to keep the name of the Props interface close to the function it serves (unless it's reused).
`ProjectConnectDataConnectorsFormProps` --> `ProjectStorageFormProps`

client/src/features/ProjectPageV2/ProjectPageContent/ProjectStorage/ProjectStorageForm.tsx
```diff
@@ -0,0 +1,265 @@
...
+            <Label className="form-label" for="mountPath">
+              Mount point
+            </Label>
+            <MoreInfo>
+              <LazyMarkdown>
+                You can either enter an absolute path (starting with `/`) or a
+                relative path (relative to your session&apos;s working
+                directory).
+              </LazyMarkdown>
+            </MoreInfo>
```
lorenzo-cavazzi: Suggestion: This is useful information; I think it makes sense to put it together with the description below ("This is the name of the folder in the working [...]"). Otherwise, it feels like there are 2 places where you give equally valuabe field descriptions

client/src/features/ProjectPageV2/ProjectPageContent/DataConnectors/ProjectDataConnectorsBox.tsx
```diff
@@ -68,13 +71,31 @@ export default function ProjectDataConnectorsBox({
...
   if (error) {
     return <RtkOrDataServicesError error={error} dismissible={false} />;
   }
 
+  if (projectStorageError) {
+    return (
+      <RtkOrDataServicesError error={projectStorageError} dismissible={false} />
+    );
+  }
+
```
lorenzo-cavazzi: Change: I understand why you put this here, but I'm not sure it's the best place. If this fails, users won't see any other data connector, which is probably not what we want. I suggest putting this in a child component

client/src/features/ProjectPageV2/ProjectPageContent/ProjectStorage/ProjectStorageLinkDisplay.tsx
lorenzo-cavazzi, line 1: A general comment here. We probably want a side panel (Offcanvas) to show up when users click. There we can display the current details -- I guess only storage size and mount point.
Otherwise, the page content makes sense 👍  I'll fully review this component once there is an offcanvas as well

lorenzo-cavazzi (commented):
The logic is already great! I added a few inline comments, but it's mostly minor things; you are already on the right track 👍 

P.S. This comment didn't fit inline: you probably want to update `DataConnectorModalHeader` as well, to change the fixed string `data connector` in the title into something like `project storage` when it makes sense.

client/src/features/ProjectPageV2/ProjectPageContent/DataConnectors/ProjectDataConnectorsBox.tsx
boucherjeanphi: Yes, I agree with you. I am considering adding a `renkuToastDanger` instead like this:
```typescript
const { renkuToastDanger } = useRenkuToast();

  useEffect(() => {
    if (projectStorageError) {
      renkuToastDanger({
          textHeader: "There was an error loading the project storage.",
        },
        {
          toastId: "project-storage-error",
        }
      );
    }
  }, [projectStorageError, renkuToastDanger]);
```
What do you think ?

lorenzo-cavazzi: That's one way of handling it. The drawback is that, currently, our toast errors disappear completely when they time out and users might miss them.

You could also just show the `RtkOrDataServicesError` together with the rest of the content. You save it to a constant, and you return it together with `ProjectDataConnectorBoxContent` later, something like:

```
const projectStorageErrorMessage =  projectStorageError ?? <RtkOrDataServicesError error={projectStorageError} dismissible={false} />;

[...]

return (
<>
    <ProjectDataConnectorBoxContent
      data={data}
      project={project}
      inaccessibleDataConnectorsCount={
        inaccessibleDataConnectorsData?.count || 0
      }
      projectStorageData={projectStorageData}
    />
    projectStorageErrorMessage
);
```

P.S: Adjust style/spacing and other details where necessary. TIP: When developing, you can easily generate an error by going to the api file and change to URL to a non-existing one

boucherjeanphi: I changed the code to what you suggested.

client/src/features/ProjectPageV2/ProjectPageContent/ProjectStorage/ProjectStorageForm.tsx
```diff
@@ -0,0 +1,257 @@
...
+  const projectStorageMaxSize =
+    storageAllowData?.max_size ?? PROJECT_STORAGE_MAX_GB;
+
+  const onSubmit = async (values: ProjectStorageFormValues) => {
+    if (!projectStorage) {
+      // Create new project storage
+      const result = await postDataConnectorsStorageMutation({
+        projectStoragePost: {
+          namespace: namespace ?? "",
+          size: Number(values.size),
```
lorenzo-cavazzi: `Number()` is an unnecessary conversion here since the type of `values.size` is already `number`. It's no big deal, but it might obscure type errors if the value’s type changes later

```suggestion
          size: values.size,
```

client/src/features/ProjectPageV2/ProjectPageContent/ProjectStorage/ProjectStorageForm.tsx
```diff
@@ -0,0 +1,257 @@
...
+      if (!result.error) {
+        toggle();
+      }
+    } else {
+      // Update existing project storage
+      const result = await patchDataConnectorsStorageByStorageIdMutation({
+        storageId: projectStorage.id,
+        "If-Match": projectStorage.etag ?? "",
+        projectStoragePatch: {
+          size: Number(values.size),
```
lorenzo-cavazzi:
```suggestion
          size: values.size,
```

client/src/components/icons/RenkuStorageIcon.tsx
lorenzo-cavazzi, line 1: We use icons from the [Bootstrap Icons library](https://icons.getbootstrap.com/) to be sure size and style are consistent, with only a few exceptions. For storage, I would drop this component and pick an icon from the list https://icons.getbootstrap.com/ , perhaps `<Hdd />` or [one of these](https://icons.getbootstrap.com/?q=disk)

client/src/components/entities/entities.types.ts
```diff
@@ -1,5 +1,6 @@
 export type EntityTypes =
   | "code-repository"
   | "data-connector"
+  | "renku-storage"
```
lorenzo-cavazzi: Nitpick: I would avoid using `renku` for these enumerations. I would go for `local-storage` or simply `storage`

client/src/features/ProjectPageV2/ProjectPageContent/DataConnectors/ProjectConnectDataConnectorsModal.tsx
```diff
@@ -196,10 +213,25 @@ function ProjectCreateDataConnectorBodyAndFooter({
...
   switchMode,
+  project,
 }: {
   mode: ProjectConnectDataConnectorMode;
-  switchMode: () => void;
+  switchMode: (mode: ProjectConnectDataConnectorMode) => void;
+  project?: Project;
 }) {
+  const permissions = useProjectPermissions({ projectId: project?.id ?? "" });
+  const canManageProjectStorage =
```
lorenzo-cavazzi: Nitpick: this is redundant and you can simply use `permissions.delete` since it will never be `true` until `permissions.arePermissionsResolved` is also `true`

client/src/features/ProjectPageV2/ProjectPageContent/DataConnectors/ProjectConnectDataConnectorsModal.tsx
```diff
@@ -245,15 +277,72 @@ export function ProjectConnectDataConnectorModeSwitch({
...
+            className="btn-check"
+            id="project-data-controller-mode-add-storage"
+            value="add-storage"
+            checked={mode === "add-storage"}
+            onChange={() => switchMode("add-storage")}
+          />
+          <Label
+            data-cy="project-data-controller-mode-add-storage"
+            for="project-data-controller-mode-add-storage"
+            className={cx(
```
lorenzo-cavazzi, line 302: Nitpick: button groups add extra borders here -- see picture. You can prevent that visual effect by adding here `border-start-0`. You should do the same for the existing "create data connector" option (it wasn't visible with only 2 elements but it is now with 3)

Image

client/src/features/ProjectPageV2/ProjectPageContent/DataConnectors/ProjectConnectDataConnectorsModal.tsx
```diff
@@ -851,5 +941,5 @@ function DataConnectorSearchSourceBadge({
       </div>
     );
 
-  return <p className={cx("mb-0", "small", "text-muted")}>{badgeText}</p>;
+  return <div className={cx("mb-0", "small", "text-muted")}>{badgeText}</div>;
```
lorenzo-cavazzi: Any reason for this change? If you make this a `div`, you can drop `mb-0`. Since it's text, I would leave `p` (not that we are always consistent with tag semantics 🙄)

client/src/features/ProjectPageV2/ProjectPageContent/DataConnectors/ProjectDataConnectorsBox.tsx
```diff
@@ -85,40 +100,55 @@ export default function ProjectDataConnectorsBox({
...
+        inaccessibleDataConnectorsCount={
+          inaccessibleDataConnectorsData?.count || 0
+        }
+        projectStorageData={projectStorageData}
+      />
+      {projectStorageError && (
+        <RtkOrDataServicesError
+          className="mt-2"
+          error={projectStorageError}
+          dismissible={false}
```
lorenzo-cavazzi: Nitpick: I think we don't need to force `dismissable` here. That's mainly meant for cases where users would see _only_ the error and dismissing it would result in a blank (and mostly unusable) page.

client/src/features/ProjectPageV2/ProjectPageContent/ProjectStorage/ProjectStorageForm.tsx
```diff
@@ -0,0 +1,257 @@
...
+          error={
+            postDataConnectorsStorageStatus.error ||
+            patchDataConnectorsStorageByStorageIdStatus.error
+          }
+        />
+        {!projectStorage && (
+          <InfoAlert dismissible={false} timeout={0}>
+            You can add a project storage to this project. This will create a
+            new storage volume that will be mounted in your sessions to avoid
+            data loss on session shutdown.
```
lorenzo-cavazzi: I would rephrase this -- but double-check the text with your build team:
"Project storage will be mounted in your sessions, jobs, and apps. Project members will have read-write access, while others will have read-only.

client/src/features/ProjectPageV2/ProjectPageContent/ProjectStorage/ProjectStorageForm.tsx
```diff
@@ -0,0 +1,257 @@
...
+                {...field}
+                className={cx("form-control", errors.mountPath && "is-invalid")}
+                data-cy="project-storage-form-mount-point-input"
+              />
+            )}
+            rules={{ required: true }}
+          />
+          <div className="invalid-feedback">Please provide a mount point.</div>
+          <div className={cx("form-text", "text-muted")}>
+            This is where the project storage will be mounted during sessions.
```
lorenzo-cavazzi: To keep consistency with the updated text in the info box.

```suggestion
            This is where the project storage will be mounted.
```

client/src/features/ProjectPageV2/ProjectPageContent/ProjectStorage/ProjectStorageForm.tsx
```diff
@@ -0,0 +1,257 @@
...
+            You can either specify an absolute path (starting with `/`) or a
+            relative path (relative to your session&apos;s working directory).
+          </div>
+        </div>
+
+        <div className={cx("d-flex", "gap-2", "justify-content-end")}>
```
lorenzo-cavazzi: You should put the buttons in a `<ModalFooter>` to keep styling consistent with the other modals (e.g. borders).
You might need to change a couple of things in the parent components as well -- E.G. in `EditProjectStorageModal` you have

```
<Modal>
    <ModalHeader
    <ModalBody>
        <ProjectStorageForm>
    </ModalBody>
</Modal>
```

But you might need to drop the `ModalBody` and add it into `ProjectStorageForm`.

There are a bunch of other example in the UI where the pattern is
```
<Form>
    <ModalBody>
    <ModalFooter>
</Form>
```

client/src/features/ProjectPageV2/ProjectPageContent/ProjectStorage/ProjectStorageForm.tsx
```diff
@@ -0,0 +1,257 @@
...
+          <Button
+            color="outline-primary"
+            data-cy="project-storage-form-cancel-button"
+            onClick={() => toggle()}
+          >
+            <XLg className={cx("bi", "me-1")} />
```
lorenzo-cavazzi: Tip: Bootstrap icons no longer require the `bi` class. You may still see it used in many places because the library was only recently updated. No changes needed, just to let you know that `<XLg className="me-1" />` would render exactly the same

client/src/features/ProjectPageV2/ProjectPageContent/ProjectStorage/ProjectStorageLinkDisplay.tsx
```diff
@@ -0,0 +1,356 @@
...
+  useEffect(() => {
+    if (result.isSuccess) {
+      executeOnSuccess();
+    }
+  }, [result.isSuccess, executeOnSuccess]);
+
+  return (
+    <Modal backdrop="static" centered isOpen={isOpen} size="lg" toggle={toggle}>
+      <ModalHeader tag="h2" toggle={toggle}>
+        Are you sure?
```
lorenzo-cavazzi: I would change text:
"Remove project storage permanently"

client/src/features/ProjectPageV2/ProjectPageContent/ProjectStorage/ProjectStorageView.tsx
lorenzo-cavazzi, line 1: Looks good 👍 
As discussed the other day during the meeting, I would add here the same infobox (or similar, if the text should adapt) we have at the creation phase, mentioning the details about visibility and other Storage info

lorenzo-cavazzi (changes requested):
Great PR! 👏 

I went through the code and added some inline comments, mainly minor cosmetic details or small rearrangements. The functionality is already on point.

lorenzo-cavazzi:
P.S. I removed the `#notest` flag in the deploy string so that it runs the acceptance tests too and surfaces errors.

Another comment: it would be great to add local Cypress tests for this feature. For that, you should check the `tests` folder and see what we do for the other data connectors; you can replicate similar tests (E.G: create, open, delete storages). It can be done in another branch

client/src/components/entities/entities.types.ts
boucherjeanphi: What do you think about `project-storage`?

client/src/features/ProjectPageV2/ProjectPageContent/DataConnectors/ProjectConnectDataConnectorsModal.tsx
boucherjeanphi: I made this change because of this small error in the console:
```
In HTML, <div> cannot be a descendant of <p>.
This will cause a hydration error.
```

client/src/features/ProjectPageV2/ProjectPageContent/ProjectStorage/ProjectStorageForm.tsx
boucherjeanphi: I changed this here: https://github.com/SwissDataScienceCenter/renku-ui/pull/4258/changes/9ab5f6b714fe433f68cf268a7d6b1addccae1913
I moved the permission check up in `ProjectConnectDataConnectorsModal` (to prevent API calls on every mode switch) and removed the `<Form>` tag (otherwise scrolling the modal does not show the ModalFooter).

lorenzo-cavazzi:
@boucherjeanphi let me know when this is ready for another check.
I see you addressed most of the points except a couple of things (the icon and the text)

client/src/components/icons/RenkuStorageIcon.tsx
boucherjeanphi: We tried to find on icon for Renku storage from the bootstrap icons library but none was really satisfying, so Jimena proposed this icon from Adobe Stock.

lorenzo-cavazzi (dismissed):
Lgtm! Great addition, some users will be very happy about this feature 👏 

As discussed, let's merge after the backend has been tagged.
P.S. Ping me when that's done. We can tag a new release together; it's an easy procedure, but it requires running an action and merging an additional commit.

lorenzo-cavazzi (approved):
Lgtm! 🚀

PR #4263, opened 2026-07-09
fix: wait for github intergation checks before showing a succesfull message

### LINKED ISSUE #4217: First time integration with Github does not work properly
See the video. This illustrates 2 problems.

1. The page says the integration is complete. But that is not the case. You have to wait for the modal to appear. Then you have to click the link in the modal which allows you to select where to install the application and what repos it should have access to. If you do not wait and just close the page using the provided close button then the integration is never completed. And it will always say that is not complete.

2. Once you do the integration once the link provided in that modal that you have to wait for leads to page with 404. This happens in enterprise cases where the user doing the integration has limited access to control where the integration is installed in.

But the biggest problem is point number 1. Because it is not clear to the user that the integration setup is not complete and they should wait for the modal with the link to show up.

lorenzo-cavazzi:
We now wait to check the integration status on GitHub before automatically closing the redirected page.

image

If the user has just activated the GitHub integration for the first time, there is a different message while invoking the APIs on GitHub and the (optional) modal with additional setup steps shows up again.

Peek 2026-07-09 15-34

/deploy renku=feat/add-jobs
fix https://github.com/SwissDataScienceCenter/renku-ui/issues/4217

### DIFF
--- a/client/src/features/connectedServices/OAuthCompletePage.tsx
+++ b/client/src/features/connectedServices/OAuthCompletePage.tsx
@@ -22,8 +22,9 @@ import { Plugin, XLg } from "react-bootstrap-icons";
 import { Link, useSearchParams } from "react-router";
 import { Button } from "reactstrap";
 
-import { ErrorAlert, SuccessAlert } from "../../components/Alert";
+import { ErrorAlert, InfoAlert, SuccessAlert } from "../../components/Alert";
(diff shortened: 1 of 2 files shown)


olevski:
I tested the CI deployment the interaction works as expected.

andre-code (approved):
lgtm 🚀

PR #4266, opened 2026-07-10
fix: add close alerts popover and show tooltip in alert icon

### LINKED ISSUE #4140: Alert icon in session toolbar always implies an error
The ⚠️ icon in the session page toolbar is always visible, regardless of whether there are actual warnings or errors. This creates a false sense of something being wrong even when the session is healthy. When there are no issues, clicking it does nothing which adds to the confusion.

**Proposed solutions**
- Use a state-aware icon (neutral by default, alert only when needed)
- Add a colored status dot badge to convey state without changing the icon
- Update tooltip copy to reflect current state, e.g. *"Session health · No issues"* vs *"2 warnings · Click to review"*

### LINKED ISSUE #4197: Close notification cloud in session iframe when clicking on the side
To replicate:
- start a session ( I used a ttyd image)
- get a notification
- try to close the little cloud
- the only way to do this is by clicking on the little triangle icon that causes the popup to appear

Expected behaviour:
- clicking outside of the red popup should close it
- maybe we should even add a little `X` in the upper right that will close the window

It is not very intuitive that clicking on the place that cause the notification to show is the  way to close it.

Seen on a feature branch for 2.18.0. But I suspect the same behavior happens on 2.17.2.

Image

andre-code:
Closes #4197 #4140 

Screenshot 2026-07-17 at 09 10 28
Screenshot 2026-07-17 at 09 24 43

/deploy

### DIFF
--- a/client/src/features/sessionsV2/SessionShowPage/SessionAlerts.tsx
+++ b/client/src/features/sessionsV2/SessionShowPage/SessionAlerts.tsx
@@ -111,7 +111,7 @@ function Alerts({ alerts }: AlertsProps) {
   if (!alerts || alerts.length === 0) {
     return (
       <>
-        <span ref={ref}>
+        <span ref={ref} tabIndex={0}>
           <Button
             className={cx(
               "bg-transparent",
@@ -126,6 +126,7 @@ function Alerts({ alerts }: AlertsProps) {
             aria-label="Session alerts"
           >
             <Bell className="bi" />
+            <span className="visually-hidden">Session alerts</span>
           </Button>
         </span>
         <UncontrolledTooltip placement="bottom" target={ref}>
--- a/client/src/features/sessionsV2/SessionShowPage/SessionAlerts.tsx
+++ b/client/src/features/sessionsV2/SessionShowPage/SessionAlerts.tsx
@@ -18,7 +18,7 @@
 
 import { skipToken } from "@reduxjs/toolkit/query";
 import cx from "classnames";
-import { Fragment, useEffect, useRef, useState } from "react";
+import { Fragment, useCallback, useEffect, useRef, useState } from "react";
 import { Bell, BoxArrowUpRight, XLg } from "react-bootstrap-icons";
 import ReactMarkdown from "react-markdown";
 import { Link } from "react-router";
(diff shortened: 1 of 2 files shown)


client/src/features/sessionsV2/SessionShowPage/SessionAlerts.tsx
```diff
@@ -95,33 +99,39 @@ function Alerts({ alerts }: AlertsProps) {
...
-
       setIsOpen(true);
     }
 
     prevAlertIdsRef.current = currentAlertIds;
   }, [alerts]);
 
-  const togglePopover = () => setIsOpen(!isOpen);
+  const togglePopover = () => setIsOpen((open) => !open);
+  const closePopover = () => setIsOpen(false);
```
leafty: `useCallback` should be used here.

client/src/features/sessionsV2/SessionShowPage/SessionAlerts.tsx
```diff
@@ -95,33 +99,39 @@ function Alerts({ alerts }: AlertsProps) {
...
-            "text-dark",
-          )}
-          data-cy="session-alerts"
-          innerRef={ref}
-        >
-          <ExclamationTriangle className="bi" />
-        </Button>
-      </div>
+      <>
+        <span ref={ref}>
```
leafty: The span element needs a `tabIndex` => https://getbootstrap.com/docs/5.3/components/popovers/#disabled-elements

client/src/features/sessionsV2/SessionShowPage/SessionAlerts.tsx
```diff
@@ -95,33 +99,39 @@ function Alerts({ alerts }: AlertsProps) {
...
+              "no-focus",
+              "p-0",
+              "shadow-none",
+              "text-dark",
+            )}
+            disabled
+            data-cy="session-alerts"
+            aria-label="Session alerts"
+          >
+            <Bell className="bi" />
```
leafty, line 131: Needs button text:


```suggestion
            <Bell className="bi" /><span className="visually-hidden">Session alerts</span>
```

client/src/features/sessionsV2/SessionShowPage/SessionAlerts.tsx
andre-code: Good point to raise, but I don't think useCallback adds much value here. `togglePopover` and `closePopover` are only used by the Button, Popover, and a native <button>, so keeping a stable function reference doesn't really give us any benefit. They also only use the stable setIsOpen setter, so there's no concern about stale state.

leafty: The current code re-creates new methods on every React render of the component, which forces children components to re-render again. Using `useCallback` allows React to optimize the render loop and not re-render children components because the function references are stable.

andre-code:
@leafty Thanks for the review! I've applied your suggested changes.

leafty (approved):

PR #4271, opened 2026-07-17
feat: improve anonymous landing page

andre-code:
### Changes
CustomizedAnonymousHome restructured:
- Header moved out of the background-image div into a full-width `#rk-anon-home-custom-header.bg-navy` wrapper, it fixes the invisible nav bug from [Discussion #4092](https://github.com/SwissDataScienceCenter/renku-ui/discussions/4092).
- Added `<BottomNav />` so the custom page matches the standard footer
- Add `sanitize={false} ` for  custom page
- Added e2e tests for customized homepage

closes #4094 

/deploy

### DIFF
--- a/client/src/features/landing/AnonymousHome.tsx
+++ b/client/src/features/landing/AnonymousHome.tsx
@@ -90,9 +90,6 @@ function CustomizedAnonymousHome(props: AnonymousHomeConfig) {
   }
   return (
     <>
-      <div id="rk-anon-home-custom-header" className="bg-navy">
-        <HomeHeader />
-      </div>
       <div
         id="rk-anon-home-section1"
         style={{
--- a/client/src/features/landing/AnonymousHome.tsx
+++ b/client/src/features/landing/AnonymousHome.tsx
@@ -101,7 +101,7 @@ function CustomizedAnonymousHome(props: AnonymousHomeConfig) {
         <div className="rk-anon-home-section-content">
           <Row>
             <Col className={cx("rk-pt-l", "rk-w-s")}>
-              <LazyMarkdown allowClasses>{content}</LazyMarkdown>
+              <LazyMarkdown sanitize={false}>{content}</LazyMarkdown>
             </Col>
           </Row>
         </div>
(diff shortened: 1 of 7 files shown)


lorenzo-cavazzi (commented):
The page structure is still different between logged in and logged out. In the image, you see on the left the logged-in page structure with `header` at the same level as the `footer`, while on the right is the logged-out page with the `header` still deeply nested in the main container

Image


Also, I see you added `allowClasses`, but the issue mentioned avoiding sanitizing, for which we can use the existing `sanitize` parameter. That's more flexible and would let admins do anything they want, not only pass classes

andre-code:
Thanks @lorenzo-cavazzi, I completely forgot about that part of the issue. I've just pushed a commit with that included.

For the sanitization, I went with `allowClasses` because passing classes was the specific case we discussed, but I agree `sanitize={false}`  is simpler and more flexible.

client/src/features/landing/AnonymousHome.tsx
```diff
@@ -88,23 +89,25 @@ function CustomizedAnonymousHome(props: AnonymousHomeConfig) {
...
+        id="rk-anon-home-section1"
+        style={{
+          backgroundImage: `url(${backgroundUrl})`,
+          backgroundSize,
+          backgroundRepeat: "no-repeat",
+        }}
+      >
+        <div className="rk-anon-home-section-content">
+          <Row>
+            <Col className={cx("rk-pt-l", "rk-w-s")}>
```
lorenzo-cavazzi: The classes here are leftovers from Renku v1. We can actually drop these additional divs altogether

```suggestion
```

lorenzo-cavazzi (changes requested):
Works well now! 👏 
Just a minor change request to remove some unnecessary divs.

@sgaist would you like to give this a try before we merge and release?

client/src/features/landing/AnonymousHome.tsx
andre-code: Sure, I have added that change in the last commit.

lorenzo-cavazzi (approved):
Lgtm! 🚀

PR #4273, opened 2026-07-17
feat: allow hibernating session when session is starting

### LINKED ISSUE #4191: Enable pausing a session that is starting or failing
It may happen that a session has been running, is paused and then cannot be resumed. For example there isn't enough compute resources or something has gone wrong. In this case it would be nice to allow pausing the session that is stuck on starting from the ui. But currently this is not possible. The pause button does not need to be prominent but it should be available in the dropdown or in the iframe for the session.

Seen on renku 2.17.2.

andre-code:
Pr to allow hibernating session when session is starting
Screenshot 2026-07-21 at 10 36 58

closes #4191 

/deploy


lorenzo-cavazzi (approved):
Lgtm! 🚀

PR #4274, opened 2026-07-17
feat: collapse project copy button

### LINKED ISSUE #4188: Collapse copy project button
In the project's page "Info" section, we have a dropdown menu and a single option to copy the project.
We can remove the dropdown and attach the copy action straight to the icon -- style it similarly to the "create" button from the other section, but with a copy icon instead of a plus

Image

andre-code:
Collapse project button.
 
Screenshot 2026-07-21 at 10 34 32

/deploy

### DIFF
--- a/client/src/features/ProjectPageV2/ProjectPageContent/ProjectInformation/CopyProjectButton.tsx
+++ b/client/src/features/ProjectPageV2/ProjectPageContent/ProjectInformation/CopyProjectButton.tsx
@@ -49,6 +49,7 @@ export default function CopyProjectButton({
         innerRef={ref}
       >
         <BootstrapCopyIcon className="bi" />
+        <span className="visually-hidden">Copy project</span>
       </Button>
       <ProjectCopyModal
         currentUser={currentUser}
--- a/client/src/features/ProjectPageV2/ProjectPageContent/ProjectInformation/CopyProjectButton.tsx
+++ b/client/src/features/ProjectPageV2/ProjectPageContent/ProjectInformation/CopyProjectButton.tsx
@@ -42,7 +42,7 @@ export default function CopyProjectButton({
   return (
     <>
       <Button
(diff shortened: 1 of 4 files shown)


client/src/features/ProjectPageV2/ProjectPageContent/ProjectInformation/CopyProjectButton.tsx
```diff
@@ -16,49 +16,48 @@
...
-      {
-        <ProjectCopyModal
-          currentUser={currentUser}
-          isOpen={isCopyModalOpen}
-          project={project}
-          toggle={toggleCopyModal}
-        />
-      }
+      <Button
+        data-cy="add-code-repository"
```
lorenzo-cavazzi: Looks like a copy/paste leftover?
I would actually change the previous `project-copy-menu-item` reference to something like `copy-project-button`; mind you should also update it in the `projectV2` spec tests.

```suggestion
        data-cy="copy-project-button"
```

lorenzo-cavazzi (changes requested):
Looks good! Let's update the `data-cy` property before merging.
I also suggest removing the `SingleButtonWithMenu` component since it's not used anywhere else and it's not a pattern we are likely to use anywhere else.

client/src/features/ProjectPageV2/ProjectPageContent/ProjectInformation/CopyProjectButton.tsx
andre-code: yes, it is.
Thanks for pointing it out. I fixed and updated the e2e tests.

lorenzo-cavazzi:
@andre-code looks good! :rocket: 

Do you mind dropping the `SingleButtonWithMenu` component? Then we can merge :+1:

lorenzo-cavazzi (approved):

PR #4286, opened 2026-07-23
feat: update no available providers message

### LINKED ISSUE #4204: Update the message to activate new integrations
We show the following message when the user is already connected to all the available integrations

Image

It turns out to be confusing; we might want to update it. If we can easily check the connections, we can explicitly say "you have already connected to all the available integrations" and distinguish the case "there are no integrations available in this RenkuLab deployment". We should add a message similar to the "contact us" card to encourage the user to reach out if they want a not-yet-existing integration.

andre-code:
PR to update the message when there are no more providers available to connect

Screenshot 2026-07-28 at 15 27 16

/deploy

### DIFF
--- a/client/src/features/connectedServices/ConnectedServicesPage.tsx
+++ b/client/src/features/connectedServices/ConnectedServicesPage.tsx
@@ -129,7 +129,7 @@ export default function ConnectedServicesPage() {
   const [isAddIntegrationModalOpen, setIsAddIntegrationModalOpen] =
     useState(false);
 
-  const { mainListProviders, modalProviders } =
+  const { startedProviders, availableProviders } =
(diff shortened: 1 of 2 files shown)


lorenzo-cavazzi (changes requested):
Nice!

Sorry for this nitpick comment, but I would change the message slightly.
Since we don't check the connection status, "Pending" counts as an active integration, and the message "connected to all the available integrations" might trick some users into believing the connections work.

What about:
> There are no other integrations to activate. If you wish to change an existing integration, you can click on the Remove or Disconnect button and connect again.

image

lorenzo-cavazzi (approved):
Lgtm! 🚀

PR #4297, opened 2026-07-28
feat: replace favicon 

### LINKED ISSUE #4276: Update Renku UI favicon assets
I've uploaded the new favicon assets to the shared Drive folder in both SVG and PNG formats.

https://drive.google.com/drive/folders/1-mZu-_4QhSUMdJFj5qyH0Gn_5hcDlhpD?usp=sharing

Available sizes:

512×512
192×192
32×32
16×16

A few notes:

- For the 32×32 and 16×16 versions, I simplified the frog into a more compact silhouette so it stays recognizable and doesn't get too distorted at smaller sizes.
- For the 512×512 version, I think we can keep the current frog logo with the dots.
- For the 192×192 version, I created two variants: dots and square.

I also reviewed the colors and used the same ones currently used in the UI logo: green #00A877 and pink #F24BA0.

It would be nice to try the different versions in the UI and see which combination works best before we settle on the final favicon set.

andre-code:
PR to update favicon files with the right Renku colors

closes #4276 

/deploy


lorenzo-cavazzi (approved):
Lgtm! 🚀

PR #4298, opened 2026-07-28
fix: UI search glitches

### LINKED ISSUE #4078: Search results glitches
After the latest updates to the search page, there are a couple of details to improve:
- Once the creation date is selected as a filter, the result number includes a strange encoding for the creation date. It should use the same definition used in the filters. E.G., in the example, instead of `<today-90d` it should be `older than 90 days`
  Image

- Groups icon occasionally wraps -- perhaps it's not group-specific but rather content-width related; in that case, it's likely just a missing `flex-nowrap`
  Image

- The logic to show only a limited number of members seem to fail when there are too many
  Image

- [nice-to-have] Groups have a creation date; it should be returned from the backend query so that the filter is usable in the UI
  Image
  P.S. we might consider adding it for users too -- sync with product team.

andre-code:
PR to fix Search UI glitches:

- Show human-readable filter labels in the results summary (e.g. "older than 90 days" instead of <today-90d)
- Stop the type icon from wrapping onto a new line in search results
- Fix the member list so long lists show a +N overflow badge instead of overflowing
- Add a Cypress test for the long member list overflow

### Out of scope
Groups creation date from the backend (nice-to-have from #4078)

close #4078 

/deploy


client/src/features/searchV2/components/MemberListRow.tsx
```diff
@@ -134,11 +162,13 @@ export default function MemberListRow({ members }: MemberListRowProps) {
...
                   }
-                : { minWidth: 0 }
+                : shouldTruncate
+                  ? { maxWidth: truncateMaxWidth, minWidth: 0 }
+                  : undefined
             }
           >
             <UserAvatar namespace={member.namespace ?? ""} />
-            <span className="text-truncate">
+            <span className={cx("text-truncate", "min-w-0")}>
```
lorenzo-cavazzi, line 171: Does `min-w-0` actually do anything? I think we mistakenly used that a couple of times but it doesn't set the minimum width to 0 🙊

lorenzo-cavazzi: I wonder if we should just add the class to `renku_additional_classes.scss`. It's actually useful in a few places and I don't find anything similar in Bootstrap

lorenzo-cavazzi (changes requested):
Great fixes! 👏 
I fear we have been using a non-existent CSS class in a few places 🤦 see inline comment. I suggest addressing that problem and merging.

P.S. The mechanism to show the names works great now! The code got a little intricate and it's hard to follow, but it works fine and includes a test, which is good 👍

client/src/features/searchV2/components/MemberListRow.tsx
andre-code: You’re right, I mixed this up with a Tailwind utility. I add a .min-w-0 { min-width: 0; } utility in renku_additional_classes.scss.

lorenzo-cavazzi (approved):
Lgtm! 🚀

PR #4310, opened 2026-08-05
chore: remove deprecated baseUrl

lorenzo-cavazzi:
Remove deprecated baseUrl from the config file

/deploy


leafty (commented):
Note: should also do in `server/tsconfig.json`

leafty (approved):

PR #4311, opened 2026-08-05
feat: get culling thresholds from parameters

lorenzo-cavazzi:
The renku-notebook `/version` endpoint was rerouted for some time, but now it no longer returns anything. This PR reads the threshold values from the parameters instead of deriving them from the (now defunct) API payload.

/deploy renku=lorenzo/add-culling-thresholds extra-values=notebooks.culling.idleSecondsThreshold.registered=14400


lorenzo-cavazzi:
image

leafty (approved):

PR #4318, opened 2026-08-10
feat: add id code to project, launcher and data connector info 

### LINKED ISSUE #4268: Add IDs in sidepanels
We do not display IDs in the ui. But they are used by admins and the cli in many cases.

Therefore it would be nice to show the id for things in the side-panel for things that have the off-canvas side panel.

For things without a side panel - i.e. projects then it should be shown in the info box shown on the side.

Image

We should also add a button to copy the ID.

For example the CLI currently needs the session launcher ID to launch jobs. But the UI never displays the ID of session launchers.

andre-code:
PR to include ID in pages and side panels.

- Identifiers are located in Info cards
- Internal Ids are in side panels and the Settings pages

closes #4268 

/deploy


olevski:
That looks great @andre-code! It will be helpful with admin tasks and checks and troubleshooting.

lorenzo-cavazzi (commented):
I like the approach but I would try to be more consistent with how we show the information (E.G. the identifier has been added to groups and users with an `@` prefix, to the data connector without the prefix, and nothing added to the project page). I would also use the name `Internal ID` or `Resource ID`

* Group and user pages: Remove the `@` prefix from the `Identifier` and show the `Internal ID` at the bottom.
* Project page: Add an `Identifier` on top (the `namespace/slug` string) and rename ID Code to `Internal ID`
* Data connector: Rename ID Code to `Internal ID`
* Session launcher: The new Card on the bottom can be called "Info" and use the same icon as in the other resources. Drop the description and leave only `Internal ID: <copy/paste>`

andre-code:
> I like the approach but I would try to be more consistent with how we show the information (E.G. the identifier has been added to groups and users with an `@` prefix, to the data connector without the prefix, and nothing added to the project page). I would also use the name `Internal ID` or `Resource ID`
> 
> * Group and user pages: Remove the `@` prefix from the `Identifier` and show the `Internal ID` at the bottom.
> * Project page: Add an `Identifier` on top (the `namespace/slug` string) and rename ID Code to `Internal ID`
> * Data connector: Rename ID Code to `Internal ID`
> * Session launcher: The new Card on the bottom can be called "Info" and use the same icon as in the other resources. Drop the description and leave only `Internal ID: <copy/paste>`

@lorenzo-cavazzi  What about including the Internal ID at the end in the **settings** page?  User page doesn't have an internal ID and also no settings page so far, so it is compatible with that as well. 

Also, @olevski do you think it would be ok if we limit the visualization to admin users only?

lorenzo-cavazzi:
We do have an ID for users
```
{
    "id":"e3d6eea9-3c69-48f2a37a-368467cb3735",
    "username":"lorenzo.cavazzi.tech",
    <...>
}
```
But I guess that's the least relevant page; we can omit that for users if it doesn't fit into the UI

lorenzo-cavazzi:
@andre-code you asked for another review but I don't see any new commits. Did you forget to push?

andre-code:
> @andre-code you asked for another review but I don't see any new commits. Did you forget to push?
@lorenzo-cavazzi  My bad! It now includes all the changes we discussed. Thanks in advance!

lorenzo-cavazzi (approved):
Lgtm! 🚀

PR #4326, opened 2026-08-13
fix: align icons properly in session launcher dropdown

lorenzo-cavazzi:
Minor visual glitch: the EnvironmentIcon icon has a larger-than-usual right margin

-----

**BEFORE**
    Screenshot_20260813_105434

-----

**NOW**
    image


leafty (approved):
LGTM. Not sure why the default was `me-2`.

PR #4338, opened 2026-08-19
feat: show group resource pools in group info

boucherjeanphi:
Display all resource pools a group is connected to in the info box on the group page.

Screenshot from 2026-08-26 15-28-42

Backend returns a 404 error if the user is not logged in / not member of the group, in which case we display the message "This group has no visible resource pools."

/deploy


client/src/features/groupsV2/show/GroupV2ResourcePoolDisplay.tsx
```diff
@@ -0,0 +1,91 @@
...
+    <GroupInformationBox
+      icon={<Cpu className="bi" />}
+      title={
+        <>
+          <span>Resource Pools</span>
+          <Badge>{data.length ?? 0}</Badge>
+        </>
+      }
+    >
+      {data.length === 0 && <p>There are no resource pools in this group.</p>}
```
lorenzo-cavazzi: * `mb-0` --> remove the additional space on the bottom
* `text-body-secondary` --> we sometimes use that to better distinguish explenatory text from actualy values

```suggestion
      {data.length === 0 && <p className={cx("mb-0", "text-body-secondary")}>There are no resource pools in this group.</p>}
```

lorenzo-cavazzi (changes requested):
I see an error when I'm not logged in. I think this information should be available to non-logged-in users (something to adjust in the backend) since groups are public. The alternative would be to not query the endpoint when users aren't authenticated, and not show the entry altogether.
Image

Actually, I see that error for all groups I haven't created, so that needs a fix before we can ship it
Image

I'll finish reviewing the feature when that's fixed.

client/src/features/groupsV2/show/GroupV2ResourcePoolDisplay.tsx
```diff
@@ -0,0 +1,85 @@
...
+      )}
+    </GroupInformationBox>
+  ) : null;
+}
+
+interface GroupV2ResourcePoolProps {
+  rp: ResourcePoolWithIdFiltered;
+}
+function GroupV2ResourcePool({ rp }: GroupV2ResourcePoolProps) {
+  return rp.name && <p className={cx("m-0", "text-truncate")}>{rp.name}</p>;
```
lorenzo-cavazzi: I'm not sure about the `"text-truncate"`. I can imagine pools with a not-so-short name and the info box has a rather short width.

I suggest using a list instead. It requires a `ul` wrapper in the parent component too

```suggestion
  return rp.name && <li>{rp.name}</li>;
```

Here is the result

Screenshot_20260828_122148

client/src/features/groupsV2/show/GroupV2ResourcePoolDisplay.tsx
```diff
@@ -0,0 +1,85 @@
...
+        </p>
+      ) : !data.length ? (
+        <p
+          className={cx("mb-0", "text-body-secondary")}
+          data-cy="group-resource-pools-empty"
+        >
+          There are no resource pools explicitly linked to this group.
+        </p>
+      ) : (
+        data.map((rp) => <GroupV2ResourcePool key={rp.id} rp={rp} />)
```
lorenzo-cavazzi:
```suggestion
        <ul className={cx("ps-3", "mb-0")}>
          {data.map((rp) => (
            <GroupV2ResourcePool key={rp.id} rp={rp} />
          ))}
        </ul>
```

lorenzo-cavazzi (commented):
Now it looks very nice 👏  And the code is good!

Only one comment about the visualization of the resource pool -- see inline

lorenzo-cavazzi (approved):
Lgtm! 🚀

PR #4340, opened 2026-08-20
fix: prevent silently disabled secret location input

lorenzo-cavazzi:
The secrets location input isn't editable and it's confusing for users, since they would try to edit it

secrets-slot

This PR renders it as disabled and makes the edit button more prominent

image

/deploy


andre-code (approved):
lgtm 🚀🙌

PR #4341, opened 2026-08-20
fix: replace PNG favicon with background-free version

andre-code:
/deploy


lorenzo-cavazzi (approved):
👍

lorenzo-cavazzi (commented):
New icons are much bigger than the name suggests. Is that intended?

Image

lorenzo-cavazzi (changes requested):

lorenzo-cavazzi:
@andre-code I'm not sure if the new icons were downgraded while resizing, but the quality seems worse than the previous ones. Here is the 32 px icon zoomed in 5 times -- left is the older, right is the newer.

The new one seems way less defined to me, no?

image

lorenzo-cavazzi:
If the PNG was resized, I would try instead to re-export from the SVG directly to the target resolution

andre-code:
> If the PNG was resized, I would try instead to re-export from the SVG directly to the target resolution

Sure, I’ll take a look with Jimena since she has the original files, and we can compare them to see what happened

jdupre81:
Hi, just responding to the question about size and quality: I think the issue is that the previous icons were not at the correct size. They were larger, which is probably why they appeared to have better quality.

In case we want to give it another try, here are the PNG and SVG files in the correct sizes.
Favicon 16px_BG_transparent
Favicon 32px_BG_transparent
Favicon 16px_BG_transparent
Favicon_32px_BG_transparent

lorenzo-cavazzi:
@jdupre81 The resolution is the same; I was comparing the latest PNG with what we are using now, not the first higher resolution upload.

I opened your PNG but it's the same. I tried opening them in an editin app.
The first version has completely filled pixels with fading to white. I think the best strategy is to replace those fading pixels with fading to transparency.

Screenshot_20260825_115028

In the new version, many pixels in the middle are partially transparent. Very few are actually 100% solid, which is what makes the final rendering look worse

Screenshot_20260825_115056

PR #4344, opened 2026-08-20
fix: remove duplicate contact us card in connected services page

### LINKED ISSUE #4343: Duplicated contact us card in integration page
The integration page show the contact us card twice.

Image

andre-code:
closes #4343 

/deploy


lorenzo-cavazzi (commented):
I think we want to remove the duplicate but still show the message once.

This is what I get now (left) vs what I got before the change (right) on an empty deployment

Image

andre-code:
> I think we want to remove the duplicate but still show the message once.
> 
> This is what I get now (left) vs what I got before the change (right) on an empty deployment
> 
> Image

Sure, I added it to be always visible.

lorenzo-cavazzi (approved):
Lgtm! 🚀

PR #4345, opened 2026-08-20
refactor: use the RenkuBadge wherever reasonable

### LINKED ISSUE #4328: Swap old Session*Badge component with RenkuBadge
We recently standardized the use of Badges to keep a consistent color scheme, but didn't update all the older components where we were setting all the classes manually.

An example is the `SessionBadge` component and the other variations (E.G: `BuildStatusBadge`), and we accidentally used them as a template for new badges. We should replace them with the newer `RenkuBadge` and avoid separately adjusting borders/colors -- see https://github.com/SwissDataScienceCenter/renku-ui/pull/4327 as an example

lorenzo-cavazzi:
Replace all the older variations of Badges with colored borders with the standardized `RenkuBadge` color schemes.

/deploy
fix https://github.com/SwissDataScienceCenter/renku-ui/issues/4328


andre-code (approved):
lgtm 🚀

PR #4358, opened 2026-09-07
fix: allow links to be opened from ttyd

leafty:
This change allows links to be clicked on when using TTYD as a frontend. For example, the `start_tunnel` script prints the device link to connect the user's GitHub account to start the vscode tunnel.

Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe#allow-popups-to-escape-sandbox

## How to test
* Current: https://dev.renku.ch/p/flora-super-group/test-global-images/sessions/01M1XSRAVTDJ4C5PY2WV4RWTGV/start
* Fixed: https://renku-ci-ui-4358.dev.renku.ch/p/flora.thiebaut/flora-test/sessions/01M1XSF4TF9SG8E6F7TZA01E9B/start

Open the session, then in the console, type:

```bash
echo "https://docs.renkulab.io/"
```

The link doesn't open on dev/prod, but it opens in the PR deployment.

/deploy


lorenzo-cavazzi (approved):
🚀

PR #4368, opened 2026-09-17
feat: update pages color background

andre-code:
PR to update page background color

/deploy


client/src/styles/renku_body.scss
lorenzo-cavazzi, line 1: Early comment on this: Bootstrap has `secondary` and `tertiary` variations of the background color, perhaps we can redefine the color value for the "tertiary" variation instead of defining a new variable name?
https://getbootstrap.com/docs/5.3/utilities/background/#background-color

